PR #5, opened 2015-03-16
Add Classifiers from Gramps pypi entry.

sam-manzi:
https://pypi.python.org/pypi/Gramps/


dsblank:
Thanks, Sam, for all of these!

PR #6, opened 2015-03-16
Year 2015

sam-manzi:
Update the year


dsblank:
Since master is purely Python 3 now, should we keep the Python2 u"" strings here?

sam-manzi:
Wasn't this generated by sphinx?  Would regenerating with sphinx under python 3 create the same result?

dsblank:
I have no idea... I was just looking at the changes. I guess we can
manually make the change here, and regenerate it later.

Thanks!

PR #12, opened 2015-04-03
Change to git.

daleathan:


dsblank:
No you know what this XML id is for? Or should it just be removed?

Nick-Hall:
The uid is a SourceForge username.  It was used by the svn2cl utility to generate a change log.  I think that we can remove it now.

The authors.xml file is also used by the About dialog, so we need to make sure it still works after the uid is removed.

dsblank:
@daleathan Can you remove these lines completely and see if the dialog still works?

daleathan:
I gather you meant that I should only removed the content of the field?

daleathan:
Looks like the build failed because of the last commit.

sam-m888:
@daleathan I've tested your PR against master and the About box does work.

daleathan:
@sam-m888 Thanks, but I can not merge the PR, that's up to you guys :)

dsblank:
@daleathan Thanks for checking the dialog. Can these UID's just be removed?

daleathan:
@dsblank It will be a few weeks before I can check.

daleathan:
No idea, I'll let you guys work this one out.

sam-m888:
Done:  by Nick-Hall  who  committed: Remove SourceForge username from authors file (This was used by the svn2cl utility to generate a change log.) commit f125a982d65c02885ba110a3613e15b3ace34557 

https://github.com/gramps-project/gramps/commit/f125a982d65c02885ba110a3613e15b3ace34557

PR #14, opened 2015-04-04
Create readme with existing instructions.

daleathan:
Copied from http://sourceforge.net/projects/gramps/files/Unstable/readme.rst

Have not managed to get this running yet.


dsblank:
Very useful! This is the year of the webapp... other devs have asked for docs like this so that they can join in on working on this, so this is perfect timing. Do you have particular interest in this part of gramps?

dsblank:
Can you rename the file to README.md? That would be more standard, I think and will stand out.

daleathan:
Yes, I'm interested in the webapp as I really don't want to use a commercial website if I have to.

Doug, am I right in guessing that because gramps master is now solely python 3 that the django 1.7 and 1.8 (LTS) and higher versions are only supported? 

https://docs.djangoproject.com/en/1.8/faq/install/#what-python-version-can-i-use-with-django

dsblank:
The webapp code hasn't been brought up to date with the latest django, nor the latest gramps database changes (or codebase changes). I'll see what needs to make it functional, as it currently exists. Then some work needs to be done to make it 100%.

daleathan:
That would be great, thank you

PR #15, opened 2015-04-09
Make place type in Place Reference Editor editable, add tooltip (maintenance/gramps41)

sebschub:
The place type field in the Place Editor is editable and has a tooltip. This simple change adds these properties in the Place *Reference* Editor (maintenance/gramps41 branch).


Nick-Hall:
Thank you.

PR #18, opened 2015-05-05
modified:   po/sl.po

beernarrd:
being new with github would like to proceed step by step with slovenian translation updates. this time po file, datehandler will follow when ready


jralls:
Cherry-picked in the sl.po commit. You messed up your master branch with that merge, I suggest that you reset and force-push to get rid of it, then see why your upstream/master was so far behind the actual master. If you need help with that, ask on gramps-devel@lists.sourceforge.net.

PR #23, opened 2015-05-11
Modification for exemple.gramps database (custom parent relationship,…

belissent:
Modification for example.gramps database 
(custom parent relationship, notes):

- Added custom parent relationship for I0044
- Added notes with ID: '_header1', '_footer1', '_custom1'

These modifications are used for the DynamicWeb report addon tests.


dsblank:
Thanks @belissent !

@Nick-Hall Does this look ok to you?

Nick-Hall:
Yes.  Adding extra examples to test new reports is useful.

One minor comment.  The first line of the commit message is too long, and the file name is misspelt.  I'm also not sure why the media path was changed, but I don't think that it is important.

PR #25, opened 2015-05-26
new framework slovenian datehandler, updated sl.po, including plural forms correction

beernarrd:
Inflections corrected in sl.po
Date parser corrected
Plural forms expression corrected. Range, Span, Estimated, Calculated precised.
Estimated and Calculated strings added.
updated sl.po with the last pot file

### DIFF
--- a/gramps/gen/datehandler/_date_sl.py
+++ b/gramps/gen/datehandler/_date_sl.py
@@ -22,9 +22,9 @@
 # Slovenian version 2010 by Bernard Banko, based on croatian one by Josip
 
 """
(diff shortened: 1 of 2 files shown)


bpisoj:
Did you see:
https://travis-ci.org/gramps-project/gramps/builds/64160562

Maybe it fails because previous failed commit from build 360?
(https://travis-ci.org/gramps-project/gramps/builds)

dsblank:
Yes, build 360 was the culprit with all of the new PlaceName changes. Fixed now.

PR #26, opened 2015-06-02
Geps/gep 032 database backend

dsblank:
This PR represents all of the changes for moving the BSDDB-specific parts into a plugin. 

Currently, there is a dialog that pops up when creating a new database. It is imagined that this will be replaced by a config setting, and that the choice of backend will be made in Preferences. For easy testing, this is left as a dialog in creating the database.

Django and DB-API backends have been moved to addons. DictionaryDB and BSDDB are installed as part of core. DictionaryDB is functional, but will be revised based on the DB-API work, which covers everything except undo/redo functions.

There are two new functions in DbState: make_database, and open_database. We may wish to make DbState a singleton, but shouldn't be a problem either way. Currently, each DbState instance will load the database plugins, if they have not been loaded yet globally.

gramps -L has been refactored to use the new abstractions. It opens each database, and can report a variety of information (including numbers of items of each type). Opening each database may be time consuming, and may require additional limitations (such as "do not open for summary").

Left to do over the next few months:

- [ ] Move upgrade from BSDDB for general use (or separate into general and specific parts)
- [ ] Add a gramps schema version to each backend
- [ ] Complete DB-API, and test thoroughly
- [ ] Add unittests for all backends; we want to make sure they are 100%


sam-m888:
@dsblank Thanks for the great effort of work that when into this.

PR #28, opened 2015-06-04
6366: Correct text on "Display tab"

sam-m888:
from "sidebar" to "Navigator"


sam-m888:
https://gramps-project.org/bugs/view.php?id=6366

PR #35, opened 2015-06-15
8047 - Add new --resourcepath option to setup.py

RossGammon:
This new --resourcepath option to setup.py allows the path to the installed location of the Gramps resources to be passed to setup.py when Gramps is being built out of the source tree in a temporary location (e.g. when packaging for a distribution).
Otherwise, the resource-path file (in gramps/gen/utils) will contain the path to the temporary build location, and Gramps fails to run from the installed location.


dsblank:
Looks good! Thanks, and glad to get this in before gramps 4.2.

PR #40, opened 2015-06-30
Slovenian translation update

beernarrd:
Updated on last pot file. The number of additions and deletions is miscalculated. Probably due to --no-wrap option of msgmerge?


Fedik:
>The number of additions and deletions is miscalculated ...

It also because: when do update then msgmerge also do update for the file info for each string, like:
screen 2015-07-01 10 23 38 617x238

romjerome:
Done on c731c68ec5c719bf5a94b9fe136d5c5c315a3b3e

PR #41, opened 2015-07-12
8663: add exception for UnicodeEncodeError.

ennoborg:


dsblank:
There may be other things to do to fix this, but this seems like it makes this function more robust, especially when there is no actual file.

PR #42, opened 2015-07-14
Make Github recognise Gramps as a Python project

sam-m888:
See https://github.com/github/linguist/issues/2462


dsblank:
We might be able to delete those, as well.

sam-m888:
Thanks

sam-m888:
@dsblank re:"We might be able to delete those, as well."

Sounds good, are they no longer needed, or can they be generated? 

Current total size of both .sql files is 13 MB.

PR #45, opened 2015-07-18
Stop using concatenation to create docstring

cjmayo:
Fixes gramps failing to start if file is byte-compiled and optimized
which sets ```__doc__``` to None.


dsblank:
Makes sense. Thanks!

PR #49, opened 2015-07-20
Slovak translation updates

ondrokrc:


dsblank:
@vperic Do you think we should wait till your concerns are addressed, or commit as-is and refine later?

Paul-Franklin:
@ondrokrc

Hi,

I was wondering whether you use a program to help you translate?

https://gramps-project.org/wiki/index.php?title=Translating_Gramps#Translating_messages

I have heard several of our translators use poedit for instance, which
is available in both Windows and linux versions.

I am told that one of the things it does is to check your
translations for things like those underscores.

I think it also does things like change the date when you
translated the file, when you write out a new translation.

(You sk.po says you translated it in "2014" for instance,
even though it also clearly says you used a modern 2015
template (gramps.pot) -- both of those dates are up in the
header information, at the top of your file.)

Thank you for helping make gramps better.

We appreciate it.

ondrokrc:
Thanks, I'll look into some tool. It's true I've forgotten about the date (but not about the underscores - see my comment above).

sam-m888:
I'm guessing this has already been merged.

PR #50, opened 2015-07-24
Move possibly erroneously placed return statements

sam-m888:
Came across these when investigating 7867 Gramplets "Quick View" does not immediately show entries


dsblank:
This looks like it never worked, since the very beginning. Anyway to test this code? Even better if you could write unittests :+1:  :smile:

PR #52, opened 2015-08-04
webkit dependency no longer required, HTML View removed

RossGammon:
Just a tidy up of README after noticing the webkit dependency was dropped by Jerome, and the NEWS file mentions that the HTML view has been removed (in the recent 4.2.0 release).

This probably should also be applied to the 4.2 branch.

### DIFF
--- a/README
+++ b/README
@@ -67,12 +67,10 @@ The following packages are optional
 
    ttf-freefont     More font support in the reports
 
-   gir-webkit       Required for the (user-downloadable) HtmlView plugin
-
    goocanvas2       Required for the (user-downloadable) GraphView plugin
 
-No longer needed in 4.1:
-   ?
+No longer needed in 4.2:
+   gir-webkit
 No longer needed in 4.0:
    pygoocanvas, pygtk, pyexiv2
 No longer needed in 3.3:


sam-m888:
Good catch,

I've also update the relevant wiki pages:
* https://gramps-project.org/wiki/index.php?title=Addon:HtmlView
* https://gramps-project.org/wiki/index.php?title=Gramps_4.1_Wiki_Manual_-_Categories#Web_Category

Thanks for noticing

--
The only thing to note is that the htmlview addon file is currently not available in 
https://github.com/gramps-project/addons/tree/master/gramps42/download
for download but is for Gramps master.

@dsblank  When you have a chance could you fix this please?

dsblank:
@sam-m888 I removed HtmlView from addons/gramps50/ in both listings:

https://github.com/gramps-project/addons/commit/6463819b0551aad67141f2ba233964b1b1595ea8

and downloads:

https://github.com/gramps-project/addons/commit/637b5c3c245850c3813f936f288eadba2fb5e8ac

Thanks!

sam-m888:
@dsblank Thank you :+1:

PR #53, opened 2015-08-07
8471: Set "visable_window" in GtkEventBox to fix transparency.

orner:
Was showing transparent background in a Compiz environment without
this option "visable_window" being explicitly set.  It is already set to False in other dialogs.


sam-m888:
thanks and welcome :+1:

PR #54, opened 2015-08-10
Expand exmaple.gramps for testing DynamicWeb report

belissent:
Modification for example.gramps database
(custom parent relationship, notes)

### DIFF
--- a/example/gramps/example.gramps
+++ b/example/gramps/example.gramps
@@ -7,7 +7,7 @@
     <researcher>
       <resname>Alex Roitman,,,</resname>
     </researcher>
-    <mediapath>/home/cristina/programs/trunk/example/gramps</mediapath>
+    <mediapath>/home/pierre/Gramps/gramps.40/example/gramps</mediapath>
   </header>
   <name-formats>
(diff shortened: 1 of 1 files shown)


dsblank:
I don't think we want to add test stuff to gramps40, which is old and in maintenance mode.

Perhaps you should add such a import file to the addon? addons-source will now look for and run tests after commits:

https://travis-ci.org/gramps-project/addons-source

except that DynamicWeb is being excluded...because it doesn't pass?

belissent:
I have never used Travis or nosetest before.
The test script for the DynamicWeb report is a standalone script and only generates the report with several configurations. I use it to generate the examples:
http://belissent.github.io/GrampsDynamicWebReport/

I suppose that it is excluded because it conflicts with nosetests.
I will look at how to make the CI tests work for the DynamicWeb report on the master branch.

dsblank:
@belissent Ok, both of those ideas (updating example.gramps on master, and making tests work with Travis) sound like great ideas. Thanks!

(DynamicWeb is looking great!)

belissent:
Cancelling pull request.

PR #55, opened 2015-08-19
Remove useless 'git clone' in .travis.yml

belissent:


dsblank:
That should be a lot faster! Thanks!

PR #57, opened 2015-08-21
8471: Set "visable_window" in GtkEventBox to fix transparency.

orner:
Was showing transparent background in a Compiz environment without
this option "visable_window" being explicitly set.  Patch for Gramps 4.2


sam-m888:
Thank you. :+1:

PR #58, opened 2015-09-05
5357 Add window position config keys for Clipboard & main window.

sam-m888:
Please test on:
- [ ] Apple Mac @jralls
- [ ] MS-Windows @bpisoj 

Tested on
- [x] Linux (Ubuntu 14.04.3)

Any feedback welcomed.

https://gramps-project.org/bugs/view.php?id=5357

### DIFF
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -23,7 +23,7 @@
 #
 
 """
-This package implements access to GRAMPS configuration.
+This package implements access to Gramps configuration.
 """
 
 #---------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


Nick-Hall:
Applications normally leave the positioning of windows to the window manager.  We should investigate setting some window manager hints instead.

gramps/gen/config.py
```diff
@@ -23,7 +23,7 @@
 #
 
 """
-This package implements access to GRAMPS configuration.
+This package implements access to Gramps configuration.
 """
```
jralls: Put editorial changes like this in a separate commit.

jralls:
Rather than hard-coding values like this we should be recording the window sizes and positions at the end of the session and restoring them when the program is next launched.

gramps/gen/config.py
sam-m888: Ok removed the editorial changes and pushed back to github and it closed !
https://github.com/sam-m888/gramps/commit/28f3d90f

sam-m888:
@jralls  Ok I'll give that a try.

sam-m888:
@Nick-Hall For hints do you mean like how some of the dialogs automatically center via

* Gtk.WindowPosition.CENTER_ON_PARENT
 and maybe for the main window:
* Gtk.WindowPosition.CENTER

PR #63, opened 2015-09-09
Feature - Split Addons check to its own tab in preferences

sam-m888:
The General tab is a bit to General. 

Make it easier to point users to the Addons tab in preferences and provide a bit information straight away.

For Gramp 5.0 only.

addons-tab-preferences-50

https://gramps-project.org/wiki/index.php?title=Third-party_Plugins


Nick-Hall:
I think that we should go further than this.  The addons deserve their own top-level menu.  We should provide a dialog that lists all available addons and allows the user to install them.  This dialog could contain the addon preference settings.

sam-m888:
That makes sense :+1: 

I note that in the following GEPS[1] that some of the screenshots(from 2010) show a "Install Addons" tab for the "Plugin Manager" but can not find an image of that tab itself, do you know if the it was the idea for them to be merged? Another option would be have the "Addon" menu option appear under "Tools" similar to how "Firefox"[2] has it listed and "Plugin Manager" be list under "Tools" also, or possible as the second two tabs eg merged together?

[1] https://gramps-project.org/wiki/index.php?title=GEPS_014:_Plugin_registration_and_management#Screenshots
[2] Firefox > Menu > Tools > Add-ons (CTRL+Shift+A)

daleathan:
:+1:

uxcq:
This really makes sense!

sam-m888:
- [ ]  Add top-level menu for Add-ons
- [ ]  Provide a dialog that lists all available addons and allows the user to install them. 
- [ ]  This dialog could contain the addon preference settings.

uxcq:
:(

uxcq:
Any chance this can be revived and added to Gramps 5.0 it would really make it easier to discover Addons :+1:  :8ball:

sam-m888:
@uxcq @daleathan Sorry, limited time and not very good with GUI stuff. I've reopened this PR and marked it WIP for now.

sam-m888:
Found a partial experimental patch I'd forgotten about!

In this I'd split out the addon check to its own menu entry and removed from preferences.

new_addons_menu_entry_and_updated_general_preferences

Nick-Hall:
Yes.  This is the way to go.

I would change the title of the dialog to "Addon Manager" and the tab name to "Updates".  Then I would create another tab called "Listing" to display all third-party addons.  Installed addons could be marked with a tick.

Perhaps this would be better in the _Edit_ menu next to _Preferences_?  I'm not sure.

uxcq:
> Perhaps this would be better in the Edit menu next to Preferences? I'm not sure.

Yes, under "Preferences" instead of cluttering the menu.

The help menu also has a "Plugin Manager" that shows the Addon's (not sure what the difference is, ones builtin I guess), couldn't they be merged into one dialog?  That would mean the help menu only has help links.

BTW Does Gramps have a check for new Gramps version anywhere?

Nick-Hall:
@uxcq The plugin manager is more of interest to power users.  It shows the status of all plugins, not just third-party addons.  We could combine it with the new dialog.

Each version has its own directory for pluguns.

sam-m888:
"Restart addon message change wording" https://github.com/gramps-project/gramps/pull/304

uxcq:
I believe you also need to add the filename for `addonmanager.py` to [POTFILES.in](https://github.com/gramps-project/gramps/blob/master/po/POTFILES.in) so it can be included in the list of source files which contain translatable strings?

prculley:
@sam-m888 I see you have assigned to me; I still cannot push to your repo so cannot update this PR directly.  
Should we just start a new one?  
Or would you like to give me some rights to your repo?

If you want to give me rights, I think you need to look at your repo settings tab, and then collaborators section. (I experimented with this so you might get an email to join my repo as a collaborator, but I cancelled it immediately, so it might not work...).

sam-m888:
@prculley Ok, invite sent.  If that does not work, start a new PR and reference this one.

uxcq:
Quite a few extra commits here...

prculley:
Sorry about the poor rebase; I thought I had the right combination, but seems not. 
Here is my proposed change, based on comments above and a few ideas of my own.

addon
addon_1

Nick-Hall:
Thia is moving in the right direction, but it really needs a tab listing all third-party addons with a tick next to the ones that are installed.

prculley:
@Nick-Hall I was already contemplating combining the two plugins tabs into a single one, with an enhanced 'status' column.  I think we already have more tabs than a user knows what to do with...
If we add in all the possible plugins found in the addons/gramps50/listings/addons-??.txt files for the languages installed, we could further enhance the 'status' and keep everything on one tab.
For enhanced status I have in mind 'New', 'Update Available', 'Installed', 'Builtin', 'Hidden', and 'Unavailable' (or 'Incompatible' or whatever is the best description for a plugin that won't load).  We might want to use a couple of these simultaneously ('Hidden' and some others).

For implementation, I think I would keep around a copy of the most recently read addons/gramps50/listings/addons-??.txt file in the user directory, to be merged with the current registered list when the addons dialog is opened..

If you or anyone has further thoughts I'd like to hear them now before I start coding ;-).

Paul-Franklin:
I haven't been following this, so perhaps my comment makes
no sense in context, but I hope the result is less than 600 pixels
high (since I assume the width will be inside a scrolled window).

Nick-Hall:
@prculley Yes.  Combining the two tabs would work well.

However, I don't think that most users are interested in the built-in plugins.  They just want to see what third-party addons are available so that they can easily install/uninstall them.  You could add a "Show built-in plugins" checkbox (unchecked by default).

Keep the status simple.  Installed (tick), Uninstalled (blank), Error (exclamation mark) should be sufficient.  From a user point of view, hidden/visible for built-in plugins is equivalent to uninstalled/installed for third-party addons.

Updates can be handled as a background task, or by clicking the "Check now" button.

prculley:
The latest edition of the Addon Manager.
I tried to make it a bit more user friendly while still supporting developers with useful information and features...

* Unified view combines Registered and Loaded views with all the available 3rd party plugins in a single view.
* Status column indicates status of Addon, Available, Installed, Built-in, Update Available, and ~~hidden~~ (via strike-through).  A Failed addon also adds 'Failed' status in red.
* All of Available, Installed, and Built-in plugins can be hidden.
* A "Remove hidden items from view" checkbox allows the user to get uninteresting (hidden) addons off the Addons pane.
* Available and Update Available addons can be installed from Addons pane, the "Install" button appears when this is possible.
* Installed addons can be un-installed from Addons pane; the "Uninstall" button appears when this is possible.
* Info button provides information on all addon types; includes 'Loaded' and failure type and stack traceback, when applicable.
* In 'debug' mode, the 'Reload', 'Load', and 'Edit' buttons appear; load and edit are greyed out when addon is not installed.
* after updates of Views or Gramplets, or uninstalls, the user is reminded to restart Gramps.

I've tried to test all the various combinations, I think it works well enough to warrant testing by others..  I'd appreciate feedback.
If this is accepted, I also plan to write some unittest code; at least some of the Gtk Listmodel code can be tested, as well as some of the install, uninstall etc. code...

@Nick-Hall I'm sorry but I just could not bring myself to simplify as much as you wanted; it seemed to me that going that far would leave useful functionality and or information on the table...  I hope this compromise is acceptable.
addonmanager_1
addonmanager_2

Paul-Franklin:
From your pictures I don't see that it will be easy for a
user to see something is a "third-party" addon, when
they file a bug report for it.  I was guessing there'd be
a column or something.

prculley:
We developers all realize that addons 'Available' or 'Installed' have to be 3rd party addons, otherwise they would be 'builtin' :wink:.  But perhaps users will not realize this.  Can you suggest a different terminology?  I'd prefer not to add another column; trying to keep the dialog a reasonable size for those users that have small screens.  Maybe a tooltip explaining the status?

Paul-Franklin:
Yes, I don't think you need another column either.

Your blue highlighted column suggests the answer.

The same as the highlighted entry has a "three-line"
status, make all the "Available" ones into a two-line
status, "Third-party, Available" with a line feed at the
comma.  That would satisfy me.

Nick-Hall:
Users will find the status column confusing.  I suggest the following changes:

1. Provide a checkbox to display built-in plugins.  Don't display them by default.
2. Remove the "Hide" button.  When a built-in plugin is uninstalled, hide it.  When it is installed, make it visible.
3. Use icons in the status column.  Installed/Visible=dialog-apply.  Error=dialog-warning.  Update Available=software-update-available.

Paul-Franklin:
If you're only going to show third-party addons by default,
then I want the column heading to say "Third-party Addons"
and maybe also text at the top of the dialog.  Both of those
would change when that "show built-in" box was checked,
but I still want "third-party" said somehow, somewhere, not
just something users won't be likely to notice (a different
icon, say).

They aren't familiar with these things, they just want
to do their genealogy.  Years of experience on the bug
tracker has shown that lots don't even understand
the concept of "third-party" addons (since all the menus
go out of their way to fully integrate them for instance).

When I point that out they usually ask "how do I know
what is a third-party addon?"  As long as this is being
heavily revised I want to solve that problem too.

Somehow.

Thanks.

Nick-Hall:
@Paul-Franklin I agree.  The average user is only really interested in third-party addons.

Perhaps we could have separate tabs for "Third-party Addons" and "Built-in Plugins"?

Paul-Franklin:
Separate tabs for both is fine with me.

Whatever Paul-C wants.

sam-m888:
Looks good 👍 

Testing shows only the following for me in a clean gramps profile with no family tree load etc selecting the  "Check for updated addons now" button shows the following in the terminal:


> 2017-03-03 15:46:11.386: WARNING: addonmanager.py: line 614: Failed to open addon status file: [Errno 2] No such file or directory: '/home/grampsdev/.gramps/gramps50/new_addons.txt'

prculley:
@sam-m888 Looks like I forgot that new addons (and the file) are not there until the user does 'Check for Updates' the first time.  I'll add a little reminder...

@Nick-Hall @Paul-Franklin As a counter point; when I put on my user hat, I often want to get rid of addons; they don't have any interest to me.  So I hide them.  I don't care where they come from when I'm doing this I just want to see them in one place to get rid of them.  Also when I'm curious if there is a capability that I am not yet familiar with, I like to peruse the descriptions.  Again, don't want to look in several different places.
A lot of our Gramplets and Quickreports are not obvious unless you know they are there, even when built-in.
I'm not personally a big fan of icons, for me they are not nearly as descriptive as words.  That said, I don't know what the icons you are suggesting look like, so I need to figure this out first.

If you feel strongly about builtins, a checkbox to stop showing builtin addons is fairly reasonable and allows for both my type and your type of 'user'.

If we keep the textual status, adding '3rd party' in front of 'Installed' or 'Available' will only add a bit of vertical height to the rows, not a big deal.

ennoborg:
@prculley I agree. I like to remove or switch off add-ons, or plug-ins, whatever they're called, and also have doubts about the use of the '3rd party' term. I think that most users don't understand that, because all add-ons are offered from what looks like a Gramps repository. I mean, there's no other site to visit, or anything like that.

I also like go through a single term for everything. We have plug-ins that are installed with Gramps, and plug-ins that can be downloaded as extras, which apparently are add-ons. It would look easier to me if they are all shown as plug-ins, that can all be switched-off if a user doesn't want to see them, or they show errors. Downloaded plug-ins can also be removed from disk, and that should be the only difference, I think.

prculley:
I added a 'remove built-in items from view' checkbox.
I changed 'Available' status to '3rd Party' and 'Installed to '3rd Party Installed'.
I added a popup to suggest that the user do an update in order to see 3rd party addons.  Only pops up until user does it first time.

I still don't think that icons are as informative as text; I particularly wouldn't know what to make of the suggested software-update-available  software-update-available icon.

addonmanager_2

Paul-Franklin:
@prculley
I like that picture.

I don't like that icon.  It doesn't suggest anything to me.
In fact it wasn't until I studied it closely that I even noticed
the little downward-pointing arrow in it.  I vote for words.
I like the various word choices in your example picture.

@ennoborg
The "third-party" designation I am fighting for is /exactly/
because people like you don't know what they are, where
they are, why they are different, and so on.  With -- as I have
stated and you have stated -- the reason being that their
availability is so tightly integrated into gramps.

I don't know if that was a policy decision or if that is
just the way the developer who set up the "plugin"
system infrastructure wanted it.  But it doesn't matter.
That is the way it is.

The problem is not that their descriptions/availability
are tightly integrated.  The problem is that /by definition/
they are not a part of "core" gramps.  Therefore, /by
definition/ "we" are not responsible for them, for their
maintenance, for their enhancement, anything.  So
I don't think they should be shown so tightly.

But of course users do not understand that, since
their names and so forth appear with no distinctive
difference in any availability list/pulldown.  (Their
names could be shown in red for instance.  Hah.)

So I am trying to force the /knowledge/ of that
difference to be wider spread.  Specifically to try
to get users to contact their authors -- as the
"policy" says they should.  Rather than just filing
a bug on the bug tracker, which of course is what
(I presume) they always do instead.

But this github comment area is the wrong place
to have a discussion about such "policy" matters.
I tried raising the issue on the list, suggesting that
many (most? all?) "third-party" addons could -- and
therefore /should/ -- be moved into "core" gramps.

But I was vociferously argued against, for days, by
one developer, and IIRC another developer (who
writes third-party addons) agreed with him.  So I
dropped the matter and don't plan to raise the
matter again.

I am just fighting to have the /distinction/ known
by a wider audience of our users.

Nick-Hall:
@ennoborg I like the idea of switching a plugin on or off.
- On = Installed / Visible
- Off = Uninstalled / Hidden

We should consider using the Gtk Switch widget.

@prculley The actual appearance of icons will depend on the theme.  I would always add a tooltip to provide extra information.  

I prefer my original suggestion of a checkbox to display core plugins rather than a separate tab, however, I don't have a strong opinion on the use text or icons.

@Paul-Franklin Yes. We need to make it clear that third-party addons are not officially part of Gramps.  They may be incomplete or of lower quality than core plugins.  Some may be aimed at developers rather than normal users and may require additional prerequisites to be installed.  As you point out, bugs should be reported directly to the plugin developer.  Bug reports on the bug tracker may not be seen by the developer for months.

sam-m888:
Additional dialog that pops up if addons listing does not exist.
2017-04-02 10_45_35-ubuntu-16 04 1-grampsdev-amd64 running - oracle vm virtualbox

prculley:
I made a few more changes based on a review by @sam-m888 (Thanks Sam), and based on his input rebased and squashed all the commits to a single one.

I think this is ready for merge into master and wider review.
I will probably try to find the time to write some unittest code for this a bit later (which may require some additional changes), and of course will fix any bugs other find.

Nick-Hall:
This PR has made good progress, but it is not ready to merge yet.  I have a few comments:

+ Error when clicking on "Info" and an uninstalled addon selected.
+ Need an ellipsis after "Addons Manager" in the Edit menu.
+ Make "Addons" the default tab.
+ Don't hide buttons.  Make them inactive instead.
+ Don't put checkboxes in the dialog button bar.
+ The meaning of strikeout is unclear.
+ Don't specify colours for text (such as "Failed").  They won't be clearly visible on all themes.
+ It is possible to hide an uninstalled addon.  Why do we need this?
+ Please run `pylint` on the new file.

prculley:
Sorry it took so long to respond, chores...
I've updated to address some of @Nick-Hall concerns.

- Error when clicking on "Info" and an uninstalled addon selected. Done
- Need an ellipsis after "Addons Manager" in the Edit menu. Done
- Make "Addons" the default tab.
If we had lawyers, I think they would say we needed to show the Updates tab first, because it had the 3rd party Addons warning.  I changed the code to only force to that tab the first time the Addon manager is opened per session, after that it opens on the previous tab.
- Don't hide buttons. Make them inactive instead. I did this for the 'normal' buttons.  The 'debug mode' buttons are still hidden when not in debug mode. 
- Don't put checkboxes in the dialog button bar.
I'm not sure where to put them that doesn't add to the size of the dialog; they are too large to put with the panel buttons without adding another line.  I'm not quite sure why you object, and am willing to entertain suggestions.
- The meaning of strikeout is unclear.
It is to indicate a 'hidden' plugin, it only shows in when showing the hidden plugins.  I thought that having the button text change from 'Hide' to 'Unhide' when cursor is on a (shown) hidden plugin was not quite sufficient, and it seemed adding 'hidden' to status column was a bit much.  Again, I'm open to suggestions.
- Don't specify colours for text (such as "Failed"). They won't be clearly visible on all themes.
The Gramps Gui guidelines says 
>  the only widget that seems safe to use color is the gtk.TreeView

and the previous plugin manager used color quite a bit (red, blue, green are used), so I thought it would be OK.  I would like to emphasize the 'failed' status somehow, am open to suggestions.

- It is possible to hide an uninstalled addon. Why do we need this?
I personally like to de-clutter the view by hiding addons (builtin or install-able) that I never think I will use.  So I think it is a good idea.
- Please run pylint on the new file.
Done.  I'm afraid I added some 'pylint disable' for some issues where pylint did not seem to be operating correctly or where I could not figure out a reasonable way around the issue.  But the score is now up over 9.4.

Nick-Hall:
@prculley Hiding buttons that are only used in debug mode is perfectly acceptable.  Tag colours are selected by the user, so they can be chosen to be readable.  The checkboxes just don't belong in the button bar - they are only used in one tab.

I suggest you try putting the third-party addons and built-in plugins into separate tabs.  This will make the distinction between them clearer, simplify the status column and remove the need for one of the checkboxes.

prculley:
@Nick-Hall I am very reluctant to splitting the view up into separate tabs; one of the main reasons I undertook this to begin with was to consolidate information from two different places.  When I am looking for a capability or feature, I don't care about the difference between built-in and 3rd party, I just want to see if it is there somewhere.  So looking in a single list was may main goal.  I was willing to adjust the view to hide builtins because I recognize that some people might think differently.

Regarding the bottom bar, you do realize that the checkboxes are only shown when the addons tab is visible?  So I'm not sure why you are so adamant that they don't belong there.
If this is a show-stopper for you, I'll add another control bar at the top of the Addons tab with these two checkboxes in it.  Even though I think this wastes space...

Nick-Hall:
@prculley Although core plugins and thrid-party addons share the same underlying plugin mechanism, we should keep a clear distinction between them.  Core plugins are available to all users and the average user will have no need to even know that they are plugins.  Third-party addons are provide extra functionality that may be useful to everyone.  These are what most users are interested in.

When widgets are visible only when a certain tab is active, it should give you a clue that they belong in the tab.

prculley:
@Nick-Hall I've moved the check boxes to the addons tab, rather than the bottom button bar.

I think that the status column provides enough of an indication that a plugin is 'builtin' or '3rd party' and I firmly prefer everything in one list, for easy of finding something.

I hope this satisfies, as I would like to get this integrated into Gramps...

Nick-Hall:
@prculley This PR needs more work before it can be merged.  I would like to see the core plugins and third-party addons in separate tabs.  The status column is still rather messy.

Paul-Franklin:
Since ISTM we have a pretty solid disagreement here, with two
people firmly preferring a different form of the UI, is there any way
we can implement it both ways?

What about something like a gramps envariable, which if set makes
it be one way and if unset makes it be the other?  That way you can
argue about what the default should be -- or flip a coin -- but it can
be the other way for the other person since he will run with the new
environment variable set, always?

Maybe something like GRAMPS_CONSOLIDATE_ADDONS or
GRAMPS_UNIQUE_ADDONS ?

I'd like to see this implemented, but if the main author wants it
done one way and the chief reviewer wants it done the opposite
way, that seems like a state which we should work on breaking,
rather than just insisting "it has to be MY way".

Any thoughts?

prculley:
I'd also love to break the stalemate.  I tried to solicit some more opinions on the mail list, as I will accept a majority on how to proceed.  For now, I will wait until some more comments appear.

I REALLY don't want to start using esoteric environment variables for UI options, not (in my opinion) user friendly at all.

sam-m888:
The real question is _what you've created better than what exist_, in my opinion it's a resounding yes for improving Gramps usability. 

My preferences is to merge and if in the Gramps 5.0.0 alpha2 release the project receives significant feedback that it is disliked then consider changing it or reverting.

Thanks for your improvements 🥇

eno93:
@sam-m888 When I go see the addons, that's what I'm expecting to see. So I suggest that by default those two checkboxes "Remove hidden items from view" and "Remove Builtin items from view"  should be checked by default or the meaning flipped altogether to say "Show hidden items" and "Show builtin items".

uxcq:
Discussion about this on the mailing list

* [OPINIONS DESIRED: Updated Addon manager PR](http://gramps.1791082.n4.nabble.com/OPINIONS-DESIRED-Updated-Addon-manager-PR-td4679778.html)

Nick-Hall:
It doesn't look like we are going to get any more comments on the list.  After reading what everyone has said, I think we need the following changes:

+ Rename the dialog to "Plugin Manager".
+ Rename the "Addons" tab to "Plugins"

+ Create a group of radio buttons ("Third-party" and "Core") for populating the list.
+ The default should be "Third-party" which should display all new available addons.

+ Remove "Hidden" status. Third-party plugins can be "Available", "Installed" or "Failed". Core plugins can be "Enabled" or "Disabled".
+ Remove strikethrough and colours from the status column.
+ I'm not sure that we need "Update available" here.  This can be handled by the updated addons dialog.

+ The "Hide" button should be changed to "Enable/Disable" and will only be active for core plugins.

+ The updated addons dialog should now only check for updates. The "What to check" and "previously notified" preferences are no longer be required.
+ Rename the "Webpage Help in Addons" button to "Help" and move it into the button bar.

prculley:
@Nick-Hall Well, I'm disappointed that you won't accept my 'features' to remove the clutter of uninteresting 3rd party addons, and that you don't think a user might want to know about an 'available update' (unless he goes to another screen and runs updates again).

But in the spirit of making some forward progress, I will fix up the PR according to your wishes.  I'm working on a couple of other things at the moment, so it might take a couple of days.

P.S. I think it is funny that we are now changing back to 'plugins' after Sam tried to get all the language switched to 'Addons'...

Nick-Hall:
@prculley Thank you.  I think most users will just want to see a list of third-party addons.  They will install those they find interesting or uninstall those they don't.

It is easiest to use the automatic checks for updates rather than the manager.  For this reason please ignore my request to remove the two update preferences.  It will still be useful to be automatically notified of a new addon.

Please note that all addons are plugins, but not all plugins are addons.  Sam and I originaly intended that the manager only handle third-party addons.

prculley:
@Nick-Hall I started to look at this again, and realized that you had also removed the 'both' option to allow a user to see all addons/plugins at one time.  Since this was my main reason for doing this in the first place, I have decided I won't work on this anymore.  After all, I am a volunteer, and am no longer required to do work I don't like or believe in.  Sorry that I raised your expectations earlier.

Nick-Hall:
@prculley OK.  We all all volunteers.  Don't feel pressured into working on anything you don't want to.

sam-m888:
Available as a third party addon for Gramps 5.0

https://github.com/gramps-project/addons-source/tree/maintenance/gramps50/PluginManager

https://gramps-project.org/wiki/index.php?title=Plugin_Manager_Plugin

@prculley Thank you

emyoulation:
a top-level menu location for Addon Manager is reasonable.  But it should probably be in the Help menu with the "Plugin Manager" because they cover similar Funtionalities: plugins. It would reinforce that they are not accidentally mislabeled duplicates of the same tool.

PR #64, opened 2015-09-10
Add Occupation in rel_graph

fobrice:
This patch is for the rel_graph plugin 
It add the possibility to include the last occupation (when available) of each person in the graph.
--
Fabrice

### DIFF
--- a/gramps/plugins/graph/gvrelgraph.py
+++ b/gramps/plugins/graph/gvrelgraph.py
@@ -115,6 +115,7 @@ class RelGraphReport(Report):
         name_format    - Preferred format to display names
         incl_private   - Whether to include private data
         event_choice   - Whether to include dates and/or places
+        occupation     - Whether to include occupation
         """
         Report.__init__(self, database, options, user)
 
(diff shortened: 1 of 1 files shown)


Paul-Franklin:
I don't have a machine I can run this on, at the moment,
but just from looking at the code I wonder why you have
code which includes the gramps-id, when you are testing
for self.occupation?

fobrice:
Sorry Paul, it was a mistake: I had bad interpretation of the code.
I have commit a new code.
--
Fabrice

Paul-Franklin:
I plan to commit this to gramps, although I have positioned
your occupation line slightly lower, in both the result and the
options dialog, as well as adding parentheses around the
occupation, as that is how that report seems to do things.

You didn't include a copyright line, with your name (and
email address as you are not a regular developer), so
I would appreciate it if you let me know.  You can email
it to me if you don't want to post it here (I guess Google
knows how to find me, but I don't like to post it).

As you may or may not know, there are at least two
"feature requests" for this enhancement.  Also, of
course I will commit it only to "master" (as opposed
to gramps42) since all enhancements go there.

Thank you for helping make gramps better.

fobrice:
Thank you Paul! You can add a line "Contributions by Fabrice <fobrice at laposte dot net>"
--
Fabrice

Paul-Franklin:
It is very confusing adding comments here on github.

Their system is adding them in places I never expected.

However, I want to credit you (and this pull request) for
https://github.com/gramps-project/gramps/commit/5a5e7a21c4e8422bfed0e45bcb78dbceaa782262
which will be in gramps 5.0.0 -- thanks.

sam-m888:
Closing as already merged.

PR #65, opened 2015-09-14
8888 'Family Tree(Manager)' add Help button (glade) and help link

sam-m888:
My attempt at modifying a glade dialog to add a Help button.

While it works and goes to the help page, I'm not quite sure why the dialog is also dismissed?

https://gramps-project.org/bugs/view.php?id=8888


sam-m888:
Committed to Gramps master only as:

https://github.com/gramps-project/gramps/commit/a6f4c7993d81433ec71b385040074f1aa5edadfa

PR #66, opened 2015-10-10
Sort HTML and CSS output (enable diff of archived exports)

belissent:


belissent:
I would not like my PR to induce any unwanted consequences...
I just thought it was a good idea because:

- The dict keys are already sorted in "stylesheet.py":
For example, see in StyleSheetList.save method:
```
line 153: for name in sorted(self.map.keys()): # enable diff of archived copies
```
It would be nice to have a consistent behavior between calling StyleSheetList.save and StyleSheetList.get_style_names

- Modifying where the methods are called would lead to more modifications

- As the dict order is not specified, ordering them could not harm

Paul-Franklin:
> Modifying where the methods are called would lead to more modifications
I am proposing that you only modify the HTML backend, so the number of
modifications would be the same, I assume, just in different places.

> As the dict order is not specified, ordering them could not harm
Ordering them always will slow down every place which uses them.
Whereas ordering them where you desire the diff has a clear cost
vs. benefit tradeoff.  As in StyleSheetList.save.

sam-m888:
Related issue:
Narrated Web Site Report: html elements emitted in different order
https://gramps-project.org/bugs/view.php?id=8950

belissent:
Thanks for the response
I agree: the issue 8950 covers this PR.
This PR is closed.

PR #67, opened 2015-10-18
Handle citation objects in glocale.trans_objclass

tomhughes:
Without this right clicking on a citation in the clipboard
gives options like "Make See Details active" which don't make
an awful lot of sense.


tomhughes:
By the way, why is the default return value `trans_objclass` set as "See details" only it doesn't seem to make much sense?

Nick-Hall:
Thanks.  Also fixed in the gramps42 branch 0baf7dc3adaa636e2e6c2d5acad912ae408713d7.

Yes.  A default of "the object" would appear to be better than "See details".

PR #68, opened 2015-10-18
Fix creation of focus change events

tomhughes:
This fixes a number of run time warnings of the form:

```
interactivesearchbox.py:195: TypeError: Passing arguments to gi.types.Boxed.__init__() is deprecated.
All arguments passed will be ignored.
```

Which appear to represent a genuine issue as the previous way of creating events indeed doesn't set the type.

### DIFF
--- a/gramps/gui/widgets/interactivesearchbox.py
+++ b/gramps/gui/widgets/interactivesearchbox.py
@@ -192,7 +192,8 @@ class InteractiveSearchBox():
         self._search_entry.grab_focus()
         self._search_entry.set_position(-1)
         # send focus-in event
-        event = Gdk.Event(Gdk.EventType.FOCUS_CHANGE)
+        event = Gdk.Event()
+        event.type = Gdk.EventType.FOCUS_CHANGE
         event.focus_change.in_ = True
         event.focus_change.window = self._search_window.get_window()
         self._search_entry.emit('focus-in-event', event)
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Alos fixed in gramps42 branch 4046c8af0f5a90709672617010eb47b9a0a7f4ff.

Paul-Franklin:
Thanks Tom!

Thanks to your model I fixed another deprecation warning
which has bothered me ever since I started running gramps42.

https://github.com/gramps-project/gramps/commit/a98daed758d7c2e4114342c5c55f0ce0d36a5729
https://github.com/gramps-project/gramps/commit/9e07b3e092b62ecfb1ccf992dd7330a67f18cbb8

Feel free to fix as many deprecation warnings as you want!
8-) 8-)

tomhughes:
@Paul-Franklin Oh I saw that but I assumed the whole point of the version check was to avoid the warning so I didn't bother changing it...

Paul-Franklin:
I don't know what that code does, or is supposed to do,
but for my system I always hit that top branch of the "if"
so I always got the warning.  But no more!  8-)

rhoit:
### Just a suggestion

I know this commit is old, but still it helped me remove the Warning message! and found one more thing.

`new()` can still be used, which still accepts the type parameters,

```
Gdk.Event.new(Gdk.EventType.FOCUS_CHANGE)
```

which helpful while sending single in one-liner, such as 'delete-event'

```
ANY_GObject_Object_WIDGET.emit("delete-event", Gdk.Event.new(Gdk.EventType.DELETE))
```

PR #69, opened 2015-10-20
Re-enable selection in the clipboard on a grab_broken event

tomhughes:
As of GTK 3.18.0 a fake button release event is no longer sent when a DnD completes:

  https://bugzilla.gnome.org/show_bug.cgi?id=749737

As a result we keep selection disabled which then means that the next drag from the clipboard fails to select the dragged object and we wind up trying to drop the old selection instead.


Nick-Hall:
Thanks.  In gramps/gui/widgets/multitreeview.py, there is similar code for the tables in quick views.  This also needs the same fix.

tomhughes:
Updated to handle both cases...

Nick-Hall:
Also fixed in gramps42 commit f30fa90.

PR #70, opened 2015-10-23
Update Russian translation

vantu5z:


Paul-Franklin:
Thank you very much for your contribution, Ivan!

We really appreciate you wanting to make gramps better.

I am not a github expert but one thing I don't like about them
is their unwillingness to give me a diff if the file is too large,
in their HTML file browser I mean.

So I downloaded your updated ru.po and that enabled me
to also run po/check_po on it.

It told me about two cases of a mismatch between an
argument in braces -- {number_of} for instance -- and
a translation which only had the %d arguments, so I
fixed both those before I committed your file.  I also
committed it to our current gramps42 branch (so it
should be in 4.2.2 when that is released).

https://github.com/gramps-project/gramps/commit/cd2db0b2671421693505fd04dacd12c05907ba0c
https://github.com/gramps-project/gramps/commit/c82b75fef824be2088841cadd49cee26cf63e41b

Thank you again.

PR #71, opened 2015-11-19
update ru.po

vantu5z:
Update Russian translation, it should fix test error problem after merging template (commit 3b16871)


vantu5z:
Include by commit 2bc4dd3

PR #72, opened 2015-11-26
9076: don't check SSL certs when fetching addons

alan-strohm:
This is a follow-on to https://github.com/gramps-project/gramps/commit/93e3c62b99832d7643eb82f2ad0ae58b0c8b1fe2

That change fixed loading the add-on list, but fetching the add-on needs the same treatment (at least for me).

Note that this is my first contribution (and first pull request, actually), so feel free to let me know if I'm not doing it right.

I tested by copying utils.py into my Gramps.app installation of 4.2.1 and verifying that I could still fetch the add-on list and install add-ons.

### DIFF
--- a/gramps/gen/plug/utils.py
+++ b/gramps/gen/plug/utils.py
@@ -173,8 +173,11 @@ class Zipfile(object):
         """
         return os.path.split(name)[1]
 
-def available_updates():
-    whattypes = config.get('behavior.check-for-update-types')
+def urlopen_maybe_no_check_cert(URL):
+    """
+    Similar to urllib.request.urlopen, but disables certificate
+    verification on Mac.
+    """
     context = None
     from urllib.request import urlopen
     if mac():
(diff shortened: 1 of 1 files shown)


jralls:
Good catch, and your patch is mostly good, too, especially noticing that the both tries with different language keys need to be checked with context. I'd obviously missed that when I first wrote the patch.

I'll make some inline comments about things that should be changed.

gramps/gen/plug/utils.py
```diff
@@ -173,8 +173,11 @@ def is_file(self, name):
         """
         return os.path.split(name)[1]
 
-def available_updates():
-    whattypes = config.get('behavior.check-for-update-types')
+def urlopen_no_cert(URL):
```
jralls: Since it's no-cert only on mac this isn't quite right as a function name. How about urlopen_maybe_no_check_cert() or urlopen_no_check_cert_on_mac()?

gramps/gen/plug/utils.py
```diff
@@ -182,6 +185,19 @@ def available_updates():
...
         context.verify_mode = CERT_NONE
+    timeout = 10 # seconds
+    fp = None
+    try:
+        fp = urlopen(URL, timeout=timeout, context=context)
+    except TypeError:
+        fp = urlopen(URL, timeout=timeout)
+    except Exception as err: # some error
+        LOG.warning("Failed to open URL {url}: {err}".
+            format(url=URL, err=err))
```
jralls: Don't catch this exception here. Put the urlopen_maybe_no_check_cert() in a try block. That prevents logging a warning just because it's the first locale try, failure of which isn't an error, saves the test of fp there, and allows us to keep the different warning messages for the addons list and the actual addon.

alan-strohm:
Thanks, John. I addressed your comments, please take another look.

jralls:
Merged and cherry-picked to master. Thanks!

PR #73, opened 2015-12-03
7763: Sourceless Javascript Files in release tarball

RossGammon:
This one removes the last hardcoded path to the local copy of jquery
and uses the online CDN provided by Google instead.

Nothing seems to be using the Amimated-DropDown stylesheet at the moment, but this should allow it to be used in the future.


dsblank:
Thanks for all of this work, and follow-up!

PR #75, opened 2015-12-05
Small fix in import script

cicl06:
The file dialogue created by clicking Family Trees -> import would not close properly when the x was pressed; rather, the user had to click cancel.


sam-m888:
Hi Bob

Welcome to Gramps :)

Thank you for the fix.

--
As per @Nick-Hall 
http://sourceforge.net/p/gramps/mailman/message/34650894/

PR #77, opened 2015-12-13
8556: ValueError: underlying buffer has been detached

dsblank:
Remove encoding on stdout/stderr.

### DIFF
--- a/gramps/grampsapp.py
+++ b/gramps/grampsapp.py
@@ -409,18 +409,6 @@ def run():
     else:
         # CLI use of Gramps
 
-        # Ensure that output is encoded correctly to stdout and
-        # stderr. This is much less cumbersome and error-prone than
-        # encoding individual outputs:
-        try:
-            _encoding = sys.stdout.encoding or sys.getdefaultencoding()
-        except:
-            _encoding = "UTF-8"
-        import codecs
-        sys.stdout = codecs.getwriter(_encoding)(sys.stdout.detach(),
-                                                 'backslashreplace')
-        sys.stderr = codecs.getwriter(_encoding)(sys.stderr.detach(),
-                                                 'backslashreplace')
         argpars.print_help()
         argpars.print_usage()
         from .cli.grampscli import startcli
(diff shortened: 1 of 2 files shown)


dsblank:
Going to go ahead and merge, as many things are broken without being able to have log messages, or show errors.

Paul-Franklin:
Since you have merged these into master you may want to
consider (in master) whether you want to once again enable
the vcard unittests, perhaps after changing debug to False
again for do_case in importvcard_test.py, since that one
now works for me (in master), although I will say that just
that one test module by itself seems to take about fifteen
minutes for me.  (That was what got me started looking
into the problem, when I noticed that the vcard tests did
an XML write to stdout, which seemed to be failing.)

dsblank:
If I remove `--exclude=vcard` from the nosetests3 line, I get the following results:

```
Ran 21374 tests in 23.620s
FAILED (SKIP=2, errors=50, failures=23)
```

Is that what you get?

Paul-Franklin:
Debugging this has turned into a time sink.

For one thing I get more errors on master than I do if I
apply the patch to gramps42 and test there.

For another I don't know how to run a Travis CI test
on github, except by making a commit, and as I have
mentioned on occasion I hardly ever run with Javascript
enabled, which apparently is needed to look at results.

So I run "python3 setup.py test" instead, if I want to run
all the tests.

But to make a more-specific statement, I suspect that
your "Remove encoding on stdout/stderr" work isn't
done yet.  At least that's my guess.

If you have some (short) unzipped gramps XML file, try:

cat foo.gramps | python3 Gramps.py -i - -f gramps

which doesn't seem to work for me.

If you don't have such a file, do a roundtrip:

gramps -e - -f gramps | gramps -i - -f gramps

and that failed for me too.

So I have a hunch that getting the vcard stuff to work
may have to come after getting the basic stdin stuff OK.

Paul-Franklin:
What happens if in importxml.py you change:

self.filehandle = sys.stdin

to

self.filehandle = sys.stdin.buffer


My tests then showed uncompressed gramps XML stdin worked.
But I didn't try any unittests, etc.

PR #80, opened 2015-12-19
add database plugins to package_core

tecknicaltom:
Make sure the database addon and the bsddb_support code is installed.
Without this, Gramps crashes on startup.


dsblank:
Good catch!

PR #81, opened 2015-12-22
Append development version number with 'alphaDATA'

dsblank:


jralls:
This will append today’s date on the version number when you run Gramps. Run the same build tomorrow and you’ll get tomorrow’s date. I doubt that that's what you want.

You need to run that code from setup.py, not from gramps itself. The same is true of git_revision.py -- it should be run by setup.py to create a static version stamp.

But wait, there’s more: The date that you ran setup isn’t really useful either, because you might have built it from a stale repository or from a detached head. The date you really want is the result of `git log -1 --pretty=format:"%cI"`, though even that won't tell you if you built it from a dirty commit.

That will get you a lot closer, but it still won't help if you've built from a tarball instead of a repo. I can't offhand think of a good way to get the commit information into a file inside of a tarball, especially one that Github's automatically generated.

dsblank:
@jralls Good points. How about a YYYYMMDDCCC where YYYYMMDD is the last log date and CCC is a count of the number of commits from that day? Thus, 20151222002 would be the value for the current log. This still won't work for a non-git repo setup. But that is about as close as we can get without specifically keeping a value.

jralls:
Huh? There are never any commits after the last log date because the last log date is the date of the last commit. Maybe you mean the last log date of the origin branch and the number of local commits after that in the local branch. That's not too hard to compute until you do a build on a  feature branch or if you've fetched but not merged and there are commits in origin that aren't in the local branch or the feature-branch's branch point. I suppose you could do YYYYMMDD-NNN+CCC in that case, indicating that the local branch is N commits behind the origin one and has C commits that haven't been pushed. In the case of a feature branch you'd also need to figure out what origin branch to query.

It blows up your sorting, though: The same commit in your local/feature branch could have different revision dates and counts just from fetching a new commit from origin.

dsblank:
I updated this PR with my suggestion, but you are right that this doesn't identify branches (in which case ordering doesn't make any sense anyway).

jralls:
Ah, I see, the nth commit of a particular date. Code is clearer than English sometimes.

Seems like Nick isn't sympathetic.

PR #82, opened 2015-12-27
Fix for a None handle lookup in relview

tecknicaltom:
If focusing on the relationship view with an empty tree loaded, an uncaught exception was raised by attempting to call get_person_from_handle with a None handle:

```
  File "dev/gramps/installed/lib64/python3.4/site-packages/gramps/gui/viewmanager.py", line 1011, in view_changed
    self.__change_page(page_num)
  File "dev/gramps/installed/lib64/python3.4/site-packages/gramps/gui/viewmanager.py", line 1024, in __change_page
    self.active_page.set_active()
  File "dev/gramps/installed/lib64/python3.4/site-packages/gramps/gui/views/navigationview.py", line 154, in set_active
    PageView.set_active(self)
  File "dev/gramps/installed/lib64/python3.4/site-packages/gramps/gui/views/pageview.py", line 304, in set_active
    self.build_tree()
  File "dev/gramps/installed/lib64/python3.4/site-packages/gramps/plugins/view/relview.py", line 205, in build_tree
    self.redraw()
  File "dev/gramps/installed/lib64/python3.4/site-packages/gramps/plugins/view/relview.py", line 465, in redraw
    self.change_person(None)
  File "dev/gramps/installed/lib64/python3.4/site-packages/gramps/plugins/view/relview.py", line 470, in change_person
    return self._change_person(obj)
  File "dev/gramps/installed/lib64/python3.4/site-packages/gramps/plugins/view/relview.py", line 500, in _change_person
    person = self.dbstate.db.get_person_from_handle(obj)
  File "dev/gramps/installed/lib64/python3.4/site-packages/gramps/plugins/database/bsddb_support/read.py", line 723, in get_person_from_handle
    return self.get_from_handle(handle, Person, self.person_map)
  File "dev/gramps/installed/lib64/python3.4/site-packages/gramps/plugins/database/bsddb_support/write.py", line 2113, in get_from_handle
    raise HandleError('Handle is None')
gramps.gen.errors.HandleError: Handle is None
```

### DIFF
--- a/gramps/plugins/view/relview.py
+++ b/gramps/plugins/view/relview.py
@@ -497,7 +497,9 @@ class RelationshipView(NavigationView):
         for old_child in self.header.get_children():
             self.header.remove(old_child)
 
-        person = self.dbstate.db.get_person_from_handle(obj)
+        person = None
+        if obj:
+            person = self.dbstate.db.get_person_from_handle(obj)
         if not person:
             self.family_action.set_sensitive(False)
             self.order_action.set_sensitive(False)


dsblank:
Looks good; thanks!

Paul-Franklin:
@dsblank
Should bugs like this be applied to gramps42 also?

tecknicaltom:
I think the problem is just that the recent refactor of the database level has introduced some new exceptions being raised where they were silently handled before. Specifically, doing a lookup of a None handle or a handle not found seem to have (correct me if I'm wrong) previously returned None, where now they raise. The calling code was written to handle the None returns, but not handle the exceptions.

tecknicaltom:
Specifically, the change made in 5c958bd7fbaea082df6f2ce427042ad0a642815e

dsblank:
@tecknicaltom is right that this is a recent change to master, but not related to the database refactor. I'm not sure it was a good idea on @Nick-Hall 's commit to raise an exception when a handle is not found. I think this will continue to cause issues as much code uses return None to test to see if an object exists.

Nick-Hall:
This code should not be used to check if an object exists.  The "has" methods should be used instead.

It this case, None is being passed to a "get" method.  We know that this does not correspond to an object before the call is made, so we can avoid the database access.

The idea behind the original commit was to avoid the None Type errors we were getting.  Eventually I suggest that we catch the HandleError exception and present a user-friendly dialog to the user.

tecknicaltom:
Thanks for the clarification @Nick-Hall and @dsblank - I was working on another change to update the documentation on the get_*_from_handle methods that say that they return None if not found. But I realized that this change caused the database classes to differ from the DbReadBase interface.

So my question is: are the DbReadBase/DbWriteBase interfaces now going to raise exceptions on these types of situations (None handle, handle not found) and that behavior should be implemented by (all of) the classes that derive from these base types?

If so, do you guys prefer code to e.g. check that a handle exists and then get the object, thus avoiding an exception, or to just try to get it and catch the potential exception?

Are these changes being worked on by somebody else and I should just hands off?

dsblank:
Yes, all backends should behave the same. I'd prefer that we don't raise exceptions, if there are known things we can test for. Thanks for your help!

Nick-Hall:
Checking for a valid handle when following a reference is not necessary.  However, where the code doesn't know if an object exists, or what type it is, then a check should be made prior to fetching the object.  Only proxies should return None.

Hopefully an exception should never be raised.  It indicates a possible database corruption that should not be ignored.  The problem in previous versions was that potentially serious errors were being hidden.

dsblank:
I prefer returning None in general when it is not found, because otherwise it takes two database accesses to get the data: once to see if it is there, and then again to get it.

Nick-Hall:
Returning None hides errors.  It is not a good idea.

Almost all of the time we are returning objects from references that should be good.  Only one database access is required.

PR #83, opened 2015-12-28
update with current template "gramps.pot" and "ru.po" files

vantu5z:
I think it is not able to merge to gramps42

### DIFF
--- a/po/gramps.pot
+++ b/po/gramps.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2015-07-28 20:56+0200\n"
+"POT-Creation-Date: 2015-12-28 14:16+0300\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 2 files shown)


Paul-Franklin:
Hi Ivan,

I have started to look into this.

Just as background, I don't know how you are making your copies
of the gramps git repository.  (I have never made a "fork" on github
for instance.)  So I don't know what the capabilities of your copy is.

But all (I think) your previous changes were made against what we
call the "gramps42" maintenance branch of the repo, not the "master"
branch of the repo.  That's what we want (gramps42).

In general, we prefer translation updates to be done in the branch
which users are using, the currently-released branch.  This is the
gramps42 branch currently.  The reason is that then the users will
see the improved translations the next time a maintenance release
happens (for instance 4.2.2 in a week or so).  The "master" branch
is only released once a year, so any improvements there will take
longer for a user to see them.  (For instance 5.0.0 in mid-2016.)

Also, we make string changes more frequently in the development
branch, adding strings, removing strings, modifying strings.  So it
is a lot more work to try to keep that up to date.  And arguably not
worth doing, since very few users test-run the "master" branch, never
mind run it in a non-English language.  (We don't make a gramps.pot
there very often either, for the same reason.)

So while there is nothing wrong with you creating a new gramps.pot
template file, the problem is that if you generate it in the "master"
branch then some of the strings will not match the strings in the
current-release (gramps42) branch.  So some strings may end up
marked "fuzzy" if they have been compared with the "master"
gramps.pot file, even if they are not really "fuzzy" when compared
with the gramps42 gramps.pot file, the one users are using.

So I don't know where you have done the "python update_po.py -p"
and the "python update_po.py -m de.po" commands (and I assume
you meant the latter to say "ru.po" instead, from what you said you did).

And therefore I don't know where you want me to commit your
https://github.com/vantu5z/gramps/commit/8fa3f7444d1683c23a484fd8414d4d5efe6ebb15
ru.po file -- to "master" or "gramps42".

The ideal thing to do -- but I don't know if you have time, or
if your repo copy allows you access to the gramps42 branch --
would be to make a new gramps42 gramps.pot file (assuming
you still want to know the latest strings) and then do the
"python update_po.py -m ru.po" using that gramps.pot and
this new ru.po file.  The fix the resulting ru.po file.

Just for reference, here are the two directories:
https://github.com/gramps-project/gramps/tree/maintenance/gramps42/po
https://github.com/gramps-project/gramps/tree/master/po

Thanks again for helping make gramps better.

Nick-Hall:
Please keep the first line of the commit message to a maximum of 70 characters.  Extra information such as the commands run can be put in the body of the commit message.

See our [committing policies](https://gramps-project.org/wiki/index.php?title=Committing_policies).

vantu5z:
Thanks you for explanation. I don't know all of finesses of git (just learn it now), try to make my commits better and clear in the future.

Paul, for this PR, it can be committed to "master" branch.
And yes, I have access to the gramps42 branch. Now try to update template from it and resolve "fuzzy" strings. Soon make PR to the gramps42 branch.

Paul-Franklin:
You are doing fine.  Better than I did when gramps changed
to git a year or two ago and I had to learn it.  8-)

As you asked, I have committed your https://github.com/vantu5z/gramps/commit/8fa3f7444d1683c23a484fd8414d4d5efe6ebb15 ru.po to master:
https://github.com/gramps-project/gramps/commit/d1e11f41724f918cb7244c1b5308f892315efc26

Thank you.

PR #87, opened 2015-12-31
Fix handling of a None handle in libsubstkeyword

tecknicaltom:
Another uncaught exception from calling get_person_from_handle with
None, this one uncovered by running test/runtest.sh to generate reports


Paul-Franklin:
Thank you.  I have committed your fix to gramps42 https://github.com/gramps-project/gramps/commit/0106661523cbb0c37d1fcfdc8ca40a324d25bd75 and master https://github.com/gramps-project/gramps/commit/4ff2cf30e2b6d40e6ef2b4c2fd8c91bc39112ece

PR #90, opened 2016-01-21
Change output order in Relationship Graph plugin

olaf-ka:
This patch adds a function sort_persons() to RelGraphReport, which will
change the processing order of persons for the output. Starting from persons
with no known parents in the database, the individual family trees are
traversed in a down-and-up fashion, so that closely related people should
appear close to each other in the output. Previously the output order was
purely given by the order of internal IDs.

Rearranging the output order will make the life of GraphViz just so much easier.
It will then manage to produce graphs with much fewer unnecessary intercrossing
arrows between generations and people and ultimately produce a much nicer
output graph.


gramps/plugins/graph/gvrelgraph.py
```diff
@@ -182,10 +182,71 @@ def write_report(self):
...
+                parent = self.database.get_person_from_handle(parent_handle)
+                if parent is None:
+                    continue
+                hasParent = True
+            if not hasParent:
+                rootNodes.append(person_handle)
+
+        # now start from all root nodes we found and traverse their trees
+        outlist = list()
+        p_done = {}
```
QuLogic: I think you want to use a `set()` here instead of a dictionary. The actual value is unused.

olaf-ka:
Won't argue with that, I know nothing about Python. I've also fixed the same thing a couple of lines further down - that's where I got the basic idea from...

If anyone could test and merge it (or comment otherwise), I'd still be very grateful.

romjerome:
Tested your last changes.
It works as expected.

Thank you for your contributions.

Also, your proposal sounds much easier than path provided
on feature request # [2635](https://gramps-project.org/bugs/view.php?id=2635).

Note, your name(s) on copyright (top of the file) is/are maybe missing on your changes?

olaf-ka:
The main scope of the mentioned feature request is the ordering of children in a family. In combination with "ordering=out" my patch should achieve the desired ordering as specified in the Gramps Family list, but to me that's just completely coincidental, and I didn't even think about it at the time of writing. But all the better if it makes more people happy than just me! :)

I haven't added my name to the copyright holders list. I obviously assumed that the patch can be redistributed under the same terms as the rest of Gramps and don't care much about the copyright for my contribution. Should it be required by some project contribution rules, I can add my name as well, though...

dsblank:
How expensive is this compared to the original? Will this still work on large trees? (The Travis failure seems unrelated to this PR).

dsblank:
How does this work on large trees? Any impact?

olaf-ka:
I've tested it on a family tree with 263 people with no noticeable impact. Would people be less reluctant to merge it if this was an option somewhere in the menu and off by default?

Paul-Franklin:
The example.gramps sample DB has about two thousand people,
if you want to try that.  Also, I understand that there are many-people
GEDCOMs out on the Internet (I think at least 100k people) -- you could
ask on gramps-users if you can't find one.

From my point of view, I'd rather have it as a settable option,
as you suggested (off by default).

Also, I'd rather see sort_persons take an argument, so it
was more obvious (rather that using self.person_handles
as now) that it depends on the previous value.

olaf-ka:
As requested, this is now a new option appearing under the ledger "Graph Style" and it's off by default. Also, sort_persons() now takes an argument rather than working on the class member.

(And for some reason even the Travis checks seem to succeed now!)

dsblank:
This looks good to me. Thanks! :cake:

QuLogic:
> (And for some reason even the Travis checks seem to succeed now!)

`master` (with which you recently merged) is no longer building against ancient Python 3.2.

sam-m888:
@olaf-ka Can you confirm that the "**Descendants of ..**." filter does not work with your patch?

While testing I only selected the new checkbox and ran the report against the following filter on example.gramps:
* Entire database  -  Linux Gramps greys out/unresponsive until completed (same with my 10K family tree) I'm willing to wait :+1: 
* Descendants of I0044 - **Report completed but creates an empty pdf. ** ( Checked and yes I0044 has descendants and checked that Gramps master and 423 creates report)
* Descendant families of I0044 - Worked
* Ancestors of I0044 - Worked
* People with common ancestor with  I0044 - Worked
* Custom made filter - **untested**

- 

Breadcrumb for me to update the Gramps 5.0 wiki when rolled over:
https://gramps-project.org/wiki/index.php?title=Gramps_4.2_Wiki_Manual_-_Reports_-_part_5#Graph_Style_2

 [] Sort persons in output from ancestors to descendants (unchecked by default) - Sort the persons in the output from ancestors to descendants, which usually improves the graph layout.

newoption-relationship graph-graph style

Paul-Franklin:
Can you also confirm that you tested with example.gramps
(as Sam did), even if you don't want to use a bigger tree?

What are the timing differences, with and without your patch?

Thanks.

dsblank:
Looks like this needs a rebase.

dsblank:
Closed due to inactivity. If you can provided a rebase and answers to above questions, please reopen. Thanks!

sam-m888:
Forgot about this one, probably should have just been marked WIP. Please keep open.

Requested:
https://www.reddit.com/r/gramps/comments/5qzlgy/how_to_fix_sibling_order_wrong_in_relationships/

sam-m888:
On issue [2635](https://gramps-project.org/bugs/view.php?id=2635) I've attached _pr90-reconstructed-diff-non-mergable.txt_ As the original repository for PR90 has been deleted I was unable to pull the PR to test, but as I could see the changes (thankfully) in the [commits and file changes, pages] I've added ( pr90-reconstructed-diff-non-mergable.txt ) that needs to be rebased and forced pushed back to PR90

sam-m888:
@olaf-ka Hi, hope your around, are you able to re-push this please? Thank you

Nick-Hall:
@sam-m888 I can fetch the original commits, but don't have permission to push them back to this PR.

prculley:
I have an interest in this as well (my tree has some issues with this sort of thing ).  Perhaps if someone would create a branch off of gramps-project (instead of a personal github) then we could all make changes?
Probably need a new PR though...

Nick-Hall:
@prculley I have pushed a branch called PR90.

prculley:
So I downloaded this from branch PR90 and tested it; it had a couple of issues where either we missed something in the recreation or someone never quite finished it up.  I got it working (pushed my changes as another commit, which should be merged with others when we decide this is ready to include).

It seems to work correctly.  It has the nice effect of making the graph do things in the order that families children are shown in Gramps.

I tested with my full tree, and with several filters that reduce my tree several different ways; all seem to work as expected (I visually compared the graphs with those where this option was turned off; took a while on the full tree...).
Even running under debug, it doesn't seem to affect performance enough to warrant an option to turn it off, in my opinion.  Just a couple of seconds for the whole creation.  My full tree of 980 people takes a 30x3 foot scroll of paper to make out details, I doubt if many people will be creating graphs much larger than this.
We could add a progress indicator to make users more comfortable with really large trees, but I think that is a different topic.
I recommend we accept this, and remove the option (too many already IMO).

Nick-Hall:
@prculley For some reason two out of four of Olaf's commits were missing from the branch.  I have pushed it again.  Does this remove the need for your commit?

prculley:
@Nick-Hall The two commits missing were a result of my own forced push; I had done an interactive rebase combining a couple of them so that only the base change (the new sort) and the added option were present as separate commits.  And then I added my fix commit...
So now it is as it was, broken again.

It appears that somewhere along the line, someone tried to change the passing of person_handles as parameter to self.person_handles, and did not quite finish it.  All my commit did was to fix that and add an assert statement to double check that the sort routine did not change the total number of people being processed (I was not sure of my understanding of his algorithm, and thought I saw a possible hole).

Do you want me to push my commit again?

Nick-Hall:
@prculley Yes please.  Remove the option if you don't think we need it, and push again.  Then I'll give it a quick test before merging.

prculley:
I removed the option and force pushed again as a single commit.
Tested and working fine for me.

Nick-Hall:
@prculley I get an assertion error when running the report with the example database and default options.

Paul-Franklin:
I don't understand github, nor its web interface (which is known
to sometimes be wrong in what it shows me), but for this p.r. it
is saying "This branch has conflicts that must be resolved."

Paul-Franklin:
Or maybe that's just for this p.r. and not the PR90 branch?

Sorry.

prculley:
I tried to replicate this; I tried both the example/gramps/example.gramps, and the data/tests/example.gramps on both my windows system (which probably doesn't have default settings), and on Linux (which I think has default settings).  In both cases everything worked fine.

You did test the PR90 branch, not the data here?

Nick-Hall:
@prculley I am testing commit bf465670515ded00ec2760e748776f02c30d69f4 with a new import of the example/gramps/example.gramps file.  The error I get is:
```
line 267, in sort_persons
assert len(person_handle_list) == len(outlist)
AssertionError
```

prculley:
@Nick-Hall I don't know what to say; I don't get the failure.  I tried that specific commit in a new checkout, used the example database, and it works.  Also tried in dbapi, in case that might be what you are using, still worked.

I had added that assert line because I was uncertain that the algorithm was going to include everyone; if it turns out that this is some sort of ordering dependency (for xyz in blah) then I'll have to give this some more thought.  Since it doesn't seem that the author is around anymore.

Would you be willing to set a breakpoint on that line, and perhaps compare the two sets of data to see what is missing?  Or at least see how many items are missing?  If I can determine if it is particular branches or people in the family tree then maybe I can get a clue what is going on.

prculley:
@Nick-Hall Never mind, it seems that somehow you are using a filter; perhaps I had set my output options in a way that was different than the default.  In any event, I found a filter that re-creates the issue and am trying to figure out what is happening now.

Nick-Hall:
@prculley Sorry about that - I was using the descendants filter.

prculley:
I found the issue.  It seems the author was expecting the report graph to have been filtered like a proxy; he was testing for None on get_person_from_handle.  Since we are using a filter apply instead, we should be testing against the filtered data.
I removed the tests for None, added the test for membership in filtered data, and did a pylint on the new code bits.
Only one commit, pushed to branch PR90, so as to be ready to integrate.

uxcq:
I should read a bit more closely that the work is being done on the PR90 branch and not here 👍 

https://github.com/gramps-project/gramps/tree/PR90

Much cleaner chart 👍

Nick-Hall:
Merged in commit 7f813e95e5995dd26833ffd7edfcfda46c79304e.

PR #92, opened 2016-01-22
Finish translation: ending ";" in keywords of desktop entry

sebschub:
openSUSE's RPM checker warns: 
> gramps.noarch: W: invalid-desktopfile /usr/share/applications/gramps.desktop value "Sukututkimus;Suvun tarina;Tutkimus;Sukupuu;GEDCOM:" for locale string list key "Keywords[fi]" in group "Desktop Entry" does not have a semicolon (';') as trailing character

This PR just corrects this. Please cherry-pick this also in master.

### DIFF
--- a/po/fi.po
+++ b/po/fi.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: Gramps422_fi_maintenance_POT_2015-10-27\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2015-10-27 09:33+0100\n"
-"PO-Revision-Date: 2015-10-27 22:04+0200\n"
+"PO-Revision-Date: 2016-01-22 10:25+0100\n"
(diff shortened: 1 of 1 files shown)


tosky:
The error from Travis is unrelated to the change. The change itself is quite trivial, does not impact the code and it's worth to fix. So I'm stepping out from my usual role of "just commit Italian translations" and taking the responsibility of merging this.

dsblank:
@tosky Thanks! Can you also apply to master?

tosky:
@dsblank sure, but should I simply cherry-pick the change, or would it be better to periodically merge the stable branch into master (as it happens in other projects)?

dsblank:
@tosky No, I don't think that we quite have that organization yet. We need to change the process a bit to be able to do that. Maybe in Gramps 5.0.0 we can do the standard. For now, you can just cherry-pick. Thanks! It is great to have new developers.

PR #93, opened 2016-01-23
7335: Cleanup code, filenames, and documentation

dsblank:
First part of cleanup, changes:

MediaObject -> Media
mediaobj -> media
mediaobject -> media


gramps/gui/clipboard.py
```diff
@@ -1644,7 +1644,7 @@ def edit_obj(self, objclass, handle):
                              self.uistate, [], ref)
                 except WindowActiveError:
                     pass
-        elif objclass in ['Media', 'MediaObject']:
+        elif objclass in ['Media', 'Media']:
```
Nick-Hall: The list is no longer required.  There are 3 other similar lines.

Nick-Hall:
Thanks.  This is a nice tidy up.

gramps/gui/clipboard.py
dsblank: Good catch! I thought I had caught all of those. Will fix and merge.

Nick-Hall:
Why commit a merge in the middle?  It makes a mess of the history.

dsblank:
Oh, I should have done a "git rebase"?

Nick-Hall:
Yes.

dsblank:
Ok, thanks. I didn't think that would show in a PR.

PR #94, opened 2016-01-24
New output options for Relationship Graph

olaf-ka:
- In Relationship Graphs, when displaying birth/death/etc. events, add an option to display date and place on the same line
- Also in Relationship Graphs, add an option to display a list of all occupation events with years and places rather than just having the option of displaying the most recent one


dsblank:
Thanks!

PR #95, opened 2016-02-05
Abstract Select API

dsblank:
Adds the following properties to the db:

- Person
- Family

etc. for each of the primary object types. Each of these is a QuerySet with the following methods:

- proxy(name, args...) - applies a named-proxy ("living", "private", or "referenced")
- filter(obj, args...) - applies a FilterObject
- map(f) - applies a function to all selected objects
- count() - returns number of matches
- limit() - set start or limit or selection
- where(fieldname=value, ...) - where clause; can use OR, NOT, and AND
- order(fieldname, ...) - order-by clause 
- select([fields...]) - returns generator
- tag(tag_name) - puts a tag on all selected items, if not tagged already

Hides the db._select implementation


jralls:
Shouldn't you consolidate the dispatch tables into a single place? ISTM having 5 identical instances and a sixth that's identical except for a few extra entries will turn into a maintenance headache downstream.

jralls:
Looks like you messed up a rebase somehow. If you merge it as-is there will be duplicate history. Try again with a rebase -i and take out the merge commit cd59aca. You don't want anything in origin/master's history showing up in the PR and right now there's 4 commits plus the stray merge.

dsblank:
Regarding the seemingly duplicated dispatch tables: No, these are private tables that must be kept with each database layer that match particular method calls. This is actually a long-standing issue that is now fixed here, found with the tests added in this PR. When filters or proxies are applied, they actually replace the database, wrapping each with another instance of the database. There may indeed need to be a refactor, but that should wait for another PR.

dsblank:
Re: rebase. I did do a rebase, but had conflicts. I can try again when I get additional time later this week. I definitely need more experience with this aspect of git.

dsblank:
Ok, I think I have this reversed back to before the duplicates occurred.

jralls:
>Regarding the seemingly duplicated dispatch tables: No, these are private tables that must be kept with each database layer that match particular method calls. 

Sounds like a job for polymorphism. Actually the whole thing looks like a job for poymorphism + the factory pattern.

jralls:
> Ok, I think I have this reversed back to before the duplicates occurred. 

Yeah, looks clean now.

dsblank:
Was unable to rebase cleanly. Closing this PR.

PR #96, opened 2016-02-07
Fix setup.py to remove the double '/' when building man pages.

woodsb02:
This was resulting in 'share/man//...' showing in the file list generated for system packages.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -180,8 +180,8 @@ def build_man(build_cmd):
                 filename = False
 
             lang = man_dir[8:]
-            src = build_cmd.build_base  + '/data/man/' + lang  + '/gramps.1.gz'
-            target = 'share/man/' + lang + '/man1'
+            src = build_cmd.build_base  + '/data/man' + lang  + '/gramps.1.gz'
+            target = 'share/man' + lang + '/man1'
             data_files.append((target, [src]))
 
             log.info('Compiling %s >> %s.', src, target)


setup.py
```diff
@@ -180,8 +180,8 @@ def build_man(build_cmd):
                 filename = False
 
             lang = man_dir[8:]
-            src = build_cmd.build_base  + '/data/man/' + lang  + '/gramps.1.gz'
-            target = 'share/man/' + lang + '/man1'
+            src = build_cmd.build_base  + '/data/man' + lang  + '/gramps.1.gz'
+            target = 'share/man' + lang + '/man1'
```
QuLogic: It would be better to use `os.path.join`, though I guess man pages will not be built on Windows, anyway.

uxcq: Yes they have been built on windows by me and by the looks of it others.

https://github.com/gramps-project/gramps/commit/fdb8704bac98808f9f6249f4f11a39f3b998012c

QuLogic: In that case, it makes even more sense to use `os.path.join` to use the right path separators, right?

dsblank:
Yes, using path.join() makes sense. Thanks!

Nick-Hall:
Be careful.  The distutils package requires some paths to be in Unix form.

Paul-Franklin:
@uxcq Just out of curiosity, how do you "build" manpages on Windows?
(I vaguely recall setup.py having a test to disable that if it is being run on Windows.)

sam-m888:
@uxcq @Paul-Franklin On windows you only need to change to the /docs/ directory and run the make.bat file eg: `make.bat html`

On master this currently spits out about 83 warnings, but it does complete.

Paul-Franklin:
@sam-m888 It seems we are talking about different conditions.  I think
you are talking about a git-downloaded gramps source tree which happens
to be on Windows.  I was referring to a Windows AIO installation, which
does have setup.py but I don't think any "docs" folder (or manpages).

dsblank:
Is this ready to commit?

Nick-Hall:
Tested, rebased, tidied up the commit message and merged in commit 7851d3bdfbba3be7e4d87657cd9839dbde747f6a.

PR #98, opened 2016-02-17
Maintenance/gramps42 - this request is invalid

niememat:
I have tried to delete a request but I have not managed. This request must be removed.


Paul-Franklin:
Hi Matti,

I don't know how to make a github clone of the gramps repo, or how to update it (or delete it and start a new one), so somebody else will have to advise you.  But this (#98) pull request has far too many unrelated changes, which somehow need to be eliminated.

You only want to update one file (I assume), the Finnish translation in gramps42,
so (1) that should be the only file which changes, and (2) it (somehow) needs to be
specified to be put into gramps42 and not "master" as this request is.

So I hope some other developer will tell you what you need to do.

niememat:
Hei Paul

Yes, I noticed I made a mistake of the request. Now I found a link that I
received a request for annulment. The request has been canceled,

2016-02-17 19:25 GMT+02:00 Paul Franklin <notifications@github.com>:

PR #99, opened 2016-02-17
3068: Wrong reference number for "same person as" in detdescendantreport

SNoiraud:
I tested this on several databases without problems.

My question is : Is the henry numbering is always good ?


SNoiraud:
AHhh!
I'm trying to test the pull request function. I needed to have two pull request for this.
One for bug 3068 and another for bug8602.
Git automatically added the second pull request to the first one. Is this normal ?
Is it possible to detach the second patch from the #99 pull request ?

dsblank:
I think I understand. You need to have two branches, one for each PR. The second commit should be removed from this branch, a new branch started, and then you can commit that one there.

PR #100, opened 2016-02-19
3068: Wrong reference number for "same person as" in detdescendantreport

SNoiraud:
I tried this on several databases. It works for me.
I have only one question : Is the henry notation always correct with this patch ?

### DIFF
--- a/gramps/plugins/textreport/detdescendantreport.py
+++ b/gramps/plugins/textreport/detdescendantreport.py
@@ -195,7 +195,11 @@ class DetDescendantReport(Report):
     def apply_henry_filter(self,person_handle, index, pid, cur_gen=1):
         if (not person_handle) or (cur_gen > self.max_generations):
             return
-        self.dnumber[person_handle] = pid
+        if person_handle in self.dnumber:
+            if self.dnumber[person_handle] > pid:
+                self.dnumber[person_handle] = pid
+        else:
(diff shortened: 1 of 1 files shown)


dsblank:
This looks like a good change. Any way to add a test that fails before change, and then passes after change?

SNoiraud:
Sorry, I don't understand the following : 
add a test that fails before change, and then passes after change ?
What test are you talking about ?

dsblank:
I'm suggesting to add a file, say "gramps/plugins/test/test_detdescendantreport.py", that contains a unittest that would fail before you add this code, and would pass afterwards. 

Actually, let me add some infrastructure for making this easier...

dsblank:
I've got some code, but it is based on the test framework in #95. I'll need to merge that first.

dsblank:
I've added the test framework now. See an example of running a report here: https://github.com/gramps-project/gramps/blob/master/gramps/plugins/test/test_reports.py#L122

sam-m888:
@eno93 Did you test that this works as expected?

eno93:
@sam-m888 Only that the henry notation was working and not that it was always correct. Test would prove this one way or the other as mentioned by @dsblank

dsblank:
Let's go ahead and merge this, and check for correctness with a test later.

PR #101, opened 2016-02-20
0743: Searching in people view when surnames are collapsed

SNoiraud:


SNoiraud:
It works correctly on 4.2.
On master I get the following problem which is not related to this patch :  
2016-02-20 10:40:39.927: ERROR: grampsapp.py: line 139: Unhandled exception                                               
  File "/home/gramps/gramps-git/gramps/gui/views/treemodels/treebasemodel.py", line 917, in do_get_value                  
    return val.encode('utf-8')                                                                                            
AttributeError: 'bytes' object has no attribute 'encode'

SNoiraud:
When we are looking for some firstname, all nodes containing this string are closed and we have the following message : "Active object not visible"
I propose to open all selected nodes instead (very easy to do).
This is true when we use the searchbar of the filter in the sidebar.

SNoiraud:
I can't apply it anymore, So I close this pull request.

PR #102, opened 2016-02-22
5449: Pedigree view crashes if you happen to choose a child as ancestor

SNoiraud:
I added the possibility to have loops in the database.
I added a new tool to find these loops.

### DIFF
--- a/gramps/gen/utils/alive.py
+++ b/gramps/gen/utils/alive.py
@@ -86,7 +86,7 @@ class ProbablyAlive(object):
         self.MAX_SIB_AGE_DIFF = max_sib_age_diff
         self.MAX_AGE_PROB_ALIVE = max_age_prob_alive
         self.AVG_GENERATION_GAP = avg_generation_gap
-        self.plist = []
+        self.pset = set()
 
     def probably_alive_range(self, person, is_spouse=False):
         # FIXME: some of these computed dates need to be a span. For
@@ -96,7 +96,7 @@ class ProbablyAlive(object):
         #        "between 1930 and 1940")
         if person is None:
             return (None, None, "", None)
-        self.plist = []
+        self.pset = set()
         birth_ref = person.get_birth_ref()
         death_ref = person.get_death_ref()
         death_date = None
@@ -274,9 +274,9 @@ class ProbablyAlive(object):
         # ago.
 
         def descendants_too_old (person, years):
-            if person in self.plist:
+            if person.handle in self.pset:
                 return (None, None, "", None)
-            self.plist.append(person)
+            self.pset.add(person.handle)
             for family_handle in person.get_family_handle_list():
                 family = self.db.get_family_from_handle(family_handle)
                 if not family:
@@ -348,9 +348,9 @@ class ProbablyAlive(object):
             return (date1, date2, explain, other)
 
         def ancestors_too_old(person, year):
-            if person in self.plist:
+            if person.handle in self.pset:
                 return (None, None, "", None)
-            self.plist.append(person)
+            self.pset.add(person.handle)
             LOG.debug("ancestors_too_old('%s', %s)".format(
                 name_displayer.display(person), year) )
             family_handle = person.get_main_parents_family_handle()
--- a/gramps/gen/utils/alive.py
+++ b/gramps/gen/utils/alive.py
@@ -86,6 +86,7 @@ class ProbablyAlive(object):
         self.MAX_SIB_AGE_DIFF = max_sib_age_diff
         self.MAX_AGE_PROB_ALIVE = max_age_prob_alive
         self.AVG_GENERATION_GAP = avg_generation_gap
+        self.plist = []
 
     def probably_alive_range(self, person, is_spouse=False):
         # FIXME: some of these computed dates need to be a span. For
@@ -95,6 +96,7 @@ class ProbablyAlive(object):
         #        "between 1930 and 1940")
         if person is None:
             return (None, None, "", None)
+        self.plist = []
         birth_ref = person.get_birth_ref()
         death_ref = person.get_death_ref()
         death_date = None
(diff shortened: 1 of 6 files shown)


gramps/gen/utils/alive.py
```diff
@@ -86,6 +86,7 @@ def __init__(self,
         self.MAX_SIB_AGE_DIFF = max_sib_age_diff
         self.MAX_AGE_PROB_ALIVE = max_age_prob_alive
         self.AVG_GENERATION_GAP = avg_generation_gap
+        self.plist = []
```
dsblank: If you made this a set() rather than a list(), then when you check for a containing person, it can happen in constant time, rather than a sequential search.

gramps/gen/utils/alive.py
```diff
@@ -272,6 +274,9 @@ def probably_alive_range(self, person, is_spouse=False):
         # ago.
 
         def descendants_too_old (person, years):
+            if person in self.plist:
+                return (None, None, "", None)
+            self.plist.append(person)
```
dsblank: You probably want to add (and check for) just their handle here, not the whole Person object, right? Otherwise you might have a large number of objects in memory.

gramps/gen/utils/alive.py
```diff
@@ -343,6 +348,9 @@ def descendants_too_old (person, years):
             return (date1, date2, explain, other)
 
         def ancestors_too_old(person, year):
+            if person in self.plist:
```
dsblank: `person.handle in self.pset`

For example.

dsblank:
Is this ready to go?

sam-m888:
Worked for me,

Suggestions:
* Add a few loops to example.gramps
* Add a help button

Minor editorial issue:
* 5 trailing whitespace need to be removed.
eg;
$ git apply ~/Downloads/102.patch
/home/102.patch:263: trailing whitespace.
        self.progress.set_pass(_('Looking for possible loop for each person'), 
/home/102.patch:271: trailing whitespace.
            GObject.TYPE_STRING)    # 4==family gid 
/home/102.patch:332: trailing whitespace.
            for pth in range(len(self.model)): 
/home/102.patch:349: trailing whitespace.
            if not family: 
/home/102.patch:419: new blank line at EOF.
+
warning: 5 lines add whitespace errors.

SNoiraud:
What is the best way to add a loop example to example.gramps ?
I created the wiki tool page for gramps 5.0.
Need to have a test case in example.gramps to make a screenshot of the tool window.
work in progress...

Paul-Franklin:
I have not tried to run this but just from a quick glance I see that
you have tool_modes = [TOOL_MODE_GUI, TOOL_MODE_CLI]
but you are doing things like importing the ProgressMeter, so if
you really want this tool to be run from the CLI you'll need to
change that to use the various progress methods in the User
class instead (self.progress = user.progress(...)).  Or else you
can remove TOOL_MODE_CLI from the .gpr.py file.

Paul-Franklin:
I am just guessing, since I haven't tried to run it, but when you ask
"What is the best way to add a loop example to example.gramps"
I would suppose adding one or two new people, say a great-grandchild
of the main person and maybe a great-grandparent too, depending on
what your tool checks for?   But that's just a guess.

SNoiraud:
I was thinking to add a new testloop.gramps in example/gramps
Do you think it's a good idea ?

I suppressed ", TOOL_MODE_CLI" in tools.gpr.py

SNoiraud:
What is the best way to remove the three last patchs ?
I don't understand what the errors are.
perhaps it is better to add another test file.

SNoiraud:
I propose to add two example files : 
 child-father-child-loop.gramps
 test_complex_loop.gramp

The help file for wiki 5.0 is done. perhaps we need some corrections.

SNoiraud:
If you agree, it's OK for me.
This is ready to go.

sam-m888:
Looks good.

The rest of the tool and report pages link to the user wiki using the name of the report/tool as shown on the menu, can we keep it consistent? 
menu-find-possible-loops
eg: I suggest either change the menu entry to "Find possible loop in the database" which I believe is to long, or rename the menu entry to say something more meaningful (in other languages also) like "Find Ancestral loops" and update the link to ?

The following warning is output(only while using the tool - one for each example database) :
```
$ ./Gramps.py 
(Gramps.py:3071): Gtk-WARNING **: Could not lookup object notrelated on signal clicked of object close
(Gramps.py:3071): Gtk-WARNING **: Could not lookup object notrelated on signal clicked of object close

Gramps Settings:
----------------
 python    : 3.4.3
 gramps    : 5.0.0-23ab7eb

 gtk++     : 3.10.8
 pygobject : 3.12.0
 pango     : 1.36.3

```

PS:I'll fix the wiki page (looks like you copied the 4.0 page instead of the 4.2 so it does not have all my changes)

Paul-Franklin:
As far as names, I'd suggest something more generic,
such as "Find Database Loops" etc.  That assumes that
the tool might be enhanced some time in the future.  But
I would not say "ancestral" as that seems limiting to me.

Also, as long as I'm offering opinions, I'd say that instead of
being in Utilities it would better belong in Family Tree Repair
(where things like Check and Repair Database are) or else in
Family Tree Processing (where things like Find Possible
Duplicate People are).  But I find the whole Tools menus and
submenus somewhat inconsistent generally, since after all
Utilities has Verify the Data and all three of those seem to
me to be roughly in the same area.  Oh well.  In the final
analysis I doubt any of it really matters, but certainly before
5.0.0 is released would be the time to shuffle things IF that
was going to happen.

SNoiraud:
> PS:[Fixed 20160412] I'll fix the wiki page (looks like you copied the 4.0 page instead of the 4.2 so it does not have all my changes)
Sorry. I saw you made modifications. Do I need to restaure some data ? is it correct for you ?

> Gtk-WARNING **: Could not lookup object notrelated on signal clicked of object
I think this is related to the glade file. Done.

> As far as names, I'd suggest something more generic, such as "Find Database Loops"
Done.

sam-m888:
@SNoiraud _Sorry. I saw you made modifications. Do I need to restaure some data ? is it correct for you ?_
Not I've already fixed it.

Tested with the each of the example files and it works :+1: 

Thanks for all your efforts

PR #104, opened 2016-03-03
fix some crashes in export preview

tecknicaltom:
These issues were all found while trying to use the "Calculate Previews"
feature of the export wizard.

gen/proxy/referencedbyselection.py: ProxyDbBase need not define the
writable functions (add_*, commit_*, del_*) in its __tables. ProxyDbBase
doesn't have methods for add_person, etc, so this was causing an
AttributeError

gen/proxy/referencedbyselection.py: Add some handle checks to prevent
lookups on None handles

gen/proxy/proxybase.py: remove redundant test

### DIFF
--- a/gramps/gen/proxy/proxybase.py
+++ b/gramps/gen/proxy/proxybase.py
@@ -591,12 +591,9 @@ class ProxyDbBase(DbReadBase):
             return attr
 
         # if a write-method:
-        if ((name in DbWriteBase.__dict__ and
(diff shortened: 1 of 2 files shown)


dsblank:
Thanks for that!

We are also working on using Proxies in other ways, so we may want to refactor this at some point to make it so that you could use a Proxy as a full database layer. But this is a good fix for this issue.

PR #105, opened 2016-03-08
Use new Travis container and Apt addon

DaAwesomeP:
What this does
 - Moves to the faster Travis container
 - Uses Travis's father APT caching/install method

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,35 +3,50 @@
 
 language: python
 python:
-    - 3.4
+ - 3.4
+sudo: false
+cache: pip
+addons:
+  apt:
+    packages:
+    - gir1.2-pango-1.0
+    - gir1.2-gtk-3.0
+    - xdg-utils
+    - librsvg2-common
+    - libglib2.0-dev
+    - intltool
+    # - python3-gobject Provided by python3-gi
+    - python3-gi
+    - python3-cairo
+    - python3-gi-cairo
+    - python3-bsddb3
+    - python3-dev
+    - python3-nose
+    - python3-mock
 
 before_install:
-    - pip install --upgrade pip
-    - pip install --upgrade setuptools wheel nose coverage codecov
+ - pip install --upgrade pip
+ - pip install --upgrade setuptools wheel nose coverage codecov
 
 install:
-    - time sudo apt-get update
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,35 +3,50 @@
...
-    - travis_retry sudo pip install pyicu==1.8
-
-    - cd $TRAVIS_BUILD_DIR
-    # $TRAVIS_BUILD_DIR is set to the location of the cloned repository:
-    # for example: /home/travis/build/gramps-project/gramps
-    - python setup.py build
-
-#before_script:
+ - travis_retry pip install --upgrade pillow
+ - travis_retry pip install pyicu==1.8
```
uxcq: Wonder why gramps developers limit pyicu to 1.8 it works for me on 1.92 ?

QuLogic:
Not sure it was necessary to reduce the indents to 1 space; seems like needless churn.

.travis.yml
dsblank: I believe that there was an incompatibility, but perhaps that is fixed.

dsblank:
Thank you for this! I've been trying to get this to the new Travis method for many months. Now, the aut-coverage should work as well.

DaAwesomeP:
Just to note, there are currently two issues with the build. This is not introduced in my updated script, but it fails to detect PyICU and Pillow.

dsblank:
I think that is because `nosetests3` is using a different python, but not sure. I think the solution will be to change that line to something like:

```python
python -m nose --nologcapture --with-coverage --cover-package=gramps --exclude=TestcaseGenerator --exclude=vcard --exclude=merge_ref_test  --exclude=user_test gramps
```

I will test/try...

dsblank:
But then, I think the apt-get python3 packages are not being used by the `python` env...

DaAwesomeP:
When I force python3, it stops finding gi or bsddb3. Pip finds it, but not nose.
https://travis-ci.org/DaAwesomeP/gramps/builds/114529604

dsblank:
Right. `python` is a different Python from that of `python3`. I believe the apt-gets are installing for `python3`. I can't get pip to install gtk for `python`.

DaAwesomeP:
Well I used `pip3`, so I would have thought that installing the dependencies with that and then running nose with `python3` would have worked.

DaAwesomeP:
Travis is running as Python 3.4, so I would also think that only Python 3 would be available unless `python` is installed afterwards.

PR #106, opened 2016-03-14
add a filter to the tree of rule filters

tecknicaltom:
I can never remember which category filters are in when making a new
rule. This change adds a filter entry above the treeview. As you type,
filters that do not contain the search text (case insensitively) are
made not visible. Categories are always visible, but will no longer
expand if all of their child nodes are hidden.

### DIFF
--- a/gramps/gui/editors/filtereditor.py
+++ b/gramps/gui/editors/filtereditor.py
@@ -455,6 +455,7 @@ class EditRule(ManagedWindow):
                         self.get_widget('rule_editor_title'),label)
         self.window.hide()
         self.valuebox = self.get_widget('valuebox')
+        self.rname_filter = self.get_widget('ruletreefilter')
         self.rname = self.get_widget('ruletree')
         self.rule_name = self.get_widget('rulename')
         self.description = self.get_widget('description')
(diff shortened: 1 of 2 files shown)


dsblank:
This looks good, and useful! The only thing I would want to add is tests. We don't have any tests of Gtk things currently, but a pointer for how we could add these, if you feel up to it:

* http://unpythonic.blogspot.ca/2007/03/unit-testing-pygtk.html

This probably won't work by itself as we call show(), but perhaps the above combined with mock. Also, we can turn on the framebuffer in our Travis tests: https://github.com/gramps-project/gramps/blob/master/.travis.yml#L41

dsblank:
Any chance of tests? Or should we apply this as-is?

tecknicaltom:
While improving the test coverage of Gramps in general is big on my personal todo list, I'm pretty swamped now and won't be able to figure out gtk testing. Whether you'd like to merge this without tests or wait until some point in the future when I can carve out some time to develop them is up to you, but it prolly won't be soon. sorry

dsblank:
No problem. Let's merge this and keep working towards testing. Thanks!

PR #107, opened 2016-03-27
Initial batch of tests of filter rules

tecknicaltom:
After finding a bug in a filter that caused it to return incomplete
results in master (not yet patched), I decided to develop tests for the
filter rules. This first batch is all listed person filter rules that
either take no parameters or take a single ID parameter. I originally
wanted to use the smaller data.gramps instead of example.gramps, but
there were more unrepresented edge cases that wouldn't allow for testing
all filters (e.g. disconnected individuals).

All test values are based on what Gramps currently returns, with some
efforts made to try to verify the results. If the example.gramps file
changes, these tests will likely have to be updated. I tried to
cross-version validate them with v4.1, but the implementation of
DictionaryDb is too incomplete and many tests fail for unrelated
reasons.

### DIFF
--- a/gramps/gen/filters/rules/test/person_rules_test.py
+++ b/gramps/gen/filters/rules/test/person_rules_test.py
@@ -1,0 +1,365 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2016 Tom Samstag
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


sam-m888:
Thanks for improving Gramps :+1:

dsblank:
Great! These tests (checking specific handles) are very specific, but I think that is something we can live with for now, especially now that we have integration testing. Thank you for these!

tecknicaltom:
I wasn't sure how else to test these functions to make sure any breaking changes are exposed. Only checking the number of filtered people could miss breakages that add and remove the same number of people. Having the tests reimplement the intended logic of the filters seems counterproductive and prone to other breakages. If you've got ideas on a way to continue with these tests in a way that doesn't mean checking specific handles, I'm very open to suggestion. It's really why I wanted to submit the PR for the tests I've done thus far instead of waiting to get better coverage.

dsblank:
I would have probably looked up the objects, and then checked the gramps_id's, as you do with handles. But this is fine for now. We're up to 26.3% !

https://codecov.io/github/gramps-project/gramps/commits

Thanks!

PR #108, opened 2016-03-29
fix the name map of tags in DictionaryDb

tecknicaltom:
DictionaryDb had a bug where the same tag could end up in the name
mapping multiple times. This was most easily seen when loading a gramps
xml into a DictionaryDb, where an extra entry with the tag name of ''
would be created and would ultimately be referencing the last-created
Tag (also referenced by its proper tag name). This change makes sure
that when editing a tag, any existing references in the name mapping are
deleted before adding.


dsblank:
Thank you... good catch!

tecknicaltom:
only caught by writing more tests... filtering people with the tag '' (empty string) unexpectedly came back with a person. trying to figure out why led me to this issue

PR #109, opened 2016-04-10
8153 GEDCOM 5.5.1 Support Custom Event on export

sam-m888:


dsblank:
This looks fine. We should add tests in gramps/plugins/test/test_reports.py (that should be renamed, as it can be used for exports and imports too). But, there isn't an example there for exports yet.

sam-m888:
@dsblank the only thing I've worked out so far is I'm terrible at writing test and is bit beyond my skill level at the moment.

Paul-Franklin:
"bit beyond my skill level at the moment"

Don't let it worry you.  When I started using gramps the only
thing I knew about Python was the name!  (Not that I know
much more now.)

dsblank:
@sam-m888 No problem... let's merge this then, and work up some tests over time.

PR #111, opened 2016-04-12
Use new-style Python 3 classes

sam-m888:
Cleanup

### DIFF
--- a/gramps/gen/plug/utils.py
+++ b/gramps/gen/plug/utils.py
@@ -396,7 +396,7 @@ def load_addon_file(path, callback=None):
 # OpenFileOrStdout class
 #
 #-------------------------------------------------------------------------
-class OpenFileOrStdout:
+class OpenFileOrStdout(object):
     """Context manager to open a file or stdout for writing."""
     def __init__(self, filename, encoding=None):
         self.filename = filename
@@ -422,7 +422,7 @@ class OpenFileOrStdout:
 # OpenFileOrStdin class
 #
 #-------------------------------------------------------------------------
-class OpenFileOrStdin:
+class OpenFileOrStdin(object):
     """Context manager to open a file or stdin for reading."""
     def __init__(self, filename, add_mode='', encoding=None):
         self.filename = filename
--- a/gramps/gen/user.py
+++ b/gramps/gen/user.py
@@ -25,7 +25,7 @@ The User class provides basic interaction with the user.
 import sys
 from contextlib import contextmanager
 
-class User():
+class User(object):
     """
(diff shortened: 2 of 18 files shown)


dsblank:
I don't think these "(object)" parts are needed in Python3. The last experiments I did showed that they were irrelevant.

sam-m888:
Just going off this [SO](http://stackoverflow.com/questions/54867/what-is-the-difference-between-old-style-and-new-style-classes-in-python) answer that mentions:

> Python 3 only has new-style classes. No matter if you subclass from object or not, classes are new-style in Python 3. It is however recommended that you still subclass from object.

You decide, if needed or not and close.

QuLogic:
I'm pretty sure that recommendation is simply for backwards-compatibility with Python 2.

sam-m888:
In that case, as Gramps is only python 3, close this.

dsblank:
I read the SO article, and the related Python2 docs. I don't even see the recommendation any more for Python2 code to subclass from object. I think @QuLogic is right regarding the original reasoning.

So, given that there are two ways to do it, and the other way is shorter and more concise, I guess we should leave it as it is.

eno93:
This probably should have been accepted if only to make these classes consistent with the rest of the code base.

dsblank:
I'm not in favor of touching a lot of code to just make it consistent with an older style. In fact, I'm not in favor of "clean up" PRs at all, unless we have tests to check to make sure we haven't broken anything.

eno93:
Is that stated anywhere for the project?

QuLogic:
If anything, you should remove `object` from the _other_ classes, since it does nothing now.

dsblank:
@eno93 That's just my personal preference. 

@QuLogic Yes, we should do that, at some point. But my main point is that if we want to spend time/energy on making the code better, it should be in writing tests. When we are done with that, then we can consider making the code more consistent, have higher pylint score, etc.

Paul-Franklin:
https://gramps-project.org/wiki/index.php?title=3.3_Roadmap#Policy_changes
says that for about five years gramps has had the policy of requiring a "minimum
pylint score of 9" and the following link says that requirement is for all "new files":
https://gramps-project.org/wiki/index.php?title=Programming_guidelines#Pylint

Although to be sure not every gramps developer follows that policy.  Just
as some developers don't write tests.  "Do as I say, not as I do."  8-)

Perhaps a sentence about generating tests should be added to the "Best
Practices" section at the bottom of that link?  Just a thought.

Nick-Hall:
I have removed the redundant inheritance from `object` in commit f688a96e9a9516e211ff753f19c9ad672fe8575e.

PR #112, opened 2016-04-12
Don't use equality to check None

sam-m888:


dsblank:
These look good, but I hesitate some making changes without tests for these sections of code. We need to get the tests in place before we have wide scale clean-up only PRs.

I'll go ahead apply this one.

PR #115, opened 2016-04-13
New WHERE clause format as Python function: replaces Django-style 

dsblank:
New WHERE clause format as Python function: replaces Django-style parameters.

Based on an idea provided by Gerald Britton.

* WHERE clause can now be a specialized Python function/lambda that takes a gramps.gen.lib primary object, and creates a boolean value.
* Requires Python meta library
* Known limitation: currently can't return boolean property; need to compare to True/False (should be fixable)
* adds a .where() QuerySet to differentiate from .filter() which is real Python code
* removes all of old Django-style where machinery
* adds tests for where functions
* If a filter has a where method, it will use it. This isn't currently used in Gramps yet.
* where clauses are also not used in gramps yet, but can be used in addons, and gramps_connect

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,9 +48,8 @@ script:
  # --exclude=TestUser because of older version of mock
  #                    without configure_mock
  # python -c "import sys; print(sys.path)"
- # need to find search path for nostests' python
- - locate bin/python3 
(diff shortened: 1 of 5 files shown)


dsblank:
Removing this; out of date.

PR #116, opened 2016-04-14
Change defaults for new db (dbapi) and read-only no-database

dsblank:
Change defaults for new db (dbapi) and read-only no-database (dictionarydb)


dsblank:
This PR is now outdated. Ideas being discussed in mailing list.

PR #117, opened 2016-04-16
7921 Move holidays.xml.in to data [refactor]

sam-m888:
> The holidays.xml.in file belongs with all the other
> similar files, in the "data" directory.

https://gramps-project.org/bugs/view.php?id=7921


sam-m888:
Thank you :+1:

PR #60, opened 2015-09-07
5357 Add window position config keys for Clipboard & main window(with editorial changes removed)

sam-m888:
Changes:
- [x] Clipboard window remembers its previous position
- [x] Main Window starts in center of screen and when moved and Gramps restarted remembers  previous positions

Please test on:
- [ ] Apple Mac
- [ ] MS-Windows

Tested on
- [x] Linux (Ubuntu 14.04.3)

**To do**
*Main Window*
- [x] Record window sizes and positions at the end of the session and restoring them when the program is next launched.
- [ ] Remove hard-coded values 
- [x] Setting window manager hints for initial startup of Gramps ( Gtk.WindowPosition.CENTER )

*Clipboard*
- [x] Record window sizes and positions at the end of the session and restoring them when the program is next launched.
- [ ] Remove hard-coded values 
- [ ] Setting window manager hints for 1st use of clipboard ( Gtk.WindowPosition.CENTER_ON_PARENT ) https://github.com/gramps-project/gramps/blob/master/gramps/gui/clipboard.py#L1390

Any feedback welcomed.

https://gramps-project.org/bugs/view.php?id=5357

Was:
https://github.com/gramps-project/gramps/pull/58

PR #119, opened 2016-04-19
Rename Main Interface Config keys[for PR-60]

sam-m888:
More meaningful names for the config keys.

Can be applied independently  of  [PR-60](https://github.com/gramps-project/gramps/pull/60)


dsblank:
Looks good to me.

PR #120, opened 2016-04-19
2066 Add dropdown list to Link Editor

sam-m888:
Shows custom entries on the Link Editors drop down list

See [#2066](https://gramps-project.org/bugs/view.php?id=2066)


dsblank:
Thanks!

PR #121, opened 2016-04-20
Add compression option on XML export

dsblank:
The write options needed to include "binary" to allow the system to write to the stream.

### DIFF
--- a/gramps/gui/plug/export/__init__.py
+++ b/gramps/gui/plug/export/__init__.py
@@ -22,3 +22,4 @@
 
 from ._exportassistant import ExportAssistant
 from ._exportoptions import WriterOptionBox
+from ._exportoptions import WriterOptionBoxWithCompression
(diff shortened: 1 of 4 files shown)


dsblank:
See https://gramps-project.org/bugs/view.php?id=9394 for details.

Paul-Franklin:
Yes, like you said on 9394 I don't think this should be committed
until it's checked on all three platforms.

dsblank:
I should also note that if one doesn't have the gzip Python library, the old module would crash with the same error that this fixes. That is, if this doesn't work, then we would have to create a different way of making .gramps files with no compression.

sam-m888:
Worked on Windows 7 Pro 64bit using Gramps 5.0.0 [13b4f93 + pr121] AIO 64 bit(Personal Testing version)

*Exported uncompressed
*Imported uncompressed 

checkbox-selection-uncompressed-xml-pr121-b9394-win764bit

dsblank:
Applied this, as we need it for further development. If there are issues on Mac, we will need to fix them.

PR #127, opened 2016-05-05
Bug 9414 fix empty Place Alternate Names on import

prculley:
The if acquisition.name is always true since Place.__init__ puts PlaceName object here. This should be
"if acquisition.name.value "
this would check for an actual assignment, not just an empty entry.

### DIFF
--- a/gramps/gen/lib/place.py
+++ b/gramps/gen/lib/place.py
@@ -577,7 +577,7 @@ class Place(CitationBase, NoteBase, MediaBase, UrlBase, PrimaryObject):
         :param acquisition: instance to merge
         :type acquisition: :class:'~.place.Place
         """
-        if acquisition.name and (acquisition.name not in self.alt_names):
+        if acquisition.name.value and (acquisition.name not in self.alt_names):
             self.alt_names.append(acquisition.name)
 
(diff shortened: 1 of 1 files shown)


dsblank:
What bug is this? 9144 appears to be a different bug:

https://gramps-project.org/bugs/view.php?id=9144

eno93:
@dsblank His branch name is `prculley:bug9414` so I guess it's this one:
https://gramps-project.org/bugs/view.php?id=9414

prculley:
Sorry for the dyslexic typo in the commit message.  It is indeed bug 9414.  I've pushed the same patch again with an amended commit message.  Is this enough, or should I kill this pull request, and resubmit?

dsblank:
No, this is fine. I was just trying to find a test condition. We are working towards having a good test framework so that someday, each commit/PR would have an associated test. The test would fail before the commit, and would pass after the commit. That is a lot to ask of a new committer. But would you be willing to accept a PR on your PR if I give an example?

dsblank:
I made a PR on your repo:

https://github.com/prculley/gramps/pulls

sam-m888:
Replaced by https://github.com/gramps-project/gramps/pull/141

PR #128, opened 2016-05-05
Bug 9415 Place Alt Names gets duplicated entries

prculley:
Code assumed that constructs like
if (new_alt_name not in self.alt_names): ...
will check for duplicates. But the 'not in' python code depends on a '==' check between items. Gramps code never defined the '==' ('__eq__') operation for alt names.


dsblank:

PR #129, opened 2016-05-06
Bug 9173, fix broken GEDCOM import PLAC:FORM handling

prculley:
The original problem was that the PLAC:FORM parsing in libgedcom.py load_place was trying to store the results of its parsing after the dbase.commit (or dbase.add). So the results were lost. I shifted the code around so that the results were available for the commit, but the routine also was trying to store 'location' results based on the using a place.handle that had not been created yet. So I then saved the 'location' results till the handle was available.


dsblank:

PR #130, opened 2016-05-06
Bug 9416 GEDCOM import PLAC:FORM in local mode doesn't work

prculley:
GEDCOM can import PLAC lines with a subsidiary FORM line which describes what the comma separated list of place items means. In its global form, GEDCOM puts these in the GEDCOM header, in the local form, the FORM line comes beneath the associated PLAC line.
In the libgedcom.py __event_place code, the FORM line is parsed and handled in a sub_state, however it is not saved in the main state.


dsblank:

dsblank:
Looks like this has conflicts now. Can you rebase? Let us know if you have questions.

dsblank:
Do any of those applied to gramps42 need to also be applied to master?

PR #131, opened 2016-05-06
Bug 9425 GEDCOM import some Place Names & Titles are blank

prculley:
When a GEDCOM import contains an event with ADDR (but NOT a PLAC), the resulting Gramps Place entry does not contain a name or title.
In the libgedcom.py __merge_address routine I created a string suitable for a name or title, and In the __event_addr routine I applied it. 
In the __event_place routine I had to fix up a side effect of the previous patch, that is to change the detection of 'no previous place'. I decided to call the place type 'Address' to make it identifieable as coming from the ADDR field of GEDCOM.  Note: 'Address' is a CUSTOM place type at the moment.  As it is already in the master .pot file, it should internationalize ok. I don't know if it makes sense to update the placetype.py to include this so it is not CUSTOM or not...


dsblank:

PR #133, opened 2016-05-08
PyGIWarning: importing without specifiying a version first

RossGammon:
Follow-on from https://gramps-project.org/bugs/view.php?id=8981
This patch was applied in Debian (gramps 42 branch) due to warnings picked
up in the build log and from testing by running gramps from the command
line. I have rebased it on master, as it will probably be a while before the
warnings become errors.


dsblank:
Looks good; thanks!

PR #134, opened 2016-05-08
9443: Gramps not appearing in Gnome Software

RossGammon:
I updated the appdata.xml file to the latest version of the specification
(https://www.freedesktop.org/software/appstream/docs/index.html), adding
some new tags in the process.

I also installed the appdata.xml file to the latest location in the
specification.

The XML has bene validated by the appstreamcli validation tool, and the
patch applied in Debian (Gramps 4.2.3).

If it looks OK, it should probably be cherry-picked to the gramps42
maintenance branch.

### DIFF
--- a/data/gramps.appdata.xml.in
+++ b/data/gramps.appdata.xml.in
@@ -1,18 +1,32 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<application>
-<id type="desktop">gramps.desktop</id>
-<licence>CC0</licence>
-<description>
-<_p>Gramps is a genealogy program that is both intuitive for hobbyists and feature-complete for professional genealogists.</_p>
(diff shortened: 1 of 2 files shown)


dsblank:
Looks good. I'll apply to gramps42, as suggested. Thanks!

tosky:
Please, fix <p> as <_p>, otherwise the tags can't be translated iirc (see po/update_po.py)

tosky:
Marker eaten correctly:  I mean, restore &lt;p&gt; as &lt;_p&gt;

dsblank:
@tosky Good catch! @RossGammon do you want to make that change, check that it still works, and make a new PR?

RossGammon:
Sorry,

The <_p> tags did not pass the appstreamcil validation test on the appdata.xml.in file. They must be <p>.

I missed the fact that they will be translated automagically and the final result will be something like <p xml:lang="uk">.

There will be a followup PR soon. Thanks tosky!

PR #136, opened 2016-05-08
Followup to PR 134

RossGammon:
Put back the <_p> tags (including the trailing ones :-) ) so that the appdata.xml file is translated again.


dsblank:
Applied to gramps42, as well. Thanks!

PR #137, opened 2016-05-09
Convert a few more opens to use the with CM

sam-m888:
The python [with](https://docs.python.org/3/reference/compound_stmts.html#with) statement

### DIFF
--- a/gramps/plugins/database/bsddb_support/write.py
+++ b/gramps/plugins/database/bsddb_support/write.py
@@ -555,7 +555,7 @@ class DbBsddb(DbBsddbRead, DbWriteBase, UpdateCallback):
 
     def __make_zip_backup(self, dirname):
         import zipfile
-        # In Windows resrved characters is "<>:"/\|?*"
+        # In Windows reserved characters is "<>:"/\|?*"
         reserved_char = r':,<>"/\|?* '
         replace_char = "-__________"
         title = self.get_dbname()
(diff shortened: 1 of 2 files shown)


gramps/gen/utils/docgen/odstab.py
```diff
@@ -335,7 +335,8 @@ def open(self,filename):
 
         try:
             self.content_xml = tempfile.mktemp()
-            self.f = open(self.content_xml,"wb")
+            with open(self.content_xml,"wb") as self.f:
```
dsblank: This is going to behave differently. Before it was open, and now it will be closed. Is that right?

sam-m888: I believe I have this one correct, as this line only test that it can open/create the file and then on line [344 ](https://github.com/gramps-project/gramps/blob/1a05aadca7f78f0ec34cda74c087e38bb15c2318/gramps/gen/utils/docgen/odstab.py#L336-L344)it actually opens the file in the original code with:
`self.f = open(self.content_xml,"w", encoding='utf-8')`

dsblank: Ok, then that makes sense. But it doesn't need to save the value in self.f then does it?

sam-m888: Looking back on the notes, `odstab.py` was identified as one of the files I should leave alone, so I'll update the PR and remove it.

PR #138, opened 2016-05-09
Remove Python2.5 Zipfile duplicate

sam-m888:
Not imported anywhere in the Gramps code I could find.


dsblank:
It is used here:

https://github.com/sam-m888/gramps/blob/RemoveUnusedZipfileCodePython25/gramps/gen/plug/utils.py#L234

sam-m888:
Well that was obvious! :( Not having a good week!

QuLogic:
Why not just replace that usage with the correct standard library version?

sam-m888:
@QuLogic I'll take a look again, later.

PR #139, opened 2016-05-09
5125 Add CSV Import support for AFN and REFN

sam-m888:
From feature request [5125](https://gramps-project.org/bugs/view.php?id=5125)

### DIFF
--- a/gramps/plugins/importer/importcsv.py
+++ b/gramps/plugins/importer/importcsv.py
@@ -52,7 +52,8 @@ ngettext = glocale.translation.ngettext # else "nearby" comments are ignored
 from gramps.gen.lib import (ChildRef, Citation, Event, EventRef, EventType,
                             Family, FamilyRelType, Name, NameType, Note,
                             NoteType, Person, Place, Source, Surname, Tag,
-                            PlaceName, PlaceType, PlaceRef)
+                            PlaceName, PlaceType, PlaceRef, Attribute,
+                            AttributeType)
 from gramps.gen.db import DbTxn
 from gramps.gen.datehandler import parser as _dp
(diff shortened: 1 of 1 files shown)


dsblank:
This will be a perfect use for a test:

1. add a new csv file in example/csv/
2. add a new test in gramps/plugins/test/test_reports.py 

Let me know if you have any questions.

sam-m888:
Ok, so to be clear I'll add the small test csv file from the issue 5125.

Here's where I get lost, in the test file you mention I only see test reports used for export and not for importing different file format, although I see a remarked section for [export_csv](https://github.com/gramps-project/gramps/blob/1a05aadca7f78f0ec34cda74c087e38bb15c2318/gramps/plugins/test/test_reports.py#L250)

Should import files and export report files be mixed in the same test file?  

I can see the importers have two [test files](https://github.com/gramps-project/gramps/tree/master/gramps/plugins/importer/test)  
 * importgeneweb_test.py
 * importvcard_test.py

sam-m888:
[>_Could be expanded to handle user-defined attributes; maybe:_](https://gramps-project.org/bugs/view.php?id=5125#c21282)

Just re-read your comment and I like this idea.

Probably be better to support custom attributes instead of add AFN/REFN and others

dsblank:
If you get the file paths right, you should be able to just add it to gramps/plugins/test/test_reports.py

Yes, that is not a well-named file, but should be fairly easy to add. I'm not sure why the other csv import isn't finished and commented out... maybe there was something wrong with csv imports at one time.

dsblank:
Do you want to close this one, and open another addressing the issues via custom attributes?

sam-m888:
Yes, that sounds like a good idea :+1:

PR #140, opened 2016-05-10
make EditPrimary, empty_object abstract

tecknicaltom:
I started fixing the issue I independently found which was fixed in 2f31096abdefb773ea4079a6e29a147c51dc3c13 - always do a pull before fixing issues. In addition to the changes in that commit, but I also went one step further in making EditPrimary abstract.

This makes EditPrimary an abstract class using abc.ABCMeta and marks
EditPrimary.empty_object as an abstract method. This moves the error
raised by forgetting to implement empty_object from when something
unsupported is done with its return (like .serialize()) to invocation
time, and provides a more obvious error.

### DIFF
--- a/gramps/gui/editors/editprimary.py
+++ b/gramps/gui/editors/editprimary.py
@@ -26,6 +26,7 @@
 #-------------------------------------------------------------------------
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
+import abc
 
 #-------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


uxcq:
[Abstract Base Classes](https://docs.python.org/3/library/abc.html)  :+1:

dsblank:
This a good idea, in general. I don't think ABCs existed when Gramps was started, and there are places that should be converted to this. Would have made my life easier over the years!

PR #142, opened 2016-05-11
Bug 9448: Merge unit test for PlaceCheck not working correctly

prculley:
While examining the merge_test.py to add test for bug 9414 and 9415, I realized the code could never fail! I believe the test was written incorrectly. It seems that during setup of the test, the author initialized the 'reference' and 'merge' by copying from the 'merged into' place. I guess this was to make them all have the same name.

Unfortunately, the copying operation in the Place initialize copied the pointer to the alt_places list, rather than the list itself. So any change made to any of the Place objects, alt_places entries, effectively occurred to all at once. So the comparison test always passes no matter what happens to alt_names.  Basically compares 'expected' to 'expected'.

Once that was corrected, I found that the test code 'expected' results for one of the affected tests was out of order and incorrectly included the primary name in the alt_names list.

After correcting this issue, I also added a test to specifically check for bug 9414.

Note: These merge tests will now fail miserably unless the PRs for bug 9414 and 9415 are also applied.

### DIFF
--- a/gramps/gen/lib/test/merge_test.py
+++ b/gramps/gen/lib/test/merge_test.py
@@ -1389,8 +1389,14 @@ class PlaceCheck(unittest.TestCase, PrivacyBaseTest, MediaBaseTest,
     def setUp(self):
         self.phoenix = Place()
         self.phoenix.set_title('Place 1')
-        self.titanic = Place(self.phoenix)
-        self.ref_obj = Place(self.phoenix)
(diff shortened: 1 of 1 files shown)


dsblank:
Great! I'll apply this one first, and let them fail. And then apply the others to make sure that it is fixed. It might be easier int the future to put the test and fix in the same PR. Thanks for these fixes!

PR #143, opened 2016-05-12
Bug 9448: Merge unit test for PlaceCheck not working correctly

prculley:
Same as previous PR with same name, but applied to gramps42


sam-m888:
Thank you :+1: 

https://github.com/gramps-project/gramps/pull/142

PR #145, opened 2016-05-12
Bug 9415 Place Alt Names gets duplicated entries

prculley:
Same as previous PR but as applied to master


sam-m888:
Thank you :+1: 

https://github.com/gramps-project/gramps/pull/128

PR #146, opened 2016-05-12
Bug 9173, fix broken GEDCOM import PLAC:FORM handling

prculley:
Same as previous PR of same name, but as applies to master


sam-m888:
Thank you :+1: 

https://github.com/gramps-project/gramps/pull/129

PR #149, opened 2016-05-12
Bug 9416 GEDCOM import PLAC:FORM in local mode doesn't work

prculley:
GEDCOM can import PLAC lines with a subsidiary FORM line which describes what the comma separated list of place items means. In its global form, GEDCOM puts these in the GEDCOM header, in the local form, the FORM line comes beneath the associated PLAC line.
Gramps import code has code which is meant to deal with this and store the PLAC line with appropriate enclosing Places.
In the libgedcom.py __event_place code, the FORM line is parsed and handled in a sub_state, however it is not saved in the main state.


sam-m888:
https://github.com/gramps-project/gramps/pull/130

PR #151, opened 2016-05-12
Bug 9430 GEDCOM import PLAC or ADDR attached Notes etc. are lost

prculley:
This is same as previous PR, but applies to master.

GEDCOM (and gramps) allows notes, sources, and media to be attached to places. In many cases these are lost.
In libgedcom.py __event_place, after parsing the sub-levels for notes etc, the sub_state was never merged back into the main state. In __event_addr, the merge was attempted, but the original code did the merge into an object called 'place' which, depending on the code path taken, was sometimes equal to state.place, and sometimes not. When not, the notes etc. were lost.


sam-m888:
Thank you :+1: 

https://github.com/gramps-project/gramps/pull/132

PR #152, opened 2016-05-13
9450: LivingProxyDB leaves data in when back references followed

dsblank:
Although these options are about replace a living person's name, the name won't show at all on certain reports, like place_report. See bug for more info.


Paul-Franklin:
Thank you.  I hope to be able to look this over in a week or so.

PR #154, opened 2016-05-19
Make trailing whitespace an error

QuLogic:
I saw some complaints about this on the mailing list, so this makes trailing whitespace into an error on Travis.

PEP8 would normally complain about trailing whitespace, but until Gramps code is fully PEP8-compliant, you can consider this an intermediary solution to ensuring there's no trailing whitespace.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_install:
 install:
  - travis_retry pip install --upgrade pillow
  - travis_retry pip install pyicu==1.8
- 
+
  # - cd $TRAVIS_BUILD_DIR
  # $TRAVIS_BUILD_DIR is set to the location of the cloned repository:
  # for example: /home/travis/build/gramps-project/gramps
@@ -47,6 +47,10 @@ script:
  # --exclude=TestUser because of older version of mock
  #                    without configure_mock
  - PYTHONPATH=meta GRAMPS_RESOURCES=. nosetests3 --nologcapture --with-coverage --cover-package=gramps --exclude=TestcaseGenerator --exclude=vcard --exclude=merge_ref_test  --exclude=user_test gramps
+ - if git --no-pager grep --color -n --full-name '[ 	]$' -- \*.py; then
+     echo "ERROR - Trailing whitespace found in source file(s)";
+     exit 1;
+   fi
 
 after_success:
  - codecov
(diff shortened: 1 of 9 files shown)


Nick-Hall:
Thanks.

The aim is to make all Gramps code PEP8 compliant.  Our [Programming Guidelines](https://gramps-project.org/wiki/index.php?title=Programming_guidelines) require that all new files have a pylint score of 9.0 or higher.  Perhaps we could add a check for this.

dsblank:
I don't mind the check for whitespace, but PEP8 is a set of stylistic suggestions, not a list of errors. There are lots of reasons one might want to stray from PEP8 recommendations. I'd rather devs spend time on writing tests than attempting to make PEP8 happy.

Nick-Hall:
Doug - I have to disagree with you here.  Coding standards are very important.  It is not just the author of the code that needs to understand it.  Good clear code helps other developers and can prevent bugs.

I notice that some of your recently added new files do not meet our requirements.  This is not really fair on other developers.  We introduced the PEP8 requirement for good reasons.  The administrators should take as much care in their code as the other contributors.

I will post to the mailing list with a list of new files that don't reach our minimum coding standards.

Writing tests is very important.  The test code should also comply with our standards.

dsblank:
Nick, I don't disagree with the spirit of what you say. But to prioritize efforts, I would put standardizing code below testing. If you have tests written, there are no bugs to track down, and you have spare time, then focus on PEP8. This is both practical and necessary. If one starts cleaning up code without tests to verify that those changes are correct, then you are just asking for the introduction of bugs. That is not fair to the original coders to introduce bugs into existing code.

tosky:
Other big python projects enforce PEP8. For example, the project under the OpenStack umbrella run tests which checks for PEP8 (so PEP8 *is* a test). If you don't need or disagree with specific code convention you can exclude them from the check, but it's definitely better to have PEP8 enforced with exceptions that not having it all.

dsblank:
Other big projects require tests for any actual code changes. I'd be in favor of that. Sure, you can make anything a test, but changing of code without testing the actual changes is asking for the introduction of bugs.

tosky:
I agree with having tests, but PEP8 is the foundation. Changes required by PEP8 are mostly cosmetics, if carefully reviewed by two or more people, should not introduce regressions. More problematic changes could be forcibly excluded with pylint exceptions for now, but not going to PEP8 at all now due to lack of tests means it won't ever be enforced.

PR #156, opened 2016-05-24
bug 9459 GEDCOM import in CLI mode with .ged file containing ANSEL en…

prculley:
…coding tries to pop up gui


dsblank:
The Travis failure is from something else I believe:

https://github.com/gramps-project/gramps/commit/6c2eac9e050537975f87610c77fba609c68d7f0c

PR #157, opened 2016-05-24
Bug 7949, crash on GEDCOM import with empty _AKA lines

prculley:


dsblank:
Does this need to be applied to master as well? Is there a test in your new tests that would show this failure?

prculley:
Yes it needs to be applied to master; I was just trying to figure out 'git
cherrypick' to see how it is best used to create the PR for master.

No test yet; if/when my tests get adopted, I would expect to include a test
as part of each PR.  Or maybe as a separate PR so they can be applied first
to watch the failures?  Not sure how much double checking you guys do, and
want to make it as easy as possible.

Paul Culley

dsblank:
I think most projects include the test and the fix in the same PR. People can try them independently if they wish. 

Yes, cherry-pick is probably the best way to go for our current workflow.

Thanks; I'll commit this now and in master.

PR #158, opened 2016-05-24
Bug 7949, crash on GEDCOM import with empty _AKA lines

prculley:
Same as for gramps42


dsblank:
I did a cherry-pick for this, so it should already be applied:

f92c35d7976ec5aa25e9fac799be58731a15b6f6

(I don't understand why github says that it has no merge conflicts, though)

dsblank:
Please re-open if it is not included.

PR #159, opened 2016-05-24
Test imports

prculley:
Prototype of import tests module.  Depending on input from email stream, there may be changes, such as location of test data subdirectory.  See [https://gramps-project.org/wiki/index.php?title=Test_Imports_module](url)  For more information.


dsblank:
Looks like this may need a "rebase". Also, looks like some errors:

https://travis-ci.org/gramps-project/gramps/builds/132658438

prculley:
Just to let you know, I'm working on these issues.
1) nose failures; Looks like nose doesn't like functions defined with the word 'test' in them unless they are tests.
2) I included a few cases that were deliberately designed to fail, as a demonstration.  I can remove or 'except' these cases.
3) the big one; I'm getting a failure that is intermittent; I have not figured out why.  About 1/3 of the time the test passes, otherwise fails.  Again it has something to do with Gramps_IDs; they are not getting duplicated, but are getting swapped.  Again, I suspect the libgedcom code for source gramps IDs ('S0004', S0005') from my MixInlineXrefNote.ged test.  Do you know a way to run the diff.py module with the bsddb?  Just to see if that is part of the issue.

dsblank:
On Wed, May 25, 2016 at 4:21 PM, Paul Culley <notifications@github.com>
wrote:

> Just to let you know, I'm working on these issues.
> 1) nose failures; Looks like nose doesn't like functions defined with the
> word 'test' in them unless they are tests.
>

That is correct, unless we --exclude them. Usually easier to just rename
the class/method.


> 2) I included a few cases that were deliberately designed to fail, as a
> demonstration. I can remove or 'except' these cases.
>

You can test for the failure, as well. You can try/except it, or check
another indicator.


> 3) the big one; I'm getting a failure that is intermittent; I have not
> figured out why. About 1/3 of the time the test passes, otherwise fails.
> Again it has something to do with Gramps_IDs; they are not getting
> duplicated, but are getting swapped. Again, I suspect the libgedcom code
> for source gramps IDs ('S0004', S0005') from my MixInlineXrefNote.ged test.
>
It might be that the ordering of the calls of your deterministic
create_id() are coming in in different orders. That's one reason I thought
using the gramps_id was so clever.



> Do you know a way to run the diff.py module with the bsddb? Just to see if
> that is part of the issue.
>
Diff.py shouldn't care, and in fact, shouldn't know about the underlying
database, unless I have forgotten some detail. What seems to be the issue
there?

-Doug

prculley:
Regarding trying to use bsddb in diff; actually in diff.py import_as_dict.  Seems that I somehow need to get the database created with a filename or something, just changing make_database("inmemorydb") to "bsddb" is clearly insufficient.  Tried hacking the code a bit to pass in a filename, but it seems to be trying to load a non-existent db and gives me grief about upgrading...  I'll have to figure out how to properly create/open to even try to see if it is a db issue.
My puzzle is why it would be intermittent; I don't think anything in libgedcom is likely to have any random element to it.  I do see that there are some calls to random in generic.py line 1483, but the debugger did not trigger when I set a breakpoint there...  Mystified.

dsblank:
You can use

>>> db = dbstate.open_database(NAME)

or

>>> db = dbstate.make_database("bsddb")
>>> db.load(PATH)

Does that help?

-Doug

prculley:
Yeah; the latter was my quick hack.  It did not work.  Trying to figure out how to create the empty db first now...

dsblank:
On Wed, May 25, 2016 at 7:09 PM, Paul Culley <notifications@github.com>
wrote:

> Yeah; the latter was my quick hack. It did not work. Trying to figure out
> how to create the empty db first now...
>
Ah, if it is a new database, I think you can make it with:

>>> db = make_database("bsddb")
>>> db.write_version(PATH)
>>> db.load(PATH)

-Doug

prculley:
Ok that did the trick.  So even with bsddb I am still getting intermittent errors; tried it 4 times, last time it swapped IDs again.  So I guess its not an db issue; but something in gedcom import...  Time for dinner now...

prculley:
Seem to have found the problem with my inconsistent source Gramps IDs.  In libgedcom, for this case, source title and handle are stored in a dict for later addition to the db.  The problem seems to be that when the dict is iterated for playback, it does not do so consistently in different runs.  The python docs say 
>"Keys and values are iterated over in an arbitrary order which is non-random, varies across Python implementations, and depends on the dictionary’s history of insertions and deletions. If keys, values and items views are iterated over with no intervening modifications to the dictionary, the order of items will directly correspond."

What it does NOT say is that the order will be the same over different runs.  Even if it did, the bit about different versions would make life difficult.  So it looks like I either need to do one of the following;
* Abandon my testing technique since it depends on consistency  (I'm not quite ready yet!) or
* Patch libgedcom to either use an OrderedDict instead of the Dict or
* Patch libgedcom to store the source GID in the DICT (source GID is actually generated and commited before storing the handle in the dict, but is later overwritten by a different GID during dict playback.  A side effect is that there are gaps in the list of source GIDs after import). 

I imagine that I will find a few more land mines like this over time.  Any thoughts?

dsblank:
On Thu, May 26, 2016 at 10:16 AM, Paul Culley <notifications@github.com>
wrote:

> Seem to have found the problem with my inconsistent source Gramps IDs. In
> libgedcom, for this case, source title and handle are stored in a dict for
> later addition to the db. The problem seems to be that when the dict is
> iterated for playback, it does not do so consistently in different runs.
> The python docs say
>
> "Keys and values are iterated over in an arbitrary order which is
> non-random, varies across Python implementations, and depends on the
> dictionary’s history of insertions and deletions. If keys, values and items
> views are iterated over with no intervening modifications to the
> dictionary, the order of items will directly correspond."
>
> What it does NOT say is that the order will be the same over different
> runs. Even if it did, the bit about different versions would make life
> difficult. So it looks like I either need to do one of the following;
>
>    - Abandon my testing technique since it depends on consistency (I'm
>    not quite ready yet!) or
>    - Patch libgedcom to either use an OrderedDict instead of the Dict or
>    - Patch libgedcom to store the source GID in the DICT (source GID is
>    actually generated and commited before storing the handle in the dict, but
>    is later overwritten by a different GID during dict playback. A side effect
>    is that there are gaps in the list of source GIDs after import).
>
> I imagine that I will find a few more land mines like this over time. Any
> thoughts?
>
I think that changing the Dict to an OrderedDict would be fine. It would
have negligible effect on speed, I suspect, and no other issues that I can
imagine.

-Doug

gramps/gen/utils/id.py
```diff
@@ -46,11 +46,19 @@
...
-rand = random.Random(time.time())
+_rand = random.Random(time.time())
+_det_id = False
+
 
 def create_id():
-    return "%08x%08x" % (int(time.time()*10000),
-                         rand.randint(0, sys.maxsize))
+    global _rand
+    if _det_id is True:
```
dsblank: Better to just leave out "is True", unless you are specifically testing between other values of True.

dsblank:
Is there a reason to leave the gramps files as compressed? If not, it would be easier to see diffs between changes on the gramps files if they are uncompressed text files. Other than that (and the in-line comment) this looks ready to go.

prculley:
Yeah; I personally find the compressed .gramps files a bit annoying, but
figured it was to reduce file sizes.

You wouldn't happen to know how to specify 'no compression' via the CLI?  I
know how to do it with GUI, but I could not find any options registered in
the config.py.
 If we want it uncompressed then I thought I would specify how to do it in
the wiki, since the .gramps files are initial created manually, and I'm
trying to fully document the process there

Regarding the id.py module; pylint did not like that module very much to
start with, and it got worse as I changed it.  But I didn't want to change
its much.  But pylint DID suggest to change my original code to "if _det_id
is True:" from "if _det_id == True:"  I keep forgetting that for python,
"if _det_id" doesn't just mean "if exists" but really means "if exists AND
is True".
Still learning.

I'm afraid I don't follow the "inline comment" part of " Other than that
(and the in-line comment) this looks ready to go.".

Paul Culley

dsblank:
Thanks for this! Glad to see the testing framework and sample files. Great addition to gramps.

QuLogic:
Hmm, this dynamic test creation seems like a good reason to use py.test...

dsblank:
@QuLogic Yes, I think using py.test could have additional benefits. This might help re-organize the built-in Python3 versus the environment Python3 in Travis. PR's accepted :smile:

PR #160, opened 2016-05-25
CacheProxyDb: adds LRU cache for handle lookups

dsblank:
Adds a LRU(10000) cache for all db.get_ITEM_from_handle() methods.

Example use in narwebpage speeds report significantly for both BSDDB and DB.API.

For more info, see:

http://gramps.1791082.n4.nabble.com/gramps-5-0-sqlite3-and-narrativeweb-performances-tt4675791.html

### DIFF
--- a/gramps/gen/proxy/cache.py
+++ b/gramps/gen/proxy/cache.py
@@ -1,0 +1,182 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (c) 2016 Gramps Development Team
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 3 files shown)


gramps/gen/proxy/cache.py
```diff
@@ -0,0 +1,180 @@
...
+
+    Does not invalid caches. Should be used only in read-only
+    places, and not where caches are altered.
+    """
+    def __init__(self, database):
+        """
+        CacheProxy will cache items based on their handle.
+
+        Assumes all handles (regardless of type) are unique.
+        Database is called self.db for consistancy with other
```
QuLogic: consistency

dsblank: Thanks! Fixed.

SNoiraud:
Whoaa! great!
It now takes about 8 minutes for my bsddb database (instead of two hours)
and about 7 minutes for the db.api database.
I think you can apply it.

PR #162, opened 2016-05-25
Put a cache on generic/dbapi

dsblank:
Caches handle and gramps_id lookups. Invalidates the cache with commit_ITEM().


dsblank:
Not sure this helps all that much.

PR #163, opened 2016-05-26
Use CacheProxyDb on reports with LivingProxyDb

dsblank:
This PR inserts the CacheProxyDb into reports after possible proxies, targeting LivingProxyDb.

### DIFF
--- a/gramps/plugins/drawreport/ancestortree.py
+++ b/gramps/plugins/drawreport/ancestortree.py
@@ -47,6 +47,7 @@ from gramps.gen.plug.docgen import (FontStyle, ParagraphStyle, GraphicsStyle,
                                     FONT_SANS_SERIF, PARA_ALIGN_CENTER)
 from gramps.plugins.lib.libtreebase import *
 from gramps.plugins.lib.librecurse import AscendPerson
+from gramps.gen.proxy import CacheProxyDb
 
 PT2CM = ReportUtils.pt2cm
 #cm2pt = ReportUtils.cm2pt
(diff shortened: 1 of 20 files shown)


Paul-Franklin:
You could of course put a new "run" into stdoptions
(run_all_three_proxies ? run_all_proxies ?) and then
change every report which did a "run" for the privacy
and then the living proxies into a "run" of the new
(all-three) one, putting the implementation of the
cache proxy into stdoptions.

Or you could just put a "run" of the cache proxy into
stdoptions (even though it isn't technically an "option")
and then you'd only need to add the "run" line to each
report (with no "import" in each report).

Just a thought.

Paul-Franklin:
I don't remember if I ever saved the value from the "run"
or the privacy option, but certainly I sometimes saved the
value from the "living" option, so if they were consolidated
that aspect would need paying attention to.

So "all-three" may be less obviously desirable.

dsblank:
In Gramps 5.1, I think all of this should be re-factored for two things:

1. We need to be able to set the order of application of proxies from the command-line and GUI. 
2. We should consider making probably_alive a property of Person

Paul-Franklin:
Upon further consideration, I think the best thing may be
to commit this p.r. now, so that its code can get into the
alpha release, since that might happen before you get
back.  ISTM that something good now is better than
something better later (especially if the "later" code
isn't really all that much "better").

dsblank:
Sounds good to me. I wanted to at least get your initial feedback, and I think you are right.

PR #170, opened 2016-06-03
bug 9124 patch.  GEDCOM doesn't accept CR as a line terminator.

prculley:
support.
Includes test files with different line terminators (modified .gitattributes to detect '.ged' as binary to avoid git messing up the tests).


.gitattributes
```diff
@@ -7,3 +7,7 @@
 # Have Github ignore js vendored files.
 # https://github.com/gramps-project/gramps/tree/master/data/javascript
 #data/javascript/*.js linguist-vendored
+
+# don't mess with line endings for Gedcom files
+*.ged binary
```
QuLogic: `binary` disable diffing, too. Are you sure that's what you want? Maybe you want to unset `text` to disable end-of-line normalization.

gramps/plugins/lib/libgedcom.py
```diff
@@ -1248,41 +1248,37 @@ def report_error(self, problem, line):
...
+        BaseReader.__init__(self, ifile, enc, __add_msg)
         self.reset()
-
-    def reset(self):
-        self.ifile.seek(0)
-        data = self.ifile.read(3)
-        if data != b"\xef\xbb\xbf":
-            self.ifile.seek(0)
+        if enc == 'UTF_8_SIG':
+            self.ifile = TextIOWrapper(ifile, encoding='utf_8_sig', errors='replace', newline=None)
```
QuLogic: Not sure why you can't just pass `enc` directly to `TextIOWrapper`. Note that encodings are case-insensitive.

gramps/plugins/lib/libgedcom.py
```diff
@@ -1562,10 +1558,15 @@ def __ansel_to_unicode(self, s):
         return ans
 
     def __init__(self, ifile, __add_msg):
-        BaseReader.__init__(self, ifile, "", __add_msg)
+        BaseReader.__init__(self, ifile, "ANSEL", __add_msg)
+        # In theory, we should have been able to skip the encode/decode from
```
QuLogic: I don't understand this comment. It's not Python 2; you _must_ encode/decode for I/O.

gramps/plugins/lib/libgedcom.py
```diff
@@ -7772,26 +7773,30 @@ def __init__(self, ifile):
     def __detect_file_decoder(self, input_file):
         """
         Detects the file encoding of the file by looking for a BOM
-        (byte order marker) in the GEDCOM file. If we detect a UTF-16
-        encoded file, we must connect to a wrapper using the codecs
-        package.
+        (byte order marker) in the GEDCOM file. If we detect a UTF-16 or
+        UTF-8-BOM encoded file, we choose appropriate decoders.  If no BOM
+        is detected, we return in UTF-8 mode it is the more modern option;
```
QuLogic: Between "mode" and "it", it seems there should be the end of the sentence.

gramps/plugins/lib/libgedcom.py
```diff
@@ -7772,26 +7773,30 @@ def __init__(self, ifile):
...
+        and anyway it doesn't really matter as we are only looking for GEDCOM
+        keywords which are only 7-bit ASCII anyway.
+        In any case, we Always return the file in text mode with transparent
```
QuLogic: Always -> always

.gitattributes
prculley: Git still does diff check with binary mode, it just doesn't show lines. (I tested this carefully).  As near as I can tell, leaving '*.ged' out of .gitattributes, leaves git to decide on its own that this is text and to do EOL changes, and possibly trailing space removal, either of which messes up the Gedcoms.

gramps/plugins/lib/libgedcom.py
prculley: The comment doesn't stand alone; for ANSEL, we earlier converted from binary to text, and here are converting back to binary again so the remaining Ansel code below can convert back to text again.  Seems like we could have skipped the first encode/decode step , but would have lost the universal newline support (or had to code it ourselves).

.gitattributes
QuLogic: Yes, I am referring specifically to telling git to [_not_ assume it's text](https://git-scm.com/docs/gitattributes#__code_text_code). It's not the same as [telling git it's binary](https://git-scm.com/docs/gitattributes#_using_macro_attributes).

gramps/plugins/lib/libgedcom.py
QuLogic: OK, can you mention the other conversion in the comment, then?

Nick-Hall:
Merged 66bf311168f4a061b6da73517cf4d7e23015036e.

PR #171, opened 2016-06-03
Bug9124 Doesn't recognize GEDCOM lines terminated with CR

prculley:
    support.

Same as for master branch, but for gramps42


Nick-Hall:
Please rebase.

PR #172, opened 2016-06-03
Bug9124 Doesn't recognize GEDCOM lines terminated with CR

prculley:
support.

Same as for master; but applied to gramps42


Nick-Hall:
Please break some of the long lines.  The maximum line length should be kept to 80 characters.

QuLogic:
I don't understand this whole double-PR thing, but see comments in #170.

Nick-Hall:
Merged be6715cd99bc4483ff7c63b1f78a456af5bdd52c.

PR #173, opened 2016-06-07
A few trailing whitespace fixes to aid compares with master

prculley:
The trailing whitespace cleanup on master made code compares with gramps42 harder; so here are a few updates for that.


Nick-Hall:
We don't normally make such cosmetic changes to maintenance branches.

prculley:
Ok; I was trying to make my life a bit easier, since I am often making changes to both gramps42 and master in these modules, performing diffs to both in the presence of the white space differences makes it harder to see the real changes. 
I thought it would be a trivial commit with no downside.

I'll withdraw the PR.

PR #174, opened 2016-06-13
bug9240 RTF files show incorrect Unicode characters

prculley:
Original code does not follow RTF spec; '\uc1' means loosely; the following Unicode character will be followed by one byte which represents the standard code page character to be used when the appropriate Unicode character is not available.

Gramps code does not put an alternate character in place, so the appropriate code is \uc0 (no bytes).  The first character of the trailing '\uc0' was sometimes recognized as the 'alternate', leading to garbage.  Some parsers caught this error and ignored it, others did not.


prculley:
The travis issue doesn't seem to be related to this PR; looks like someone changed something to do with "death" the person missing from the list has a death event, but no death age.

PR #175, opened 2016-06-13
bug 7565 and 9521; missing newline in table and illegal RTF

prculley:
Initial issue was 7565; messed up multi-line text in tables.  Needed to add newline when paragraph ends, but only if within a cell, not at end.  When testing the code, I found bug 9521, where we did not separate the paragraph formatting from the cell formatting per RTF spec.
Code was changed to defer paragraph formatting till after cell formatting was established.
Original 7565 also mentioned '\line' found in output; this was a mistake on someone's part where he tried to put a tag in place and it got interpreted as text.

Tested by running all text reports in both RTF and PDF versions with a fairly extensive database that used most all features.  Manually compared for content using two versions of Windows Wordpad (Win7 & Win10), two versions of Msoft Word (2007, 2010), and LibreOffice Writer v5.0.5.2.

### DIFF
--- a/gramps/plugins/docgen/rtfdoc.py
+++ b/gramps/plugins/docgen/rtfdoc.py
@@ -185,43 +185,53 @@ class RTFDoc(BaseDoc, TextDoc):
         # build paragraph information
 
         if not self.in_table:
-            self.file.write('\\pard')
+            para_type = '\n\\pard'
+        else:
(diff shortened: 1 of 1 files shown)


prculley:
Travis failure is due to [https://github.com/gramps-project/gramps/commit/6890fc43248dbd7154afc05950515f6972419f57](url)

PR #176, opened 2016-06-15
bug 9499, CSV fails to recognize utf-8 encoding output by Gramps CSV export

prculley:
Fixed CSV export to output utf-8 with BOM; this involved removing the horrible hack that allowed python2.x to deal with unicode, and replacing it with a simple file open. Also sorted the places before putting them out to CSV, since it seems we don't always get the same order of places on scan of the database, which makes the file compare on testing difficult.
Fixed CSV import to check for presense of BOM and open in utf-8 if so, otherwise open with default (os specific) encoding.
Wrote a test file to make sure that CSV export changes will be detected in the future.
Changed .gitattributes, since it wanted to mess with the CRLF on the CSV files.


gramps/plugins/export/exportcsv.py
```diff
@@ -317,9 +251,9 @@ def write_csv(self, *items):
     def export_data(self):
         self.dirname = os.path.dirname (self.filename)
         try:
-            self.g = open(self.filename,"w")
-            self.fp = open(self.filename, "wb")
-            self.g = UnicodeWriter(self.fp)
+            self.fp = open(self.filename, "w", encoding='utf_8_sig',
```
QuLogic: I'm not sure this should be unequivocally enabled; pretty much nobody uses a BOM on Linux...

prculley: How else can a program automatically detect that the input is utf-8?  (I am just trying to get Gramps to accept its own export, which someone made utf-8 some time ago.)

prculley:
Hmmm. Looks like date format for the Travis run is different than the one for my test systems.  And CSV export uses that.  Need to figure out how to make that the same for both...

gramps/plugins/export/exportcsv.py
QuLogic: This change is on the output side though. The BOM only came about on Windows because they picked UTF-16 which could have two orders, and then it percolated to UTF-8, too. But on Linux, no-one bothers with a useless BOM because the whole OS went UTF-8 wholesale.

Nick-Hall:
This needs a rebase.

PR #177, opened 2016-06-16
Bug9499m , CSV fails to recognize utf-8 encoding output by Gramps CSV export

prculley:
Fixed CSV export to output utf-8 with BOM; this involved removing the horrible hack that allowed python2.x to deal with unicode, and replacing it with a simple file open. Also sorted the places before putting them out to CSV, since it seems we don't always get the same order of places on scan of the database, which makes the file compare on testing difficult.
Fixed CSV import to check for presense of BOM and open in utf-8 if so, otherwise open with default (os specific) encoding.
Wrote a test file to make sure that CSV export changes will be detected in the future.
Changed .gitattributes, since it wanted to mess with the CRLF on the CSV files.


prculley:
I still don't know how to "rebase" correctly it seems; I always seem to have to delete and repush the origin branch to make it work without a bunch of weird conflicts.
If anyone can tell me what is wrong with this sequence, I'll be quite grateful:
git fetch upstream
git checkout master
git merge upstream/master
git checkout mybranch
git rebase master
git push origin mybranch

Paul-Franklin:
I am only glancing at your files in my web browser, which makes
it hard to look at everything, and to see everything.  But I wonder
what is the purpose of the file "exp_sample_csv.gramps" as it
seems to be very similar to the existing "data.gramps" file?

Also, I can't tell for sure but from your comment "the CRLF
on the CSV files" I wonder whether your files have a CRLF
and you need them to remain that way?  If so let me warn
you that linux developers don't normally maintain such line
endings when they modify a file.

Also, as I am not a Windows user I am not really familiar
with all this BOM stuff, but I wonder whether some of the
things you are proposing shouldn't be mentioned on the
gramps-devel list first?  (I vaguely recall translators who
sent along translation files with BOM headers which then
had to have the BOM headers removed before putting
them into the repo, as (I think) something broke otherwise.
(But it wasn't me and this is from (vague) memory so I
could be wrong.)

I certainly agree with the previous comment that a BOM
shouldn't be written all the time, on linux and Mac too.
If you have to write one, why not put it in an "if win()"
(or maybe you already have, since as I said I am only
glancing at what you propose)?

Thanks.

Paul-Franklin:
If you don't get an answer to your git question here, ask it on
the list also.  Not every gramps developer reads these messages.

QuLogic:
> git push origin mybranch

`git push -f origin mybranch` - you need to force push to replace the existing branch.

prculley:
I'm OK with not putting a BOM on UTF-8 for Linux systems.  I note that LibreOffice didn't care either with or without it.  But I certainly don't know what other utilities might do with it.
But Windows side of the world does want the BOM on UTF-8 https://msdn.microsoft.com/en-us/library/windows/desktop/dd374101(v=vs.85).aspx.
And I certainly think that Gramps should be able to import its own export files.  On either Linux or Windows.  I would never had noticed this issue if someone had not forced the CSV exporter to ALWAYS export in utf-8; if they had left things up to Python, then it would have been UTF-8 (or the LANG environment, which is the same for my system) on Linux, and cp1252 for Windows.  And unless I had some pretty unusual characters (which my initial test file did not), I would never had seen this issue.

Its easy to test for a BOM, in the Gramps code, and fall back to Python default if not present.  So that I can deal with either on my Windows system.

I don't really expect CSV to be used for interchange between users, so as long as it works on a single system, I'm OK with that.

As for exp_sample_csv.gramps; it is -similar- to data.gramps, but it contains some international characters to help highlight encoding issues.  data.gramps is strictly 7-bit ASCII.  Someone edited out all of the international characters at some time in the past.

As for posting to the list; I did put up a bug tracker issue https://gramps-project.org/bugs/view.php?id=9496, which received no comment.  I thought that might get more attention than the list; apparently I just have to be more patient.  But then, posting a PR does seem to get attention. ;-)

In this case, I guess a CSV's line endings don't really matter much; I was just trying to avoid the usual Windows issues when only LF was present.  For these test files not a big deal.  The *.ged in PR 170 are a different matter; in that case we are actually testing different line endings.

Nick-Hall:
@prculley Bug reports will often be missed by developers.  It is best to post to the list for important discussions.

Your PR looks good, but use `constfunc.win()` to test for a Windows OS.  Also using the constant `codecs.BOM_UTF8` will make your code self-documenting.

Since utf-8 is specified for the export, you should probably specify utf-8 for the import unless the BOM is detected.

prculley:
I've used the codecs.BOM_UTF8 in my next commit. 
I disagree about always specifying utf-8 for import; that may be ok for Linux where QuLogic says "But on Linux, no-one bothers with a useless BOM because the whole OS went UTF-8 wholesale.", but on Windows most spreadsheets will output in the OS specified codepage (for me its CP1252).  So if no BOM, I prefer to use the default Python input.  Which, for my Windows is CP1252, and for my Linux is LANG=en_US.UTF-8 which will import the CSV with no error.  I assume that for international use other Linux users get other language strings, but are still UTF-8.  And I know that other windows users get other code pages appropriate to their locality.

Paul-Franklin:
I have started a thread about the Windows BOM and
UTF-8 files and so on in the gramps-devel mailing list.

I would like to please ask all the participants in the various
pull request discussions going on here at github to move
their discussion to to the gramps-devel list, perhaps by
summarizing things there?

Thanks.

Nick-Hall:
@prculley Yes, you are correct.  I was thinking that users would be importing Gramps exports, but in reality they could well be importing CSV files modified or created by another program.

SNoiraud:
I don't agree to add BOM automaticaly. This is only a windows problem.
I prefer we add an option to choose if we want BOM or not.
I know there are many issues on Linux and Unixes with this kind of header.
The only thing which can be done automaticaly is to see if we have a BOM when importing files.

prculley:
Are you saying that the 'if win, add BOM' kind of logic is not acceptable?  If I understand your comment, that you want to add an option for users to choose, then I really disagree.  I think most users would not have any idea what a 'BOM' or 'utf-8' is.  I realize that developers have no trouble with these esoteric terms and concepts, but I suspect that the vast majority of users would not.

Nick-Hall:
Merged in commit 495e7801d79f6c8c6c37c4d5e941fafd0b0cd829.

PR #182, opened 2016-06-28
fix LOG.warn() deprecation warning

prculley:
LOG.warn() -> LOG.warning()


prculley:
Trailing whitespace in a different commit is failing Travis

PR #183, opened 2016-06-28
Gdk.Cursor.new() deprecation fixes

prculley:
Gdk.Cursor.new(CURSOR) -> Gdk.Cursor.new_for_display(Gdk.Display.get_default(), CURSOR)
Had to break up a lot of long lines, and in some places I defined the cursor at the top of the Class.  

I tested each change by setting a breakpoint at the set_cursor() call which used the defined cursor, and exercising Gramps until the breakpoint triggered.  Found a few self created bugs that way.

### DIFF
--- a/gramps/gui/dbman.py
+++ b/gramps/gui/dbman.py
@@ -110,6 +110,7 @@ BACKEND_COL = 7
 RCS_BUTTON = {True : _('_Extract'), False : _('_Archive')}
 
 class Information(ManagedWindow):
+
     def __init__(self, uistate, data, parent):
         super().__init__(uistate, [], self)
         self.window = Gtk.Dialog()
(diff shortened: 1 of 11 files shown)


prculley:
Trailing whitespace in a different commit is failing Travis

prculley:
Note to self: I have -got- to figure out how to run nosetests on my AIO windows installation...

Nick-Hall:
Merged in commit 55c99fe718d562e48a8ea996696462ea92220485.

Paul-Franklin:
"Note to self: I have -got- to figure out how to run nosetests on my AIO windows installation...'

I'd like to suggest you possibly consider augmenting the wiki with
what you learn as you go along, doing development on Windows
I mean.

You can put it where you want but if you don't want to create
a new page you could add an appendix (or several) to the
wiki page on how to make an AIO from an old AIO, on Windows.

For instance, is there pylint in the AIO or did you have to
find it and then install it?  What about the Python3 unittests
module?  I vaguely recall Josip mentioning that he only
included things in the AIO which were needed for a user to
run gramps, not everything which was needed to develop
gramps on Windows (or test gramps on Windows I suspect).

I think it would be useful to have the information you are
learning preserved for posterity -- and available if we ever
get other Windows users who want to contribute too.

It's more work than just doing it I understand, but I hope
you'll consider doing it.

Thanks.

PR #184, opened 2016-06-30
bug 9501 CSV import/export messes up places

prculley:
When trying to export a CSV file, and import it again into a new tree as a test, I discovered that the places were completely scrambled. Event places had changed to completely different locations and the places list had some of the GrampsIDs wrong or even missing.
Carefully looking at the csv file itself, it appears to be created correctly, the failure is on import.

This does not appear to be an issue if the csv is imported back into the same tree that exported it, in other words a merge.

Two simple patches seem to correct this poor behavior.
1) change order of section output in exportcsv.py; put places first.
2) use place id as the GrampsID for new places when the syntax matches.

P.s. bug 9548 was also fixed in this patch since it is in the same three lines of code in the importer.


prculley:
All right; this won't work until the new export tests PR is accepted and I can get the right files here.

PR #185, opened 2016-07-05
Bug 9561, fix for intermittent plugin recognition via CLI

prculley:
Email discussion on 
http://gramps.1791082.n4.nabble.com/Plugin-issue-intermittent-error-td4676070.html
During CLI loads of plugins, sometimes the plugin information is lost.  Seems that during load of subsequent plugin, the init code of the plugin triggers imports which eventually try to initialize the GUI plugin manager.  That in turn, clears out some of the data needed by the initial CLI plugin process, as part of a 'hidden plugin' initialization.
I believe that the 'hidden plugin' initialization was not necessary, other code ensures that any 'hidden' plugins remain hidden during GUI and CLI use.

### DIFF
--- a/gramps/gui/pluginmanager.py
+++ b/gramps/gui/pluginmanager.py
@@ -78,7 +78,7 @@ class GuiPluginManager(Callback):
         Callback.__init__(self)
         self.basemgr = BasePluginManager.get_instance()
         self.__hidden_plugins = set(config.get('plugin.hiddenplugins'))
-        self.__hidden_changed()
+        # self.__hidden_changed() # See bug 9561
 
     def load_plugin(self, pdata):
         if not self.is_loaded(pdata.id):


bmcage:
You assign to self.__hidden_plugins so clearly the hidden plugins changed and hidden_changed should be called. 
However, this is the init method, so hidden did not exist before, and hence it is not so much hidden_changed, but instead hidden_created, which does not need hidden_changed :-)

The question to ask is if 
self.emit('plugins-reloaded')
should happen after init. Apparently not as this is creation of the plugin manager class.

Please remove the line completely instead of just making it a comment line.

prculley:
Thanks for responding; I though I would leave the change as a comment to assist if some unexpected issue would turn up in the future.  But if you also agree that the self.emit('plugins-reloaded') is not necessary during init (my testing did not find issues for what it is worth), I will remove it.
It looked like you were the author, and I know it has been a while, but I figured you were still the best expert...

dsblank:
We can leave the comment in for now, and the github commit will contain links to this conversation.

PR #186, opened 2016-07-05
Rename test files so that unrelated tests don't conflict with import tests

prculley:
This PR changes the import test module and related test files so that unrelated test files can be stored in the same data/tests directory.  At the moment, import test files begin with 'imp_', and export test files begin with 'exp_'.
see http://gramps.1791082.n4.nabble.com/gramps-files-in-master-td4676025.html
and http://gramps.1791082.n4.nabble.com/Fwd-unittests-td4676084.html


Paul-Franklin:
Merged.  Thanks.

PR #187, opened 2016-07-05
Fix filter tests for changed sample.gramps

prculley:
This changes the filter tests to accommodate the changes to the sample.gramps in
https://github.com/gramps-project/gramps/commit/a60ae79ca06daba8da17d5177e806e70c9777d4d
We can now move a copy of the entire example.gramps and all its dependent media files to data/tests or leave it where it is for now.

### DIFF
--- a/gramps/gen/filters/rules/test/event_rules_test.py
+++ b/gramps/gen/filters/rules/test/event_rules_test.py
@@ -193,7 +193,7 @@ class BaseTest(unittest.TestCase):
         Test HasDayOfWeek rule.
         """
         rule = HasDayOfWeek(['2'])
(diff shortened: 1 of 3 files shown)


Paul-Franklin:
Merged.  Thanks.

prculley:
Was one error, where I missed one; you were a bit quick off the line...

prculley:
Still have not gotten nose tests to run properly on my Windows development system.  It will run an individual named test file fine, but not find and run groups of tests from a directory.  And it is a pain and time-consuming to transfer everything to my linux vm to run them there; or maybe I'm lazy.

PR #188, opened 2016-07-05
Missed the place filter rule when updating for changed example.gramps

prculley:
I'm not perfect so please check to make sure -why- Travis fails before committing my PRs, I tend to use the PR system to check my own work, since nose testing is not working right on my Windows system.

Thank You


Paul-Franklin:
Merged.  Thanks.

PR #190, opened 2016-07-07
Fix Bug 9569 Vcard export order is not consistent

prculley:
On some occasions, the Vcard export output file, exported from a db imported from the same '.gramps' file, puts out data in a different order.

The Python3 code does not provide ordering guarantees on some of its data constructs from run to run, and I have been told that the db.iter_person_handles() also does not guarantee consistent ordering of its return. The results can apparently change from run to run depending on memory layout conditions, garbage collection timing etc.

So we sort the data before export to correct the issue.


Paul-Franklin:
Merged.  Thanks.

PR #191, opened 2016-07-08
Bug 9479 Gedcom import loses spaces in text fields from FTM

prculley:
GEDCOM specifies that when using a CONC continuation line, any spaces that fall at the line break should be placed at the front, one space after the 'CONC' tag.
FTM (Family Tree Maker) has historically been doing the opposite; it places spaces at the end of the line, where it is not easily seen in most editors, and where it is in violation of the GEDCOM spec. I've quoted the lines below so the space is clear.

<pre>  "2 NOTE The FTM way to do a CONC line with space at end --> "
  "3 CONC <--"
  "2 NOTE The GEDCOM spec way to do CONC line, with space at beginning -->"
  "3 CONC  <--"
  "2 NOTE Be nice if we supported FTM way to do CONC so "
  "3 CONC users would not have to spell check and edit their whole "
  "3 CONC database for these kind of errors." </pre>

With Gramps current code, the user end ends up with words run together like this:
<pre>  "The FTM way to do a CONC line with space at end --><--"
  "Be nice if we supported FTM way to do CONC sousers would not have to spell check and edit their wholedatabase for these kind of errors."</pre>

It is possible to support both the GEDCOM specified way, and FTM way with a minor code patch. I think that this enhancement is very worthwhile.
My import tests have had a test in place for a while that exposes this, I just had to remove the exception 'imp_test_FTM_CONC.difs' file to make it pass.


gramps/plugins/lib/libgedcom.py
```diff
@@ -781,7 +781,10 @@ def __readahead(self):
                 # There will normally only be one space between tag and
                 # line_value, but in case there is more then one, remove extra
                 # spaces after CONC/CONT processing
-                data = data[:2] + (data[2].strip(),) + data[3:]
+                # Also, Gedcom spec says there should be no spaces at end of
+                # line, however some programs put them there (FTM), so lets
```
QuLogic: let's

dsblank: @prculley @QuLogic found a typo; can you fix that?

PR #202, opened 2016-07-16
Bug9552 Export to '.wft' file crashes if Unicode characters are present on Windows

prculley:
The exportftree.py module, like other modules, opens the file output for text write, leaving everything else at default. So on my Windows, it ends up in cp1252, which does not support all Unicode characters.

After some research I found  [http://freecode.com/projects/familytree_cgi/](url) which is a barely maintained bit of software by Ferenc Bodon and Simon Ward.  It currently supports input via Gedcom, excel, and csv files.

Reason I had a hard time finding it is that it does not use '.wft' files.

Our exporter puts out a '.wft' file that is in fact a csv with (some of) the fields that their importer can read. A user would have to rename the wft file to csv or txt to actually use it.

It appears that at some point the authors decided that the csv import was unwieldy, they now prefer excel, or their own '.ser' format for the working format and have converters for the others to that for use.

Their csv import code is still there in the ExtendedSimonWardFormat.pm module; it appears that at some point they decided to convert to utf-8 for their default.

So if we continue to want to support this, we should put out utf-8 by default as well.

### DIFF
--- a/data/tests/exp_sample.wft
+++ b/data/tests/exp_sample.wft
@@ -1,0 +1,44 @@
+Anna Hansdotter;;;;;2/10/1864-29/9/1945
+Keith Lloyd Smith;Lloyd Smith;Janis Green;;;11/8/1966
+Hans Peter Smith;Gustaf Smith;Anna Hansdotter;;;17/4/1904-29/1/1977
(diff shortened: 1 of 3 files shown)


dsblank:
Sounds reasonable; looks good. Thanks!

PR #203, opened 2016-07-17
CSV backport to gramps42 for bug 9501 and 9499

prculley:
Same as PR201 and PR177 except for gramps42 branch
P.S. I was tempted to just copy the master files, except I saw all the whitespace errors, so I just did the changes instead.


dsblank:
I see... not "errors" but difference. Thanks for the extra work to make the diff easier to read, and the history of the file cleaner. Looks good!

BTW, one can edit patch/diff files. So, if you get a patch/diff from one branch with a bunch of whitespace differences, you can edit the patch/diff to just remove those sections, and then apply the patch/diff.

PR #206, opened 2016-07-30
bug 9549 Chinese characters are not rendered properly in pdf reports

prculley:
This PR allows the user to change the way fonts are handled for the GUI and report backends on Windows systems.  The default rendering uses a PANGOCAIRO_BACKEND called 'win32'.  That code seems to use some older windows api for font selection, and doesn't seem to do a very good job with far eastern Unicode characters for the fonts typically installed in western Windows machines.
By changing the backend to 'fontconfig', the results are much better.  Whoever coded that up did a significantly better job in this area.  I believe that 'fontconfig' is the usual default for Linux machines, or at least on my Ubuntu.
See bug 9549 for some uploaded pdf files showing the difference.  Also should clear bugs 7574, 9240, 9611, and possibly a few others, when the user utilizes the option.

None of this has any effect on Linux systems ('if win()' around all code).

This adds a new Preferences General item 'Use alternate Font handler for GUI and Reports (requires restart)' that appears ONLY on Windows systems.  It also adds the 'preferences.alternate-fonthandler' config item for use by the CLI.

### DIFF
--- a/gramps/grampsapp.py
+++ b/gramps/grampsapp.py
@@ -46,6 +46,7 @@ from .gen.const import APP_GRAMPS, USER_DIRLIST, HOME_DIR
 from .gen.constfunc import mac
 from .version import VERSION_TUPLE
 from .gen.constfunc import win, get_env_var
+from gramps.gen.config import config
 
 #-------------------------------------------------------------------------
 #
@@ -473,6 +474,9 @@ def main():
         resource_path, filename = os.path.split(os.path.abspath(__file__))
         resource_path, dirname = os.path.split(resource_path)
         os.environ['GRAMPS_RESOURCES'] = resource_path
+    if win() and ('PANGOCAIRO_BACKEND' not in os.environ) and \
+            config.get('preferences.alternate-fonthandler'):
+        os.environ['PANGOCAIRO_BACKEND'] = "fontconfig"
     errors = run()
     if errors and isinstance(errors, list):
         for error in errors:
--- a/gramps/grampsapp.py
+++ b/gramps/grampsapp.py
@@ -46,7 +46,7 @@ from .gen.const import APP_GRAMPS, USER_DIRLIST, HOME_DIR
 from .gen.constfunc import mac
 from .version import VERSION_TUPLE
 from .gen.constfunc import win, get_env_var
-from gramps.gen.config import config
+from .gen.config import config
 
 #-------------------------------------------------------------------------
 #
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -294,6 +294,7 @@ register('paths.quick-backup-filename',
 register('preferences.date-format', 0)
 register('preferences.calendar-format-report', 0)
 register('preferences.cprefix', 'C%04d')
+register('preferences.alternate-fonthandler', False)
 register('preferences.default-source', False)
 register('preferences.tag-on-import', False)
 register('preferences.tag-on-import-format', _("Imported %Y/%m/%d %H:%M:%S"))
--- a/gramps/gui/configure.py
+++ b/gramps/gui/configure.py
@@ -66,6 +66,7 @@ from gramps.gen.plug.utils import available_updates
 from .plug import PluginWindows
 from gramps.gen.errors import WindowActiveError
 from .spell import HAVE_GTKSPELL
+from gramps.gen.constfunc import win
 _ = glocale.translation.gettext
 
 #-------------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


bpisoj:
I think that this (fontbackend) should be an option in gramps.ini and not hardcoded like that.
It also should be changeable/selectable from both gui and cli session.


BTW
Patch is built on assumption that pango have fontconfig backend support and that is not always true because of the size and speed, some builds left them out, other delay load them etc.

Both backend use same technique of custom aliases list for font/glyph substitution which in his current state does not works for all versions of Windows OS.
CJK scripts are known to be a problem on Linux also as they require complex checking rules.

prculley:
I am not too worried about the presence of 'fontconfig' support in the Pango build;
1.  There is an "if win()" on the operation, so it only effects the Windows AIO installations, which ARE built with 'fontconfig'.
2. if fontconfig was not present, the env variable would be ignored.
3. if it becomes a problem, we can have users set the env variable themselves back to 'win32'.  So it is not really 'hard wired'.
I've personally tested this on Win7 and Win10 on two computers each with no noticeable issues.  Of course I am only running the US English versions of Windows.

I'm looking forward to other comments on this.

Paul-Franklin:
If a comment from the man who makes the Windows AIO
bundle -- and who controls what is inside every library file
he includes in that bundle -- doesn't influence you, then
I don't know what will.

By your arguments, this patch is not needed at all, since
/if/ a user complains we can tell them what envariable to set,
and since the present situation (apparently) works fine for
all non-CJK users, they don't need this change at all.

"If it ain't broke, don't fix it."

If you're going to fix it, do it right, integrate it into
the existing gramps user-interface structure, don't
take some quick-and-dirty route instead.

prculley:
I've added the requested GUI and CLI config options, see first note.

gramps/grampsapp.py
```diff
@@ -46,6 +46,7 @@
 from .gen.constfunc import mac
 from .version import VERSION_TUPLE
 from .gen.constfunc import win, get_env_var
+from gramps.gen.config import config
```
Nick-Hall: Use a relative import.

Nick-Hall:
Do we need a GUI configuration options for this?  Will users understand what it means?

Perhaps Windows builds should just select the best option available.

prculley:
So Josip, Enno, and Paul F, voted for the GUI option, I personally would prefer not to have one.  But due to limited testing (it seems I'm the only Windows user around here), I was willing to go along.
Perhaps I should set the default to what I think works better?  Rather than what it has been in the past?

Paul-Franklin:
Well, not only do I prefer a GUI option, I can recall bugs on the tracker
where users were asking to be able to change the font from within gramps.

Windows users in at least some cases.  So I certainly had that in the back
of my mind when I suggested the GUI interface.

Although I haven't downloaded it and tried it out, so I don't know
where in Edit=>Preferences it shows up.  Maybe it's not perfect yet?

In fact, since I am not sure how many font "backends" are
possible (in linux and Mac as well as Windows) I am not sure
about the "if win()" conditional -- although I will defer to Josip
or other people who understand such things.  But if there are
more than two (possible, I don't mean instantly available now,)
I would suggest that maybe the Boolean option might instead
be a combo box with named choices, to allow for future expansion.

p.s. I haven't tried to find mentions of font problems on the bug
tracker, but if people felt that was needed I suppose I could dig
them up.  Mind you, I am not saying this patch is the same thing,
but it seems to me it is a step in the right direction.

enlite93:
> p.s. I haven't tried to find mentions of font problems on the bug tracker,

Choosing default font through the preferences
https://gramps-project.org/bugs/view.php?id=3722

prculley:
I would prefer to keep this about the backend choice, not font selection.
Keep in mind we are talking about the code used to figure out what to do when the selected font does not include all the needed Unicode characters; the backend code has to choose an alternate font with as close a match to the selected font as possible.
At the moment, there seem to be three backends; 'coretext', used for MAC systems, 'fontconfig' ('fc' as synonym), and 'win32'.  Even though you would think that 'win32' would be the right choice for Windows systems, it appears that the 'fontconfig' does a better job for MY Windows system (and some others who have tried this change).
So this patch allow us to change from 'win32' to 'fontconfig' as an 'Alternate Font Handler', and only for Windows systems.  The GUI item shows up as the first line of the Edit/Preferences/General dialog box.  The code changes are pretty simple if you look at them.
If you look at [https://gramps-project.org/bugs/file_download.php?file_id=7779&type=bug](url) at the icon for the Tom Tester, you can see an example of what folks see when 'win32' is the choice.

Paul-Franklin:
"I would prefer to keep this about the backend choice, not font selection."

I will not argue with you about what might be considered a semantic
difference, especially one which most users will not understand.

I merely think the choice might be better represented as a list
of strings, than a Boolean 0/1.  So that if it is ever changed to
a list of strings later, for whatever reason, the user's "old" choice
will still be easily read, not have to be deduced, etc.

I am merely trying to say that ISTM my suggested way could
be easily expanded, should it ever become necessary.  If it
never becomes necessary, then the two choices will work fine.

As opposed to two Boolean choices which have to be made
three, should that contingency ever happen.  If if never does,
then of course Boolean choices will work fine too.

Do what you want.  It's your patch.

prculley:
Not sure how to write tests that work for code that only runs in Windows env...

PR #209, opened 2016-08-01
update da.po from attached patch file in bug 9609

prculley:
[https://gramps-project.org/bugs/view.php?id=9609](url)


romjerome:
Done. Thanks!

PR #210, opened 2016-08-01
Bug9515

kulath:
Fix for 0009515: DBAPI: gramps.plugins.database.dbapi_support.sqlite.Sqlite.try_execute ignores errors. Check if the table 'person' exists to determine whether the database is new (the check is different for the different DBAPI backends). This removes the need for try_execute.

Also improve pylint score for sqlite.py to 10.0

I have tested the changes to sqlite by opening an existing sqlite database and by creating a new one. This tests both outcomes of the new method table_exists. I haven't been able to provoke an exception in gramps.plugins.db.dbapi.dbapi.DBAPI.create_secondary_indexes_table (for creating a table when it already exists).

Similar tests need to be run on mysql.py and postgresql.py

### DIFF
--- a/gramps/plugins/db/dbapi/postgresql.py
+++ b/gramps/plugins/db/dbapi/postgresql.py
@@ -96,7 +96,7 @@ class Postgresql:
 
     def table_exists(self, table):
         self.cursor.execute("SELECT COUNT(*) FROM information_schema.tables "
-                            "WHERE table_name='%s';" % table)
+                            "WHERE table_name=?;", [table])
         return self.fetchone()[0] != 0
 
     def close(self):
--- a/gramps/plugins/db/dbapi/postgresql.py
+++ b/gramps/plugins/db/dbapi/postgresql.py
@@ -94,15 +94,10 @@ class Postgresql:
     def rollback(self):
         self.connection.rollback()
 
-    def try_execute(self, sql):
-        sql = self._hack_query(sql)
-        sql = sql.replace("BLOB", "bytea")
-        try:
-            self.cursor.execute(sql)
-        except Exception as exc:
-            self.cursor.execute("rollback")
-            #print("ERROR:", sql)
-            #print(str(exc))
+    def table_exists(self, table):
+        self.cursor.execute("SELECT COUNT(*) FROM information_schema.tables "
+                            "WHERE table_name='%s';" % table)
+        return self.fetchone()[0] != 0
 
     def close(self):
         self.connection.close()
--- a/gramps/plugins/db/dbapi/dbapi.py
+++ b/gramps/plugins/db/dbapi/dbapi.py
@@ -167,14 +167,18 @@ class DBAPI(DbGeneric):
             code = compile(file.read(), settings_file, 'exec')
             exec(code, globals(), settings)
         self.dbapi = settings["dbapi"]
-        self.update_schema()
+
+        # We use the existence of the person table as a proxy for the database
+        # being new
+        if not self.dbapi.table_exists("person"):
+            self.update_schema()
 
     def update_schema(self):
         """
         Create and update schema.
         """
         # make sure schema is up to date:
-        self.dbapi.try_execute("""CREATE TABLE person (
+        self.dbapi.execute("""CREATE TABLE person (
                                     handle    VARCHAR(50) PRIMARY KEY NOT NULL,
                                     given_name     TEXT        ,
                                     surname        TEXT        ,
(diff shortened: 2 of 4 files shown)


gramps/plugins/db/dbapi/postgresql.py
```diff
@@ -94,15 +94,10 @@ def commit(self):
...
-        sql = sql.replace("BLOB", "bytea")
-        try:
-            self.cursor.execute(sql)
-        except Exception as exc:
-            self.cursor.execute("rollback")
-            #print("ERROR:", sql)
-            #print(str(exc))
+    def table_exists(self, table):
+        self.cursor.execute("SELECT COUNT(*) FROM information_schema.tables "
+                            "WHERE table_name='%s';" % table)
```
dsblank: Don't use % on sql statements; use ? and follow with args (see other examples of execute). Otherwise, you can end up with bad errors and SQL injection hacks.

dsblank:
Other than the one comment I made about % vs ?, I think that this is an improvement. We can probably come up with a better check to see if a table exists for all DB-API implementations, but this will work for now.

QuLogic:
I don't think Postgresql uses `?`, but `%s`: http://initd.org/psycopg/docs/usage.html#the-problem-with-the-query-parameters
However, sqlite _does_ use `?`, but that line has not been corrected yet.

kulath:
Sorry, I'm not sure what more to do.

I can, and have, tested sqlite.py. I don't see that it makes any difference whether sqlite uses ?, because the %s is replaced before sqlite3 ever sees it.

I have changed Postgresql, because that was requested, but again, I don't see that it makes any difference because the %s is replaced before postgresql sees it, Also, the code in Gramps postgresql.py.execute has code to replace ? or %s (I forget which), but I don't call that code, I just call the underlying postgresql.

The existing code in Grapms dbapi.py uses both %s and ? apparently fairly randomly.

Could someone please test the latest change I have made?

Tim.

dsblank:
@QuLogic The ? is a DB-API execute standard. If one is going to be replacing standard SQL items (table names, field names, field values) then one should use ? and [values].

QuLogic:
There are multiple replacement options; it depends on the backend, and can be found using the [paramstyle attribute](https://www.python.org/dev/peps/pep-0249/#paramstyle). Psycopg2 uses 'pyformat', not 'qmark'.

dsblank:
@QuLogic Yes, I know. I wrote the three different backends to use '?' as the standard and then replace it with '%s' where necessary.

dsblank:
@kulath I think we should go ahead and commit this, and continue to refine.

QuLogic:
I think you're referring to [this `_hack_query` function](https://github.com/gramps-project/gramps/blob/c217eaae4d574e75636c44784ec317cee3c44af6/gramps/plugins/db/dbapi/postgresql.py#L47), I don't think that applies, because [this line](https://github.com/kulath/gramps/blob/c217eaae4d574e75636c44784ec317cee3c44af6/gramps/plugins/db/dbapi/postgresql.py#L98) is run directly on the cursor.

dsblank:
@QuLogic Ah, I see what you are saying. I'll test both on mysql and postgresql when this is applied and adjust as needed.

Paul-Franklin:
It is not clear to me what the status of this is.

Doubtless partly (wholly?) because I don't know SQL.

Are all three of you satisfied?  Has there been a resolution?
Is more work needed?  Or only a commit?  Etc.

Thanks.

p.s. I care because this is one of the gating items for the
5.0.0-alpha2 release.

dsblank:
I think this is ready for merging. I'll test mysql and postgresql after merge, and refine from there.

Paul-Franklin:
OK.

So are one of you (the originator, the reviewer) going to commit it,
or shall I?

Thanks.

kulath:
Paul,

I'd be delighted if you would commit it; I think it might need to be rebased before being commited, and I would have to research how to do that.

Thanks,

Paul-Franklin:
I didn't think it needed any rebase, since the "history" of the
four files showed the last changes were before your 8/1 pull
request.  So I just hit the "merge" button here (after having
turned on their stupid Javascript momentarily).

kulath:
Paul,

Unfortunately, just merging without a rebase creates a rather convoluted github history. As you can see from the attached, instead of simply adding my three changes 3461316, fcf1c1d and c217eaa on top of the old master head (d89e79b), it creates a loop in the history starting at 4f1eed7. At least I think that a rebase is supposed to prevent this sort of complex history, but I could be wrong.

[I have lots of sympathy for you, because we both found getting to grips with git very hard!]

github history

Paul-Franklin:
Well, I'm sorry it didn't work out.

OTOH you are ahead of me since I haven't the faintest idea how to
create such a list-chart as you just posted.

But If either one of you had committed it more than a week ago,
when the discussion apparently finished, I wouldn't have worried
that nothing was happening and stuck my oar in.   8-)   8-(

QuLogic:
There's nothing wrong with splits in the graph; that's how git is designed.

PR #213, opened 2016-08-02
Support for FTM and others Custom Gedcom Event Tags on import

prculley:
This PR provides support for Custom Event Tags used in Gedcoms.  These tags are typically of the form '_xyz' and are attached to Individuals and Families.  Gedcom allows this type of extension and it has been extensively used by a number of programs.  The PR converts the ones I have been able to identify to 'friendly' names and stores them as Gramps Events with Custom Event types.
The PR also supports tags I have not identified by storing the tag itself ('_xyz') as the Gramps Custom Event type (currently Gramps drops this information).
The '_JUST' tag is also supported, it is Source Citation 'Justification'; I stored this as a comment with the Gramps custom comment name "Citation Justification".

Note: I considered adding these tags to Gramps standard tags lists, but decided not to for now.  

I will commit this myself after a few days unless comments suggest changes.

### DIFF
--- a/data/tests/imp_CustTags.ged
+++ b/data/tests/imp_CustTags.ged
@@ -1,0 +1,82 @@
+0 HEAD
+1 SOUR FTM
+2 VERS Family Tree Maker (22.0.0.1410)
+2 NAME Family Tree Maker for Windows
+2 CORP Ancestry.com
+3 ADDR 360 W 4800 N
+4 CONT Provo, UT 84604
+1 DEST GED55
+1 DATE 01 MAR 2016
+1 CHAR UTF-8
+1 FILE D:\Family Tree Maker\imp_FTM_LINK.ged
+1 SUBM @SUBM@
+1 GEDC
+2 VERS 5.5
+2 FORM LINEAGE-LINKED
+0 @SUBM@ SUBM
+1 NAME The Subm /Tester/
+1 ADDR 123 Main St., Winslow, PA, 12345
+2 ADR1 123 Main St.
+2 CITY Winslow
+2 STAE PA
+2 POST 12345
+0 @I0@ INDI
+1 NAME The /Tester/
+2 SOUR @S6@
+3 PAGE 777, record for The Tester
+3 _JUST Because it’s good
+1 BIRT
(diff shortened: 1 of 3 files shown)


sam-m888:
Works really well, tested using your sample gedcom and my enhanced TMG GEDCOM (about 50k custom tags!) and both work.

I'll accept this if you rebase the conflicting line "+TOKEN__JUST = 134" because of your previous change in  https://github.com/gramps-project/gramps/blame/ed4c9fa14865926cc6ed8059f627c59e134791da/gramps/plugins/lib/libgedcom.py#L272 I believe the number 134 is not significant as you use it for "TOKEN__PRIM = 134"


```
diff a/gramps/plugins/lib/libgedcom.py b/gramps/plugins/lib/libgedcom.py	(rejected hunks)
@@ -269,6 +269,7 @@
 TOKEN__FSFTID = 131
 TOKEN__PHOTO = 132
 TOKEN__LINK = 133
+TOKEN__JUST = 134
 
 TOKENS = {
     "HEAD"         : TOKEN_HEAD,    "MEDI"         : TOKEN_MEDI,

```

sam-m888:
Just reread your comment and now see that I'd have to add TMG's custom tags to [CUSTOMEVENTTAGS](
https://github.com/gramps-project/gramps/pull/213/files#diff-5b5a55dbd3970ca1b521eb2300312c36R567)

prculley:
Sam and others; if you have a list of other custom Gedcom tags, as used by modern software (there are many older softwares with a lot of tags) and a good definition of their usage and syntax, I would be happy to see about getting them incorporated.  Keep in mind that the ones fixed in this PR are (mostly) event type tags, some of the other tags (_DATE, _SDATE) are probably not events, and would need additional understanding on my part to deal with them properly.

sam-m888:
Hi Paul, 

Thanks for the clarification, I'll do a bit more research on those custom GEDCOM tags for TMG.  

In the meantime if you have a moment please rebase the conflicting line, so this PR can be accepted.

sam-m888:
Thanks you

PR #214, opened 2016-08-02
Bug 8128 for importgedcom and libgedcom; dialog parent=

prculley:
Modified and both patches tested to work on Windows.


Paul-Franklin:
Thanks.

PR #220, opened 2016-08-23
Fixes for the PHON, FAX, EMAIL and WWW Gedcom tags to support Gedcom v5.5.1.

prculley:
This PR updates support for the PHON, FAX, EMAIL and WWW Gedcom tags to support Gedcom v5.5.1.

For the Gedcom SUBM section, when the preference "Add default source on Gedcom import" is true, the FAX and WWW tags are **added** to the internet tab of the added repositories; 'SUBM (Submitter)'.  The EMAIL continues to be stored on the internet tab. The PHON tag continues to be present on the Address tab for that repository.  The 'Database owner' continues to have only the EMAIL and PHON data.  May want to fix this when we next look at data model.

For the Gedcom CORP section, when the preference "Add default source on Gedcom import" is true, the FAX, EMAIL, and WWW tags are **added** to the internet tab of the added repository; 'Business that produced the Gedcom'. The PHON tag continues to be present on the Address tab for that repository.

For normal repositories, the Gedcom FAX number is **added** to the repository internet tab.

Gedcom allows up to three PHON (and FAX, EMAIL, WWW) entries.  Since we are currently storing Phone numbers in the Address tab for repositories, only a single (**first found**) phone number is supported.  **Added a warning** message if additional phone numbers are found.  FAX, EMAIL, WWW tags stored on internet tab support multiple entries.

For events, the PHON tag is **moved** from the attached Place's Code field to an Attribute of the Event.  The FAX tag is **added** as an Attribute of the event, the EMAIL and WWW tags remain as attributes of the event.  These Gedcom tags need NOT be part of an address, they will be recognized even if they stand alone.

When the FAX (**new**), PHON, EMAIL and WWW (currently supported) Gedcom tags are attached directly to a person (against Gedcom standard), they is stored in the Internet tab for that person.  An Address attached directly to a person (against Gedcom standard), continues to be stored in the Address tab for that person.

I also had to **change** the test_imports and gen/merge/diff modules to support testing with 'Add Gedcom default source on import' preference turned on.  Nice side effect of testing some more lines of libgedcom code.

I also **changed** test_imports to export the '.gramps' file for any failing test to assist the developer in comparing the expected and failing files for diagnostics.


sam-m888:
Thank you

PR #221, opened 2016-08-26
Fixes for the PHON, FAX, EMAIL and WWW Gedcom tags to support Gedcom v5.5.1.

prculley:
This PR updates support for the PHON, FAX, EMAIL and WWW Gedcom tags to support Gedcom v5.5.1.
Same as PR220 except for Gramps42 branch. 

Note: I argue that this is enough of a 'bug' to be considered for Gramps42 branch, others may differ.

For the Gedcom SUBM section, when the preference "Add default source on Gedcom import" is true, the FAX and WWW tags are **added** to the internet tab of the added repositories; 'SUBM (Submitter)'. The EMAIL continues to be stored on the internet tab. The PHON tag continues to be present on the Address tab for that repository. The 'Database owner' continues to have only the EMAIL and PHON data. May want to fix this when we next look at data model.

For the Gedcom CORP section, when the preference "Add default source on Gedcom import" is true, the FAX, EMAIL, and WWW tags are **added** to the internet tab of the added repository; 'Business that produced the Gedcom'. The PHON tag continues to be present on the Address tab for that repository.

For normal repositories, the Gedcom FAX number is **added** to the repository internet tab.

Gedcom allows up to three PHON (and FAX, EMAIL, WWW) entries. Since we are currently storing Phone numbers in the Address tab for repositories, only a single **(first found)** phone number is supported. **Added a warning message** if additional phone numbers are found. FAX, EMAIL, WWW tags stored on internet tab support multiple entries.

For events, the PHON tag is **moved** from the attached Place's Code field to an Attribute of the Event. The FAX tag is added as an Attribute of the event, the EMAIL and WWW tags remain as attributes of the event. These Gedcom tags need NOT be part of an address, they will be recognized even if they stand alone.

When the FAX **(new)**, PHON, EMAIL and WWW (currently supported) Gedcom tags are attached directly to a person (against Gedcom standard), they are stored in the Internet tab for that person. An Address attached directly to a person (against Gedcom standard), continues to be stored in the Address tab for that person.

### DIFF
--- a/gramps/plugins/lib/libgedcom.py
+++ b/gramps/plugins/lib/libgedcom.py
@@ -262,7 +262,7 @@ TOKEN_LONG = 122
 TOKEN_FACT = 123
 TOKEN_EMAIL = 124
 TOKEN_WWW = 125
-TOKEN_URL = 126
+TOKEN_FAX = 126
 TOKEN_ROLE = 127
 TOKEN__MAR = 128
 TOKEN__MARN = 129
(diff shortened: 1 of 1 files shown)


enlite93:
> Note: I argue that this is enough of a 'bug' to be considered for Gramps42 branch, others may differ.

Yes, I'd argue that also..

Will this make it into [Gramps 4.2.4](https://gramps-project.org/wiki/index.php?title=Template:Releases/4.2.4) as I don't see it listed, it would be sad if it didn't?

sam-m888:
Matching PR https://github.com/gramps-project/gramps/pull/220 for Gramps master:

PR #223, opened 2016-08-29
Bug 9665 Gedcom import exception processing fixes

prculley:
This PR changes the methods to always use the exception line's Gedcom 'level' as the indicator to skip any lexically subordinate lines.  For the widely used '__ignore()' method, It also performs a check on the token for the line and modifies the message to indicate "not supported", versus "not understood", the distinction being that valid Gedcom may generate "not supported" messages, and invalid (or customized Gedcom) would generate "not understood".
A call to __not_recognized() that was not executable was removed.

The exception processing for the Gedcom parser (when unexpected Tags or defined tags in unexpected places are found) should report the error, and skip any lexically subordinate lines (as these would be dependent on the exception line to make sense).
Current code tries to track the 'state' of the Gedcom processing (which is mostly the same as the lexical level).  But not always.  In the cases where the lexical level and state level differ, the authors have tried to compensate by manually adjusting the level used in exception processing.  In a few cases they did not get it right.

### DIFF
--- a/data/tests/imp_notetest_dfs.ged
+++ b/data/tests/imp_notetest_dfs.ged
@@ -1,0 +1,227 @@
+0 HEAD
+1 SOUR GEDitCOM
+2 NAME GEDitCOM
+2 VERS 2.9.4
+2 CORP RSAC Software
+1 SUBM @SUBMITTER@
+1 SUBN @SUBMISSION@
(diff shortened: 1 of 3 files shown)


prculley:
An unrelated commit has whitespace errors causing travis to fail...

PR #224, opened 2016-08-29
Copy data.gramps to test directory and point db_test at it

prculley:
This PR makes a copy of the example/gramps/data.gramps and example/gramps/example.gramps files into the tests directory and shifts the tests to use them, so that changes can be made to the example directory files without crashing the tests.

Note there are still a few other references to the example/gramps/data.gramps file in other tests, however they appear to be pretty much ignoring the content, so should not effect the results.

Requested by Paul Franklin...


prculley:
An unrelated commit has whitespace errors causing travis to fail...

PR #225, opened 2016-08-29
Translation of new and fuzzy strings.

paulohpmoraes:
Translated the new strings, revised fuzzy tranlatons and changes some terms and translations to correct mispelled words, wrong translations and uniformization.


romjerome:
Thanks!

PR #226, opened 2016-08-29
Translated untranslated and fuzzy strings.

paulohpmoraes:
Translated untranslated and fuzzy strings. Some changes, corrections and uniformization.


romjerome:
Thank you!

I will merge it.

PR #227, opened 2016-08-30
0009626: Google map API now requires a key

phejl:


prculley:
Do we have any wiki help on the use of this 'key'?  Or where/how to get it?

phejl:
I'm not aware of a wiki. Should it be created? Instructions on getting a key: https://developers.google.com/maps/documentation/javascript/get-api-key

prculley:
Just trying to think like a user; I certainly would not be able to guess where to get such a thing based on the Gramps dialog alone...
`
        self.__googlemapkey = StringOption(_("Google maps API key"),"")
        self.__googlemapkey.set_help(_("The API key used for the Google maps"))
`
I would think that few users will ever figure this out without some detailed instructions on our part; and maybe not then (I still have not figured it out after 15min of looking at Google's help files).  I could not even get their 'Cloud Platform Console' to open on my Firefox browser, the first time I tried.  The second time with Chrome browser worked but put up a page full of options that I did not guess where to go next.  The first thing on the list talks about a "$300 credit and 60 days...", not exactly encouraging.

phejl:
You are certainly right. Honestly I'm not sure if user without any knowledge of web technologies will ever be able to succeed. The same would apply to "character set encoding" setting. I doubt user without html/web background know what's going on. Such inexperienced user has the option to not to touch anything as the default is OpenStreetMap now.

For the rest of it. Have you followed the doc? The first link of "Get an API key" leads to Google API Console (not Cloud Platform Console) where is just field with project name and create button. In second step there are two fields with key name and allowed urls and on the third click you get the key.

phejl:
Certainly we may create a wiki for Google map usage.

eno93:
Or just stick with the Free OpenStreetMap

phejl:
That's the new default so you don't have to do anything to generate proper web map. The Google is an option for users who prefer it.

sam-m888:
Tested, works well.

New default of "OpenStreetmap"
narweb-placemapoption-defaults

With Google selected
narweb-placemapoption-google-selected

 [ ] Instructions needed to get **Google Maps Javascript API key** (Possibly change tooltip to match Googles description of what the key is? ) https://developers.google.com/maps/documentation/javascript/get-api-key
get a key_authentication-google maps javascript api

phejl:
I'm afraid there is no way to do a real validation. We might check length and allowed characters (if we are able to find out allowed ones).

prculley:
Perhaps we could include a link to the appropriate Google page (or our wiki on the topic) in the box somewhere.  Don't know if you can put it in the tooltip, but that would be nice.

phejl:
I'll try to figure it out.

uxcq:
> I'm afraid there is no way to do a real validation. We might check length and allowed characters (if we are able to find out allowed ones).

I'd say don't bother and just mention it in the manual, because if people really want to use Google they will work it out.

sam-m888:
> @phejl I'd say don't bother and just mention it in the manual, because if people really want to use Google they will work it out.

This is probably the best compromise. I'll merge the change.

PR #229, opened 2016-09-02
Gedcom import, remove old, dead, code, no longer used

prculley:
About 6 years ago, a Gedcom import customization file (gedcom.xml) was removed from the distribution.  Since that time, this code has been unused.  It originally parsed the file and allowed various changes to Gedcom handling based on the results.
We currently have better industry wide Gedcom standard support, and at the moment, the few customizations still needed, based on who wrote the Gedcom, are handled directly by the import code.

This will improve the test coverage, as well as reduce the very large line count on the libgedcom module.

### DIFF
--- a/gramps/plugins/lib/libgedcom.py
+++ b/gramps/plugins/lib/libgedcom.py
@@ -1066,207 +1066,6 @@ _MAP_DATA = {
 
 #-------------------------------------------------------------------------
 #
-# GedcomDescription
-#
-#-------------------------------------------------------------------------
-class GedcomDescription:
-    def __init__(self, name):
-        self.name = name
-        self.dest = ""
-        self.adopt = ADOPT_STD
-        self.conc = CONC_OK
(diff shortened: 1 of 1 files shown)


dsblank:
Looks good to me! Thanks!

gramps/plugins/lib/libgedcom.py
```diff
@@ -5574,15 +5371,11 @@ def __event_type(self, line, state):
...
-                    name = EventType((EventType.CUSTOM, val))
-                else:
-                    try:
-                        name = EventType((EventType.CUSTOM,
-                                                 line.data))
-                    except AttributeError:
-                        name = EventType(EventType.UNKNOWN)
+                try:
+                    name = EventType((EventType.CUSTOM,
+                                             line.data))
```
QuLogic: Can you fix the alignment here? It might even fit on one line now.

PR #230, opened 2016-09-03
9642 Adjust GedcomStageOne notes Parsing

sam-m888:
Minor fix.

In libgedcom.py's class GedcomStageOne Parsing of Tags noticed that the zero level NOTE tag is the only one that leaks its content.

Which is probably not what is desired, especially when:
* you have alot of NOTE's in your GEDCOM file

and the whole purpose of GedcomStageOne is a quick first pass over the GEDCOM!

https://gramps-project.org/bugs/view.php?id=9462

### DIFF
--- a/gramps/plugins/lib/libgedcom.py
+++ b/gramps/plugins/lib/libgedcom.py
@@ -8122,7 +8122,7 @@ class GedcomStageOne:
             self.lcnt += 1
 
             try:
-                data = line.split(None, 2) + ['']
+                data = line.split(None, 3) + ['']
                 (level, key, value) = data[:3]
                 level = int(level)
                 key = key.strip()


prculley:
Not sure why you think this was bad, but doesn't seem to effect anything (all my tests pass).  For my own curiosity, I tried to see if it made any difference to the memory footprint or timing.  I loaded my own Gedcom (22,017 notes).  The memory footprint was actually slightly larger with your fix (171.7MB vs 171.4MB, probably just normal variation).
The run time was .1% longer as well with your fix, also probably within normal variation (I ran it three times each way and averaged the results, after dropping the first since it would have included python compile time.  The variations between runs was around 0.22%...

prculley:
More fun (geek for sure);
Just tried using a few strategically placed print(time.time()), and report % of total time for import for:
0.006% Stage 0
1.4%     Stage 1
93.08% Stage 2
3.3%     Cleanup
2.16%   Cross reference check
Without your patch.
With your patch, the stage 1 went to 1.405% (9ms longer out of 16544ms total).

sam-m888:
@prculley Very interesting, thanks for the speed check.

I found the issue while I was looking to see if I could implement [GEDCOM STS](http://www.tamurajones.net/GEDCOM_STS.xhtml)  and had to work around this issue, so it tripped me over! 

If correctness is'nt good enough then I' don't know!  Either way at the end of the day accept or don't accept this.

uxcq:
> If correctness is'nt good enough then I' don't know! 

You noticed it and reported it, that more than alot of people do, good to see someone reading the code. It all helps improve Gramps.

uxcq:
> With your patch, the stage 1 went to 1.405% (9ms longer out of 16544ms total).

@prculley 9ms longer! for the for the sake of correctness... :+1:  add those 9ms!

prculley:
>  Either way at the end of the day accept or don't accept this.

Seems odd for the newbie developer to be accepting PRs, but as this doesn't have much user noticeable effect, here goes...

PR #231, opened 2016-09-04
Gedcom import Notes get NoteType appropriate to object

prculley:
Gramps has NoteTypes available for Notes; this fix change the Gedcom import to utilize them (They were mostly set to 'General' previously, now set to 'Media', Person', Family' etc.).
Also some refactoring; I modified the __media_ref_note and __event_inline_note processing to utilize the general __parse_note code, rather than having code (and fix) duplication.

__parse_note also was modified to eliminate a no longer used input parameter (level).

Added Gedcom SUBM/NOTE support; note added to the default source (when default source is enabled via preferences).

For Gedcom CHAN/NOTE (Notes on a Change time which we cannot support), modified the code to create the appropriate error message, instead of dropping the note entirely.

### DIFF
--- a/data/tests/imp_CustTags.gramps
+++ b/data/tests/imp_CustTags.gramps
@@ -3,7 +3,7 @@
 "http://gramps-project.org/xml/1.7.1/grampsxml.dtd">
 <database xmlns="http://gramps-project.org/xml/1.7.1/">
   <header>
-    <created date="2016-08-02" version="5.0.0-alpha1"/>
+    <created date="2016-08-29" version="5.0.0-alpha1"/>
     <researcher>
       <resname>The Subm /Tester/</resname>
       <resaddr>123 Main St.</resaddr>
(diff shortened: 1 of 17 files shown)


uxcq:
Always a pleasure to see GEDCOM support improve, Thanks :+1: 

Any idea when Gramps 5.0.0 will be released?

Paul-Franklin:
Always a pleasure to see more volunteers to help gramps improve.
8-)

uxcq:
@Paul-Franklin Your welcome, I do as much as I can :+1:

sam-m888:
@uxcq  _Any idea when Gramps 5.0.0 will be released?_

You can follow the discussion on the mailing list:
http://gramps.1791082.n4.nabble.com/4-2-5-and-5-0-0-td4676792.html

sam-m888:
Thank you.

PR #232, opened 2016-09-05
9676 CSV import fails for certain types of cross references

prculley:
The CSV wiki [https://gramps-project.org/wiki/index.php?title=Gramps_4.2_Wiki_Manual_-_Manage_Family_Trees:_CSV_Import_and_Export]
shows that a user can use cross references within the CSV file to refer to persons, places, and families. However, for some formats of the cross references, the CSV import fails.
The formats that are enclosed in "[ ]" (like "[I0001]" or "[F0003]" work fine. Formats that look like "I1" fail.
It appears that the cross reference definitions are stored temporarily after being processed to make them look like Gramps IDs, if they look something like Gramps IDs to start with.
When searching for these IDs later, the code DOES NOT do the same processing to make them look like Gramps IDs. So searches fail when IDs are close enough to GIDs to be modified ('I1', 'F01', 'P27' etc.) and pass when the xrefs look like something else ('a1', 'xyz' etc.).

A secondary bug is that the code mistakenly processed place IDs with the person ID code. So I changed "self.db.id2user_format(id_)" to "self.db.pid2user_format(id_)" in a couple of places.

A third bug was that some case variations of the CSV row titles were not recognized. So I changed the dict that contains row titles to reduce redundancy and adjusted its access to always user lower case.


dsblank:
Great fixes! Why do the test CSV files have all of the commas (empty columns)? They should not be needed, and no human would write those.

prculley:
The extra commas are a side effect of using Microsoft Excel to edit the CSV.  If there is ever any data in those columns, then Excel adds the comma.  Since the 'Person' row has the most entries, the other rows are extended to match the number of columns.

dsblank:
@prculley I see; thanks for the explanation. I just was making sure that that wasn't a new requirement, and that it will work without them.

prculley:
Seems that my attempt to switch to case agnostic comparison on CSV row header strings was a mistake.  Apparently some of the translations are different (in at least one language) for different case variations.
In the Dutch translation "husband" -> "echtgenoot", but "Husband" -> "Man", and a similar issue for "wife".  So I put back the translatable strings that had alternate cases.

sam-m888:
Thank you

PR #235, opened 2016-09-08
bug 9682 fix Check & repair so it won't crash with bad 'get_from_handle'

prculley:
Was trying to add a new test case to test case generator (TCG) and new fix to Check & Repair (C&R) when I discovered that C&R crashes on db created by TCG when 'Generate database errors' option was used.
A commit [https://github.com/gramps-project/gramps/commit/5c958bd7fbaea082df6f2ce427042ad0a642815e](url) done Dec 2015 changed the way the Db '_get_from_handle()'  responds to an invalid handle; it used to return a 'None', now it causes an exception.
Since TCG creates invalid handles, the C&R tool crashed on the exception.
This PR uses the 'try: ... except:' syntax to catch the exceptions and handle them as the original code did when it got a 'None' return.

There doesn't seem to be any easy way to test C&R code, so I manually tweaked TCG on master branch to include a 'seed()' on the random number generator and used my 'Deterministic ID' plugin so it would produce the same result each time.  Then tweaked the db code so it would return 'None' again, to create and check a TCG db from master branch.
I then cleared the changes to Db (so it would produce exceptions), added the 'seed()' to TCG in my test branch and repeated the create and check a TCG db.  Finally, I compared the C&R plugin outputs from both the original slightly tweaked master, and the PR version. I also checked an xml export from each against each other; they matched except for change dates.
P.S. I'm going to have to figure out a way to freeze these change dates for testing; they make manually comparing .gramps files quite annoying...

### DIFF
--- a/gramps/plugins/tool/check.py
+++ b/gramps/plugins/tool/check.py
@@ -71,6 +71,7 @@ from gramps.gui.plug import tool
 from gramps.gui.dialog import OkDialog, MissingMediaDialog
 from gramps.gen.display.name import displayer as _nd
 from gramps.gui.glade import Glade
+from gramps.gen.errors import HandleError
 
 # table for handling control chars in notes.
 # All except 09, 0A, 0D are replaced with space.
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Yes.  In previous versions a broken link did not cause an exception to be thrown.  Users were left unaware about potential database corruption.

In most cases the exception can be caught in top level code and a suitable dialog presented to the user.  This should recommend running the Check and Repair tool.

I would expect the Check and Repair tool to catch these exceptions.  The alternative would be to check that the handle exists using one of the "has handle" database methods.

PR #236, opened 2016-09-14
bug 9697, fix Test Case Generator to avoid GUI progress meter in CLI mode

prculley:
Used the user.progress version of progress meter instead of GUI GTK progress meter.  That way it works right in either GUI or CLI mode.


sam-m888:
Thank you

PR #237, opened 2016-09-15
Update Check & Repair tool & Test Case Generator tool to fix bad alternative Place names

prculley:
The Check & Repair (C&R) tool has been enhanced to fix Place alternate names entries that are blank, or duplicates of each other or primary name.

In order to test the modified C&R, the Test Case Generator (TCG) has been enhanced to create bugs in alt place names like those fixed in C&R above.  I used this to manually create a db and then watch C&R correct the created bugs.

I also used an old version of Gramps (4.2.2) to generate a db with these bugs based on an import of a Gedcom file that exercised the place code (and found the bugs to begin with).  And used C&R to fix them while under observation.

TCG contains some quite useful code already in that it creates bugs that C&R can fix.  In order to use TCG as a test tool, however, it needs to be repeatable.  TCG uses a lot of random number generation to create test cases, names etc.  So I modified TCG to use a private copy of the random number generator, and to expose that random generator where a unit test code can find it to arrange a seed.  Under normal conditions, TCG is not seeded, so remains as random as usual.

I also did a very little pylint cleanup on lines I was already modifying, so the diff shows more in-a-line changes than the search & replace for random/randint/choice accounts for.

I also created a unittest module 'test_tools.py' which currently uses TCG to create a database that C&R can repair.  This module loads a small '.gramps' file to start things off, then seeds TCG's random generator for repeat-ability, and uses TCG to create the db.  Finally I use Gramps again to run C&R.  At each step, expected results are captured and compared.

P.S. I expect that this PR will conflict with an earlier TCG PR (236); I will rebase and correct after that one is accepted.

### DIFF
--- a/gramps/plugins/test/test_tools.py
+++ b/gramps/plugins/test/test_tools.py
@@ -1,0 +1,168 @@
+#! /usr/bin/env python3
+"""
+Gramps - a GTK+/GNOME based genealogy program
+
+Copyright (c) 2016 Gramps Development Team
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 2 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
+"""
+
+import os
+import sys
+import codecs
+import unittest
+import random
+
+from gramps.test.test_util import Gramps
+from gramps.gen.const import DATA_DIR
+from gramps.cli.user import User
+from gramps.gen.utils.id import set_det_id
+from gramps.gen import const
+from gramps.gen.utils.config import config
+
+TREE_NAME = "Test_tooltest"
+TEST_DIR = os.path.abspath(os.path.join(DATA_DIR, "tests"))
+const.myrand = random.Random()
+
+
+def call(*args):
+    """ Call Gramps to perform the action with out and err captured """
+    if __debug__:
+        print ("call: %s", args)
+    gramps = Gramps(user=User(auto_accept=True, quiet=True))
+    # gramps = Gramps(user=User(auto_accept=True))
+    out, err = gramps.run(*args)
+    # print("out:", out, "err:", err)
+    return out, err
+
+
+def check_res(out, err, expected, do_out=True):
+    """
+    compare the output with the expected
+    """
+    # return all([(line in out) for line in expect])
+    retval = True
+    for line in expected:
+        if line in (out if do_out else err):
+            continue
+        else:
+            print("*** Expected: '%s', not found" % line)
+            retval = False
+    if not retval:
+        if do_out:
+            print("*** The following was the actual stdout:\n%s"
+                  "*** The following was the stderr:\n%s"
+                  "*** End of stderr" % (out, err))
+        else:
+            print("*** The following was the actual stderr:\n%s"
+                  "*** The following was the stdout:\n%s"
+                  "*** End of stdout" % (err, out))
+    return retval
+
+
+class ToolControl(unittest.TestCase):
+    """
+    These tests run some of the tools used to maintain Gramps db
+    """
+    def setUp(self):
+        self.db_backend = config.get('database.backend')
+        call("--config=database.backend:bsddb", "-y", "-q", "--remove", TREE_NAME)
+
+    def tearDown(self):
+        config.set('database.backend', self.db_backend)
+        call("-y", "-q", "--remove", TREE_NAME)
+
+    def test_tcg_and_check_and_repair(self):
+        """
+        Run a 'Test Case Generator' and 'Check & Repair Database' test.
+        Note that the 'Test Case Generator" uses a lot of random numbers to
+        generate its test cases.  This makes it less than ideal for a
+        predictable unit test.  Still it contains good code for testing the
+        'Check and Repair' tool.  So I force the random functions to be
+        predictable by seeding it with a fixed number.  I also used the
+        'Deterministic ID' function to make the usual db handle generation
+        stop using random numbers and potentially reduce Gramps version to
+        version issues.
+        """
+        # the TCG creates bad strings with illegal characters, so we need to
+        # ignore them when we print the results
+        try:
+            sys.stderr = codecs.getwriter(sys.getdefaultencoding()) \
+                (sys.stderr.buffer, 'replace')
+            sys.stdout = codecs.getwriter(sys.getdefaultencoding()) \
+                (sys.stdout.buffer, 'replace')
+        except:
+            pass
+        tst_file = os.path.join(TEST_DIR, "data.gramps")
+        set_det_id(True)
+        # the following line assumes that TCG has run through init code, where
+        # it puts 'myrand', a 'Random' class object, into the 'const' module so
+        # we can access it here.
+        const.myrand.seed(1234, version=1)
+        # print(const.myrand.random())
+#         out, err = call("-s")
+#         expect = ["bsddb"]
+#         check_res(out, err, expect, do_out=True)
+        out, err = call("-C", TREE_NAME, "-q",
+                        "--import", tst_file)
+        expect = ["Opened successfully!",
+                  "data.gramps, format gramps.",
+                  "Cleaning up."]
+        self.assertTrue(check_res(out, err, expect, do_out=False))
+        self.assertEqual(out, "")
+        out, err = call("-O", TREE_NAME,
+                        "-y", "-q", "-a", "tool", "-p",
+                        "name=testcasegenerator,bugs=1,persons=0,"
+                        "add_linebreak=0,add_serial=0,"
+                        "long_names=0,lowlevel=0,person_count=20,"
+                        "specialchars=0")
+        expect = ["Opened successfully!",
+                  "Performing action: tool.",
+                  "Using options string: name=testcasegenerator,bugs=1",
+                  "Cleaning up."]
+        self.assertTrue(check_res(out, err, expect, do_out=False))
+        expect = ["person count 41"]
+        self.assertTrue(check_res(out, err, expect, do_out=True))
+        out, err = call("-O", TREE_NAME,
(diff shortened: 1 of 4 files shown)


prculley:
Looks like I have an issue that did not show up in my Windows or Linux tests, but shows up in Travis...

daleathan:
Could it be this error in the Travis output?

```
ERROR:.:Failed to start tool.
  File "/home/travis/build/gramps-project/gramps/gramps/gui/plug/tool.py", line 275, in cli_tool
    options_class = clt.option_class, name = name, callback = None)
  File "/home/travis/build/gramps-project/gramps/gramps/plugins/tool/check.py", line 186, in __init__
    checker.check_parent_relationships()
  File "/home/travis/build/gramps-project/gramps/gramps/plugins/tool/check.py", line 1024, in check_parent_relationships
    fgender = self.db.get_person_from_handle(father_handle
AttributeError: 'NoneType' object has no attribute 'get_gender'
```

prculley:
In response to daleathan;
Yes, that is the error.  The mystery is that it occurs under Travis, but not under either my Windows or Linux installations.  The 'NoneType' should not happen.  A commit [https://github.com/gramps-project/gramps/commit/5c958bd7fbaea082df6f2ce427042ad0a642815e] done Dec 2015 changed the way the Db '_get_from_handle()' responds to an invalid handle; it used to return a 'None', now it causes an exception.
I modified the C&R tool to deal with the exceptions, it works fine on my local installations, but for some reason, on Travis, I am getting 'NoneType'.  I've carefully looked over the code, and cannot see how this can occur, so I'm beginning to suspect that there might be some sort of bug in Python itself.  Travis is using Python 3.3.6, My Linux uses python 3.4.3 and my Windows uses 3.5.0.  I've been putting some extra logging statements into the code in an attempt to narrow the issue down, since I don't know how to use a debugger on Travis.
I'm also looking into how to add another version of python to my Linux, maybe I can get it debugged that way.

dsblank:
It could depend on the DB type. Currently DB-API returns None in this case, and BSDDB raises an exception.

prculley:
Ouch; That seems like a bug in DB-API.  Is there a way to tell what type of DB Gramps uses under Travis?  I thought we had set the default to BSDDB.

dsblank:
I really think that crashing in the middle of general code is a bad idea. But, if we really think that https://github.com/gramps-project/gramps/commit/5c958bd7fbaea082df6f2ce427042ad0a642815e is a good idea, we can make DB-API do the same. You could also check for both cases since we want check and repair to be robust. The default for Gramps 5.0 is BSDDB. But, if you change it, it will be the other. You can force it by setting the config.

prculley:
Well, when I noticed that C&R crashed on bsddb (with bad DB generated by TCG) because of the change from 'None' returns to exceptions, I submitted the PR that 'fixed' it to deal with exceptions.  I suppose I could figure out a way for C&R to deal with either, although I suspect it would be ugly, with a lot of code repetition.
But I really think that we should have all our DBs respond as nearly the same as possible, even to 'exceptional' conditions.  Perhaps we should take this to the list; it is a little off topic for my issue.

dsblank:
Let's just change DB-API for now.

prculley:
So how did Travis get set up to use db-api?  I just added a check in my code, and lo and behold, it IS using db-api!!!
Doesn't Travis start with the Gramps defaults every time?  Apparently not...

dsblank:
Perhaps some test code changes config?

prculley:
That is probably it; I guess I will have to get more defensive in my programming to make sure everything is set up the way I expect.  I'll look around to see if i can spot this and figure out a way to deal, at least for this test.

Will you fix the db-api to return the "HandleError" exception instead of 'None'?  Or should I submit a PR?  I'd like to get this fixed up in the next couple of days.

dsblank:
Yes, I'll fix the DB-API not raising an exception difference later today. Thanks!

Nick-Hall:
The database should raise a HandleError if a record is expected but not found.  This indicates a possible database corruption.  In such cases, the standard database error dialog should be presented to the user.

However, it is also acceptable for a developer to catch the exception and deal with it differently.  The check and repair tool is one example of where this may be useful.  In this case the developer knows the error is not real. For example, they may be checking the type of a handle, but there are normally better ways to do this.

prculley:
I suspect that db_test.py may have left Travis configured for dbapi.
Is there a way to ensure that unit test, nose etc. execute tests in a particular order?  If so, I would prefer that we set db_test.py to run the default DB last.
On another topic, we may want to figure out a way to run all of the unit tests with both (all?) of our supported DB backends.  Would help catch issues like this...

dsblank:
There are a few settings that can change that can effect output, and DB-API is just one. We may want to have a way of resetting config values to the "factory settings" to ensure known settings. We don't have that yet. Would be good to add something like "config.reset()".

Regarding testing both types of database backends:

1. The plan is to remove one of the backends, so we could put off testing for a bit
2. gramps/plugins/db/test/db_test.py does test both backend types

dsblank:
#239 Changes the way that DP-API handles getting objects where the handle doesn't exist. And it also adds code that catches one "error" in a filter. Please check to make sure that that exception to the condition is correct.

Nick-Hall:
It is best not to catch the HandleError in the filter.  This will hide broken links which are real errors and the user should be made aware of.  However, you do need to avoid getting the family from the database where fam_id is None.  We know that no family will exist in this case.

Instead, the error should be handled at a higher level.  For example, in filters, this will be around the model constructor in views where the database error dialog will be invoked to inform the user.

Of course, we hope to never find any broken links.

dsblank:
@Nick-Hall Ok, I reverted the try/except in the filter. I tried finding other examples of such handling of HandleError, but I couldn't find any. Do you want me to make an issue in the tracker? Or perhaps @prculley knows how to fix this? 

BTW, I think this "error" can happen very easily: simply make an export containing a person whose parent family isn't selected for export. That would imply that any code that tries to get that parent family is going to crash, or is protected by try/except.

Nick-Hall:
The HandleError should not be thrown unless there is a broken link in the database.  That should hopefully never happen, and was why the exception was introduced.

The filter still needs an if statement to catch the case when the handle is None.  This could be because the filter is run through a proxy or perhaps there is no parent family.

We should also check that the export code catches the error and presents a suitable error message.  This should be done through the User class because an export can be run from the CLI.

dsblank:
The filter rule error was just not ignoring a None handle. Fixed that, which made all of the tests pass. #239 has been merged. Thanks!

dsblank:
@prculley I added a config.reset() in #240 to get the config back to original settings. We should call this before any operation that could have different outcome based on changed settings. There are at least three settings that can be set by users affecting imports:

1. default marriage type
2. add tags on import
3. database backend type

Example:

```python
config.reset()
```

prculley:
Thanks guys; I appreciate the timely fixes, you work much faster than I do.  I will start getting my code cleaned up to use your fixes now.

prculley:
When dbapi is in use, is there some kind of automatic backup that occurs when Gramps is closed?
I'm seeing a cli percentage count in my output that is not there under bsddb.  There is another difference I see that I don't understand as well, more debugging...

dsblank:
On Thu, Sep 22, 2016 at 5:52 PM, Paul Culley <notifications@github.com>
wrote:

> When dbapi is in use, is there some kind of automatic backup that occurs
> when Gramps is closed?
> I'm seeing a cli percentage count in my output that is not there under
> bsddb. There is another difference I see that I don't understand as well,
> more debugging...
>
Yes, as a precaution, DB-API makes a backup after each session where
something changes. We're being careful with this, and want to make sure we
don't lose anything, and that we have info for tracking down any losses.

dsblank:
@prculley So, I already had written a config.reset() :) But it works as I described:

```python
config.reset()
```

prculley:
So it appears that TCG as it stands now is NOT usable with dbapi.  It looks like dbapi.py is trying to do something in 'update_secondary_values' that has the side effect of an exception on an error that TCG deliberately creates.  This occurs when TCG attempts to create a family with an invalid father_handle.
Since TCG deliberately makes a number of such errors, I expect that this will take some re-thinking to handle well.
Have filed a bug report [https://gramps-project.org/bugs/view.php?id=9714](url)
A side note; this particular exception seems to get ignored or lost during this test...  It only shows up when I run the GUI.  Or watch it happen in the debugger.

dsblank:
Ok, thanks for the bug. I'll take a look.

dsblank:
I added some notes on the issue:

What is a good, simple way to replicate this?

I guess a basic question is: should we allow the database to add data that corrupts it? This is bad data, and DB-API correctly crashes when trying to add it. This isn't a bug, but a feature.

Another possibility is to skip this test for DB-API because there is no way that it could ever be created.

prculley:
The bug replicates pretty easily for me in the GUI; just run TCG with dbapi as backend and select  'Generate Database Errors'.

I understand the tension between doing checking during db add/update versus being able to create a bad db for testing the db repair code.

I'm tending toward more checking during db creation.  Although I do think it might be nice if we had a db variety or mode that allowed for testing the C&R code.  If we deprecate bsddb (which works for now), perhaps 'in memory db' could have reduced checking?  I certainly don't want to add more work to the db backend, just for this test.
I also wonder what will happen if someone has a corrupted *.gramps' file from an older db; will that become impossible to import if we have too many checks?
Another possibility would be to have the db be forgiving of errors, perhaps with a popup message, but to repair on the fly for some of the easier issues (like bad handles).
`While trying to (add|change) the (object), a reference to (object) was invalid.  The reference has been removed.`

dsblank:
Of the items you mention, this is the most important:

> I also wonder what will happen if someone has a corrupted *.gramps' file from an older db; will that become impossible to import if we have too many checks?

If you find that DB-API cannot import an export, then that is indeed an issue. But perhaps a different issue. If that is the case, let's fix that and that may lead to a solution that will address this issue too.

Right now, DB-API and BSDDB have exactly the same interfaces. I'm hesitant to diverge in order to allow corrupted data to be inserted. (You could drop down to a lower level if you really want to poke in bad data.)  The Test Case Generator is pretty old code, and some of these cases just may not be able to be created in DB-API, and may not appear in the real world with BSDDB any more. Which means there is nothing to fix if it can't be created to begin with. 

FYI (in case you hadn't seen this) unittest does allow the skipping of tests based on particular criteria. See skipIf:

https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures

dsblank:
@prculley I added a config option in 68bd1e9038e0593d8c24f05b2b29c98635a5baa2 to be able to turn off autobackup:

config.set("database.autobackup", 0)

This is currently only used by DB-API, and is on by default.

dsblank:
That's a pretty large increase in coverage!

sam-m888:
Thank you

PR #238, opened 2016-09-16
Add Export Options filter to Vcard Export

sam-m888:
While looking at issue [9701](https://gramps-project.org/bugs/view.php?id=9701#c49426) I noticed that  vcard export does not have "Export options" to filter the export like the other export formats.


daleathan:
Very handy

Thanks for adding :+1:

PR #240, opened 2016-09-22
Add a config.reset() to restore to factory settings

dsblank:
Now you can:

```python
config.reset()
```
which will reset to the defaults, ignoring all previously set values, and loaded values from config files.

### DIFF
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -34,6 +34,7 @@ This package implements access to GRAMPS configuration.
 import os
 import re
 import logging
+import copy
 
 #---------------------------------------------------------------
 #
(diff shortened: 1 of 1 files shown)


prculley:
Did you mean to add a second reset() routine?  There was one there just above where you added yours.

Paul-Franklin:
Insofar as I understand it, this p.r. might satisfy
https://gramps-project.org/bugs/view.php?id=7961

dsblank:
@prculley No, I missed that. I'll take a look. Thanks!

dsblank:
Closing, as I already had a method named reset that did exactly this, only better.

PR #242, opened 2016-09-24
quiet down the import tests

prculley:
I've found that following the Travis log it is sometimes hard to see the real issues for all the 'noise'.  And I've learned a bit more on how to capture output and only present it when something fails.
So this PR modifies my import_test.py module to produce less normal output.


dsblank:
Nice!

sam-m888:
Thank you

PR #244, opened 2016-09-25
Make VCARD import errors into user dialog instead of logger.warning

prculley:
The VCARD import previously reported import issues via the logging.warning mechanism.  This is not always visible to the user, so he cannot determine if or what went wrong.  In fact, for the detected Vcard errors, the user got no error messages at all via the GUI.

This PR changes the VCARD import to report errors via the user.info dialog mechanism, similar to the way that Gedcom import works.  In addition, the line numbers in the input file of the specific error are added.

A side effect is a bit less noise on the Travis log...

I've tested with both good and bad Vcard files via the GUI and CLI, the reports look good to me.  Travis continues to run the standard test_imports, which includes a Vcard import.

### DIFF
--- a/gramps/plugins/importer/importvcard.py
+++ b/gramps/plugins/importer/importvcard.py
@@ -46,7 +46,7 @@ LOG = logging.getLogger(".ImportVCard")
 #-------------------------------------------------------------------------
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
-ngettext = glocale.translation.ngettext # else "nearby" comments are ignored
+ngettext = glocale.translation.ngettext  # else "nearby" comments are ignored
 from gramps.gen.errors import GrampsImportError
 from gramps.gen.lib import (Address, Date, DateError, Event, EventRef,
         EventType, Name, NameType, Person, Surname, Url, UrlType)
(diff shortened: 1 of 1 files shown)


Paul-Franklin:
It is not clear to me how often vcard input happens, nor how
much effort should be put into it -- not only by you, but by
others -- but since you obviously care enough to do the work
to generate this p.r., I will point out that one of the differences
between logging.warn (or .info, etc.) messages and user
messages is that user-oriented messages have traditionally
been translated.  Some of your messages are, some aren't.

prculley:
I actually was not going to add work to translators; just tried to conserve strings already in place.  The original code was not consistent.  And I did add the two error result strings, marked as translatable.
But you have a good point about the GUI/CLI user output; it should probably all be translated.

Paul-Franklin:
Yes, well, as I said, it's up to you what you want to do, or not.

I too don't like to make needless work for translators (which was
what I meant by "by others").  You could argue it either way.  We
could wait until somebody complains for instance.  Historically, it's
also true that some developers felt that anybody using CLI was (by
definition I guess) a wizard -- and thus probably spoke English, so
any "print" statement was never translated.  But slowly developers
who felt otherwise started translating "user strings."  It is still the case
that most logging messages are not translated though, IIRC.  And
probably shouldn't be, IMHO.

Paul-Franklin:
I think you should translate the whole of the first string, not
just concatenate the elements.  The minor reason is that
possibly, perhaps, maybe in some language the order would
be slightly different.  But the more important reason is that
some punctuation marks get "translated" as they are done
differently in some languages.  Specifically, the colon is
handled differently in French (a space is put ahead of it
as I recall).  Since there are two variables, that will mean
naming them both (and putting them into a dictionary not
a tuple), which again will allow the translator to reorder
them, if he needs to, for his language.

Thanks for flagging them for translation.  I think it's
the right thing to do, for user-facing strings.

sam-m888:
Thank you

PR #245, opened 2016-09-29
added an option for changing the thumbnail size from normal to large in gvfamilylines.py

MJBMZ:
hi,

i added an option for changing the thumbnail size from normal to large in gvfamilylines.py.

github and coding is new for me, so i hope this is the right way to contribute the change.

matthias

### DIFF
--- a/gramps/plugins/graph/gvfamilylines.py
+++ b/gramps/plugins/graph/gvfamilylines.py
@@ -51,7 +51,7 @@ from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
 from gramps.gen.lib import EventRoleType, EventType, Person, PlaceType
 from gramps.gen.utils.file import media_path_full
-from gramps.gen.utils.thumbnails import get_thumbnail_path
+from gramps.gen.utils.thumbnails import (get_thumbnail_path, SIZE_NORMAL, SIZE_LARGE)
 from gramps.gen.plug.report import Report
 from gramps.gen.plug.report import utils
 from gramps.gen.plug.report import MenuReportOptions
@@ -236,6 +236,12 @@ class FamilyLinesOptions(MenuReportOptions):
                                        'should appear relative to the name'))
         add_option('imageonside', self.image_location)
 
+        self.image_size = EnumeratedListOption(_('Thumbnail size'), SIZE_NORMAL)
+        self.image_size.add_item(SIZE_NORMAL, _('Normal'))
+        self.image_size.add_item(SIZE_LARGE, _('Large'))
+        self.image_size.set_help(_('Size of the thumbnail image'))
+        add_option('imagesize', self.image_size)
+
         # ----------------------------
         add_option = partial(menu.add_option, _('Family Colors'))
         # ----------------------------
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Almost.  You should have created the pull request directly from your MJBMZ-patch-1 branch.  This would have avoided the extra commit that we don't need.

gramps/plugins/graph/gvfamilylines.py
```diff
@@ -72,6 +72,9 @@
            {'name' : _("Colored outline"), 'value' : "colored"},
            {'name' : _("Color fill"), 'value' : "filled"}]
 
+SIZE_NORMAL = 0
```
Nick-Hall: Import these constants from `gramps.gen.utils.thumbnails`

gramps/plugins/graph/gvfamilylines.py
```diff
@@ -236,6 +239,12 @@ def add_menu_options(self, menu):
                                        'should appear relative to the name'))
         add_option('imageonside', self.image_location)
 
+        self.image_size = EnumeratedListOption(_('Thumbnail size'), 0)
+        self.image_size.add_item(0, _('Normal'))
```
Nick-Hall: Use the constants here instead of 0 and 1.

gramps/plugins/graph/gvfamilylines.py
```diff
@@ -809,9 +819,16 @@ def write_people(self):
                     media = self._db.get_media_from_handle(media_handle)
                     media_mime_type = media.get_mime_type()
                     if media_mime_type[0:5] == "image":
-                        image_path = get_thumbnail_path(
-                            media_path_full(self._db, media.get_path()),
-                            rectangle=media_list[0].get_rectangle())
+                        if self._imagesize == SIZE_NORMAL:
```
Nick-Hall: The `if` statement is not necessary here.  Use `size=self._imagesize` in the call to `get_thumbnail_path`.

MJBMZ:
Thank you Nick, i changed the parts.

Nick-Hall:
In testing, I noticed that this option should be made inactive when thumbnails are not selected.  The thumbnail location option already does this, so it should be easy to copy the code.

Paul-Franklin:
As long as you still have changes to make, you might as
well change "EnumeratedListOption(_('Thumbnail size'), 0)"
so that the zero becomes a SIZE_NORMAL too.

MJBMZ:
I have read about merging or squashing these different commits, but i'm not sure if this is something i can do or where. can i do something to clean this up?

prculley:
read up on "git rebase -i yourcommithash".  If you have your text editor wet up to start when git needs it, then it will pop up a text file with the commits after the one you specify.  You can then use a 'squash' or 'fixup' to merge several commits into a single one and you can update the commit message as well.

MJBMZ:
thanks prculley, worked fine (after installing git and learning how to use it, this was the first file i ever edited so i used the github webui before...)

Nick-Hall:
Thanks.  I just made one minor change.  Merged a8a3c0865d52390c08dc518b98922f1f1a86a4c4.

PR #248, opened 2016-10-07
bug 8993; fix Gedcom import in some alternate languages; improper date parsing

prculley:
While importing Gedcom file with Swedish language settings, the import failed.  It failed when parsing a date.  Turns out that the DateParser class (which -should- be a base class for other languages, and is in English) was setting the month, date, year order based on the current locale.  For Gedcom the 'locale' is supposed to always be US English.

I've run a few tests and it seems to me that this fix is safe, but please look over the patch carefully as I am unable to fully test in many languages...
I think this will also clear https://gramps-project.org/bugs/view.php?id=9138


Paul-Franklin:
I have modified _dateparser in master, so your patch
will probably need rebasing.

Why not add a third "elif" to specifically check for the
"dmy" string you are using in libgedcom, and then
change the "match" one to just "elif match" (since by
then you will have explicitly checked for the other
three possibilities)?

Other than that, I think your patch should be OK.

Paul-Franklin:
I modified gramps42 also, I forgot to say.

By the way, I haven't checked but it might be a good
idea for you to check that GEDCOM export always
creates dates using the American date format, if that
is what it supposed to do (rather than using the user's
locale).  Just a thought.

Paul-Franklin:
"""
if dmy_type is None and match:
    self.dmy = (match.groups() == ('d', 'm', 'y') or
                match.groups() == ('d', 'b', 'y'))
    self.ymd = (match.groups() == ('y', 'm', 'd') or
                match.groups() == ('y', 'b', 'd'))
    # note ('m','d','y') is valid -- in which case both will be False
elif dmy_type == 'mdy':
    self.dmy = False
    self.ymd = False
elif dmy_type == 'ymd':
    self.dmy = False
    self.ymd = True
else:
    self.dmy = True
    self.ymd = False
"""

I think I like this if/elif chain better.  I think it
is more obvious what the choices are.

What do you think?

Paul-Franklin:
'''
if dmy_type is None and match:
    self.dmy = (match.groups() == ('d', 'm', 'y') or
                match.groups() == ('d', 'b', 'y'))
    self.ymd = (match.groups() == ('y', 'm', 'd') or
                match.groups() == ('y', 'b', 'd'))
    # note ('m','d','y') is valid -- in which case both will be False
elif dmy_type == 'mdy':
    self.dmy = False
    self.ymd = False
elif dmy_type == 'ymd':
    self.dmy = False
    self.ymd = True
else:
    self.dmy = True
    self.ymd = False
'''

I think I like this if/elif variant better.
I think it makes the choices more obvious.

What do you think?

Paul-Franklin:
___
if dmy_type is None and match:
    self.dmy = (match.groups() == ('d', 'm', 'y') or
                match.groups() == ('d', 'b', 'y'))
    self.ymd = (match.groups() == ('y', 'm', 'd') or
                match.groups() == ('y', 'b', 'd'))
    # note ('m','d','y') is valid -- in which case both will be False
elif dmy_type == 'mdy':
    self.dmy = False
    self.ymd = False
elif dmy_type == 'ymd':
    self.dmy = False
    self.ymd = True
else:
    self.dmy = True
    self.ymd = False
___

I think I like this if/elif variant better.
I think it makes the choices more obvious.

What do you think?

Paul-Franklin:
```
if dmy_type is None and match:
    self.dmy = (match.groups() == ('d', 'm', 'y') or
                match.groups() == ('d', 'b', 'y'))
    self.ymd = (match.groups() == ('y', 'm', 'd') or
                match.groups() == ('y', 'b', 'd'))
    # note ('m','d','y') is valid -- in which case both will be False
elif dmy_type == 'mdy':
    self.dmy = False
    self.ymd = False
elif dmy_type == 'ymd':
    self.dmy = False
    self.ymd = True
else:
    self.dmy = True
    self.ymd = False
```

I like this if/elif variant better.
I think it makes it more obvious.

What do you think?

Paul-Franklin:
Upon further thinking about this, I have just realized that
it /might/ be the case that no modification to the base
_dateparser.py module is necessary at all.

So first, let me say that this is only a speculation or a
hunch on my part and I haven't tried to test my idea.

But getting on to what the idea is, it might well be the
case that given that my modification to DateParser a
few days ago now allows a subclass to modify the "fmt"
variable -- once I made it be a class variable -- so in
your case you might be able to get away with just
defining "fmt" in GedcomDateParser to be '%d/%m/%Y'
-- or whatever the American "fmt" is (I haven't checked) --
since (I believe) that then the self.fmt you just defined
in GedcomDateParser would be used in DateParser (to
make that self.dmy test for instance).

At least, the ability for a subclass to redefine it was key
to my fix for the Czech bug I fixed a few days ago, and
that worked so I don't know why something similar would
not work in the GedcomDateParser class also.

But as I say I haven't tried it.  This is "fresh off the press"
as it were, hot off the neurons in my head.  8-)

Take a look at my recent commit https://github.com/gramps-project/gramps/commit/ac00bd18430269247889a9402e453be3e32a19c1

kulath:
The "American" format is usually month day year, whereas the GEDCOM format is day month year.

ennoborg:
Tim is right, and month should always be the abbreviated name, according to the calendar that's in use, i.e. French, Gregorian/Julian, or Hebrew. GEDCOM dates like 10/10/2016 are plain wrong, and even though 10/16/2016 could be interpreted without warning, I think that it should not.

Paul-Franklin:
All right, I'm not a GEDCOM user and don't pretend to
know much about it, so I got the format string suggestion
wrong.  But don't be confused by the date parser's "fmt"
string and real GEDCOM dates.  The date parser string
is just the format (according to some standard, I suppose)
for how dates are supposed to be displayed (and parsed)
in any particular locale.  The gramps date parser code
looks at that "fmt" string in order to decide how to process
dates, how to parse them, how to display them.  It's the
starting point.  So my suggestion/hope/hunch is that all
that is needed is to set (in GedcomDateParser) the "fmt"
string to be whatever the GEDCOM standard calls for.
(Although my personal guess is that (in "fmt") only the
order of the three things really matters.)  HTH.

prculley:
I decided to adjust my patch to your idea Paul.  Looks like it performs as expected.  No more problems with Swedish LANG settings.
Note that I did NOT set the fmt to dmy, but to mdy.  This is because in that is the setting it has always had in us-en; where I presume the majority of testing has been done.  A weird side effect of the way the code works is that even with this setting, dates with the format "23 JUN 1999" are parsed correctly, because the "JUN" is correctly interpreted as the month.  Also, with this setting, I think that dates in the Gedcom files that look like "6/23/1999" are correctly parsed (even though they are not valid Gedcom).

Another submitted bug I have not thought through was when Gedcom file dates like "23/6/1999" are found; at the moment, these fail as they are not easily distinguishable.  So what I have now acts the same as always in us-en, and works exactly the same in other languages, as is proper with Gedcom.

Paul-Franklin:
Good to know.  Thanks for the feedback.

PR #250, opened 2016-10-10
Some more changes to gvfamilylines.py

MJBMZ:
- changed to show all location types which makes sense (See: https://gramps-project.org/bugs/view.php?id=9082)

- changed display of dates and locations to be more responsive (See additional FeatureRequest at bottom of this one: https://gramps-project.org/bugs/view.php?id=7788) (hope you like it, for me it looks much cleaner)

- added option for rounded corners for men additional to women

- corrected five to four indented spaces before arguments to get_thumbnail_path (sorry for that)

### DIFF
--- a/gramps/plugins/graph/gvfamilylines.py
+++ b/gramps/plugins/graph/gvfamilylines.py
@@ -134,10 +134,15 @@ class FamilyLinesOptions(MenuReportOptions):
                                        '"family lines".'))
         add_option('removeextra', remove_extra_people)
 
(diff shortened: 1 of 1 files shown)


Nick-Hall:
These are all good suggestions. However, it would be better if you made them separate pull requests.

The fixes for locations and the indented code can be accepted quickly, but the other two changes may need to be discussed on the list.

MJBMZ:
Ok, will make some separate PRs. 

What is the right order for the "other two changes", first PR and then discuss or first discuss and then PR?

PR #251, opened 2016-10-16
Update Tool 'Check & Repair' and 'TestCaseGenerator' to detect and repair duplicate Gramps IDs

prculley:
A couple of past bugs in the Gedcom importer have left databases with potential duplicate Gramps IDs.  So I decided to update Check and Repair to detect and fix these.  I also updated Test Case Generator to create duplicates in each category where we have GIDs.  And the test_tools to check that the correct number of problems were found as a test.

Imagine my surprise when I ran C&R against my personal database and found yet another situation where duplicate GIDs are present... Sigh, looks like more debugging and another PR...


Paul-Franklin:
https://gramps-project.org/bugs/view.php?id=6323
https://gramps-project.org/bugs/view.php?id=6574

Perhaps others, too.

sam-m888:
Thank you

PR #252, opened 2016-10-19
bug 9755, fix duplicated Gramps IDs on Gedcom import

prculley:
Found this when running the updated 'Check & Repair' tool which now scans for duplicate Gramps IDs.

Gramps tries to use Gedcom @xref@ as Gramps IDs. But they may already be in use, so the importer must track the xref to GIDs in a "swap" table.
In this case, the Gedcom contains a cross reference to a xref that is defined later. The importer examines the xref at first encounter and does a 'find_next_gramps_id', and a check of the 'swap' table to see if it is in use, determines that it is ok to use, saves it, but does not 'commit' it yet.
Later, another Gedcom line wants to create a note, but does not check the swap table, finds the same GID, uses it and commits.
Finally when the xref definition is found the importer uses the GID from the swap table (Which is now duplicated) to commit the xref note.

This bug has been there since before 2007; showing how easily a contributor can mess up if he doesn't fully understand the code. Particularly since the comments don't really explain this stuff very well.

I used an updated imp_sample.ged file to test for this; that file also has quite a few other changes to increase the code coverage, which is why there are so many changes.

### DIFF
--- a/data/tests/imp_sample.ged
+++ b/data/tests/imp_sample.ged
@@ -21,6 +21,7 @@
 1 NAME Anna /Hansdotter/
 2 GIVN Anna
 2 SURN Hansdotter
+2 NOTE Hans daughter? N0000
 1 SEX F
 1 BIRT
 2 TYPE Birth of Anna Hansdotter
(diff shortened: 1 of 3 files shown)


sam-m888:
Thank you

PR #253, opened 2016-10-21
Feature 9082, show all location types (which makes sense) in Familylines-Report

MJBMZ:
- changed gvfamilylines.py to show all location types which makes sense (See: https://gramps-project.org/bugs/view.php?id=9082)

--> Please have a look at the location types and order if you agree with them, thanks.


Nick-Hall:
Fixed indent aab0984dfbe4745a56edc761a7cdeec80e38de3a.
Implemented feature request d333f2eee75194c43ddbf27902aa500dc3b64207.

PR #254, opened 2016-10-22
8250 and 9736: avoid to close database or quit

SNoiraud:
```
           when exporting and solves quickview
           behind export assistant.
```

### DIFF
--- a/gramps/gen/plug/docgen/basedoc.py
+++ b/gramps/gen/plug/docgen/basedoc.py
@@ -60,7 +60,7 @@ class BaseDoc(metaclass=ABCMeta):
     such as OpenOffice, AbiWord, and LaTeX are derived from this base
     class, providing a common interface to all document generators.
     """
-    def __init__(self, styles, paper_style):
+    def __init__(self, styles, paper_style, parent=None):
         """
         Create a BaseDoc instance, which provides a document generation
         interface. This class should never be instantiated directly, but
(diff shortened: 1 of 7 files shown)


prculley:
This appears to solve the 'hidden window has focus' problem.  Although it appears that it is now possible to partially hide the export window with the quickreport, and then re-select the export window to give it focus (but not make it move to top).  Perhaps I am being overly concerned; I prefer the window with focus to be on top, except in rare cases like editor find dialogs.

Why don't you leave it for now, I'll rethink this a bit.  Besides, I think I might try to add a progress meter (rather than or in addition to a busy cursor) to give the user more feedback on long exports.  And for the 'Backup' menu item.

SNoiraud:
The busy cursor is the only thing we can do there. I use it to avoid to close the database or quit gramps.
I need to do that because there are several phases in the export assistant. Between each tab of the export assistant, we can close the database. The worse case is between the filtering and the real export, and in this case, we have a crash because the dummy database doesn't accept all function that a real database accept.
I am not sure the progress meter will change that. It will be too late.

SNoiraud:
> I prefer the window with focus to be on top, except in rare cases like editor find dialogs.

If you click on the quickview report, I think this is the quickview which must have the focus and not the export assistant.
Anyway, these quickviews are only informal and must be closed to continue the export.

SNoiraud:
I'm not sure this is the good solution. We always have possibilities to have crashes.
We need to avoid the tools (check and repair, rebuild indexes, ...), In the edit menu (Add..., Edit..., Remove, Merge) for each pageview (we always can switch between views).
I think the modal mode is the best way to avoid all these possibilities.
I'll make a new PR with the modal mode.

SNoiraud:
The following PR seems to be better than this one
https://github.com/gramps-project/gramps/pull/267

I close this PR.

PR #256, opened 2016-10-24
Improve test coverage in the Gedcom importer

prculley:
This PR is primarily to improve test coverage.  The goal was to test as much of the standard code (as used in normal imports) as possible.  Some unusual conditions are also added, but most exceptions remain untested.

It was necessary to use Mock to keep the test_imports routine from failing on some dates, this in turn required a few changes to the Python imports in the 'unknown.py' module.  I have not figured out how to properly Mock patch standard time.time in an individual module yet unless it imports via the 'from time import time' sytax.  So I changed the module to use that syntax, rather than importing the entire time module via 'import time'.

I also added in the 'tag on import' to the tests when the '_dfs' is part of the file name; this causes Gedcom (or others) to tag imported data (and increase test coverage of that feature).

I commented out an 'assert' statement in libgedcom.py that got executed with the added test coverage; it prevented the test from completing.  Apparently the author thought that that particular code path should not ever be executed.

I found several (minor) issues with the Gedcom importer by doing this process, I will submit bugs and PRs to address these as time permits.

### DIFF
--- a/data/tests/imp_PhonFax_dfs.ged
+++ b/data/tests/imp_PhonFax_dfs.ged
@@ -10,14 +10,20 @@
 3 FAX (801) 705-7001
 3 EMAIL help@ancestry.com
 3 WWW http://www.ancestry.com
+2 DATA FTM Source data
+3 COPR Copyright 2999
+2 DATE 1 JAN 1111
 1 DEST GED55
 1 DATE 01 MAR 2016
 1 CHAR UTF-8
+1 CHAR UTF-8
+2 VERS 1.1
 1 FILE D:\Family Tree Maker\imp_FTM_LINK.ged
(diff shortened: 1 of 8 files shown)


prculley:
Oops; looks like there is some difference between Windows and Linux implementations of time.strftime...

prculley:
Looks like it works; improved coverage from 83 to 92%.

sam-m888:
Can you rebase this please? It seems you included some files from https://github.com/gramps-project/gramps/pull/252 


> Conflicting files
> data/tests/imp_sample.ged
> data/tests/imp_sample.gramps
>

prculley:
Rebase complete...

sam-m888:
Thank you :+1:

PR #260, opened 2016-10-26
Update LICENSE to contain correct license text

uxcq:
I'm guessing Gramps really isn't released under the GNU Free Documentation License Version 1.2, November 2002, maybe the documentation but not the program.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,397 +1,339 @@
-		GNU Free Documentation License
-		  Version 1.2, November 2002
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
 
-
- Copyright (C) 2000,2001,2002  Free Software Foundation, Inc.
- 51 Franklin St, Fifth Floor, Boston, MA  02110-1301, USA
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc., <http://fsf.org/>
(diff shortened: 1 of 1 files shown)


Nick-Hall:
The GPL2 license is in the [COPYING](https://github.com/gramps-project/gramps/blob/master/COPYING) file.  Do we need both a COPYING and LICENSE file?

uxcq:
Well that is interesting did not see the COPYING file! I only created this PR as Github displays the license based on the content of a file called LICENCE!  So totally up to you if you need both but one of then is preferred by Github.

https://github.com/blog/2252-license-now-displayed-on-repository-overview

License now displayed on repository overview

QuLogic:
It should [check COPYING as well](https://github.com/benbalter/licensee/blob/master/docs/what-we-look-at.md).

uxcq:
> It should check COPYING as well.

@QuLogic Then why hasn't it?

QuLogic:
Yes, LICENSE has the [highest priority](https://github.com/benbalter/licensee/blob/master/lib/licensee/project_files/license_file.rb#L20-L29).

Nick-Hall:
Old LICENSE file removed in commit aaf4bee03dbdcb3383b71421defa87c5c5a13061.

PR #261, opened 2016-10-26
Fix bug 9733, CSV import Fails, 9676, failure for certain types of cross references etc.

prculley:
This PR updates Gramps42 branch CSV import to match master.  I've done this because there are a couple of bugs to the CSV importer that have been corrected in master but not in gramps42, and we are seeing complaints.  It was easier to just copy the master source to the gramps42 branch than to try to cherry-pick and merge.  The master has some extra fixes (some pylint, log.warn() -> log.warning() ) that came along for the ride.
I'm assuming we will have a 4.2.5 release.

Note: I have a personal patched up version of the gramps42 branch with the import tests that the public branch does not have; this passes in gramps42.

### DIFF
--- a/gramps/plugins/importer/importcsv.py
+++ b/gramps/plugins/importer/importcsv.py
@@ -44,15 +44,15 @@ LOG = logging.getLogger(".ImportCSV")
 
 #-------------------------------------------------------------------------
 #
-# GRAMPS modules
+# Gramps modules
 #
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Yes. There will be a 4.2.5 release.

sam-m888:
Thank you

PR #262, opened 2016-10-29
Bug 8887 Add PEDI and _FREL/_MREL Gedcom Tag Export support

prculley:
The bug complained that certain combinations of child relationship to parents was not reflected in the exported Gedcom.
Gedcom can report one parent adopting, or both parents fostering or birth or stepchild, but not one parent fostering and one birth.

When one or both of the parents are adopting, Gedcom allows the creation of an ADOP event on the INDI (individual) to explain the parental relationship
0 INDI
+1 ADOP {1:1}
+2 <<INDIVIDUAL_EVENT_DETAIL>> {0:1}\* p.34
+2 FAMC @XREF:FAM@ {0:1} p.24
+3 ADOP <ADOPTED_BY_WHICH_PARENT> {0:1} p.42

Gramps does this with the 'ADOP Y' syntax if no real adoption event is present.

Gedcom also uses the PEDI tag to report on pedigree in the FAMC link subordinates.  Gramps did not formerly utilize this tag.
I've noted that some of the extensions to Gedcom use the _FREL and _MREL tags to report odd situations where this can occur. We even deal with this on import.

0 INDI
+1 FAMC @XREF:FAM@
+2 PEDI <PEDIGREE_LINKAGE_TYPE> {0:1}
+2 STAT <CHILD_LINKAGE_STATUS> {0:1}
+2 <<NOTE_STRUCTURE>>
+2 _FREL <PEDIGREE_LINKAGE_TYPE> {0:1} (Fathers linkage Custom extension)
+2 _MREL <PEDIGREE_LINKAGE_TYPE> {0:1} (Mothers linkage Custom extension)

This PR that extends our export to add the PEDI tag when both parents are doing the same thing, and the _FREL _MREL tags when the parents to child relationships are different.

I updated the test_exports test files to test this case and to expand the code coverage for Gedcom as much as possible.  In this process I also spotted a bug in exporting a FAX tag and fixed it.
In order to (finally) get the Travis to work when media are present in the database, I had to change the resourcepath code to end up with an absolute path.  Effects situations where GRAMPS_RESOURCES are set to relative paths, like ".".  Now this will be expanded internally to the actual absolute path in use.


sam-m888:
Thank you

PR #263, opened 2016-10-30
Bug9767 Fix LDS editor for 'Family' related issues

prculley:
Several problems fixed in LDS ordinance editors:

1) ~~On newer versions of GTK+ the Family select ICON was not visible.  It appears that Gramps never assigned an icon to the Family select button.  As near as I can tell the only reason this ever worked was that there was a 'feature' (bug?) in GTK+ that allowed the most recently defined image ICON value to be used in object images if there was none defined at all.  I suspect that this 'feature' disappeared in more recent versions of GTK+ (the icon was visible in GTK+ 3.16.7, but is gone in 3.18.2 and newer.).~~
Already fixed in https://github.com/gramps-project/gramps/commit/22c1bd4e1c61de62df099b355e473e91afecfa90

2) The Person LDS editor was trying to hide the 'Family' line whenever any LDS type was not SEAL_TO_PARENTS, but this didn't always work.  The if-elif set missed the CONFIRMATION value; which then left the 'Family' line showing or not, depending on the last LDS type shown.

3) On the Person LDS editor when the dialog was first activated, the 'Family' line was always shown, because after executing the 'hide', command the last thing the init code did was 'show' everything again.

4) On the Family LDS editor, the parent line was always shown, it should not have been, because the only option was 'Sealed to Spouse' and there is no need for the 'Family' there.

Sorry, no automated test for this; that is a much larger project for another day.  Suggestions are appreciated.  I manually tested on Windows and Linux for each issue listed above; seems to work.

### DIFF
--- a/gramps/gui/editors/editldsord.py
+++ b/gramps/gui/editors/editldsord.py
@@ -179,6 +179,9 @@ class EditLdsOrd(EditSecondary):
         self.parents_label = self.top.get_object('parents_label')
         self.parents = self.top.get_object('parents')
         self.parents_select = self.top.get_object('parents_select')
+        image = self.parents_select.get_child()
+        image.set_from_icon_name('gtk-index', Gtk.IconSize.BUTTON)
+        self.parents_select.set_tooltip_text(_('Select Family'))
 
         self.priv = PrivacyButton(
             self.top.get_object("private"),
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Bug (1) in your list was my fault.  It was introduced in commit 75009f0fe0ed6ecdc575c522f6a14a9d6262a666.  I removed the stock icon, but didn't replace it with a named icon.

It is probably neater to set the icon and tooltip in the glade file.

prculley:
I considered modifying Glade file but I am not exactly experienced in Glade (raw newbie...).  I was unable to find the icon 'gtk-index' in my Glade list of Icons; so was unsure how to handle it there.  And none of the icons that were there looked similar.  So I copied what was done for the Place selector icon I found elsewhere.

Nick-Hall:
I have fixed the icon and tooltip in gramps42 17c51b3e76bc08bdfe8bc85e715e3e9d1872ee37 and master dd0eedc77218d7cab0b9e32177054746cb9bdac4.

sam-m888:
Thank you

PR #264, opened 2016-10-30
bug 9765; fix citation tree view for HandleError exceptions

prculley:
A relatively recent change makes attempts to look up objects via db.get_*_from_handle(), when the handle is not present in the db, cause an exception, where it used to just return a None.

The citation tree view has both sources and citations in the same list; so when a list line is selected for an operation, the code has to try both sources and citations to figure out which is actually selected.  When the code attempted to lookup both source and citation with the same handle, one always failed with an exception.

This PR corrects this by using try/except syntax in place of the None tests.

A secondary bug found was that following a merge, the tree view was not updated and an attempt to select and do something one of the 'merged out' entries would fail.  I forced a view update following a merge.

### DIFF
--- a/gramps/plugins/view/citationtreeview.py
+++ b/gramps/plugins/view/citationtreeview.py
@@ -51,7 +51,7 @@ from gramps.gui.views.listview import ListView
 from gramps.gen.utils.db import (get_source_and_citation_referents,
                                 get_citation_referents)
 from gramps.gui.views.bookmarks import CitationBookmarks
-from gramps.gen.errors import WindowActiveError
+from gramps.gen.errors import WindowActiveError, HandleError
 from gramps.gui.ddtargets import DdTargets
 from gramps.gui.dialog import ErrorDialog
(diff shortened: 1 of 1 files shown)


gramps/plugins/view/citationtreeview.py
```diff
@@ -555,6 +541,23 @@ def get_handle_from_gramps_id(self, gid):
         else:
             return None
 
+    def get_source_or_citation(self, handle, check=True):
```
QuLogic: Should it be private?

sam-m888:
Do you need to respond to QuLogic's review, before this can be accepted?

prculley:
I had not intended to respond, Gramps code has so many methods that -should- be private, but are not, that I didn't think we were particularly concerned about this.  I actually don't think about it at all during coding.

If, as a policy, we are going to start enforcing this, then I think we have some real work in going through all the code to get it to comply.

QuLogic:
Marking one _new_ method as private is not the same thing as enforcing it across the code base.

But even a "no" is better than silence...

sam-m888:
Thank you

PR #267, opened 2016-11-04
8250 and 9736: avoid all other operations during export.

SNoiraud:


SNoiraud:
This is an alternative to https://github.com/gramps-project/gramps/pull/254
The solution will be PR254 or this one PR267.
Personally, I prefer this one.
In this case, We have no more possibility to have a crash during export.

Paul-Franklin:
I prefer this one too.

prculley:
I'm probably acting like a clueless user here, but when I open the quickreport from the export dialog (by pressing one of the "number of people" buttons), and then open another quickreport from there (right-click then "see the filter details"), and then open a person (or whatever) to edit (right-click then "see the person details")... I can get the windows to be quite confusing.  The 'person' I opened ends up behind the quick reports, but doesn't seem to have keyboard focus.
In writing this comment, I switched windows and when I went back to Gramps, nothing visible on the screen would respond.  The only dialog with the focus had gotten restacked beneath another dialog.  Jumping back and forth between application windows Gramps would sometimes change the stacking of its dialogs, so I eventually got the one with focus on top again.  But this would be VERY confusing to a user.

I'm beginning to think that perhaps we should disable the quickreport functions completely.  Or recode them to execute within the export dialog window and limit how far you can go (don't let user try to edit person etc.).

This is all made worse on Msoft Windows because there is no key that switches between windows within a Gtk application (I think Linux uses Ctrl-tab for this).  A somewhat experienced user might realize what happened and be able to switch between Gramps windows on Linux, but not on Windows...

SNoiraud:
OK. I see the problem.
Perhaps in this case, we could avoid editing. Our current action is to export, not to edit.
I never use these possibilities and I don't know if some people use it.

We could perhaps comment the Quickviews report as you said in the last comment.
If someone ask for these quickviews, we shall avoid editing.

SNoiraud:
It can be useful to see what family(ies) or what person(s) are exported, so the solution seems to be to avoid the object edition.
I'm working on this.

prculley:
I think you are on the right track.  Just for fun I pulled your patch into my Linux VM to see what it looked like there.
I open the "Summary counts" quickreport from the export dialog (by pressing one of the "number of people" buttons), and then open another "Filtering on Person" quickreport from there (right-click then "see the filter details").
So far, so good.
I then used alt-tab to switch to a different application.  And did it again to switch back to Gramps; however I noticed that there were two Gramps windows in the Kubuntu alt-tab switching list, one for each open quickview.  So (acting like a dumb user) I deliberately selected the "Summary counts" quickview.  Since that one is larger, on my screen it covered the other one.  I attempted to close it with the mouse, no response.  The other quickview has the focus but is hidden beneath.

Now a Linux user has the option of alt-tab to try to locate the other quickview.  A Windows user does not, and on Windows, which quickreport ends up on top seems to be random.

P.S. I accidentally hit on a way out, even for Windows users; since the hidden quickview has the focus, the 'Esc' key will close the hidden dialog (although that is not obvious visually).  At that point the visible quickreport has the focus and responds to keyboard or mouse again.  A user who doesn't use the keyboard for pushing buttons (like me) will not figure this out...

If we could just convince the window manager to keep the window with the focus on top this would all be a non-problem.

SNoiraud:
> I then used alt-tab to switch to a different application. And did it again to switch back to Gramps; however 
> I noticed that there were two Gramps windows in the Kubuntu alt-tab switching list, one for each open 
> quickview. So (acting like a dumb user) I deliberately selected the "Summary counts" quickview. Since 
> that one is larger, on my screen it covered the other one. I attempted to close it with the mouse, no 
> response. The other quickview has the focus but is hidden beneath.

I don't have this kind of problem. Whatever I select with alt-tab, I have always the good window on top.
I don't use KDE.  What versions do you use ? (I use xfce)

prculley:
Keep in mind that I am barely able to use Linux, I do most all of my work on Windows.  My Linux appears to be Ubuntu 15.10 with essentially no upgrades.  So whatever that comes with.  I see I mispoke earlier when I mentioned Kubuntu, that was another VM, not the one I used for this test.

Anyway, after looking at the code for a bit, I suspect that I might see a solution, maybe; I mentioned it in my last two paragraphs above (you might have missed these as I edited the last comment over a period of time).  If you are tired of dealing with this, I can try to finish figuring this out, but it appears that you might be better with this GUI stuff than I am, or at least have a lot more experience.

SNoiraud:
Describe your idea about your solution. I'll try to apply it.

prculley:
Here is what I tried so far, the idea was to make the second quickreport always end up on top of the first and the export dialog.  It kind of works, in that at least if you switch windows to another application and back, the most recent quickreport ends up on top.  But it is still possible for the user to mouse select the export dialog or the previous quickreport, neither of which has focus.  I also tried to make them modal, but apparently you can only have one modal window; if you have more, they stop acting modal.

<pre>--- a/gramps/gui/plug/quick/_quicktable.py
+++ b/gramps/gui/plug/quick/_quicktable.py
@@ -163,7 +163,8 @@ class QuickTable(SimpleTable):
         run_quick_report_by_name(self.simpledoc.doc.dbstate,
                                  self.simpledoc.doc.uistate,
                                  'filterbyname',
-                                 'Inverse %s' % obj_class)
+                                 'Inverse %s' % obj_class,
+                                 parent=self.simpledoc.doc.uistate.window)

     def on_table_doubleclick(self, obj):
         """
@@ -279,7 +280,8 @@ class QuickTable(SimpleTable):
                 run_quick_report_by_name(self.simpledoc.doc.dbstate,
                                          self.simpledoc.doc.uistate,
                                          'filterbyname',
-                                         handle)
+                                         handle,
+                                         parent=self.simpledoc.doc.uistate.window)
         return False # didn't handle event

     def on_table_click(self, obj):</pre>
I think what is needed is to make the current most recent window modal, and only that window.  That would require some code to remove modality from the parent, and put it on the current quickreport, and then to reverse this when closing a quickreport.  In combination with your fix to not allow editing, that SHOULD keep the window stack correct and only the top one focused.  I'll think about this some more and see if I can figure out a way to do this.

SNoiraud:
After looking your two patches, they are already there but hidden. Perhaps I forgot something.
In fact, run_quick_report_by_name call run_report.
run_report call TextBufDoc. This is where I put the parent when I call DisplayBuf.
sel.parent is set in the BaseDoc

The only problem we have is the modal mode.
Perhaps we need to create two methods in uistate like set_export_mode 
For example : 
uistate.set_modal_mode(window, True) # To set the window modal.
In this case, we add the window to a modal_list.
uistate.set_modal_mode(window, False) # which remove the modal mode for this window
we remove the window from the modal_list  and set the modal mode for the last window in the list.
The first element on this list will be the parent for all modal windows.

I use the present window method which force the window to be on top.

We don't need anymore the parent window in the quickview reports and in basedoc.

I push this.

prculley:
Looks like we are heading in the same direction.  I am/was working on a similar procedure.  It is clear that we must have only a single modal window at one time, at least within a group that have parent/children relationship.  My approach is to integrate this into the ManagedWindow class.  I did this and it worked, except that it became clear that some of our code does not properly use that class; in particular the 'track' parameter to that class init is not being set in activating the quickreports (and probably others like the filter editors and calculate previews progress popup).
I've been spending my time getting this fixed...
So it looks like we might have competing PRs for the same problem ;-)

When I get a chance I'll try out your fix and compare to mine and see if one or the other covers more cases...

SNoiraud:
This PR modify only the following files : 
gramps/gui/displaystate.py
gramps/gui/managedwindow.py
gramps/gui/plug/export/_exportassistant.py
gramps/gui/plug/quick/_quicktable.py
gramps/gui/plug/quick/_textbufdoc.py

SNoiraud:
Paul, Did you find a better solution ?

prculley:
PR275 offers an alternate solution.

eno93:
Alternate solution: https://github.com/gramps-project/gramps/pull/275

PR #268, opened 2016-11-08
bug 9771; fix missing right-click menu items on gallerytab

prculley:
Decided to remove (almost) all the right-click menu icons and make sure that text was present.  The missing text was the cause of the initial bug.

This icons were not being shown due to a Gtk bug in the deprecated Gtk.ImageMenuItem() method.

I did a sweep search of all our code for these right-click menus and tried to patch everything I could find.  The Addons were also searched and will be the subject of another PR.

After icons were removed, I also removed the left side space where the icons used to reside via
`self.menu.set_reserve_toggle_size(False)`

The reason I say 'almost' all menu icons were removed is that the Gramps left sidebar (dropdown version) also used menu icons.  I judged that we would like to keep these.  So the dropdownsidebar got fixed with icons and labels in a box inside of the menu item.

During testing of the updated code, I also observed some additional issues, which I also patched.

1) bookdialog right-click menu got a real menu seperator instead of a large empty space.
2) in pedigreeview, widget `self.get_style().font_desc` was deprecated two ways, I changed to
    `self.get_style_context().get_font(Gtk.StateFlags.NORMAL)`
  unfortunately the `get_font()` is also deprecated, (more recently), but I could not figure out a way to correct this that worked.
3) in pedigreeview, `Gtk.ArrowType` is also deprecated; I switched this out for icons.
4) in pedigreeview, the `window.get_hadjustment()` was pointing to the wrong level of the Gtk hierarchy, a ViewPort rather than the ScrolledWindow.
5) in relview (relationship view) the right-click menu and tooltip were originally labeled with
    `label=_("Edit %s") % name`
At first glance, I thought a persons name started with "Edit Smith, John" so I changed it to
    `label=_("Edit Person (%s)") % name`
so it now looks like "Edit Person (Smith, John)".  The new text string was already in the .pot file so no new translation work for that.
6) in configure.py I added an optional parameter to the add_slide method to allow for a wider slider; I observed it was not possible to slide this on one of the configure dialog boxes for an addon because it was so narrow.

### DIFF
--- a/gramps/gui/configure.py
+++ b/gramps/gui/configure.py
@@ -437,7 +437,7 @@ class ConfigureDialog(ManagedWindow):
         return combo
 
     def add_slider(self, grid, label, index, constant, range, callback=None,
-                   config=None):
+                   config=None, width=1):
         """
         A slider allowing the selection of an integer within a specified range.
         :param range: A tuple containing the minimum and maximum allowed values.
@@ -455,7 +455,7 @@ class ConfigureDialog(ManagedWindow):
         slider.set_value_pos(Gtk.PositionType.BOTTOM)
         slider.connect('value-changed', callback, constant)
         grid.attach(lwidget, 1, index, 1, 1)
-        grid.attach(slider, 2, index, 1, 1)
+        grid.attach(slider, 2, index, width, 1)
         return slider
 
     def add_spinner(self, grid, label, index, constant, range, callback=None,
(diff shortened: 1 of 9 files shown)


sam-m888:
Much better

2016-11-12 12_08_05-ubuntu-14 04 1-grampsdev-amd64 running - oracle vm virtualbox

sam-m888:
Tested 1 to 5 and assume that that 6 works. As well as fixing 9771's issue.

Thank you

PR #222, opened 2016-08-28
Feature: Gep-030 FanChart2Way

sam-m888:
...New FanChart consisting of both ascendants and descendants.

It can be checked-out there :
https://sourceforge.net/u/bubblegum00/gramps/ci/geps/gep-030-FanChart2Way/~/tree/

<snip>

I find it quite handy, so please let me know if you have any comments or suggestions.

NB: I heavily cleaned up the FanChart code on my way.

Regards,

Bastien Jacquet

https://sourceforge.net/p/gramps/mailman/message/32908110/
........................................................
Refactor fanchart for further modification

Fix radial text pos_start radian alignment

Simplify father's and mother's details getter

Rename draw_gradient to draw_gradient_legend

refactor prepare_background_box

Add option not to flip names

Add a display_format(self, person, num) function as proposed by Benny Malengier on 2012-12-13
It just returns name_display.name_formats[num][_F_FN](person.get_primary_name%28%29)

Add option to show the name on two lines

Allow for variable radius depending on generation

Switch to WORD_CHAR wrapping of name (ie word, and char if 0-length word)

Move rescaling tentative inside wrap_truncate_layout

Fix person_under_cursor bugs

Refactor root angle computation

Refactor code positionning the fan

Refactor personpos_at_angle

move implementation of person_under_cursor outside of FanChartBaseWidget class

Change draw_person to take angles in radians

Use same structure for innerring as for outerring

Uses cursor_to_polar and cursor_on_tranlation_dot

Slightly change person_under_cursor logic to return an "address" in the fan

Uses radian_in_bounds to compare angles modulo 2 PI

Fixup test on cursor over inner ring

Fix Center size for FanchartDesc

Fixup fanchart check up to last generation

Give same signature to draw_person

Refactor the common code of self.draw_person in a single function in Base class

Fix center box comment

Refactoring inside celladdress

Remove manual central box drawing since done with draw_person

Fixup draw_person color for duplicates

Use draw_person for central person too

Make __compute_angle and __rec_fill_data public for use in FanChart2Way

Add 2Way View

Rewrite create_map_rect_to_sector to allow bottom-outside-oriented text-arc

Allow to automatically right upside-up bottom arc-text

Correct icons for Fanchart2Way

Small code refactoring

Refactor code of fanchartdesc to use self.rootangle_rad

Rename change_slice to toggle_cell_state

Fanchart2Way code formating and changes

Small refactoring of fanchartdesc innerring fill data

Remove the name from the local temporary data structure

[...]

PR #269, opened 2016-11-09
fix name_displayer.add_name_format so it won't create duplicates

prculley:
I found a small bug caused in part by the fan chart code [https://github.com/gramps-project/gramps/pull/222](https://github.com/gramps-project/gramps/pull/222) Seems if you run the three types of fan chart (normal, descendant, 2-way) you end up with three copies of new name formats (Edit/Preferences/Display Name format: then Edit the name formats) and see 'Surname' and ''Given Suffix" more than once. I think the bug has been there forever,

This checks if the name format is already present, and returns the number of the format if so.  Other wise add as before.

### DIFF
--- a/gramps/gen/display/name.py
+++ b/gramps/gen/display/name.py
@@ -433,6 +433,9 @@ class NameDisplay:
         self.set_default_format(self.get_default_format())
 
     def add_name_format(self, name, fmt_str):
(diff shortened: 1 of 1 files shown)


sam-m888:
Thank you

PR #270, opened 2016-11-13
Re-enable selection in MultiTreeView on a drag-end event

tomhughes:
In 201f4a5e517ae5d84fbbde95113ecdf9c7e7bb62 we made grab_broken re-enable selection because GTK 3.18 and later no longer send a fake button release event.

The sending of grab broken is an implementation detail of DnD on X based systems however, and at least some GTK backends do not generate this, wayland being one example.

So we now use drag-end instead which is always sent to the source when a DnD operation completes.


eno93:
@Nick-Hall Does this make sense to you?

Nick-Hall:
Yes, it make sense to me.  Merged da018c184283fb2f358f1f45cf295a2ddda2985b.

PR #272, opened 2016-11-14
bug 9787; fix right-click menus crash caused by PR268

prculley:
PR268 removed icons from right-click menus in a lot of places.   I did not spot that some of the changes were to classes that were inherited. So I need to edit the inheriting classes as well where they overrode the method.


bpisoj:
Menus are too narrow for my taste.
If "reserve_toggle_size" is off can we do some padding manually?

PR #273, opened 2016-11-15
Fix translation confusion on 'View' noun/verb

prculley:
I previously changed the right-click menu on the gallery tab from icons to text; I apparently did not know that the word 'View' could be confusing in other languages (to look at, vs. something to see).  So tweaked the string to include a hint to translators.


vantu5z:
Looks good for me.

romjerome:
Thanks!

PR #274, opened 2016-11-16
9378: Quick Backup: remember Media (include/exclude) setting

SNoiraud:


gramps/gui/viewmanager.py
```diff
@@ -1352,6 +1352,11 @@ def quick_backup(self, obj):
                                                                 _("Exclude"))
         include.connect("toggled", lambda widget: self.media_toggle(widget,
                                                                     file_entry))
+        include_mode = config.get('preferences.quick-backup-include-mode')
+        if include_mode:
+            include.set_active(True)
+        else:
+            exclude.set_active(True)
```
dsblank: I think that should be False

eno93:
This works for me 👍

sam-m888:
Thank you

PR #275, opened 2016-11-16
8250 and 9736: avoid all other operations during export and fixes for export popup windows hidden.

prculley:
This is an alternate solution to PR267.  It builds on Serge Noirauds idea of making the export dialog modal, and further adds code to ensure that only a single window at a time can be modal.

The modal request is integrated into the ManagedWindow class.

Note: It is clear that multiple modal windows in a parent/child relationship does not work well in Gtk; For future work, I hope that developers use the ManagedWindow class for new windows (instead of just popping up a Gtk Window or dialog), then the fixes in our class can deal with the problem.

I also spent most of my time trying to figure a reasonable way to make all our quick reports use the ManagedWindow class the way it was intended; the original authors apparently did not feel that was too important.

This involved making sure that the 'track', used by ManagedWindow class was properly located and entered as a parameter when opening a new ManagedWindow.

I also changed code to ensure that not only quick reports, but filter edits, were properly tracked as both are opened from export dialog.

And I changed the code so that the second level quick report doesn't pop up a right-click menu (Serge already prevented that menu from doing anything, but I judged it better to avoid the menu altogether rather than having menu entries not work).  This involved a bit of refactoring in _quicktable.

I've tested every Gui combination I can think of in the export dialog, as well as testing that quick reports in general still seem to work correctly.

BTW, I see that Serge has updated his PR since I last tested, I will test that as well and report on my results a bit later (I may run out of time as have some non-Gramps things to take care of).

### DIFF
--- a/gramps/gen/plug/docgen/basedoc.py
+++ b/gramps/gen/plug/docgen/basedoc.py
@@ -60,7 +60,7 @@ class BaseDoc(metaclass=ABCMeta):
     such as OpenOffice, AbiWord, and LaTeX are derived from this base
     class, providing a common interface to all document generators.
     """
-    def __init__(self, styles, paper_style):
+    def __init__(self, styles, paper_style, track=[]):
         """
         Create a BaseDoc instance, which provides a document generation
         interface. This class should never be instantiated directly, but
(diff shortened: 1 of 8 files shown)


Paul-Franklin:
While I do not have time to test this, I commend your idea of
migrating as many windows and dialogs to a ManagedWindow
as possible.  Getting existing ones to work better is a plus.  8-)

Thanks.

SNoiraud:
It works for me.

Nick-Hall:
Nice work @SNoiraud and @prculley !

I expect that quick reports started as very simple reports invoked only from the list views.  They were used in the export later when the preview was added.

sam-m888:
My testing showed that it works as advertised.

@prculley You mention `**I will test that as well and report on my results a bit later**` so I will not merge until you report back.

prculley:
Looks to me like @SNoiraud has closed his PR for the other approach to this, so I think that testing it now is not necessary.  I guess he thought this was ok...

sam-m888:
Thanks 👍

PR #276, opened 2016-11-22
Create module for database utilities

Nick-Hall:
Move all database utilities into a single module.

### DIFF
--- a/gramps/cli/arghandler.py
+++ b/gramps/cli/arghandler.py
@@ -46,7 +46,7 @@ import re
 from gramps.gen.recentfiles import recent_files
 from gramps.gen.utils.file import rm_tempdir, get_empty_tempdir
 from .clidbman import CLIDbManager, NAME_FILE, find_locker_name
-
+from gramps.gen.db.utils import make_database
 from gramps.gen.plug import BasePluginManager
 from gramps.gen.plug.report import CATEGORY_BOOK, CATEGORY_CODE, BookList
 from .plug import cl_report, cl_book
@@ -501,7 +501,7 @@ class ArgHandler:
                 else:
                     self.imp_db_path = get_empty_tempdir("import_dbdir")
                     dbid = config.get('database.backend')
-                    newdb = self.dbstate.make_database(dbid)
+                    newdb = make_database(dbid)
                     newdb.write_version(self.imp_db_path)
 
                 try:
--- a/gramps/cli/clidbman.py
+++ b/gramps/cli/clidbman.py
@@ -49,6 +49,7 @@ from gramps.gen.plug import BasePluginManager
 from gramps.gen.config import config
 from gramps.gen.constfunc import win
 from gramps.gen.db.dbconst import DBLOGNAME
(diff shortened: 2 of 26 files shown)


prculley:
Moving some of this around is going to effect at least some of our addons...

Nick-Hall:
I created a PR to get comments.  My main concern was that `import_as_dict` was in the wrong place.  I moved the other utilities at the same time.

@prculley Do you think that I should abandon this?

eno93:
Good changes, keep going ?? 

One function **import_from_filename** I was using that's a candidate to be moved here was removed 473d3aab48841a4516ca861553e13efcc364e02a by you and I have to revert on my personal tree, which is used to import a file while using gramps with the ipython/Jupyter Notebook by @dsblank , it's how I found your project via [reddit](https://www.reddit.com/r/gramps/comments/39dbew/gramps_50_from_outside_gramps_gtk_ipython_notebook/) and how I use gramps mostly to explore the api. 

Or is there another way in ipython/jupyter I can do the same thing?

BTW the link in that reddit is gone/outdated and you can see the notebook [here](https://athena.brynmawr.edu/jupyter/hub/dblank/public/jupyter.cs/Genealogy/Gramps%205.0,%20Getting%20Started.ipynb)

uxcq:
@eno93 What did you do to get around https://github.com/gramps-project/gramps/commit/c9ae621d612bd798b99dc2ef303f579e70b42553 being removed?

prculley:
The reason I commented on addons possibly using some of the items being moved (or for other API deprecations) is that I think it would be a courtesy to go through the addon sources and make the same changes there.  If we don't then there will be a lot of bugs filed as folks discover that addons don't work.  And not every addon author watches for every commit for changes that might affect their addon.

Nick-Hall:
Only two addon files will need to be changed.

Nick-Hall:
@eno93 Thanks for the updated link.

The problem with `import_from_filename` is that it really belongs in the `cli` not the `gen` module.  In fact the `import_new_db` method in the `CLIDbManager` class does something very similar.

I would like to keep this command line functionality working, but it may be better to use `CLIDbManager`.

Nick-Hall:
@uxcq The usual way of using transactions is:

    from gramps.gen.db import DbTxn
    with DbTxn('', db) as trans:
        db.commit_person(person, trans)

kulath:
Yes, I think it is a very good idea to remove utilities from where they don't belong. One of my complaints was that make_database had nothing to do with dbstate, so moving it (and similar things) out is good.


One separate thing I noticed was that having code in _init_ modules made it very difficult to get decent pylint scores, but maybe you have sorted that now. (I make this comment, because the _init_ module here seems to have more code in it than usual).

kulath:
The comment on make_database could do with improving, to make it clear that it creates and returns a Database class instance, but that the database itself is not actually created or populated  (on disk or anywhere else).

eno93:
Thanks for https://github.com/gramps-project/gramps/pull/277  👍

PR #278, opened 2016-12-01
bug 9811; Fix for Relative media path incorrectly calculated for Windows

prculley:
This issue (apparently matching base and target initial media paths, did not collapse correctly) was created as a side effect of supporting environment variable expansion in the base media path.
The base media path got 'os.path.normcase()' (on Windows this makes it lower case, since Windows paths are case insensitive). The actual new media path did not get that treatment. So if the initial parts of the path were not all loser case, they would not compare and got the '../..' treatment.

From Bug report:
I have the Base path for relative media path set in preferences as:
C:\Users\Public\Genealogy\Media

While adding a new media file, with Convert to relative path checked, this is what results:
..\..\..\..\Users\Public\Genealogy\Media\Census\1860 ME, Oxford Co, Andover 7.jpg

### DIFF
--- a/gramps/gen/utils/file.py
+++ b/gramps/gen/utils/file.py
@@ -137,12 +137,16 @@ def relative_path(original, base):
     base_list = [_f for _f in base_list if _f]
     target_list = [_f for _f in target_list if _f]
     i = -1
+    # base path is normcase (lower case on Windows) so compare target in lower
+    # on Windows as well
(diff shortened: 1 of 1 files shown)


sam-m888:
After testing the Reporter states:

> [The patch appears to work.](https://gramps-project.org/bugs/view.php?id=9811#c50090)

sam-m888:
Thank you

PR #279, opened 2016-12-01
fix for bugs 9768 and 9778, nonetype error when merging events or families

prculley:
These are another example of the issues caused when we decided to raise 'HandleError' whenever a 'None' handle was input. To correct the issue we've been either testing for 'None' before making the call, or surrounding the call with try/except.


sam-m888:
Thank you

PR #280, opened 2016-12-01
bug 9818 allow merging of families with one or more parents in common

prculley:
Family merge fails with "Cannot merge people" with "Spouses cannot be merged. To merge these people, you must first break the relationship between them."

This can occur if a user tries to merge two families where at least one of the parents is common (same parent person handle) between the two families.

This results in the current code trying to merge a person with himself.

The family merge code calls the person merge code for each parent; the writer of the person merge code did not expect anyone to try to merge the person with himself (cannot be done by GUI).

So a simple test to make sure a parent to be merged is himself before calling the merge code can eliminate the problem.

### DIFF
--- a/gramps/gen/merge/mergefamilyquery.py
+++ b/gramps/gen/merge/mergefamilyquery.py
@@ -134,15 +134,23 @@ class MergeFamilyQuery:
 
         with DbTxn(_('Merge Family'), self.database) as trans:
 
-            phoenix_father = self.database.get_person_from_handle(self.phoenix_fh)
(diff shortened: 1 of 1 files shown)


sam-m888:
Needs rebasing,

prculley:
This PR was superseded by fixes applied in [https://github.com/gramps-project/gramps/commit/e584704e7be0e2f77d8815137395cc3ea7d089b2#diff-7b86ced4afd5da122c611e83c6680e02](url)
so I am closing it.

PR #281, opened 2016-12-04
Feature:  New global [Add]>[New] menu with Keybindings

sam-m888:
Possible New global [Add] menu with Keybindings.

Came across this older experiment of mind, placing here for testing, comment and suggestions?

Really like using the keyboard and not the mouse.  (Many of the keybindings may look familiar to TMG users)

Only the following menu options work ( Person / Family / Event / Place / Source / Citation and new source / Repository / Media / Note )

Keybindings work in any category, although from my notes I hadn't tracked down why I get:
/gramps/gui/grampsgui.py:412: Warning: g_value_get_int: assertion 'G_VALUE_HOLDS_INT (value)' failed
  Gtk.main()

From my notes I Hadn't worked out:
* How to Open a "New Event" with an "Event type" set to something other than the default?

How to do the following:
_Husband       (for the current active person)
_Wife          (for the current active person)
Sp_ouse        (for the current active person ?? auto selects wife or husband based on setting)
_Father        (for the current active person)
_Mother        (for the current active person)
_Brother        (for the current active person)
Sis_ter        (for the current active person)
Siblin_g        (for the current active person ??? new person with unknown gender)
_Son        (for the current active person)
_Daughter        (for the current active person)
_Child        (for the current active person ??? new person with unknown gender)
_Unrelated     (new unrelated person - possibly same as new _Person?)

Screenshot of the **Add** Menu 
possible new global add menu with keybindings

Related feature request:
  * [2732: keyboard shortcut for adding an existing source/event/place...](https://gramps-project.org/bugs/view.php?id=2732)

Existing documentation:
[Gramps_4.2_Wiki_Manual_-_Keybindings](https://gramps-project.org/wiki/index.php?title=Gramps_4.2_Wiki_Manual_-_Keybindings)

[10041](https://gramps-project.org/bugs/view.php?id=10041)

### DIFF
--- a/gramps/gui/viewmanager.py
+++ b/gramps/gui/viewmanager.py
@@ -91,6 +91,12 @@ from .aboutdialog import GrampsAboutDialog
 from .navigator import Navigator
 from .views.tags import Tags
 from .actiongroup import ActionGroup
+from gramps.gen.lib import (Person, Surname, Family, Media, Note, Place,
+                            Source, Repository, Citation, Event, EventType,
+                            ChildRef)
+from gramps.gui.editors import (EditPerson, EditFamily, EditMedia, EditNote,
+                                EditPlace, EditSource, EditRepository,
+                                EditCitation, EditEvent)
 from gramps.gen.db.exceptions import DbWriteFailure
 from .managedwindow import ManagedWindow
 
@@ -127,6 +133,23 @@ UIDEFAULT = '''<ui>
     <separator/>
     <menuitem action="Abandon"/>
     <menuitem action="Quit"/>
+  </menu>
+  <menu action="AddMenu">
+    <menu action="AddNewMenu">
+    <separator/>
+    <menuitem action="PersonAdd"/>
+    <separator/>
+    <menuitem action="FamilyAdd"/>
+    <separator/>
+    <menuitem action="EventAdd"/>
+    <separator/>
+    <menuitem action="PlaceAdd"/>
+    <menuitem action="SourceAdd"/>
+    <menuitem action="CitationAdd"/>
+    <menuitem action="RepositoryAdd"/>
+    <menuitem action="MediaAdd"/>
+    <menuitem action="NoteAdd"/>
+    </menu>
   </menu>
   <menu action="EditMenu">
     <menuitem action="Undo"/>
@@ -574,6 +597,27 @@ class ViewManager(CLIManager):
         self._action_action_list = [
             ('Clipboard', 'edit-paste', _('Clip_board'), "<PRIMARY>b",
              _("Open the Clipboard dialog"), self.clipboard),
+            ('AddMenu', None, _('_Add')),
+            ('AddNewMenu', None, _('_New')),
+            ('PersonAdd', None, _('_Person'), "<Alt>p", None,
+             self.add_new_person),
+            ('FamilyAdd', None, _('Famil_y'), "<Alt>y", None,
+             self.add_new_family),
+            ('EventAdd', None, _('_Event'), "<shift>e", None,
+             self.add_new_event),
+            ('PlaceAdd', None, _('_Place'), "<shift><Alt>p", None,
+             self.add_new_place),
+            ('SourceAdd', None, _('_Source'), "<shift><Alt>s", None,
+             self.add_new_source),
+            ('CitationAdd', None, _('_Citation'), "<shift><Alt>c", None,
+             self.add_new_citation),
+            ('RepositoryAdd', None, _('Repositor_y'), "<shift><Alt>y", None,
+             self.add_new_repository),
+            ('MediaAdd', None, _('_Media'), "<shift><Alt>m", None,
+             self.add_new_media),
+            ('NoteAdd', None, _('_Note'), "<shift><Alt>n", None,
+             self.add_new_note),
+            #--------------------------------------
             ('Import', 'gramps-import', _('_Import...'), "<PRIMARY>i", None,
              self.import_data),
             ('Tools', 'gramps-tools', _('_Tools'), None,
@@ -1354,6 +1398,87 @@ class ViewManager(CLIManager):
             ClipboardWindow(self.dbstate, self.uistate)
         except WindowActiveError:
             return
+
+    # ---------------Add new xxx --------------------------------
+    def add_new_person(self, obj):
+        """
+        Add a new person to the database.  (Global keybinding)
+        """
+        person = Person()
+        #the editor requires a surname
+        person.primary_name.add_surname(Surname())
+        person.primary_name.set_primary_surname(0)
+
+        try:
+            EditPerson(self.dbstate, self.uistate, [], person)
+        except WindowActiveError:
+            pass
+
+    def add_new_family(self, obj):
+        """
+        Add a new family to the database.  (Global keybinding)
+        """
+        family = Family()
+        try:
+            EditFamily(self.dbstate, self.uistate, [], family)
+        except WindowActiveError:
+            pass
+
+    def add_new_event(self, obj):
+        """
+        Add a new custom/unknown event (Note you type first letter of event)
+        """
+        try:
+            event = Event()
+            event.set_type(EventType.UNKNOWN)
+            EditEvent(self.dbstate, self.uistate, [], event)
+        except WindowActiveError:
+            pass
+
+    def add_new_place(self, obj):
+        """Add a new place to the place list"""
+        try:
+            EditPlace(self.dbstate, self.uistate, [], Place())
+        except WindowActiveError:
+            pass
+
+    def add_new_source(self, obj):
+        """Add a new source to the source list"""
+        try:
+            EditSource(self.dbstate, self.uistate, [], Source())
+        except WindowActiveError:
+            pass
+
+    def add_new_repository(self, obj):
+        """Add a new repository to the repository list"""
+        try:
+            EditRepository(self.dbstate, self.uistate, [], Repository())
+        except WindowActiveError:
+            pass
+
+    def add_new_citation(self, obj):
+        """
+        Add a new citation
+        """
+        try:
+            EditCitation(self.dbstate, self.uistate, [], Citation())
+        except WindowActiveError:
+            pass
+
+    def add_new_media(self, obj):
+        """Add a new media object to the media list"""
+        try:
+            EditMedia(self.dbstate, self.uistate, [], Media())
+        except WindowActiveError:
+            pass
+
+    def add_new_note(self, obj):
+        """Add a new note to the note list"""
+        try:
+            EditNote(self.dbstate, self.uistate, [], Note())
+        except WindowActiveError:
+            pass
+    # ------------------------------------------------------------------------
 
     def config_view(self, obj):
         """


uxcq:
Yes, this the missing menu Gramps needs, works really well and faster than a mouse of course! 👍 

Have a look at [editfamily.py](https://github.com/gramps-project/gramps/blob/master/gramps/gui/editors/editfamily.py) for some ideas.

Nick-Hall:
>  I hadn't tracked down why I get:  /gramps/gui/grampsgui.py:412: Warning: g_value_get_int: assertion 'G_VALUE_HOLDS_INT (value)' failed Gtk.main()

This is a Gtk3 problem.  See bug [#8029](https://gramps-project.org/bugs/view.php?id=8029).  I attempted to implement the suggested workaround, but it looks like I missed a `set_text` somewhere.

Nick-Hall:
This is a good idea and would be easy to implement.

* _Person_ and _Unrelated_ are the same.
* _Husband_ and _Wife_ are not really needed.  The gender of the current person can be used with the _Spouse_ option.
* Why would you want to add a _Family_ without adding a spouse or child at the same time?
* How would we add a _Child_ to a person with multiple marriages?
* A new _Citation_ could use an existing source.
* When creating a new _Event_, the default event type is intelligent.  However, setting a default is easy.

Some of this functionality already exists in the relationship view.

sam-m888:
Went through and reduced the options on the initial version of the menu, so it can be added sooner than later.  And mostly because I feel I lack the experience to do this correctly.

initial-add-menu-reduced-options

> * _Person_ and _Unrelated_ are the same.

Ok, removed unrelated.

* _Husband_ and _Wife_ are not really needed.  The gender of the current person can be used with the _Spouse_ option.

Ok, removed husband and wife. Experimented for half an hour to get spouse from relview.py working (I didn't!) but got lost working out self.get_active for relview.

> * Why would you want to add a _Family_ without adding a spouse or child at the same time?

?  Normally I'd goto the Families category and add a new family and from it add the family ? I've left it in the menu, say if you believe otherwise.

> * How would we add a _Child_ to a person with multiple marriages?

Hmm. good point, I was thing along the lines of using a selector that show each of the multiple spouses to be selected.

> * A new _Citation_ could use an existing source.

Made it just "Citation" as I can see that from the new citation you can already add an existing source or create a new source.

> * When creating a new _Event_, the default event type is intelligent.  However, setting a default is easy.

Ran out of time for at this point to explore, hopefully when I look again it really is that easy :)

@uxcq @Nick-Hall Thank you both for your suggestions.

prculley:
Kind of a side note;  did you realize that this new menu item (and all other top level menus) are using deprecated APIs?  They have deprecated since GTK 3.10.  I wonder if we should be putting an update in this area on our GEPS list?

Nick-Hall:
@sam-m888 To add a new event with a given event type use:
```
event = Event()
event.set_type(EventType.BIRTH)
```
Then invoke the editor as usual.

Nick-Hall:
@prculley The UIManager class is now deprecated.  We could define the menu bar in a glade file instead, but this probably needs more thought.  Feel free to create a GEP.

sam-m888:
@Nick-Hall Thanks that worked added back BMDB menu entries 👍 

@prculley No did not realise that, won't what ever it changes to still be a menu?

daleathan:
Very handy :) Will this make it into alpha 2?

sam-m888:
> Very handy :) Will this make it into alpha 2?

Thanks,   Possibly but no guarantees.  

This PR is ready to go, as is with the reduced options shown in the screenshot below, the other options (Add: Spouse / Father / Mother / Brother / Sister /  Sibling / Son / Daughter / Child ) will have to wait until I figure out how to get the current active person( as it seems my knowledge of OOP is not very strong!)

new-add-menu-gramps-master

Nick-Hall:
@sam-m888 To get the handle of the active person use:
```
handle = self.uistate.get_history('Person').present()
```

Nick-Hall:
@daleathan We should be able to get this into alpha-2.

sam-m888:
@Nick-Hall Thanks that makes more sense, I was thinking it was the "Make Active Person" from the Edit Person dialog!

sam-m888:
Decided to separate the [New] and [Existing] in the menu, so that it is more obvious was is going to happen when you select the menu.

new add menu submenu new

Before I look at doing it what do you think about:

* Adding just between the "New" and "Partner(Spouse)" menu entries the Active Person's name and ID [maybe try to adapt how the bookmark code shows the person entry  or the status bar, or is there a better way?] 

If that can be done then to clearly show the actions below are related to the active person eg Spouse to Child;  move them under there own [New] menu entry (which can be expanded to allow adding the other first [New] menus option that apply only to the active person? And then add a separate menu entry for [Existing] eg: Parents/../Sibling/..Child  that would bring up the selectors for each?

New menu layout examples:

Menu:
******* Add > New > (as per image)
******* Add > (to existing "Active Person") "to: [I0049] Smith, John" > New > (as per image except it means "new action/record for the active person)
******* Add >to: "[I0049] Smith, John" > Existing > (as per image) Spouse(Partner) etc...

Or just leave it alone, if this menu change @prculley mentioned above change  how the menu work totally?

enlite93:
Looks interesting 👍 

If you do the existing person menu like you mention, it has potential to be applied to the other existing records in the same way.

Then maybe later under the Edit menu incorporate similar options to delete records and merge? :)

uxcq:
@sam-m888 I'd like to suggest that you split this into two parts  and commit the first(Add>New), while you work on the second(Add to existing)?
(1) Adding [New] records  (which already I cannot do without)
(2) Add to Existing Active Person

sam-m888:
Possibly, if I have sometime tomorrow I'll will look at doing that.

I'm actually stuck on the second part at the moment.

Nick-Hall:
@sam-m888 What help do you need?

pat49:
Really good, so obvious in hindsight!

uxcq:
Like how you changed **add>new>event**  to be unknown/custom instead of being a repeat of the birth event 👍 

Thank you for splitting the patches.

sam-m888:
@Nick-Hall Do you believe the first part https://github.com/gramps-project/gramps/pull/281/commits/a4e1a008e7db93496c85f98dcfd562d13f5c0811 is ready to be added?

Nick-Hall:
@sam-m888 Would the events be better in a sub-menu?  How about a baptism option?

sam-m888:
> @sam-m888 Would the events be better in a sub-menu?

@Nick-Hall Will give events as a 4th level sub-menu a try over the weekend. 

Shouldn't we avoid. too many sub-menu levels from the usability perspective?

>  How about a baptism option?

I can add a baptism entry (Note that I stopped myself initially from doing this believing that the standard "BMDB"'s would be enough)

sam-m888:
Added Baptism as an option and moved all event into separate 4th level sub-menu.

new-add-new-menu-4th-level-events

It does look better group under it's own menu.

ennoborg:
I don't like this, because it suggests that baptism is a common event where in fact it is not. Most other programs, including good old PAF, use Christening instead.

uxcq:
The defaults for the  **New Event** window in Gramps v4.2.5 only shows Christening  if you go to the Religious fly out.  Is the default in Gramps wrong or should Christening also be listed in that top section?

defaults for new event - gramps4 2 5

You might want to read the following [Wikipedia:Infant baptism](https://en.wikipedia.org/wiki/Infant_baptism)  which mentions _"Most Christians belong to denominations that practice infant baptism"_

Nick-Hall:
@uxcq The default in Gramps is _Baptism_ which is categorised as a [Life Event](https://github.com/gramps-project/gramps/blob/master/gramps/gen/lib/eventtype.py#L137).  This has been the case since v3.4.

We should probably include all Life Events in the menu for consistency.

Nick-Hall:
@sam-m888 That looks good!

Do we really need the top-level menu?  Is creating an event without the active person as a participant useful?  All people added in a future person sub-menu, except for _Unrelated_, will be related to the active person.

ennoborg:
Baptism is wrong, because in the GEDCOM definition only Christening is associated with the naming of a child: "CHR {CHRISTENING}:=The religious event (not LDS) of baptizing and/or naming a child.". Baptism is just the water thing.
My preference is supported by the fact that PAF, being the 1st LDS approved program, has Christening as a fixed life event in the person editor window, after Birth, and before Death and Burial, and Ancestral Quest, which PAF was derived from does the same. On other words, to me it looks like Christening is the de-facto standard. It is also shown as the closest equivalent of birth in the individual event definition:
n [ BIRT | CHR ] [Y|<NULL>] {1:1}
+1 <<INDIVIDUAL_EVENT_DETAIL>> {0:1}* p.34
+1 FAMC @<XREF:FAM>@ {0:1} p.24
Note that this is not about religion, but about traditions in genealogy, and the use of Christening is a real tradition, set in programs much older than Gramps, and in GEDCOM itself. And I think that users will appreciate it if their data appears in the proper place in other software. All the books where I have found 'baptisms' are in fact Christian church books, so based on the number sources alone, there is no doubt in my mind that Christening is the proper event.
I acknowledge that outside the Christian world, it may be better to let the user decide what life events should be shown.

sam-m888:
[9300: Gramps assumes wrong type of christening i GEDCOM output.](
https://gramps-project.org/bugs/view.php?id=9300)

Nick-Hall:
@ennoborg This is getting off-topic for this PR.  In English, baptism and christening have slightly different definitions - neither of them is wrong.  Both Gramps and GEDCOM support both event types.  I would suggest recording them as described in the source document.

makesson:
> @daleathan We should be able to get this into alpha-2.

Any idea when alpha2 will be released as the exiting alpha1 is from june last year?

Sorry if this is off topic.

sam-m888:
> We should probably include all Life Events in the menu for consistency.

@Nick-Hall  Do you mean an entry for each?  
And an additional option for "Christening", may as well as it is a submenu?

> Do we really need the top-level menu? Is creating an event without the active person as a participant useful? All people added in a future person sub-menu, except for Unrelated, will be related to the active person.

Having problem parsing this? I believe you mean don't make the "To existing: active person" a separate sub-menu right?

@makesson Probably makes sense to rename [Add][New]  to [Add][Unrelated]  as its implied that an Unrelated person would be new!


Yes it is a little noisy here, please attempt to focus on the Feature please.

ennoborg:
IMO, the choice of life events created a random (and unfounded) distinction between baptism and christening, and all other the religious events, so IMO a correct solution should be to take either baptism out, or put all the others in. The current solution will lure users into adding baptisms that don't show up as life events in programs like Ancestral Quest or the FamilySearch Tree, and will hence lead to lots of frustrations. This is even more so for other languages like my own, in which Baptism and Christening are translated to Doop and Doopsel, and those are terms that are seen as the same on Dutch wikipedia. In other words, Dutch users, and probably German users too, will add non standard Baptism events without even knowing it.
I define standard here as defined by GEDCOM and the LDS, where Christening is the proper event, for what Dutch call a 'doop'.

kulath:
Just a minor point; I would change the menu items to:
Add -> Add new
Add -> Add to existing Active Person new

I know that this means the word Add appears twice, once in the top level menu, and once at the start of the submenu, but I think this makes it clearer what is going on.

It emphasises that you are going to add a /new/ object (you don't get the option to choose an existing object). Just starting the wording with "to existing..." is rather unclear - what are you doing to the existing person. Also I speculate that the existing wording may be rather difficult to translate. The current ":Active Person" seems to me to be a bit difficult to parse.

kulath:
If you are going to have a separate 4th level sub-menu for the events, I don't see any reason not to include all the possible events, with the same structure as for the new event window (even if this means a 5th level sub-menu). Otherwise, how do you justify including some events but not others?

I don't think it is necessary to have shortcuts for all the events - even if that does raise questions as to how those that do have shortcuts are chosen.

(Just a view - I am quite prepared to be talked out of it).


A separate point: we need to check that the shortcuts work OK in Windows and Mac (is it possible to input the shortcuts, and are they already dedicated to some OS function?)

uxcq:
> I know that this means the **word Add appears twice**, once in the top level menu, and once at the start of the submenu, but I think this makes it clearer what is going on.

@kulath In that case for discussion(brain storming), why not just rename the top level menu instead to something like the word **[Action]** (maybe use a better word than this?) then it allows you to include other actions, like [Merge], [Remove (reference)],  [Delete]  Also consider if you can add things like, "Set Home person", "Set Active Person" and "Tag"( is it really an edit operation?),  How about including a menu for the drag & drop actions eg something like [Clone(object)], or duplicate?

Nick-Hall:
@kulath I have no objection to adding more event types.  Using the same structure used in the event editor is an option.

@ennoborg Eventually, I would like to give the user the ability to define their own event categories.

Nick-Hall:
@sam-m888 Is it useful to allow an event to be created that is not linked to the active person?  The user would probably want to link it to a person as the next step.

kulath:
@Nick-Hall "Is it useful to allow an event to be created that is not linked to the active person?"

I definitely think it is useful to support _some_ objects that are not linked to the active person. For example Place or Source, so that the user can think about creating these separately from their subsequent use. If _some_ objects can be created separately, then I think it is better to allow all of them so the user is not puzzled about why some and not others. Hence I think you need both
Add -> Add new
Add -> Add to existing Active Person new

Nick-Hall:
@kulath For places, it would be useful to create a new place enclosed by the active place.

makesson:
> @uxcq Can a version of this be added soon at least.

👍 🥇 

@uxcq Suggest maybe make a new comment instead of editing your old comment.

sam-m888:
I'll make one final attempt at this (next week) with the following change:

* Removing the Events submenu altogether and only show "Event" with the default keybinding of "Shift+alt+E" which will bring up the default "New event" dialog where the event type box is in focus and you just type the first letter and use the up/down keyboard arrows to select the event you want as well as just overtyping for custom events (as per this screenshot).

2017-04-30 13_28_00-ubuntu-16 04 1-grampsdev-amd64 running - oracle vm virtualbox


Note if this not acceptable then this PR is up for adoption only because I have it working as I like for my personal branch in the way mentioned above.

I'd really hope this can be shared with normal Gramps users in the Gramps 5 alpha2.

eno93:
> Removing the Events submenu altogether and only show "Event" with the default keybinding 

@sam-m888 I'm in favour of this, especially as it gets around the [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding) going on above!

> you just type the first letter and use the up/down keyboard arrows to select the event you want as well as just overtyping for custom events (as per this screenshot)

Learned something new, this should be mentioned in the user manual with the screenshot example 👍

sam-m888:
Reduced menu as mentioned previously:

reduce-menu-final

eno93:
> Reduced menu as mentioned previously:

Thank you 👍

Paul-Franklin:
https://gramps-project.org/bugs/view.php?id=10055#c51829

PR #285, opened 2016-12-07
bug 8333; fix merge issue with Person Tree View

prculley:
During the debug process I found;
The Merge process creates a db transaction with a person delete, and a different person update (and a lot more stuff).
When the transaction closes, is when we get the signals from the db:
person_update
person_delete

When the person_update signal arrives, the treebasemodel code does an update_row_by_handle method, which in turn does
        self.delete_row_by_handle(handle)
        self.add_row_by_handle(handle)

during the delete_row_by_handle, some uistate.change_active callbacks are made to update the active person to the previous person in the list (which was the deleted person in the merge). However, the handle for the previous person is now invalid, since that delete commit happened already. So the active_person gets set to an invalid handle. And the various gramplets get assertions.

When these errors are dismissed by user, the add_row_by_handle eventually runs and everything returns to normal.

I fixed treebasemodel update_row_by_handle method to not do a 'change_active' during the delete. I've tried this and it seems to work for this case and every other treeview case I could come up with.
I sure wish we had a GUI test code suite so I could be sure I got this right.

I also noted that the People views for columns "number of children" and "number of marriages" did not update when marriages or children were added or removed.  So I added callbacks to the libpersonview module when something changes in families so that these can be updated correctly.

During testing of this change I noted additional bugs and fixed them:
* another handle error in listview
* bug 9832 Place sort issue when enclosing place name changes; the updated name might end up at a different position on the tree; I changed to rebuild the tree when this might happen.
* bug 9833 CitationTree sort issue when source name changes; the updated name might end up at a different position on the tree; I changed to rebuild the tree when this might happen.

### DIFF
--- a/gramps/gui/views/listview.py
+++ b/gramps/gui/views/listview.py
@@ -57,7 +57,7 @@ from .navigationview import NavigationView
 from ..actiongroup import ActionGroup
 from ..columnorder import ColumnOrder
 from gramps.gen.config import config
-from gramps.gen.errors import WindowActiveError, FilterError
+from gramps.gen.errors import WindowActiveError, FilterError, HandleError
 from ..filters import SearchBar
 from ..widgets.menuitem import add_menuitem
 from gramps.gen.const import CUSTOM_FILTERS
@@ -707,7 +707,13 @@ class ListView(NavigationView):
         """
         selected_ids = self.selected_handles()
         if len(selected_ids) > 0:
-            self.change_active(selected_ids[0])
+            # In certain cases the tree models do row updates which result in a
+            # selection changed signal to a handle in progress of being
+            # deleted.  In these cases we don't want to change the active to
+            # non-existant handles.
+            if hasattr(self.model, "dont_change_active"):
+                if not self.model.dont_change_active:
+                    self.change_active(selected_ids[0])
 
         if len(selected_ids) == 1:
             if self.drag_info():
@@ -804,10 +810,15 @@ class ListView(NavigationView):
             lookup_handle = self.dbstate.db.get_table_metadata(nav_type)['handle_func']
             for handle in selected_ids:
                 # Still exist?
-                if lookup_handle(handle):
+                # should really use db.has_handle(nav_type, handle) but doesn't
+                # exist for bsddb
+                try:
+                    lookup_handle(handle)
                     # Select it, and stop selecting:
-                    self.change_active(handle)
-                    break
+                except HandleError:
+                    continue
+                self.change_active(handle)
+                break
 
     def _button_press(self, obj, event):
         """
--- a/gramps/gui/views/treemodels/treebasemodel.py
+++ b/gramps/gui/views/treemodels/treebasemodel.py
@@ -283,6 +283,7 @@ class TreeBaseModel(GObject.GObject, Gtk.TreeModel, BaseModel):
         cput = time.clock()
         GObject.GObject.__init__(self)
         BaseModel.__init__(self)
+
         #We create a stamp to recognize invalid iterators. From the docs:
         #Set the stamp to be equal to your model's stamp, to mark the
         #iterator as valid. When your model's structure changes, you should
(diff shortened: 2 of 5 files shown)


Nick-Hall:
The node is deleted and then added again because it may change position in the tree.  In the person tree view this will happen if the surname is changed.  There is no easy way to see what change was made when a _person-update_ signal is received.

If you modify the treebasemodel, then you should also test the place and citation tree views.

You may have also noticed that the design is not good.  Usually the model would listen for database signals, not the view.

prculley:
Nick-Hall;  It figures I would miss that case (changed surname) in my testing.  I did do a number of other tests in place and citation tree view as well as I noted that they inherited treebasemodel.

I am open to suggestions on how to correct the original bug.  I considered trying to get the displaystate.set_active history updated (with handle_remove) during the delete_row, but could not find what seemed to be a reasonable place to do that.  As you noted the different layers of code don't always make doing some of the cross layer calls easy.

I also considered shutting down callbacks until after the merge, and then emitting a "database-changed" to just redo everything, but this seemed inelegant at best.

And last but not least, I considered updating **_all_** the gramplets to accept invalid handles and ignore them.  Since the bug also appears in at least the places tree view, I think I would have to do this with **_all_** of them.

prculley:
I update the initial comment to say what I changed.
I did a more careful testing of all three (PersonTree, PlaceTree, CitationTree) views.  
Tried merging all elements that can be merged while in these views (persons, places, source, citation).  Tried adding each element, and tried deleting each element.
Based on Nick's comments I also tested for renames of the various elements of the tree.
I found three additional bugs and patched them as well.
Still hope I did not miss any 'interesting' cases...

uxcq:
> I sure wish we had a GUI test code suite so I could be sure I got this right.


_To get automated tests of GTK+ code running on a headless server is Xvfb (virtual framebuffer X server). It provides the xvfb-run -a command which creates a temporary X server without the need for any real display hardware._

https://pygobject.readthedocs.io/en/latest/testing.html

kulath:
I also encountered this problem (merge A with B works, but merge B with A fails), and also noticed that the wrong way round results in no entry being highlighted when the merge completes. I tried to investigate months ago, and got as far as discovering that the problem was that the entry was accessed after it had been deleted, but couldn't work out how to fix it.

Paul, I very much admire your work, but in this case, I'm afraid I don't like the solution. It seems to me that the setting of the "dont_change_active" flag is too remote from it's use. So, the flag may be set in module A, which calls B which calls C which uses the flag. The problem I have seen in cases like this is that if someone calls B directly they don't know about the flag, or what it should be set to. I may not have got this situation quite right, but my point is about the use of a remote flag in general.

When I looked at it, I wondered whether it would be possible to fix it by testing whether the handle pointed to an existing object before using it. Sorry if you had already tried that and it didn't work.

prculley:
I originally tried to fix this by not deleting and then re-adding the object.  But Nick pointed out that certain cases (rename of sorted item) would not work if I did that.

I was not very happy with this fix for the same reason you are.  But given the way that the modules are separated I could not see an easy way to fix it otherwise.
By making the parameter a named one, and optional, I thought that most users of the 'change_active' would not be affected.

I looked into 'change_active' code detect a bad handle itself, but that code is in the generic list handling module and doesn't know what kind of handle it is.  Would have required hacking around to find out.  Also would have added yet another db access in this (very often called) path.

I also looked into making the users of the 'change_active' callback insensitive to the bad handle.  When I realized that every gramplet and the status bar (and probably other users) would have to be changed, I abandoned that approach.  I prefer to keep fixes 'surgical' if possible.

The fix I have does seem to work, but I am open to suggestions on a better way.

kulath:
Paul, I'm entirely happy with your explanation; I hadn't previously understood that getting change_active to do the detection would have a significantly adverse effect on performance (because it is called so often).

I had been working on fixing gramplets etc. to work properly when there is no database (including if you close the current database, or delete the current database or exit the manager without selecting a database). This would be a slightly related thing, but I don't have a timetable for it! It does entail fixes in a number of places, if only to add a handler for 'no database'.

PR #287, opened 2016-12-09
bug 9831; fix HandleError when using finddupes tool to merge

prculley:
Another in the many HandleErrors that occur because we chose to make an assertion instead of just returning None from db.get_*_from_handle...

### DIFF
--- a/gramps/gui/merge/mergeperson.py
+++ b/gramps/gui/merge/mergeperson.py
@@ -76,8 +76,8 @@ class MergePerson(ManagedWindow):
     Displays a dialog box that allows the persons to be combined into one.
     """
     def __init__(self, dbstate, uistate, handle1, handle2, cb_update=None,
-            expand_context_info=True):
-        ManagedWindow.__init__(self, uistate, [], self.__class__)
+            expand_context_info=True, track=[]):
+        ManagedWindow.__init__(self, uistate, track, self.__class__)
         self.database = dbstate.db
         self.pr1 = self.database.get_person_from_handle(handle1)
         self.pr2 = self.database.get_person_from_handle(handle2)
--- a/gramps/gui/merge/mergeperson.py
+++ b/gramps/gui/merge/mergeperson.py
@@ -75,8 +75,8 @@ class MergePerson(ManagedWindow):
     """
     Displays a dialog box that allows the persons to be combined into one.
     """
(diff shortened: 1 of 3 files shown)


Nick-Hall:
I would expect the deleted handles to be in `self.dellist`.  Try adding the following after line 591:
```
if p2key in self.dellist:
                continue
```

prculley:
Switched to Nick's suggested fix.
However, I don't understand why the objects need to be checked for None in line 604.  I'm pretty sure that this code should never be subjected to a filter.  The 'None' check had allowed it to work despite not completely killing all elements from the dellist, until we started asserting HandleError.

Also found that the windows could get 'unstacked' with MergePerson behind ShowMatches after I switched to a different application.  Found that the ManagedWindow was not set up correctly, and fixed it.

Nick-Hall:
> However, I don't understand why the objects need to be checked for None in line 604. I'm pretty sure that this code should never be subjected to a filter. The 'None' check had allowed it to work despite not completely killing all elements from the dellist, until we started asserting HandleError.

The check isn't needed.  Only proxies now return None, and this code doesn't use proxies.

It fact, it masked an error which you have now fixed properly.

Nick-Hall:
This simple fix is now spread over two commits.  We should have one commit for the HandleError fix and another for the window stak order.

prculley:
I'm pretty sure I got the changes unscrambled into their separate commits...  Sorry for my sloppiness, comes of doing things in a hurry.

Nick-Hall:
Thanks.  I have merged one of the two commits.

prculley:
Fixed for translation...

Nick-Hall:
It may be better to add `track` after `dbstate` and `uistate` to make the interface similar to that of the edit dialogs.  I have pushed a commit.

We could easily make a similar change to the other merge dialogs.  They do not appear to be used in third-party addons.

prculley:
I guess you prefer positional parameters; I thought that to avoid potential issues using an optional named parameter was safer.
Since you believe there are no other users, I have no objection, it seems to me to be a matter of style.

Nick-Hall:
I have made the changes in commit 13b53e35d41e0bcf43a3e0be24789904189bbd1f.

It is more a matter of consistency rather than style.  There is nothing wrong with using keyword arguments.

PR #288, opened 2016-12-09
bug 9478; fix quick search exception when nothing in searched list

prculley:
Our quick search (start typing a name when a list is active) can fail if the list is empty.  Not usual for someone to try to search an empty list, but can happen by mistake.
Added check for None on node to fix.

### DIFF
--- a/gramps/gui/widgets/interactivesearchbox.py
+++ b/gramps/gui/widgets/interactivesearchbox.py
@@ -440,6 +440,8 @@ class InteractiveSearchBox:
         search_column = self._treeview.get_search_column()
         is_tree = not (model.get_flags() & Gtk.TreeModelFlags.LIST_ONLY)
         while True:
+            if not cur_iter:    # can happen on empty list
+                return False
             if (self.search_equal_func(model, search_column,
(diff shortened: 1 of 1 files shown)


Nick-Hall:
> Also changed deprecated GoObject methods to GLib to reduce deprecation warnings. Pretty sure this is safe since other code in Gramps has already done this.

@prculley This is safe.  Gtk 3.10 requires GLib 2.37.5 or later.

Nick-Hall:
The removal of the deprecation warning should be in a separate commit.

prculley:
Nick;
Do you mean that I should separate the deprecation changes to a separate commit within this PR, or as part of a different PR?  I can do it either way.

I guess I've been lazy, when I see an issue while working on a different but related issue, I tend to fix it right away and test everything together; seems like it is easier and (possibly) more thorough.

If I have to set aside some changes for later and a different PR, I have to re-familiarize myself with the module again, and how to test it to make sure I did not screw up (which I do with depressing regularity).

I do understand from a documentation viewpoint that separate commits are useful, so will try to keep the pylint and deprecation stuff away from bug fixes.

PR #289, opened 2016-12-12
Shrink size of Break Lock (and other QuestionDialogs) by setting default width to 760.

prculley:
bug 9235:  Original 'Break Lock' dialog was greater than 1280 wide, since the text was long and had no line breaks.  Rather than add line breaks (and require new translations) I decided to set default dialog size to 760; looked at other dialogs and this seems to work OK.


uxcq:
Believe I found another too large dialog, uncheck the "Suppress warning when adding parents to a child" in Edit/preferences/warnings to see what I mean.

PR #290, opened 2016-12-13
bug 8510 fix for special characters in Family Tree names

prculley:
If special characters are present in Family Tree names, then the automatic backup fails, since the Family Tree name is part of the filename for the automatic backup.  

Default file names for reports are also effected.

The fix is not totally robust, since there are a number of other special cases regarding filenames (Try using 'con' or 'prn' on Windows, for instance), but should avoid most usual cases of trouble.
By placing the name substitution where I did, the user has an opportunity to see the change immediately and edit it himself.  It affects both 'New' and 'Rename' are corrected.

### DIFF
--- a/gramps/gui/dbman.py
+++ b/gramps/gui/dbman.py
@@ -36,6 +36,7 @@ import copy
 import subprocess
 from urllib.parse import urlparse
 import logging
+import re
 
 #-------------------------------------------------------------------------
 #
@@ -543,6 +544,8 @@ class DbManager(CLIDbManager):
         If the new string is empty, do nothing. Otherwise, renaming the
         database is simply changing the contents of the name file.
         """
+        # kill special characters so can use as file name in backup.
+        new_text = re.sub(r"[':<>|,;=\"\[\]\.\+\*\/\?\\]", "_", new_text)
         #path is a string, convert to TreePath first
         path = Gtk.TreePath(path=path)
         if len(new_text) > 0:


uxcq:
You may want to look at the robust solution Django project came up with.
```

def slugify(value, allow_unicode=False):
    """
    Convert to ASCII if 'allow_unicode' is False. Convert spaces to hyphens.
    Remove characters that aren't alphanumerics, underscores, or hyphens.
    Convert to lowercase. Also strip leading and trailing whitespace.
    """
    value = force_text(value)
    if allow_unicode:
        value = unicodedata.normalize('NFKC', value)
        value = re.sub(r'[^\w\s-]', '', value, flags=re.U).strip().lower()
        return mark_safe(re.sub(r'[-\s]+', '-', value, flags=re.U))
    value = unicodedata.normalize('NFKD', value).encode('ascii', 'ignore').decode('ascii')
    value = re.sub(r'[^\w\s-]', '', value).strip().lower()
return mark_safe(re.sub(r'[-\s]+', '-', value))
```

https://github.com/django/django/blob/master/django/utils/text.py#L419
```

def get_valid_filename(s):
    """
    Returns the given string converted to a string that can be used for a clean
    filename. Specifically, leading and trailing spaces are removed; other
    spaces are converted to underscores; and anything that is not a unicode
    alphanumeric, dash, underscore, or dot, is removed.
    >>> get_valid_filename("john's portrait in 2004.jpg")
    'johns_portrait_in_2004.jpg'
    """
    s = force_text(s).strip().replace(' ', '_')
return re.sub(r'(?u)[^-\w.]', '', s)

```
https://github.com/django/django/blob/master/django/utils/text.py#L237

Mentioned in the following stackoverflow [answer](http://stackoverflow.com/a/295466) and comments it''s pointed out that:

> the positive side of this approach is that it doesn't just remove non-alpha characters, but attempts to find good substitutes first (via the NFKD normalization), so é becomes e, a superscript 1 becomes a normal 1, etc.

another comment points out that 

> For a non-web application. I've rewritten the function to retain German (and other non-ascii) letters as is: re_slugify = re.compile('[^\w\s-]', re.UNICODE) def slugify(value): value = unicodedata.normalize('NFKD', value) value = unicode(re_slugify.sub('', value).strip()) value = re.sub('[-\s]+', '-', value) return value

prculley:
I did look at this and several other methods during my review of the code.  I selected the regex as a simple method to clear out unacceptable characters.  At least on my systems (Windows and Linux), Unicode text characters and spaces are allowed in filenames, so no apparent need to convert these to ASCII, or remove spaces.  Spaces at either end are already dealt with.

If there is some feature of the check that I am missing please let me know.

Nick-Hall:
Merged 9c92bb85519bb091b443f726149a3cbf46adfae1.

PR #292, opened 2016-12-17
bug 9286; fix styledtext editing of LINK with URL as Visible element

prculley:
If Gramps note contains text like 'www.google.com' and the user selects that text and uses the Link edit to make a styled LINK, then right-click on that doesn't allow editing of the underlying LINK.  Also the right-click 'Open Link' or 'Copy Link Address' only works with the visible text, not the underlying link.

I changed the priority of the 'cursor hover' automatic link detection so that if it is a styled LINK, it is detected over the visible text that matches the link detection regex.

During testing I also noticed that the tooltip did not go away after leaving the LINK, it would reappear if the cursor stopped anywhere in the text area of the note editor box.  
So I fixed that as well.

When testing the code I got some GTK deprecation warnings on the line:
`self.get_root_window().get_pointer()`
Since the value returned by that line was never assigned (and did not have side-effects I could find in the GTK docs), it appeared to be non-functional code, so I deleted it.

### DIFF
--- a/gramps/gui/widgets/styledtexteditor.py
+++ b/gramps/gui/widgets/styledtexteditor.py
@@ -291,20 +291,18 @@ class StyledTextEditor(Gtk.TextView):
             iter_at_location = iter_at_location[1]
         self.match = self.textbuffer.match_check(iter_at_location.get_offset())
         tooltip = None
-        if not self.match:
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Merged 1e8ab3ec6990e6bc43b4922a7d1b6f78eaa55056.

PR #293, opened 2016-12-17
bug 9235: Shrink size of Break Lock (and other QuestionDialogs) by setting default width to 760.

prculley:
bug 9235: Original 'Break Lock' dialog was greater than 1280 wide, since the text was long and had no line breaks. Rather than add line breaks (and require new translations) I decided to set default dialog size to 760; looked at other dialogs and this seems to work OK.
 uxcq says:
`Believe I found another too large dialog, uncheck the "Suppress warning when adding parents to a child" in Edit/preferences/warnings to see what I mean.`

So I also set default MessageHideDialog size to 760.

### DIFF
--- a/gramps/gui/glade/dialog.glade
+++ b/gramps/gui/glade/dialog.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.18.3 -->
+<!-- Generated with glade 3.20.0 -->
 <interface>
   <requires lib="gtk+" version="3.10"/>
   <object class="GtkDialog" id="hidedialog">
@@ -97,6 +97,7 @@
                 <property name="margin_bottom">12</property>
                 <property name="hexpand">True</property>
                 <property name="wrap">True</property>
+                <property name="max_width_chars">80</property>
               </object>
               <packing>
                 <property name="left_attach">1</property>
@@ -760,6 +761,7 @@
                     <property name="visible">True</property>
                     <property name="can_focus">False</property>
                     <property name="label" translatable="yes">label</property>
+                    <property name="max_width_chars">80</property>
                   </object>
                   <packing>
                     <property name="expand">False</property>
@@ -778,6 +780,7 @@
                     <property name="label" translatable="yes">label</property>
                     <property name="use_markup">True</property>
                     <property name="wrap">True</property>
+                    <property name="max_width_chars">80</property>
                   </object>
                   <packing>
                     <property name="expand">False</property>


Nick-Hall:
Rather than setting the dialog width, it may be better to set the maximum width of the message label using `set_max_width_chars`

prculley:
As I understand it `set_max_width_chars` is used to truncate the message label to a certain size, and restrict it to a single line.  See [https://books.google.com/books?id=NxaTd1iwNcgC&pg=PA121&lpg=PA121&dq=gtk+set_max_width_chars&source=bl&ots=p_bkPfK2V_&sig=SCnRL4d2Ipd4Mr_3jUr0pUvNh1o&hl=en&sa=X&ved=0ahUKEwi146O4pfzQAhVY42MKHcQVAC0Q6AEIKzAD#v=onepage&q=gtk%20set_max_width_chars&f=false](url)
I'm pretty sure that is not what we want.

Nick-Hall:
No. The `width-chars` and `max-width-chars` properties apply to both wrapped and ellipsized labels.  See the [text layout](https://developer.gnome.org/gtk3/stable/GtkLabel.html#label-text-layout) section in the documentation for more details.

bpisoj:
I agree with Nick!
We don't know users computer|screen|gtk settings and amount of actual text displayed varying by used translation so it is better to limit width with max-width-chars.

Somewhere else they will call it more "mobile first & responsive" approach.

Python GObject Introspection API Reference
https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Label.html

prculley:
Since I am new to GTK and still learning, I tried it your way as well.
With a setting of 80, the 'max-width-chars' did indeed limit the box size to about the same size as when I set up the dialog 'default_width' to 760.  The dialog had a lot more than 80 characters of width in it with the font we are using...

Both the methods have the dialogs still manually resizable.  
*   With default_width, the text wraps with both larger and smaller dialog boxes.  
*   With 'max-width-chars', the text wraps to smaller dialog sizes, but not larger.  The dialog looks odd when resized larger with the text off-center to the left.  But I doubt that users will resize this at all...

Both methods start to the maximum screen size, when the screen size is too small for the dialog size.  In both cases the text is wrapped properly.  (I tested this in my Linux VM by changing the VM screen size).  So if anyone ever tries to run this on a really narrow screen, it still works.  I had the VM down to about 400px wide and it was getting hard to make the rest of Gramps visible (the right Filter pane had shrunk to nothing).

In my opinion, for this application, the methods are effectively the same.  So please tell me why I would want to use one over the other.

Nick-Hall:
The user will not normally want to resize a dialog; they will read the question and make a choice.  Narrower columns tend to be easier to read than wider ones, but setting a fixed dialog width can look too wide for short messges.

Setting a maximum label width is probably better.  It can actaully be set quite low.  The message label will expand to the width of the title label.

prculley:
Ok.

PR #296, opened 2016-12-20
Fixes for some Gtk deprecations

prculley:
_textbufdoc (used for QuickReports etc.): 
* flags= and buttons= parameters are deprecated for Gtk.Dialog; used as initializers

interactivesearchbox 
* Had GObject items that have been moved to GLib (Gtk 3.10 requires GLib 2.37.5 )

_windows (Plugin manager) 
* flags= and buttons= parameters are deprecated for Gtk.Dialog; used as initializers
* positional arguments are deprecated in initializers
* Using the action_area directly is deprecated; use add_button instead (required a check for Gtk.ResponseType to get to the right routine).

dialog (Gramps dialogs)
* positional arguments are deprecated in initializers (parent -> transient_for=parent)
* flags= parameters are deprecated for Gtk.Dialog; used as initializers
* type= parameters are deprecated for Gtk.Dialog; used as initializers

### DIFF
--- a/gramps/gui/plug/_windows.py
+++ b/gramps/gui/plug/_windows.py
@@ -77,6 +77,9 @@ def display_message(message):
     """
     print(message)
 
+RELOAD = 777    # A custom Gtk response_type for the Reload button
+
+
 #-------------------------------------------------------------------------
 #
 # PluginStatus: overview of all plugins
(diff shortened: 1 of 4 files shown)


gramps/gui/plug/_windows.py
```diff
@@ -282,9 +283,7 @@ def __init__(self, dbstate, uistate, track=[]):
         if __debug__:
             # Only show the "Reload" button when in debug mode
             # (without -O on the command line)
-            self.__reload_btn = Gtk.Button(label=_("Reload"))
-            self.window.action_area.add(self.__reload_btn)
-            self.__reload_btn.connect('clicked', self.__reload)
+            self.window.add_button(_("Reload"), Gtk.ResponseType.NONE)
```
Nick-Hall: It may be better to assign a custom response value rather than `Gtk.ResponseType.NONE`.

prculley:
Changed the Gtk reponse_id for the Plugin Manager 'Reload' but to a custom type per request.  And added another commit for the dialogs module.

PR #297, opened 2016-12-22
bug 9478; fix quick search exception when nothing in searched list

prculley:
Same as PR288 except for Gramps42

Our quick search (start typing a name when a list is active) can fail if the list is empty. Not usual for someone to try to search an empty list, but can happen by mistake.
Added check for None on node to fix.


sam-m888:
Thank you

Paul-Franklin:
Should this go into master also?  Did I miss it?

sam-m888:
https://github.com/gramps-project/gramps/commit/ce884b65787a1447d32d20a5fb317901b1114d39

PR #298, opened 2016-12-22
bug 9470; Don't allow selecton of Active or Home person until actually commited

prculley:
"Make Active Person" or "Make Home Person" gives error from Edit Person Dialog on a new person.

What is happening is that until you hit 'OK' the person is not really in the database (not committed). So when you attempt to "Make Active Person" you are setting up other portions of the UI to try to use something not in the database. Thus handle not found errors.


sam-m888:
Thank you

PR #299, opened 2016-12-23
[Experiment] Relationships Category view the default instead of Dashboard.

sam-m888:
Found another older experimental patch to improve Gramps usability and rebased it on gramps master.

Would ideally work with [PR281's](https://github.com/gramps-project/gramps/pull/281) new global Add menu.

**Made** Relationships Category view the default instead of Dashboard.

 * Includes a new "Start here - Welcome to Gramps!" gramplet in the relview sidebar with brief instructions.

 * Changed the default for showing the edit buttons in relview to always show independent of toolbar.
 
 * Added a new menu and toolbar icon to "Add a new Person" 

**TODO**[ ] (Currently only adds a new person and does not switch relview to the newly created person?)

  * Reordered the categories in the Navigator so that List are grouped below and non list are at the top. (ideally would like a separator or label "List" to appear in-between "Geography" and "People") This makes it easier for people to understand the at a glance the difference between the top 4 categories and those below.

Also changed default for  "preferences.use-last-view" so that relview would be the first view shown on startup.

TODO:
Instead a of blank view in relview when:
 - [  ]  No Family Tree (database) is open, show a message to "Create a new family tree or open an existing one"
 - [  ]  Empty Family Tree is open, show a message or icon to "Add a new Person"
        
Feed back appreciated.

Merry Xmas to all.

### DIFF
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -187,10 +187,21 @@ register('geography.use-keypad', True)
 # needed), for instance to four 'interface.clipboard' variables --
 # so do a recursive grep for "setup_configs" to see all the (base) names
 register('interface.dont-ask', False)
+#Default Navigator views (sets order shown)
+#register('interface.view-categories',
+#         ["Dashboard", "People", "Relationships", "Families",
+#          "Ancestry", "Events", "Places", "Geography", "Sources",
+#          "Citations", "Repositories", "Media", "Notes"])
+# New order 1
+#register('interface.view-categories',
+#         ["Dashboard", "Relationships", "Ancestry","Geography",
+#          "People", "Families", "Events", "Places", "Notes",
+#          "Media", "Citations", "Sources", "Repositories",])
+# New order 2
 register('interface.view-categories',
-         ["Dashboard", "People", "Relationships", "Families",
-          "Ancestry", "Events", "Places", "Geography", "Sources",
-          "Citations", "Repositories", "Media", "Notes"])
+         ["Relationships", "Dashboard", "Ancestry","Geography",
+          "People", "Families", "Events", "Places", "Notes",
+          "Media", "Citations", "Sources", "Repositories",])
 register('interface.filter', False)
 register('interface.fullscreen', False)
 register('interface.grampletbar-close', False)
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -197,9 +197,12 @@ register('interface.dont-ask', False)
 #         ["Dashboard", "Relationships", "Ancestry","Geography",
 #          "People", "Families", "Events", "Places", "Notes",
 #          "Media", "Citations", "Sources", "Repositories",])
-# New order 2
+# New order 2 (Make relationships the default view #Todo work out how to keep
+#              dashboard 1st | Group list views below non list views eg
+#              Geography etc # Todo add separator or label btwn list & non list
+#              views
 register('interface.view-categories',
-         ["Relationships", "Dashboard", "Ancestry","Geography",
+         ["Dashboard", "Relationships", "Ancestry","Geography",
           "People", "Families", "Events", "Places", "Notes",
           "Media", "Citations", "Sources", "Repositories",])
 register('interface.filter', False)
@@ -276,8 +279,8 @@ register('preferences.private-record-text', "[%s]" % _("Private Record"))
 register('preferences.private-surname-text', "%s" % _T_("[Living]"))
 register('preferences.rprefix', 'R%04d')
 register('preferences.sprefix', 'S%04d')
-register('preferences.use-last-view', False)
-register('preferences.last-view', '')
+register('preferences.use-last-view', True)
+register('preferences.last-view', 'relview')
 register('preferences.last-views', [])
 register('preferences.family-relation-type', 3) # UNKNOWN
 register('preferences.age-display-precision', 1)
--- a/gramps/plugins/gramplet/gramplet.gpr.py
+++ b/gramps/plugins/gramplet/gramplet.gpr.py
@@ -328,6 +328,20 @@ register(GRAMPLET,
          )
 
 register(GRAMPLET,
+         id="WelStartRelview",
+         name=_("StartRelview"),
+         description = _("Gramplet showing a welcome message"),
+         status = STABLE,
+         fname="welcomerelview.py",
+         height=300,
+         expand=True,
+         gramplet = 'WelcomeRelview',
+         gramplet_title=_("Start here - Welcome to Gramps!"),
+         version="1.0.0",
+         gramps_target_version=MODULE_VERSION,
+         )
+
+register(GRAMPLET,
          id = "What's Next",
          name =_("What's Next"),
          description = _("Gramplet suggesting items to research"),
--- a/gramps/plugins/gramplet/welcomerelview.py
+++ b/gramps/plugins/gramplet/welcomerelview.py
@@ -1,0 +1,139 @@
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2007-2009  Douglas S. Blank <doug.blank@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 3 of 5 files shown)


sam-m888:
A few image of before an after for comparison.

Left hand is Current Dashboard default / Right side is new Relview
reorder-navigator-and-make-relationships-category-default-view-01-side-by-side

Current relview
reorder-navigator-and-make-relationships-category-default-view-03-original-dashboard-default

New relview with sidebar Gramplet and "Add new Person" icon (also Global "Add" menu from PR281 is shown.
reorder-navigator-and-make-relationships-category-default-view-08-test-data

New relview (again) with sidebar Gramplet(expanded so you can see full message)  and "Add new Person" icon (also Global "Add" menu from PR281 is shown.
reorder-navigator-and-make-relationships-category-default-view-09-expanded-with-pr281-add-menu

uxcq:
> Merry Xmas to all.

Merry Christmas, Have a good day.

At a glance this PR looks good, I will try it after the festivities are over and  provide feedback. 👍

enlite93:
Wonderful :)  SEason greetings to you all.

daleathan:
Another good pr. Merry Xmas.

> TODO[ ] (Currently only adds a new person and does not switch relview to the newly created person?)

Believe the following is related to this issue.

 *  [**9856: master no longer "remembers" the selected person between views**](https://gramps-project.org/bugs/view.php?id=9856)

Nick-Hall:
I like the idea of encouraging new users to use the Relationships view.  I have developed viewers based on this view that display information in more useful ways.

Paul-Franklin:
I would like to make some comments on this proposal,
and have a discussion about what new users should
be shown, but I am very loath to do it here, on github.

As I have stated several times, I don't believe these
comments are searched by Google (so they can't be
found again).  I also don't believe that they are put
in any sort of mirror (and archive) the way Nabble
mirrors and archives the gramps-devel (and also
the gramps-user, etc.) mailing lists.  (And I believe
at least one gramps developer interfaces to the
gramps-devel mailing list via Nabble, so IMHO it
is wrong to exclude people like that.)

It hasn't worked in the past, when I have made a
plea here on github, to transfer a discussion to
the gramps-devel mailng list, but I am trying yet
again.

uxcq:
Starting in this view is a good idea.

Combined with the new add menu from pr281 and a few tweaks to make the view change to the newly entered person I can see this as a big win for Gramps.

> Includes a new "Start here - Welcome to Gramps!" gramplet in the relview sidebar with brief instructions.

Especially like the new get started Gramplet (can this be extended to other views, say one for each and on the dashboard split it in two so dashboard has it's own specific one) if possible make it an option to have them displayed or instead of embedding then in the right sidebars obscuring the filters have them pop up to the side(no overlaying the gramps main window, if enough display space) with an option like the _tip of the day_ so they can be dismissed or pushed to the back ground?

> Reordered the categories in the Navigator

Initially I wondering what gramps was attempting to present to me and believe that changing the order of the categories(left hand side bar) into logical groups and labeling them for what they [List] are seems sensible to me and should be easier to get across to new users as a concept.   Even just changing the tooltips to be a bit more more descriptive would go a long way eg: Citations > tooltip = "List of entered Citations"

> Instead a of blank view in relview when:

Probably a good idea for any category/view to have a default message displayed when no record of that is available. eg: Person view(list or grouped) have a single line (grayed out and non select-able) that states something like _"No records available, add one using the toolbar or menu above"_

@Nick-Hall @Paul-Franklin Can I suggest if you have concerns with comments from github not being archived, contacting github and ask for a possible feature? 

> developed viewers based on this view that display information in more useful ways.

@Nick-Hall Where can these views be found?

@sam-m888 These are the type of minor changes that will help change the perception that gramps is difficult to use, and should be in alpha2 before released for comment from real newbie users(which I don't consider myself). Thank you.

If you need me to clarify anything please ask.

Happy New year to all from the Gramps project.

Nick-Hall:
@uxcq I made a prototype back in April 2014. It attracted very little interest and is no longer publicly available.

My current version uses non-standard navigation code to move between views.  I'll investigate the possibility of combining the viewers into a single plugin.

pat49:
Making Gramps start from the Relationships category,  instead of starting on a dashboard is very good thing.  

By the way really don't like that it's called the "Relationships" view,  as it more a person centric view but I can't think of a better name, note that the Dashboard used to be called the "Gramplets" category until I asked it be renamed, so things can change.

Agree about the differences between the list categories and other categories and making it more obvious at a glance. I recall requesting that reports and tools be added to the navigator also (Already appear on the toolbar as out of place icons, would be better integrated.). BTW on the navigator can you move that category selection thing into preferences (believe I requested something similar)

Happy new year Grampers!  

Make Gramps easy enough for Grandma's friends to use (mine that is) :)

Well back to lurking and ancestry.com I go really felt I need to add my bit for what it's worth (Just a touch to much testosterone here :( )

sam-m888:
@uxcq @Nick-Hall @pat49 Thanks for the feedback.

kulath:
Would it be possible to change the "Start Here" gramplet so that it came up with different messages when
(1) there is no database open
(2) there is nobody in the database
(3) there is no active person

(there are a few circumstances where there are people in the database, but no one is selected).

Nick-Hall:
We could create a _Start Here_ page where the user could enter:
* A family tree name
* Their name and birth details
* The names, birth and possibly death details of their parents.
* The names and birth details of their children.

Gramps could then create a database and switch to the relationship view.

kulath:
@Nick-Hall I think your comprehensive "Start Here" page is too prescriptive and limiting. What if the user is inputting the royal family, or wants to start with an early ancestor rather than themselves.

On initial startup, Gramps should present a Researcher dialogue box (with a check-box to say don't show this again) to allow the user to input his personal data. (It is a little known fact that this data is used to initialise the Database owner information stored in the database when it is created). This should be the first thing that is input by the user. (Text at the top of the box should say how this is used, and that it can be input later for a database by Tools -> Family Tree Processing... -> Edit Database owner information...).

After that, I think instructions alongside the Relationships window works just fine.

Nick-Hall:
@kulath The _Start Here_ page would be aimed mainly at new users building their own family tree, starting with themselves and working backwards in time.  However, the concept of creating mulitple objects from a single page could be useful even for experienced users.

Imagine entering details of a marriage event into a single data entry page and then having the citation, event, family and places created for you.  Perhaps it would be useful to enter the baptism information of several children from a parish register into a single page.

I am considering developing the forms addons in this direction.

uxcq:
@sam-m888 I notice that you added one of the commit to issue [8723](https://gramps-project.org/bugs/view.php?id=8723) for the missing buttons on relationshipView, would it make more sense where possible to split this PR up and commit the parts that don't change gramps functionality greatly.

sam-m888:
> would it make more sense where possible to split this PR up and commit the parts that don't change
> Gramps functionality greatly?

@uxcq I'll look at splitting this up into smaller PR's as mentioned this was just an experiment 👍

sam-m888:
@uxcq I've split most of them out except for:

* the first patch to "Make the Relationships Category view the default instead of Dashboard." which is simple enough to apply if wanted.

* the last patch "To show a message in the relationship view if the familytree is empty/new or if no family tree has been created.  (never finished)

emyoulation:
I tried a more minor change to make an initial Dashboard screen more useful for newbies and for receiving new shared .gpkg Trees.  

This proposal changes the assortment of default (improved) Dashboard Gramplets:

The most current proposal elsewhere suggest merely improving the "Welcome to Gramps!" Text but not the set of Gramplets.  This is good as far as it goes. But the "What's Next?" and "To Do" gramplets  have more use for an ongoing basis.

Here, a crudely hacked "What's Next?" gramplet recognizes that the Tree has not yet been loaded and gives direction of the 1st steps for ANY user of Gramps: load or create a tree, then add or import data.
 
image

After creating a blank tree, "What's Next?" suggests the next step:
image

After importing the sample.ged included with the Gramps distribution. It suggests setting a Home Person (which incidentally also sets an Active Person)
image

If instead, the user had: loaded a Gramps tree, that tree had people, there was an Active Person, there was a Home Person; then the original "What's Next?" content finally has the context to work.  And the "To Do" Gramplet shows the highest priority "To Do" note in the Tree:
image



 

**Note that the hard-coded "Welcome to Gramps!" text is a misery** to update, markup and translate. It creates silo code that is not re-used elsewhere. (The "Tip of the Day" has similar problems.) Instead, the plugin should leverage the Note viewer (or a GitHub Markup viewer) and reference a .gramps XML (or JSON) file of Note objects (each having a language code suffix for a common note type. e.g., "Welcome-fr" french or "Welcome"/"Welcome-en" default/english) in the same folder with a Fall-upwards to an add-on folder and be overridden by a similarly language coded "Welcome" type note in the loaded Tree. The add-on folder would allow the text to evolve independently of the major upgrade cycle. The "Welcome" type note in the Tree would allow an introduction of the research.

PR #300, opened 2016-12-23
Bug 7928; Make merging people with complex families more friendly

prculley:
If families are complex, skip the family merge portion of the person merge with a warning message rather than abort the persons merge.
Users will be able to merge people successfully even when they have complex families, there may be some extra families attached to them which could perhaps be merged, but would be unsafe for Gramps to do on its own.  The warning message lets them know and suggests how to fix.

### DIFF
--- a/gramps/gen/merge/mergepersonquery.py
+++ b/gramps/gen/merge/mergepersonquery.py
@@ -116,13 +116,15 @@ class MergePersonQuery:
         """
         if trans is None:
             with DbTxn(_('Merge Person'), self.database) as trans:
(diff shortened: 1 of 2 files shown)


prculley:
Added the fix commit for bug 7928; it needed to patch my PR as well as the separate issue.

Nick-Hall:
@prculley Thanks.  In future, it is probably better to start a separate PR for each bug.

Please keep the first line of the commit messages to 70 characters of less.  It is easier to read in GitHub.  See our [Committing policies](https://gramps-project.org/wiki/index.php?title=Committing_policies).

PR #302, opened 2016-12-28
bug 8128; Fix transient parent for popup 'Gramps Warnings'

prculley:
An obscure feature is the ability to popup a display of recent warnings (normally in debug log) whenever the little icon in the left of the Gramps status bar appears.  This was popping up in the wrong place, so needed to fix for 'transient parent'.
To test, you need to generate loggable warnings.  I generated a new tree, used the 'tools/debug/Generate testcases' with the 'Generate database errors' check set, and then used 'tools/Family Tree Repair/Check & Repair' to have something logged.
The icon in the status bar then shows up and can be clicked.

The Glade file was causing the external transient parent setting to be ignored at first, so I did some editing and updated it; not sure exactly which setting caused the issue, but it works now.

### DIFF
--- a/gramps/gui/displaystate.py
+++ b/gramps/gui/displaystate.py
@@ -317,13 +317,14 @@ from .logger import RotateHandler
 
 class WarnHandler(RotateHandler):
 
-    def __init__(self, capacity, button):
+    def __init__(self, capacity, button, parent=None):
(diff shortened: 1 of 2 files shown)


sam-m888:
Thank you 👍

PR #303, opened 2016-12-29
bug 8128; Fix 'Info' popup from Manage Family Trees' dialog for parent

prculley:
This PR updates the 'Manage Family Trees' dialog and its 'Info' popup for correct transient parent.

Originally we had an unusual case of the unmanaged 'Manage Family Trees' dialog spawning the ManagedWindow 'Info' dialog.  A clean fix seemed difficult.

I decided to fix this the 'right' way, rather than hack it.  So I upgraded the 'Manage Family Trees' dialog to become a ManagedWindow.  Along the way I also added some comments to the ManagedWindow docstring to help remind me and other developers how to use it properly.

I also made some minor changes to the 'Info' dialog so it would respond properly.
I tested everything except the 'Archive' function; I suspect it requires some additional programs (rcs) I don't have installed.  I don't think this would be effected by my changes.

P.S. getting the transient parent set right is pretty obvious on a system with two (or more) screens; if Gramps is running on a secondary screen, and the new window pops up on the primary screen, it is wrong.

### DIFF
--- a/gramps/gui/dbman.py
+++ b/gramps/gui/dbman.py
@@ -113,12 +113,11 @@ RCS_BUTTON = {True : _('_Extract'), False : _('_Archive')}
 
 class Information(ManagedWindow):
 
-    def __init__(self, uistate, data, parent):
-        super().__init__(uistate, [], self)
+    def __init__(self, uistate, data, track):
+        super().__init__(uistate, track, self, modal=True)
         self.window = Gtk.Dialog()
         self.set_window(self.window, None, _("Database Information"))
         self.setup_configs('interface.information', 600, 400)
-        self.window.set_modal(True)
         self.ok = self.window.add_button(_('_OK'), Gtk.ResponseType.OK)
         self.ok.connect('clicked', self.on_ok_clicked)
         s = Gtk.ScrolledWindow()
@@ -132,8 +131,6 @@ class Information(ManagedWindow):
             model.add((key, str(value),), key)
         s.add(treeview)
         self.window.vbox.pack_start(s, True, True, 0)
-        if parent:
-            self.window.set_transient_for(parent)
         self.show()
 
     def on_ok_clicked(self, obj):
@@ -142,7 +139,8 @@ class Information(ManagedWindow):
     def build_menu_names(self, obj):
         return (_('Database Information'), None)
 
-class DbManager(CLIDbManager):
+
+class DbManager(CLIDbManager, ManagedWindow):
     """
     Database Manager. Opens a database manager window that allows users to
     create, rename, delete and open databases.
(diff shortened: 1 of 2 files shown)


Paul-Franklin:
Well, I haven't tried to run this but I really really appreciate what
it is going to do.  So let me thank you.

Some comments come to mind, however.  I vaguely recall that
modal_call is only used once, inside gramps, so if that's true
you might put eliminating it on your list.  And I also vague recall
that the docstring for modal_call suggests/recommends that
it be used everywhere, in the future, so if it does say something
like that you might consider eliminating those words, besides
adding your "deprecated" warning.

On another matter, what about modal widgets which are
spawned?  Ones which appear to be a dialog that is.  Do
they need to (somehow, I haven't a clue) be made into a
ManagedWindow also?  I have in mind the color-changing
widget (I think it's called Gdk.RGBA).  I am working on 
making both "tags" dialogs into ManagedWindows and the
subject of that widget came up.  So I'd appreciate some
clarification, and if you want, some related documentation
in the docstrings in ManagedWindow.

One specific question I have about this p.r. though is
about the FTM's name.  Does it still get correctly set,
since I noticed your set_window call had None for the
title string?

But like I said originally, this is great work!  Thanks!!

Paul-Franklin:
Also, I recall a comment you made on some bug which
said (IIRC) that two modal windows were all right, if they
had the same parentage (or something like that).  Am
I remembering it correctly?  If so, do you want to say
something about that, too?  For instance the color
changing widget in the OrganizeTags dialog comes
up centered over its calling window, even if I've moved
that around first, which implies to me that the color
one has code in it to collect the parent of its caller
and then use it.  But (a) I don't know if I'm right, and
(b) I don't know if all Gtk/Gdk widgets do that, if I
am right.

Nick-Hall:
@Paul-Franklin The tag editor uses a Gtk.ColorButton widget which calls the standard colour selection dialog.

Nick-Hall:
@prculley Thanks.

Care must be taken when defining a managed window using multiple inheritance.  For example:
```
class MyWindow(Gtk.Window, ManagedWindow):
...
self.show()
```
In this case the wrong `show` method will be called, and as a result, `set_transient_for` is never called.

This is not a problem in your patch because `CLIDbManager` does not contain a `show`method, but I thought it was worth mentioning.

prculley:
Nick;
Thanks for the warning; I just found another bug in this (forgot to call ManagedWindow.close()...).  When I tried to use 'self.close()' I got hit with this or something like it.

Paul;
Thanks for the quick merge (I think) ;-)  After i came back from lunch I noticed that some weird menu entries had appeared in the 'Window' menu; seems I forgot to close my newly converted ManagedWindow.
Another commit forthcoming.

In my opinion, it is not ideal to have two or more modal windows at once.  When they are parent/child, and have the transient parent _correctly_ set, it appears that the child stays on top.  But the parent can still be activated via the mouse (not greyed out).  It doesn't seem to actually respond to mouse/keyboard events so far as I can tell, but it doesn't look right and could be confusing to users.  

Looking at the Tag color picker, it appears to me (by the way it works, I didn't look at the code) that it has a correct transient parent but suffers the two modal windows problem.

Paul-Franklin:
@prcully I did not do any merge of this (303).

PR #304, opened 2016-12-29
Restart addon message change wording

sam-m888:
As not all addons are Views.

### DIFF
--- a/gramps/gui/plug/_windows.py
+++ b/gramps/gui/plug/_windows.py
@@ -1217,7 +1217,7 @@ class UpdateAddons:
                      "%s %s" % (ngettext("{number_of} addon was installed.",
                                          "{number_of} addons were installed.",
                                          count).format(number_of=count),
-                                _("You need to restart Gramps to see new views.")),
+                        _("You need to restart Gramps to use the new addons.")),
                      parent=self.window)
         else:
             OkDialog(_("Done downloading and installing addons"),


gramps/gui/plug/_windows.py
```diff
@@ -1217,7 +1217,7 @@ def install_addons(self, obj):
                      "%s %s" % (ngettext("{number_of} addon was installed.",
                                          "{number_of} addons were installed.",
                                          count).format(number_of=count),
-                                _("You need to restart Gramps to see new views.")),
+                        _("You need to restart Gramps to use the new addons.")),
```
Nick-Hall: It is not necessary to restart Gramps after installing most addons.  Views are a special case.

sam-m888: Note this was part of https://github.com/gramps-project/gramps/pull/63 and I split it out here.

Also I was just about to split out the  "Change config keys to mention addon" part leaving only the split addons part should I continue?

Nick-Hall: Perhaps we could change the message to "If you have installed a view, you will need to restart Gramps"?

Nick-Hall: I don't mind you renaming config keys to make their use clearer.  Please continue.

sam-m888: > Perhaps we could change the message to "If you have installed a view, you will need to restart Gramps"?

Ok, but slightly modified to match what the user actually sees ''**Gramps View**",  eg:
 "If you have installed a "Gramps View", you will need to restart Gramps"

And if I can work out where that label is also display in the "Available Gramps updates for addons"  dialog "**(requires restart)**" beside it eg:

available gramps updates for addons

Is there a Global flag, that indicates Gramps requires a restart?

sam-m888: > I don't mind you renaming config keys to make their use clearer. Please continue.

Change config keys to mention addon
https://github.com/gramps-project/gramps/pull/306

Nick-Hall: @sam-m888 There is no global restart flag.

sam-m888: > @sam-m888 There is no global restart flag.

Thanks

For now I've only pushed the wording change mentioned above as https://github.com/gramps-project/gramps/commit/af674df82e33fc5e1ae76dc50e9cacf99e1666c4

PR #306, opened 2016-12-30
Change config keys to mention addon

sam-m888:
Renaming config keys to make their use clearer


sam-m888:
Split from https://github.com/gramps-project/gramps/pull/306

PR #307, opened 2016-12-31
bug 8128; Fix common error/warn dialogs to deal with modal parent

prculley:
Unmanaged dialogs are often called for errors, warnings, info etc.
On Windows, If called from a modal parent window, they can be a bit
confusing, since the parent window appears to still be active.  This
fix makes the parent window non-modal for the duration of the dialog,
and then returns it to its original state.

During testing of this fix (all dialogs changed were tested), I realized that the Windows Gtk implementation definitely is not as polished as the Linux one.  I did not find any of these modal issues I'm trying to fix on Linux, only Windows.  This fix appears to have no effect on Linux at all.

### DIFF
--- a/gramps/gui/dialog.py
+++ b/gramps/gui/dialog.py
@@ -73,6 +73,9 @@ class SaveDialog:
         label2.set_use_markup(True)
         if parent:
             self.top.set_transient_for(parent)
+            parent_modal = parent.get_modal()
+            if parent_modal:
+                parent.set_modal(False)
         self.top.show()
         response = self.top.run()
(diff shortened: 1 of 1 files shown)


Nick-Hall:
> During testing of this fix (all dialogs changed were tested), I realized that the Windows Gtk implementation definitely is not as polished as the Linux one. I did not find any of these modal issues I'm trying to fix on Linux, only Windows. This fix appears to have no effect on Linux at all.

Yes.  These are Windows related issues.  Thanks for fixing this - we don't have many developers that use Windows.

PR #308, opened 2016-12-31
bug 9872: fix Windows issue; Book dialogs don't stay on top

prculley:
They are pushed to background if background window tooltip appears.
Also, The book list display was not popping up above the dialog
(no transient parent).

Turns out that the dialog, although a ManagedWindow, was not calling the ManagedWindow.show(); only the Gtk.Window.show().
This meant that the ManagedWindow transient parent code was not being executed.

The book list display transient parent was theoretically being set in the Glade file; however this functionality appears to be broken, it sets the parent to _some_ window, but I have no idea which one, certainly not the correct parent. Also the Glade file had set the book list display to 'Centered' which seems to somewhat override the transient parent location and instead centers on the screen (the center of dual screens in my case), rather than centered above the dialog.

### DIFF
--- a/gramps/gui/glade/book.glade
+++ b/gramps/gui/glade/book.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.18.3 -->
+<!-- Generated with glade 3.20.0 -->
 <interface>
   <requires lib="gtk+" version="3.10"/>
   <object class="GtkWindow" id="top">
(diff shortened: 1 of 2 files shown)


Paul-Franklin:
I am very familiar with the book dialog (and had been aware
of some of the things you noted, such as the Glade problems
(although while I knew the window came up in the center of
the physical screen I didn't think of people like you with two
physical screens), but I hadn't known how to solve them).

As far as I'm concerned your changes could be committed
instantly.

Paul-Franklin:
A related thought suggests doing a recursive grep for
that offending Glade line, as I am pretty sure it appears
in other .glade files too.

prculley:
Already did that, turns out that 'book.glade' is the only place where that occurred.  Checked addons source as well.

Paul-Franklin:
What do you know.

Well, perhaps my memory is faulty -- always high on
the list of probabilities,

But I'm glad that doesn't exist in gramps anymore!

Thanks.

PR #315, opened 2017-01-03
bug 8128; fix progress meters for transient parent

prculley:
Update:
This is a partial fix for the current progress meters; it adds an attempt to find the current Gtk Window to the meters if the parent is not set.  It doesn't always work, in that if the user activates another application before the progress meter gets put up, the code cannot find a parent.  This occurs during debugging (when the debugger is the other parent).

I've abandoned the earlier attempt to do an end run around and find the Gramps main window via a stored global variable.  So some comments below are no longer relevant.

This started as a fix for the deeprelationship filter using a progress meter.  Another PR will address that more effectively.

### DIFF
--- a/gramps/gui/utils.py
+++ b/gramps/gui/utils.py
@@ -152,6 +152,12 @@ class ProgressMeter:
         self.__dialog.vbox.set_spacing(10)
         self.__dialog.vbox.set_border_width(24)
         self.__dialog.set_size_request(400, 125)
+        if not parent:  # if we don't have an explicit parent, try to find one
+            for win in Gtk.Window.list_toplevels():
+                if win.is_active():
+                    parent = win
+                    break
+        # if we still don't have a parent, give up
         if parent:
             self.__dialog.set_transient_for(parent)
             self.__dialog.set_modal(True)
--- a/gramps/gui/widgets/progressdialog.py
+++ b/gramps/gui/widgets/progressdialog.py
@@ -494,13 +494,17 @@ class GtkProgressDialog(Gtk.Dialog):
         :type title: string
         """
         Gtk.Dialog.__init__(self)
+        parent = None
         if len(window_params) >= 2:
-            self.set_transient_for(window_params[1])
-        else:
+            parent = window_params[1]   # we got an explicit parent
+        else:                           # try to find an active window
             for win in Gtk.Window.list_toplevels():
                 if win.is_active():
-                    self.set_transient_for(win)
+                    parent = win
                     break
+        # if we still don't have a parent, give up
+        if parent:
+            self.set_transient_for(parent)
         if len(window_params) >= 3:
             flags = window_params[2]
             if Gtk.DialogFlags.MODAL & flags:


Paul-Franklin:
Makes sense to me.

Nick-Hall:
This is not a nice fix.  The DisplayState does not belong in the constants.

We should pass an instance of the User class to rules.  This is a good opportunity to get rid of another gui import in the gen module.

prculley:
So that I don't make 'mistakes' like this again, perhaps you could point me to an architectural document or other resource that explains why I would want to use the 'User class' (which looks like a lot of work to me, if I even understood how it could help here).

I generally try to keep my fixes as 'surgical' as possible, as testing even small fixes is prone to missing important cases.  I think adding 'User' to all the filters and their manager would not be very surgical...

Paul-Franklin:
Maybe he means that the GUI User instance has a
uistate (which got tested for, I think in the GEDCOM
importer, to help decide if a GUI dialog should exist).

The docstrings for the User class hint at its intended
(I think) use: as a generic API which can be used from
both CLI and GUI reports (and tools, and maybe importers
and exporters, I am not sure).  So a "user.warn" could be
in a report (or tool, etc.) and do different things, yet the
calling code would be the same.  There's "progress"
code in the (three) User classes also, but I don't know
if it is widely used.  I don't think it's as general as the
"real" Gtk ProgressMeter for instance (sublabels come
to mind -- I think that's the reason the Pro-Gen importer
still has a Gtk ProgressMeter and not a User one).

HTH.

Nick-Hall:
@prculley I should have explained better.  Unfortunately there isn't much documentation.  The file organisation is described in [GEPS 008](https://gramps-project.org/wiki/index.php?title=GEPS_008:_File_Organization).  You can see that the `gen` module should not import from any other top-level modules and should not be dependent on Gtk.  There is a comment in the rule pointing out that the `User` class should be used instead of a progress dialog.

The filters can be used by both gui and non-gui code.  For example, if a filter using this rule was used in a report called form the command line, then using a Gtk progress bar would cause a problem.  We had problems of this type in the past.  See [bug #4646](https://gramps-project.org/bugs/view.php?id=4646).

The `User` class allows us to display warnings and progress information or ask simple questions without knowing if a gui environment is present.  It is useful in reports, tools and import/export plugins.  We can also use it in the `gen` module.

If you look at the `apply` method in the `GenericFilter` class, you will see a `cb_progess` callback which pre-dates the `User` class.  It is called once for each object filtered.  This allows the implementation of the progress bar to be defined by the caller.

The problem dialog is actually in a `prepare` method of a rule. These are called once when a filter is applied.  A possible solution would be to pass a `User` class in the filter `apply` method which could pass it to the rule `prepare` method.

Eventually we could replace the `cb_progess` callback, but this would be a bigger job.

Paul-Franklin:
Yes, 4646 contains a lot of useful history, IMHO.

IIRC I believe the User class was made in the process
of all the changes which 4646 required.  (Which of course
are still not done, for things like these filters as well as
many of the tools (IIRC).)

Nick-Hall:
Is this still needed if we add `uistate` to the view models in PR #318?

prculley:
I think this fix is still useful; there are other cases where both of these progress meters get popped up.  And one is still popped up by the filter (until I decide how to fix it).  I may eventually get all the various cases fixed, but this will keep the users happier until that happens.

Nick-Hall:
OK.  We should aim to provide an explicit parent window for dialogs, but until then this is a sensible fallback.

PR #316, opened 2017-01-05
bug 9877; fix performace with people tree view and some filters

prculley:
When using deep relationship filters on the person tree view with a
large tree, the tree redraw can take a long time (hours).  The
DeepRelationshipPathBetween 'prepare' method was getting called for
every person in the preceding filter. And then it also ran through
every person in the db within the 'prepare' method.  Resulting in time
as a square function of number of people.

Change tree view code so that 'prepare' method is called once, and 'apply'
method is called once with list of handles to scan, rather than both
called once per handle.

End result, tree view update in seconds or less, rather than hours.
Tested with several different filters, including manual selections from filter gramplet (name, event, sex etc.).  Results same as before this patch, except faster on those that do a lot of work in the 'prepare' stage.

### DIFF
--- a/gramps/gui/views/treemodels/treebasemodel.py
+++ b/gramps/gui/views/treemodels/treebasemodel.py
@@ -591,16 +591,20 @@ class TreeBaseModel(GObject.GObject, Gtk.TreeModel, BaseModel):
         pmon.add_op(status_ppl)
 
         self.__total += items
+        assert not skip
+        if dfilter:
+            for handle in dfilter.apply(self.db,
+                                        cb_progress=status_ppl.heartbeat):
+                data = data_map(handle)
+                add_func(handle, data)
+                self.__displayed += 1
+        else:
+            with gen_cursor() as cursor:
+                for handle, data in cursor:
+                    status_ppl.heartbeat()
+                    add_func(handle, data)
+                    self.__displayed += 1
 
-        with gen_cursor() as cursor:
-            for handle, data in cursor:
-                if not isinstance(handle, str):
-                    handle = handle.decode('utf-8')
-                status_ppl.heartbeat()
-                if not handle in skip:
-                    if not dfilter or dfilter.match(handle, self.db):
-                        add_func(handle, data)
-                        self.__displayed += 1
         status_ppl.end()
         status.end()
 


ennoborg:
It looks like it breaks the deep connections gramplet:

2017-01-08 20:26:30.358: WARNING: _gramplet.py: line 338: Gramplet gave an error: Verre relaties
  File "/home/enno/gramps/gramps/gen/plug/_gramplet.py", line 322, in _updater
    retval = next(self._generator)
  File "/home/enno/.gramps/gramps50/plugins/DeepConnectionsGramplet/DeepConnectionsGramplet.py", line 232, in main
    relatives = self.get_relatives(current_handle, current_path)
  File "/home/enno/.gramps/gramps50/plugins/DeepConnectionsGramplet/DeepConnectionsGramplet.py", line 102, in get_relatives
    person = self.dbstate.db.get_person_from_handle(person_handle)
  File "/home/enno/gramps/gramps/plugins/db/bsddb/read.py", line 738, in get_person_from_handle
    return self._get_from_handle(handle, Person, self.person_map)
  File "/home/enno/gramps/gramps/plugins/db/bsddb/read.py", line 701, in _get_from_handle
    raise HandleError('Handle %s not found' % handle.decode('utf-8'))
gramps.gen.errors.HandleError: Handle c99dda79fa361f7d27fffc31023 not found
Continuing after gramplet error...

I don't get this error in the gramps42 branch.

prculley:
I am unable to recreate this issue.  It is possible that the db is corrupted, please run Check & Repair and try again.  Also see if the issue can be recreated in the 'People list' view; my code changes should have had no effect there.

Be aware that we have been finding a lot of 'HandelError' bugs; in the Gramps 4.1.x branch, these returned 'None' and a lot of code has tests for that and could ignore them.  In the 5.x branch these were set up to create an exception.  So underlying issues that 4.1.x could ignore are now needing to be dealt with.

If you can recreate, perhaps you would file a bug report and include a test tree (.gramps file) that we can use to test with.

ennoborg:
I ran check & repair, but got no errors, but I understand your comment about 'HandleError' bugs. If you plan to back-port this to 4.2, I can test this single change on that later. If you don't, I think it's better that I try a back-port to 3.4, because that's what I still use here.

ennoborg:
I need to apologize. The error that I get appears in master, and your code is not in there yet. In other words, the deep connections Gramplet has a problem with handles already, and to my regret I don't have much time for further analysis yet. Maybe later, sorry!

Nick-Hall:
Calling `apply` instead of `match` is the correct solution.  However, you now loop through the table at least twice.

This needs more thought.  I don't think this will be acceptable for users with large databases.

prculley:
I'm not certain I understand your concern; do you mean how I run through the db getting the list ready for the filter, and then run through results to update rows?  (The specific filter that is causing the trouble actually does another run within 'prepare', but that was how it was architected).

I think this is very similar to what the list view already does; and it performs OK for my 900+ db (update under a second unless debugging).

Nick-Hall:
The current code only loops through the table once, although it will call `prepare` for each object.  I have only done a few quick tests, but your patch seems to add about 25-30% to the load time when no filter is applied.  This will not be popular with users with large databases.  Remember that the models are also used in the selectors.  However, even for simple filters it does reduce the load time slightly.  I tested with the _IsMale_ and _IsFemale_ filters.

I suggest modifying the patch slightly so that it only uses a single loop when no filter is applied.  You could also consider applying the filter to all objects, and ignoring the skip list when adding the rows.  The skip list is only used in the selectors, and will only contain a small number of items.

romjerome:
> your patch seems to add about 25-30% to the load time when no filter is applied. 

Not tested yet, maybe this can fix the current slow down on [Person Selector if there is no filter](https://gramps-project.org/wiki/index.php?title=GEPS_041:_New_Selector#Performance_results)?

e.g., select a person from _Associations_ tab on _Person Editor_

Nick-Hall:
@romjerome The selectors use fast filters, which work like searches, rather than standard filters.  This patch will not improve the load time.

prculley:
@Nick-Hall I've revamped the patch like you asked.  For my 977 person tree (488 men)  There was some variation on the times so this is an average over 4 runs.
Original master Treeview/Patched Treeview/List view times:
 None filter - .06/.05/.03
 Men filter  - .265/.25/.21
 Everyone filter - .28/.28/.21
 DeepRelationship - hours/1.87/1.87

Nick-Hall:
@prculley Thanks.  That looks better.

The skip list is only used in selectors which don't use filters.  It may be better to apply the filter using the following code:
```
for handle in dfilter.apply(self.db, cb_progess=status_ppl.heartbeat)
    if handle not in skip:
        data = data_map(handle)
        add_func(handle, data)
        self.__displayed += 1
```

prculley:
@Nick-Hall Thanks; Gramps has so many features that I am somewhat leery of removing code functionality; never certain if I missed finding the weird case that makes use of that.  I had found the use of skip list in selectors for families (skipping parents while selecting children etc.), but I might have missed other cases...

If skip list is only used in selectors, then we really don't need to include in the 'rebuild_filter' at all...
And since we can drop skip list, we save some memory and another loop through the handle list...
As long as we are optimizing, I don't think we need the test for str handle anymore either...  Seems to work fine with both bsddb and dbapi.

PR #317, opened 2017-01-05
Fix the debug module (profiling code) for current python libraries

prculley:
Looks like this has not been used recently, the 'hotshot' module is no longer available (at least in my distro).  So modified to use the cProfile library instead.


Nick-Hall:
Merged b7155b47ed98913ba00ea9be092a4a55bc57ca80.

PR #318, opened 2017-01-07
Bug9876; fix two progress bars occupy the same space

prculley:
The person tree view pops up a progress bar (GtkProgressDialog) on slow updates, with a useful indication of progress.

When the 'Relationship path between <person> and people matching <filter>' filter is used, a second progress bar (ProgressMeter) pops up which overlays the first. The 2nd bar conveys no useful information, it just flashes.

This fix moves the person tree view progress down to the Gramps Status bar (where the progress bar was already present for other operations).

Currently the other progress (ProgressMeter) pops up in the center of the screen, so both are visible.

In order to do this cleanly, I needed uistate in the TreeBaseModel; this rippled through all the dependent classes as well as the FlatBaseModel and its dependent classes for compatibility.

If (or when) this and PR316 are accepted, I will also finalize a fix for the deeprelationshippath filter and its progress bar, to get rid of its dependency on gramps.gui.utils ProgressMeter

### DIFF
--- a/gramps/gui/selectors/baseselector.py
+++ b/gramps/gui/selectors/baseselector.py
@@ -268,11 +268,10 @@ class BaseSelector(ManagedWindow):
 
         #reset the model with correct sorting
         self.clear_model()
-        self.model = self.get_model_class()(self.db, self.sort_col,
-                                            self.sortorder,
-                                            sort_map=self.column_order(),
-                                            skip=self.skip_list,
-                                            search=filter_info)
+        self.model = self.get_model_class()(
+            self.db, self.uistate, self.sort_col, self.sortorder,
+            sort_map=self.column_order(), skip=self.skip_list,
+            search=filter_info)
 
         self.tree.set_model(self.model)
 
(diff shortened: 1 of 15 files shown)


Nick-Hall:
I like the idea of displaying view progress in the status bar.  For consistency, we should do this in all cases not just for filtered views.

prculley:
I intend to try to shift more (pop-up) progress bars to the status; just takes time, and I wanted to put this PR out as a trial balloon, so to speak.

Nick-Hall:
The text next to the progress bar reads "Progress Information" rather than "Loading items...".

The progress bar when using the search bar rather than the filter sidebar should also be moved to the status bar.  What other progress bars are you considering moving?

prculley:
I can change the "Loading Items" (too many sources of title...).

The search bar progress status is more problematic.  Firstly I have a heck of a time getting it to pop up at all.  I found it can be started from the place selector, or the citation selector.  Both of these have search bars (Why doesn't people selector have a search bar?).
But Even with my newly created 5500 person db (tools/debug/generate test cases), the bar doesn't come up unless I run in debug mode (~1/4 normal processing speed).  The display updates in under the 2 sec popup timeout limit.
And then the popup has a 'cancel' button which allows for the screen update to be cancelled.  I'm not sure if this can/should be integrated into the status bar.
And lastly, when brought up from a selector, it seems possible that the  Gramps main window might be covered up by the editor and/or selector windows.  So if I moved this progress to the status bar, it might not be seen.

As to what other progress bars I might move to status area; I would consider a candidate any that are popped up with only the main window showing.  The import and backup progress are already there, and with the treeview progress moved, I am unaware of any others at the moment.  Except for the DeepRelationshipPath filter that started this whole line of examination...  I have some thoughts on that one that will be the subject of another PR.

Nick-Hall:
@prculley Use `time.sleep(0.005)` in the loop to trigger the status bar for testing.  Sort on a different column to force the view to rebuild.  Hide the sidebar to reveal the search bar in the views.

It is probably possible to cover the status bar when loading any view.  I don't know why it is possible to cancel a search, but not a filter.

prculley:
So the latest tweak adds a commit which uses the main window status bar for the progress indicator even when using selectors (tree view) or the non-filter sidebar tree views.  Also fixed the text "Loading Items" in the status bar.

Now to see if I can do something about the horrible performance (redraw time) in the selectors with tree view (several bugs on tracker about this)...

PR #324, opened 2017-01-11
upgrade Update Addons to ManagedWindow

prculley:
This change was initiated to correct status bar and popup dialogs that were coming up in the wrong place under Windows.
Upgrade to Managedwindow allows for size/position config, correct popup of subordinate dialogs, and better window management in relation to other dialogs.
The Glade change allows the transient parent function to work.
Note that UpdateAddons remains a modal dialog.

During Gramps startup, it is possible to start two modal dialogs;
the 'Family Trees' and the 'Update Addons' (when preferences is set to check on startup).  Both are modal. This presented some issues on Windows, at least.

So the second commit upgrades the Window manager and ManagedWindow to support this case.  If this occurs, then the most recent modal window is left that way, a previous modal window is temporarily downgraded to non-modal, and the newest window is set 'transient' for the previous modal window, even if it was not the parent.
This means that there can be only one modal window, and it is always on top and active.
During testing for this change, I ran into a conflict with the Gtk.Window.present and the ManagedWindow.present methods, so I renamed the ManagedWindow version.

I've put both commits in the same PR as they can be best tested together.

### DIFF
--- a/gramps/gui/plug/_windows.py
+++ b/gramps/gui/plug/_windows.py
@@ -1072,16 +1072,16 @@ class ToolManagedWindow(tool.Tool, ToolManagedWindowBase):
 # UpdateAddons
 #
 #-------------------------------------------------------------------------
-class UpdateAddons:
+class UpdateAddons(ManagedWindow):
 
-    def __init__(self, addon_update_list, parent_window):
+    def __init__(self, uistate, track, addon_update_list):
         self.title = _('Available Gramps Updates for Addons')
 
+        ManagedWindow.__init__(self, uistate, track, self, modal=True)
         glade = Glade("updateaddons.glade")
-        self.window = glade.toplevel
+        self.set_window(glade.toplevel, None, None)
         self.window.set_title(self.title)
-        self.window.set_size_request(750, 400)
-        self.window.set_transient_for(parent_window)
+        self.setup_configs("interface.updateaddons", 750, 400)
 
         apply_button = glade.get_object('apply')
         cancel_button = glade.get_object('cancel')
@@ -1135,13 +1135,11 @@ class UpdateAddons:
                 pos = iter
         if pos:
             self.list.selection.select_iter(pos)
-        self.window.run()
 
-    def close(self, widget):
-        """
-        Close the dialog.
-        """
-        self.window.destroy()
+        self.show()
+
+    def build_menu_names(self, obj):
+        return (self.title, " ")
 
     def select_all_clicked(self, widget):
         """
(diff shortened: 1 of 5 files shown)


prculley:
It figures that the Travis test would find a case I had not considered; using a ManagedWindow, without ever calling ManagedWindow.set_window().  Easy to fix with an otherwise unneeded initialization.

gramps/gui/plug/_windows.py
```diff
@@ -1072,16 +1072,16 @@ def __init__(self, dbstate, uistate, options_class, name, callback=None):
...
         self.title = _('Available Gramps Updates for Addons')
 
+        ManagedWindow.__init__(self, uistate, track, self, modal=True)
         glade = Glade("updateaddons.glade")
-        self.window = glade.toplevel
+        self.set_window(glade.toplevel, None, None)
         self.window.set_title(self.title)
-        self.window.set_size_request(750, 400)
-        self.window.set_transient_for(parent_window)
+        self.setup_configs("interface.addonscheck", 750, 400)
```
Nick-Hall: Perhaps we should rename this class to `AddonsCheck`?

gramps/gui/plug/_windows.py
```diff
@@ -1135,13 +1135,17 @@ def __init__(self, addon_update_list, parent_window):
...
             self.list.selection.select_iter(pos)
-        self.window.run()
 
-    def close(self, widget):
+        self.show()
+
+    def build_menu_names(self, obj):
+        return (self.title, " ")
+
+    def close_btn(self, widget):
```
Nick-Hall: Is this method needed?  Why not just connect to `self.close`?

prculley: ManagedWindow.close does not accept a 2nd parameter (obj)...

gramps/gui/plug/_windows.py
prculley: Are you suggesting that we actually rename the _class_?  Or that I should rename the config file "addonscheck" to "updateaddons"?  I did not think that there had to be a direct correspondence between the config interface element name and the class (at least there doesn't seem to be a direct link in a number of cases).

Nick-Hall: There is no direct correspondence, but it may cause confusion in the future if we make them different.  It would be easier to rename the config variable.  Did you have a resaon for using _addonscheck_?

gramps/gui/plug/_windows.py
Nick-Hall: Yes it does.  See [line 569](https://github.com/gramps-project/gramps/blob/master/gramps/gui/managedwindow.py#L569).

Nick-Hall:
Rebased.

PR #325, opened 2017-01-12
Fix AddMedia to use updated ManagedWindow modal parameter

prculley:
and Remove now unused ManagedWindow modal_call method.

This eliminates one of the ways that ManagedWindow classes can implement a modal window.  The use of an initialization parameter (modal=) on the ManagedWindow is preferred.


Nick-Hall:
Thanks.

PR #328, opened 2017-01-16
Fix 'Tip of the day' dialog to stay on top of main window

prculley:
Previously never had a transient parent because it failed to call
ManagedWindow.show()


Paul-Franklin:
Fixed in https://github.com/gramps-project/gramps/commit/fb8d529e5ba2e930ec0a0981acb437437cad278a

We were clearly thinking similarly.  8-)

(Sorry.  Didn't mean to steal your p.r..)

prculley:
That will teach me not to pull the master more often ;-).

Paul C.

PR #330, opened 2017-01-18
Change Add,Remove to New,Delete

ewongbb:
Re: bug [9229](https://gramps-project.org/bugs/view.php?id=9229).

Work In Progress as I think I might have others to fix; but would like to have some comments on this.

### DIFF
--- a/gramps/gui/views/listview.py
+++ b/gramps/gui/views/listview.py
@@ -207,9 +207,9 @@ class ListView(NavigationView):
 
         self.edit_action = ActionGroup(name=self.title + '/ChangeOrder')
         self.edit_action.add_actions([
-                ('Add', 'list-add', _("_Add..."), "<PRIMARY>Insert",
+                ('Add', 'list-add', _("_Create..."), "<PRIMARY>Insert",
                     self.ADD_MSG, self.add),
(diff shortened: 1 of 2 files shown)


prculley:
Just out of curiosity, why?  I personally think the current terminology is quite clear.

Paul-Franklin:
@ewongbb What language do you run gramps in?

ewongbb:
@prculley : this is in re: bug 0009229, and I'm using en-US

Nick-Hall:
@ewongbb Changing _Remove_ to _Delete_ is a good idea in this case.  We use _Remove_ in the editors when we remove an object from a list without actually deleting it.

I'm not so sure about changing _Add_ to _Create_.  The bug report suggests _New_.

I suggest that you work on these two changes separately starting with _Delete_.

Nick-Hall:
Delete renamed to Remove in commit 22e724366039ab104f14fbfdd63ca2c8be58d3fe.

The term "Delete" should be used when an object is deleted from the database. The term "Remove" should be used when an object is removed from a list, but not actually deleted.

I don't see much point in renaming "Add" to "New". It could be confusing for users who are happy with "Add" and "Share".

PR #331, opened 2017-01-18
bug 9886; fix citation tree view quick report right click

prculley:
Another HandleError issue;  Also, the Quick report that started was
actually the wrong one, due to a typo in the quickview.gpr.py.

I also rediscovered the bug 6352; and its relationship to quickreports.
It seems that you cannot run a quick report on a source from citation tree view.  Even if it looks like you can, it really ran on the last citation you looked at, not the source.  If you never look at a citation in this view since startup, the quick report menu item doesn't even appear.


sam-m888:
Thank you

PR #332, opened 2017-01-19
Make logging consistent in grampsapp.py

ewongbb:
re: bug [6376](https://gramps-project.org/bugs/view.php?id=6376): Logging in grampsapp.py is mixed up.

### DIFF
--- a/gramps/grampsapp.py
+++ b/gramps/grampsapp.py
@@ -33,8 +33,6 @@ import signal
 
 import logging
 
-LOG = logging.getLogger(".")
-
 from subprocess import Popen, PIPE
 
 #-------------------------------------------------------------------------
@@ -118,15 +116,17 @@ if win():
         elif not os.path.isdir(HOME_DIR):
             os.makedirs(HOME_DIR)
         sys.stdout = sys.stderr = open(logfile, "w")
+
 stderrh = logging.StreamHandler(sys.stderr)
 stderrh.setFormatter(form)
 stderrh.setLevel(logging.DEBUG)
 
 # Setup the base level logger, this one gets
 # everything.
-l = logging.getLogger()
-l.setLevel(logging.WARNING)
-l.addHandler(stderrh)
+
+LOG = logging.getLogger(".")
+LOG.setLevel(logging.WARNING)
+LOG.addHandler(stderrh)
 
 # put a hook on to catch any completely unhandled exceptions.
 def exc_hook(type, value, tb):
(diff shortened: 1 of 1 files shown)


Paul-Franklin:
As I am not an expert in logging, I would like to see comments
from both Josip and John Ralls, about their respective bundles.

jralls:
@ewongbb do you understand that multiple calls to logging.getlogger with the same path ('.' in this case) return references to the same object? In other words, you're not using a different logger for the locale messages, you're just using a different name for the same logger.

But I don't understand why you think that those messages even need a separate logger. They're debug messages, intended to gather information from the user when she reports a problem to help diagnosis. What would be the benefit of having to set debug on two loggers to include the locale information?

kulath:
Have you managed to setup one or more test scenarios where all these logging messages are output. I suspect that some of the logging is coded in a way that it can't actually be output.

jralls:
@kulath, the local environment stuff is output only by the CLI, so if you say
```
python Gramps.py -l --debug=".locale"
setup debugging .locale
2017-01-20 16:08:50.840: DEBUG: grampsapp.py: line 421: Encoding: UTF-8
2017-01-20 16:08:50.840: DEBUG: grampsapp.py: line 422: Translating Gramps to en
2017-01-20 16:08:50.840: DEBUG: grampsapp.py: line 423: Collation Locale: en_US
2017-01-20 16:08:50.840: DEBUG: grampsapp.py: line 424: Date/Time Locale: en_US
2017-01-20 16:08:50.840: DEBUG: grampsapp.py: line 425: Currency Locale: en_US
2017-01-20 16:08:50.840: DEBUG: grampsapp.py: line 426: Number-format Locale: en_US
2017-01-20 16:08:50.841: DEBUG: grampsapp.py: line 430: Using LANG: en_US.UTF-8
2017-01-20 16:08:50.841: DEBUG: grampsapp.py: line 435: Using LANGUAGE: en
Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.
List of known Family Trees in your database path

/Volumes/SSD/Google Drive/grampsdb/5610726a with name "CrashTest42"
/Volumes/SSD/Google Drive/grampsdb/548e3383 with name "GCRalls42"
/Volumes/SSD/Google Drive/grampsdb/55d0495f with name "GrampsExample41->42"
/Volumes/SSD/Google Drive/grampsdb/548e33f2 with name "Jones42"
/Volumes/SSD/Google Drive/grampsdb/55e38e8a with name "Ralls-42"
```

Your question did lead me to do a little messing around. The "l" logger, deleted by 7a5c055, exists to set the formatter and handler for all of the other loggers in the program. Setting them on logger "." doesn't do that, so with this change all logging in Gramps except the uncaught exception hook is broken because without a handler the log messages go to /dev/null.

Starting Gramps in GUI mode prevents --debug=logger from working, so the only way I know to get the locale messages and the one other log message, "A GUI is needed, set it up", for which @ewongbb felt it necessary to separate out LOG and LOG_LOCALE to display is to call ```LOG.setLevel(logging.DEBUG)``` after instantiating it.

After all of which I conclude that we don't want this change and that bug 6376 should be closed "not a bug", perhaps with the explanation of what the "l" logger is for and why it doesn't need to be called. Sorry, @ewongbb, I know you were looking for a simple problem to get your feet wet and it's kind of discouraging to get shot down on your first try.

eno93:
By the looks of it this can be closed as "No Change required" same with the Gramps issue right?

PR #333, opened 2017-01-20
9911: Provide an option for custom colors in Hourglass graph

sam-m888:
[9911](https://gramps-project.org/bugs/view.php?id=9911)  The Relationship and the Family Lines graphs provide options to color the persons/families (or their outline) (on the "Graph Style" tab), and makes those colors customizable. The Hourglass graph has the same color option but the colors are hardcoded and cannot be customized.

hourglass-graph-graph-style-tab-grampsmaster


sam-m888:
Thank you

PR #334, opened 2017-01-20
Improve progress indicator for deep relationship filter

prculley:
and improve gui.utils.ProgressMeter to help it find a parent window.

The deep relationship filter now actually shows progress, not just an activity indicator.

I've spent way too much time trying to figure out a way to fix up the plumbing for this one rarely used filter.  So here is a fix that at least makes it work for the person tree view, exports, and reports in both GUI and CLI modes.

The fix is not ideal in that it does NOT use the User class or similar.  That would take extensive architectural reworking in filters, proxy db, db, reports, export, and views.  After spending almost a week unsuccessfully trying to figure out limited changes to most of these, I am retreating, at least for now.

This work was an offshoot of the [8128](https://gramps-project.org/bugs/view.php?id=8128#c49001) transient parent bug.

In the research for this I did find a couple more bugs which I CAN deal with...

### DIFF
--- a/gramps/gen/filters/rules/person/_deeprelationshippathbetween.py
+++ b/gramps/gen/filters/rules/person/_deeprelationshippathbetween.py
@@ -45,7 +45,6 @@ def filter_database(db, progress, filter_name):
     """Returns a list of person handles"""
 
     filt = MatchesFilter([filter_name])
-    progress.set_header(_('Preparing sub-filter'))
     filt.requestprepare(db)
 
     progress.set_header(_('Retrieving all sub-filter matches'))
@@ -102,6 +101,8 @@ def find_deep_relations(db, progress, person, path, seen, target_people):
     if handle in seen:
         return []
     seen.append(handle)
+    if progress:
+        progress.step()
 
     return_paths = []
     person_path = path + [handle]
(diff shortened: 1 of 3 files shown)


SNoiraud:
Perhaps we should add the progressmeter to all filters. With my big database, I often wait the filtering and I don't know what happens. Perhaps it will become easier to solve this problem ?

Nick-Hall:
@SNoiraud The generic filter [`apply`](https://github.com/gramps-project/gramps/pull/334) method provides a progress callback, but it is not often used.

A better fix would be to pass a `User` instance to the `apply` method which could be used for both the main loop and the `prepare` methods.  This wouldn't involve any major architectural reworking.

prculley:
@Nick-Hall 

> This wouldn't involve any major architectural reworking.

My confidence in in undertaking this is not high; I personally think it is more difficult than that.  So I'll leave it to someone else.

I think it would be a good idea to accept this as it is until someone has the time to 'do it right'; users would get the benefit in the meantime.

Nick-Hall:
> My confidence in in undertaking this is not high; I personally think it is more difficult than that. So I'll leave it to someone else.

@prculley OK.  I'll have a look at this one.

Nick-Hall:
I have converted filters to use the `User` class.  See PR #338.

Progress information is now available for all filters.

prculley:
With changes implemented by Nick; this becomes mostly irrelevant.  So am closing.

PR #336, opened 2017-01-22
Improve Addon management performance

prculley:
I do a lot of debugging and Gramps runs pretty slow under the debugger.  However, it seemed that the startup and view initialization was remarkably slow for what was happening.  So I decided to take a look to see if any optimization was possible.

I discovered that:
1 - the plug in manager PluginRegister spent a lot of time (seconds) in one tiny routine, get_plugin().  It was a 'list comprehension', a loop through a list to find a registered plugin by its id.  The inner loop was executed several hundred thousand times.
2 - The list of registered plugins had two copies of everything, making the loop in (1) twice the size necessary.
3 - The plug in manager 'reg_plugins()' was called twice for each plugin directory, resulting in two copies of the registration, two scans of the directories etc.
4 - The db manager was calling 'reg_plugins()' to allow getting the dummydb plugin, before the CLI called it normally.
5 - While scanning for plugins, the 'reg_plugins()' was also running through all the '__pycache__', 'po' and 'locale' directories (unnecessarily), despite code that looked to be trying to exclude them.
6 - Each directory was scanned twice, once by the os.walk, and once again by the os.listdir.  Combined with (3) above each directory got scanned a total of 4 times.

This PR fixes the above.
- I changed the db manager to load the dummydb directly, moving the dummydb to the gen.db.  I judged that since it was a basic part of Gramps, loaded when no other db was loaded, it did not need to be a plugin.  This eliminated one of the passes through the plugin registration.
- I added a dict to allow direct lookups of plugins by id, thereby avoiding many thousands of passes through the loop in (1).
- I fixed up the os.walk call so it would actually ignore the 'po' and 'locale' directories as intended, and added the '__pycache__' to the directories to ignore.
- I passed in the list of files that os.walk accumulated in each directory to the _pluginreg scan_dir routine rather than scanning for the files again.
- I removed some dead code (related to __registereddir_set) that did nothing useful.

The original code had the ability to have duplicate plugins or at least duplicate IDs for them; the updated code will generate an assertion if this occurs (in debug mode).

During unit testing, I found that since portions of Gramps are run abnormally, multiple registrations could occur, so I added a test to see if plugins in a directory were already scanned and skipped the directory walk when that was the case.

When running through by debugger Gramps starts quite a bit faster, and the person view initialization with a lot of Gramplets went from 5-6 seconds to less than a second.  I mad no attempt to measure normal startup and running, but I expect that users with slower machines will notice an improvement.


gramps/gen/plug/_pluginreg.py
```diff
@@ -1123,7 +1128,7 @@ def scan_dir(self, dir, uistate=None):
         extlen = -len(ext)
         pymod = re.compile(r"^(.*)\.py$")
 
-        for filename in os.listdir(dir):
+        for filename in filenames:
             name = os.path.split(filename)[1]
```
QuLogic: Is this function called from anywhere else? This should be a no-op since the result of `os.walk` is _just_ the file names.

prculley: Actually, os.walk (in the calling module) yields a 3 tuple: (dirpath, dirnames, filenames).  I passed the filenames as well as the original dirpath through to _pluginreg in the line above.
And no, the scan_dir() method in _pluginreg is not called by anything else in the current Gramps or Gramps Addons.

QuLogic: Yes, but I mean that every value in `filenames` is a single file in `dirpath`. There is no subdirectory, so this `split` does nothing.

uxcq:
This makes Travis-ci [fail](https://travis-ci.org/gramps-project/gramps/builds/194311850#L1837) for the webcal and narratove web report plugins.

prculley:
I am aware that Travis is failing; I am attempting to find a reason. These plugins don't fail under the GUI or CLI when run individually on my system.  And under my Linux (the only place I have ever gotten nose testing to work) the test_reports module passes as well.  So something different is going on.
It is difficult to debug when the only place I see a failure is Travis...

romjerome:
Is it planned to remove '[gramps/plugins/db/dummydb.gpr.py](https://github.com/gramps-project/gramps/pull/336/files#diff-d4a8de2da2889e97c354e20f3114536f)'? 
Maybe this could explain travis issue?

gramps/gen/plug/_pluginreg.py
prculley: got it...

prculley:
Finally finished debugging Travis issues.  Travis runs portions of Gramps many times; this was causing several problems.
First is that many of the tests start in a way that did not create the user directories (.gramps/css, gramps/gramps50/plugins etc.).  If these got run first, (before one which did create user directories) and they depended on the presence of these directories, then they failed.  So I added a check for and creation of these directories to the test utilities so that application level tests would have them, regardless of test order.

Another problem was that I had assumed that the plugin scan would only happen once; under test, this was definitely not the case.  If I aborted the scan completely on additional passes, then, in combination with the first problem, some of the plugin data might not get created (the load_on_reg webstuff file).  If I allowed it to run multiple times, it failed because there would be lots of copies of data attached to the webstuff pdata.  So I had to fix so that even with multiple runs, it would only have one copy of data (I initialized the pdata.data entry every time).  That way, when it ran in a later test with correct user directories present, it would end up with correct data, during the specific test that needed it.

I had to fix up the test utilities to allow prior console output, even if an exception occurred, so I could see what was happening as I debugged.  And I added a number of LOG.warning lines (now commented out) to see what was going on in the Travis listing.  I left them in as comments so if anyone ever has to do this again, he can undo the comment to see some debug info.  Did not want to 'pollute' the Travis listing and logs all the time.

sam-m888:
Works for me and looks good, but as it touches something as major as the plugin/addons system I'd like at least one more set of eyes over this to test then merge if that ok @Nick-Hall ?

kulath:
@prculley Re: dummydb "loaded when no other db was loaded". My long term intention is that when no database is loaded, there should really be no database loaded, not that dummydb is loaded. dummydb is really just a workaround because some parts of Gramps don't recognise the possibility that there is no database, and/or don't properly respond to the no-database (or similar) signals. I had been working my way through parts of Gramps to fix this and eventually remove dummydb.

(I don't like workarounds - the code should do the right thing. I had previously added dummydb and changed the code from using a real empty database to using dummydb so that the code was a bit cleaner.)

I don't think this affects anything that you are doing, but as you mentioned dummydb, I thought I would let you know. Calling it directly probably makes it easier to remove.

prculley:
@sam-m888 Why the WIP label?  I think this is ready to go...

sam-m888:
Your right maybe we need another label like "Waiting on Review"?

Nick-Hall:
@sam-m888 I can see people who have requested a review from me if I use a filter.  However, a new label may also be useful.

Nick-Hall:
Sorry I clicked the merge button by mistake.  It would have been neater to rebase first.

sam-m888:
@Nick-Hall Ok, I've added the "**Review Required**" label for the future. Thanks

PR #337, opened 2017-01-23
Allow user to define expand_tree to SelectPerson

ewongbb:
This is for [bug 9738](https://gramps-project.org/bugs/view.php?id=9738).
Based on master

### DIFF
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -187,6 +187,7 @@ register('geography.use-keypad', True)
 # needed), for instance to four 'interface.clipboard' variables --
 # so do a recursive grep for "setup_configs" to see all the (base) names
 register('interface.dont-ask', False)
+register('interface.expand-tree', False)
 register('interface.view-categories',
          ["Dashboard", "People", "Relationships", "Families",
           "Ancestry", "Events", "Places", "Geography", "Sources",
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -187,7 +187,7 @@ register('geography.use-keypad', True)
 # needed), for instance to four 'interface.clipboard' variables --
 # so do a recursive grep for "setup_configs" to see all the (base) names
 register('interface.dont-ask', False)
-register('interface.expand-tree', False)
+register('interface.expand-selector', False)
 register('interface.view-categories',
          ["Dashboard", "People", "Relationships", "Families",
           "Ancestry", "Events", "Places", "Geography", "Sources",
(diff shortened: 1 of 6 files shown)


gramps/gen/config.py
```diff
@@ -187,6 +187,7 @@ def emit(key):
 # needed), for instance to four 'interface.clipboard' variables --
 # so do a recursive grep for "setup_configs" to see all the (base) names
 register('interface.dont-ask', False)
+register('interface.expand-tree', False)
```
Nick-Hall: I would prefer 'interface.expand-selectors'.

gramps/gui/configure.py
```diff
@@ -1108,6 +1108,13 @@ def add_formats_panel(self, configdialog):
                           row, 'interface.grampletbar-close', stop=3,
                           extra_callback=self.cb_grampletbar_close)
         row += 1
+
+        # When using Select <item>, do the trees expand by default?
+        self.add_checkbox(grid,
+                          _("Expand trees by default."),
```
Nick-Hall: Change to "Expand selectors by default".   The trailing full stop is not needed.

prculley:
I tested this and it seems to work, mostly.

However, I don't get what you were trying to do with the gramps/gui/dialog SaveDialog.  It looks like a leftover of some experiment, and it doesn't work.

Also, in relationship view, the children selector is not covered by this.

Another thought, @Nick-Hall suggested the change 'Expand selectors by default'; I think that since not all tree like selectors are expanded (place tree selectors), you might use something like 'Expand the entries in dialogs when choosing from existing people'.  It might be nice to add a tooltip that explains that for larger trees this slows down the dialog.

prculley:
@ewongbb Are you still willing to work on this? 

If not, let me know and I will try to get it finished up (may have to close this and start another PR).

 If so, I think it would be better to move the 'expand-selector' check from the editfamily.py to the selectperson.py module.  That way it will work with all the code that uses the person selector.  As it is, I think there are at least two other places that use this selector and don't get the option with your current code.

I think it would also be useful to reinstate the code that was removed in commit 10629b1 so that when users don't have it expanded it behaves like earlier versions of Gramps (and expands to current object only).

Nick-Hall:
@prculley Yes.  You will need the code that expands the selected path, if you no longer expand all rows.  It is also old code that can be written more concisely for Gtk3.

prculley:
@Nick-Hall I did some testing of a variation of this PR (switching between List view and Tree view instead of expanded vs. collapse Tree).  I inserted the removed code from commit 10629b1, only to discover that it doesn't seem to be used.  So I suspect that I did not get all of the changes.  Do you know where other commits relating to this might be?

BTW, I like the List view instead of expanded Tree; with 16000 persons, List view opened in 0.9 sec, the collapsed tree opened in 2.7 sec, the current code expanded tree took 8.5 sec.  I understand why users with large trees are complaining.

sam-m888:
Closed see #384

PR #338, opened 2017-01-26
6085: Convert filters to use the User class

Nick-Hall:
The filter apply method can now be passed a User instance
instead of a progress callback.

The prepare progress will no longer be displayed without the main
filter progress.

### DIFF
--- a/gramps/gen/filters/_genericfilter.py
+++ b/gramps/gen/filters/_genericfilter.py
@@ -29,7 +29,6 @@ Package providing filtering framework for GRAMPS.
 # Gramps imports
 #
 #------------------------------------------------------------------------
-from ..const import GRAMPS_LOCALE as glocale
 from ..lib.person import Person
 from ..lib.family import Family
 from ..lib.src import Source
@@ -40,6 +39,8 @@ from ..lib.repo import Repository
 from ..lib.media import Media
 from ..lib.note import Note
 from ..lib.tag import Tag
+from ..const import GRAMPS_LOCALE as glocale
+_ = glocale.translation.gettext
 
 #-------------------------------------------------------------------------
 #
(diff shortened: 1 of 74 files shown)


prculley:
I tried this out and it still has some more work to do, I think.  I tested with the DeepRelationShip filter: my favorite ;-).
On my machine, the progress popup from people tree view starts on the main window, but then jumps to my secondary window, and I get two "GtkDialog mapped without a transient parent. This is discouraged." messages.

When I tested with a report I got no progress at all.
When I tested with an export, I also got no progress.

P.S. I had gotten about as far as this with a similar attempt of my own, when I realized how much more there was to do (proxy db etc.)...  I had pushed 'user' into the _rule base class ('rule.user = user' in caller), I decided to let interested filters see if a self.user was present, rather than edit all the individual filters that way.

Anyway, I can assist on the first issue, if you would like to attack the plumbing for the second two.

Nick-Hall:
The commit only updated code that previously used the progress callback to show the progress of the main filter loop.

I have now included other filters.  See 1e5405495373728adbeeb29bda3d1bae92bcf787, 2f2b619762b763939966351bf01bb28515bcbe4f and 68fbebe7d18b97b7e3906da9e8bd7299bdb41b63.

prculley:
And keep in mind that when you plumb up the elements that are required for reports and exports, that this all started with the need to get a window parent to the new window.  Using user class is not, by itself, sufficient to take care of this.  For views and reports, I think that the Gramps main window is parent, so the ultimate progress indicator (and other popup dialogs, if any) can get this from user.uistate.window.

But this isn't sufficient for exports, where the parent window is the export dialog.  Depending in how you plumb this, it MAY be possible to create a new user instance (with a copy of uistate, rather than a reference) and then adjust the uistate.window to the export dialog, although that is kind of a hack too.

If you are going to pass all the parameters down, you probably need to include the parent parameter as well.

I don't know if there are any other users of filters out there, but if so, we need to take a look at them as well.

These considerations are why I regarded this as a non-trivial fix...

prculley:
When this settles down a bit, I have a tweak for the DeepRelationshipPath filter; currently it calls step_progress a lot more times than the total.  Which makes it appear that the progress freezes when (almost) finished.
If the step is moved out of the 'for family_person' loop up a few lines to just after the tests which 'return', the step count is much more accurate.  It can still be off; too few steps if the family tree contains a lot of non-relatives, and too many steps if there are a lot of loops in the tree, but generally much closer and more user friendly.

Nick-Hall:
@prculley The main objective of this fix was to convert the filters to use the `User` class.  This removes another gui import listed in bug [#6085](https://gramps-project.org/bugs/view.php?id=6085).

The first commit converted six reports and the tree views.  These were the only calls that previously used the old progress callback mechanism.  The prepare progress will now only display with the main loop progress.  Since your last comment I enabled the exports, the complete individual and place reports and the sort events tool.

At the moment, the progress meter uses the main application window as its parent.  This doesn't cause a problem using my OS, but if there is an issue with Windows, a parent window can be specified in a `User` instance.

PR #340, opened 2017-01-27
9914: Add [Arrowhead direction] option to Hourglass graph

sam-m888:
Bring a bit more consistency to the three graph reports. Adds the [Arrowhead direction] option to Hourglass graph.

hourglass-arrowhead-direction-option-drop-down-list


uxcq:
Tested this and it worked. 

@sam-m888 BTW the second example image is duplicate of the 1st.

sam-m888:
@uxcq Thanks for testing. I've removed second image.

PR #343, opened 2017-01-30
Bug 9913: Fix Glade file loading to avoid loading and leaving extra toplevels

prculley:
This involved a list of fixes.
Change Glade class so that it operates in two modes; if no toplevel parameter is given, load whole file, if given, load only the specified toplevel object.  Thanks to @bpisoj (Josip) for this idea and initial patch.

As usual this created its own problems.  I went through each glade file that had more than one toplevel object and tested via the GUI.
- Some glade files needed additional toplevel objects to perform correctly; I added a new parameter to the Glade class initialization; 'also_load'.  When loading with a specified toplevel, this parameter provides a list of additional objects to load.
- Some glade files included with multiple objects just needed to have the calling file adjusted with a toplevel parameter.
- Some glade files included dead objects.  I removed the unused objects.
- Some glade files were completely unused; I deleted them.

To assist in this effort, I created a Gramplet that reports on number of Gtk windows and lists them.  This Gramplet is not included here, I don't think it is ready for prime time...

After this was done, I still found a couple of cases where calling up windows that include glade files still left extra windows in memory when the expected dialog was dismissed.
The owner editor and report dialog needed a bit of attention to properly destroy leftover glade objects.

I am not sure I found everything, but will keep my little Gramplet around for a while to see if I spot any more issues.

P.S. I also spotted a couple more transient parent issues; I tried to break these out into a separate commit...


sam-m888:
Worked for me when tested (used the leak gramplet to check)

Needs another set of eyes to check this over before being accepted.

Paul-Franklin:
Note that glade files are in POTFILES.in also (so when
exportcsv.glade is removed that will need editing too).

Paul-Franklin:
What happens when you kill ownereditor.py with
a delete-event signal (and thus call MW's self.close)?
Is everything still OK?

prculley:
@Paul-Franklin Thanks for the catches; the delete-event signal turned out to be more difficult than I expected; again the ManagedWindow code had connected to that signal and was calling close for me so the connect in the Glade file wasn't getting executed at all.
Turned out the easiest way to deal was to tell ManagedWindow to destroy my Glade object for me...
Also had some fun rebasing with your own patches...

Paul-Franklin:
There are some things I don't understand.

How can you remove the three signals from book.glade
while leaving those three signals in _bookdialog.py (and
the three methods they then started?

Paul-Franklin:
In _styleeditor.py, now that you no longer call set_window
how can you transfer the modality from the "other modal
window"?

Paul-Franklin:
In _styleeditor.py, now that you no longer call set_window,
does the new dialog still get the name "Style editor"?  How?

Paul-Franklin:
In ownereditor.py, wouldn't it have been easier to
define your new method as "close" and then to call
ManagedWindow.close() at the end of it?

Paul-Franklin:
In glade.py, should you change

:type  toplevel: string

to instead say "string or None"?

Paul-Franklin:
In glade.py, I don't see that the new self.toplevel_given is ever used?

Paul-Franklin:
"The owner editor and report dialog needed a bit of
attention to properly destroy leftover glade objects."

Should I see some change to the "report dialog" here?

prculley:
@Paul-Franklin Again, thanks for the careful review.

> How can you remove the three signals from book.glade while leaving those three signals in bookdialog.py (and the three methods they then started?

I had seen some warnings from Gtk around these and clearly did not think this through well enough or test it adequately either.  Fixed, along with a problem where the 'delete-event' (using 'x' in titlebar to close) did not cleanly close the booklist dialog.

> In _styleeditor.py, now that you no longer call set_window how can you transfer the modality from the "other modal window"?
In _styleeditor.py, now that you no longer call set_window, does the new dialog still get the name "Style editor"? How?

That was poor editing on my part when dealing with the merge conflict created on your change to a ManagedWindow... Sigh

> In glade.py, should you change 
:type toplevel: string
to instead say "string or None"?
In glade.py, I don't see that the new self.toplevel_given is ever used?

Fixed, and fixed (was a leftover from some experiments).

> Should I see some change to the "report dialog" here?

Should have said report dialog _papermenu.py PaperFrame class...

> In ownereditor.py, wouldn't it have been easier to define your new method as "close" and then to call ManagedWindow.close() at the end of it?

Possibly, but I re-examined OwnerEditor again; I got nervous about using Python's 'delattr()' to deal with the Glade object, (even though it appeared to work) so adjusted the code to use Gtk 'destroy()' instead.  Then could leave 'close()' as the ManagedWindow version with no override.

Paul-Franklin:
Thank you for making the changes.

Everything looks fine to me as long as you remember to
edit POTFILES.in when these are all committed.

prculley:
> Everything looks fine to me as long as you remember to
edit POTFILES.in when these are all committed.

Already in the most recent commit push for this PR...

Paul-Franklin:
"Already in the most recent commit push for this PR..."

That could be, but if so github isn't showing it to me,
when I look at each commit of yours.  But I am willing
to put that down to some anomaly of github, maybe
due to my unwillingness to turn on Javascript for them.
They always say "added some commits Jan 30, 2017"
even for your newer ones for instance.  If you say you've
already done it, that's fine with me.  Thanks.

prculley:
@Paul-Franklin Its in commit https://github.com/gramps-project/gramps/pull/343/commits/ef647b068069f504f0c3af3aeda070277d99b518, at the end when examined on the web page.

Paul-Franklin:
As I said, it's a github anomaly, since when I look at that
commit all it shows me is the 335 lines of exportcsv.glade
being deleted.  One file, 335 lines, even though the top
of that web page says "6 files" and "1247" lines in red
(being deleted).

sam-m888:
Thank you

PR #344, opened 2017-02-01
Add spouses to the child listing in DAR and DDR

bertcarnell:
Add spouses to the child listing in the Detailed Ancestor Report (DAR) and the Detailed Descendant Report (DDR) for issues #9926, #8941, and #5698.

### DIFF
--- a/gramps/plugins/textreport/detancestralreport.py
+++ b/gramps/plugins/textreport/detancestralreport.py
@@ -12,6 +12,7 @@
 # Copyright (C) 2011      Tim G L Lyons
 # Copyright (C) 2013-2014 Paul Franklin
 # Copyright (C) 2014      Gerald Kunzmann <g.kunzmann@arcor.de>
+# Copyright (C) 2017      Robert Carnell <bertcarnell_at_gmail.com>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


sam-m888:
Tested both reports and works, Thank you 👍 

I've attached test results to the relevant issues.

DAR:
dar_content_tab_list-spouses-of-children-gramps-master

dar-comparison-of-results

DDR:
ddr_content_tab_list-spouses-of-children-gramps-master

ddr-comparison-of-results

sam-m888:
Waiting on results of both:
-----------
* test/det_ancestor_report.sh 

-----------
* test/det_descendant_report.sh

bertcarnell:
I will fix the "Ignoring unknown options".

prculley:
I'll bet you waited (or are still waiting) a long time!  Looked to me like
those scripts would run Gramps over 8000 times, and produce as many files
to be examined.  Hope you have some time to do so 😀.

sam-m888:
>> I will fix the "Ignoring unknown options".

Thank you.

>> I'll bet you waited (or are still waiting) a long time!

@prculley Yes it took all night! :)

BTW had worked out before running that it was going thru all possible
combinations of the report, but not that it was that many combinations!

bertcarnell:
I am working on unit tests that will make checking the files easier for some subset of the 8000 cases and improve the code coverage I hope.  The purpose of the shell script seems to be a mostly operational check since no output is checked.  I may be able to make that better as well.  Once I have better unit test coverage, I'd like to refactor out the common parts of the DAR and the DDR to make maintenance easier.  Should I continue that work under this pull request, or would you rather just accept these simple changes first?

prculley:
I think it would not be too hard to add an 'options' list using the
gramps/plugins/test/test_reports addreport method.  It would then be easy
to adjust options in the report_list for your report, perhaps to add a
couple of variations (all options, every even option, and every opp
option), that would give pretty darn good coverage.

The way it is now, the only default options are tested...

Paul Culley

sam-m888:
>  Should I continue that work under this pull request, or would you rather just accept these simple changes first?

Lets keep it simple and accept this PR.

Thank you

PR #346, opened 2017-02-01
A new Soundex filter for matching 'name' with people using Soundex

prculley:
A response to [request](http://gramps.1791082.n4.nabble.com/Spelling-variations-and-the-SoundEx-algorithm-td4678729.html) on Gramps developers email list.

### DIFF
--- a/gramps/gen/filters/rules/person/_hassoundexname.py
+++ b/gramps/gen/filters/rules/person/_hassoundexname.py
@@ -1,0 +1,86 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2017  Paul Culley <paulr2787_at_gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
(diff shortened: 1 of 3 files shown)


gramps/gen/filters/rules/person/_hassoundexname.py
```diff
@@ -0,0 +1,92 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2002-2007  Donald N. Allingham
+# Copyright (C) 2007-2008  Brian G. Matherly
```
QuLogic: New file should get a reasonably new copyright date?

prculley: I should probably take credit as well...

gramps/gen/filters/rules/person/_hassoundexname.py
```diff
@@ -0,0 +1,92 @@
...
+    allow_regex = False
+
+    def apply(self, db, person):
+        for name in [person.get_primary_name()] + person.get_alternate_names():
+            if self._match_name(name):
+                return True
+        return False
+
+    def _match_name(self, name):
+        if self.list[0] and self._soundex_match(0, name.get_first_name()):
```
QuLogic: This is a bit repetitive; can be written as:
```python
if self.list[0]:
    if self._soundex_match(0, name.get_first_name()):
        return True
    elif self._soundex_match(0, name.get_surname()):
        return True
    ...

    for surn in name.get_surname_list():
        ...

return False
```
(and remove the `self.list[0]` check in `_match_surname`.)

QuLogic: Actually, you always pass `param_index=0`, so why not just send `self.list[0]` directly to `_soundex_match`?

sam-m888:
Works for me.

General filter > **Soundex match of People with the <name>** 
soundex match of people with the _name_ filter

 - [ ] Update documentation when it rolls over.

sam-m888:
Related feature request [1768: improve soundex tool. Integrate better with gramps](https://gramps-project.org/bugs/view.php?id=1768)  which asks for:

> 2. Allow finding surnames with the same soundex code. E.g. have an option to select a surname and then do a "find surnames with same soundex code" This could be a quick view and should also be possible in the soundex tool.

sam-m888:
Thank you

PR #347, opened 2017-02-02
Feature request [3146:Dynamic labels on Family Editor] & [9899:Family Group Report shows unmarried parents as husband and wife]

romjerome:
This pull request should cover (at least):
* [3146](https://gramps-project.org/bugs/view.php?id=9899): Family Group Report shows unmarried parents as husband and wife
* [9899](https://gramps-project.org/bugs/view.php?id=3146): Dynamic labels on Family Editor


prculley:
How hard would it be to add a test case for this to our test suite?  Seems like we could improve coverage at least for gen.plug.report.utils.  We already test Family Group report, so I suspect that we would only need to add one or more families with non-traditional parent arrangements to example.gramps.data.gramps.

Of course then we would have to update other report test results to avoid test failures, but that is pretty easy from the test results.

romjerome:
@prculley, 

Yes. Thanks!
You are right. I ignored test suite, but I was also further on [feature request 9899](https://gramps-project.org/bugs/view.php?id=9899) because we have two types (groups) of labels and strings, using generic words like "Father", "Mother", "Husband", "Wife".

There is one group, more global, for header column, view, GUI, or some reports like NarrativeWeb and Tag textual reports, and a second one, more specific, related to one family (father.handle and mother.handle). This pull request makes the focus on the second group, specific to an active family.

Just proposed a "light" version!

Maybe some tweak changes for handling uppercase, but I am not certain that having uppercase is a common way for Japanese, Chinese, Hebrew, Arabic, etc ... ?

romjerome:
> Seems like we could improve coverage at least for gen.plug.report.utils

True, some values were added into docstring (comment)...
Also, got problems for strings based on "Right To Left" locales like Arabic or Hebrew, as my IDE forced one order during edition! Yes, it was rather for tests, based on current translation strings for locales without localized relationship handler.

I do not understand the codecov stuff.
Are these lines on the docstring, the reported issue by codecov?

romjerome:
> How hard would it be to add a test case for this to our test suite?

Oh, strange, my commit also modified rights on one test file! (mode)
 gramps/test/regrtest.py 100755 → 100644 
I have an idea why this occured ... just wonder why on this file???

romjerome:
> if len(family.get_child_ref_list()) > 0:

This test seems incomplete for returning either "father"
or "mother" strings! We might also check parents -> child
status if one parent is a step-parent... This is more global
because all children whatever status with parents
are listed as children on Family Editor.

So, one corner case with two parents, one child, where
one "biological" parent and the second one a "step-parent".

True, "parents" is a generic word and we cannot make labels
too specific for only one child <-> parents relation.

Anyway, it is better than current labels.

sam-m888:
Would it please be possible to separate these two as it would make it easier to test.

I've attempted to test _9899: Dynamic labels on Family Editor_  and have the following results so far:

* New Family Editor comes up with both participants shown as "**Unknown**" (personally I'd rather you call them _participant_1 and _participant_2 as that is who they are.
new-family-editor-semi-dynamic-labels-unknown-default

Issues so far:
When you create a new female person in the 1st participant position and don't enter the label remains as _unknown_ until to press ok for the whole dialog:
new-family-editor-semi-dynamic-labels-female-before-pressing-ok

Returning to the dialog you can see that the label not shows that the female is reported as a **Man** in the 1st _participant_ position :
new-family-editor-semi-dynamic-labels-female-before-after-pressing-ok

When you enter a Male as the first _participant_ and _unknown_ for gender on the 2nd _participant_ you see the **partner** label (Should be an uppercase **P**) 
new-family-editor-semi-dynamic-labels-partners-male-and-unknown

Suggestion the New Family editor needs a few changes:
* Each _participant_ should have a gender symbol shown (like used in the Relationship view)
* Be able to handle any gender in either field and still create a validly reported  relationship
* labels are currently semi-automatic and only change on the next opening of the dialog

Did not test _3146: Family Group Report shows unmarried parents as husband and wife_

romjerome:
> Would it please be possible to separate these two as it would make it easier to test.

The first idea was to use the same code and logic for GUI and reports.
Few differences between 'gramps/gen/plug/report/utils.py'
and 'gramps/gui/utils.py'. They should be related. If need, I could separate them.

romjerome:
> New Family Editor comes up with both participants shown as "Unknown" (personally I'd rather you call them participant 1 and participant 2 as that is who they are.

I did not want to use the same type of label as the current one (Father/partner1 - Mother/partner2). Does it [really fir well](https://gramps-project.org/bugs/view.php?id=3146#c47769)?
Maybe an empty string is also possible?

romjerome:
> When you create a new female person in the 1st participant position and don't enter the label remains as unknown until to press ok for the whole dialog

Ah yes, I missed the gender check on this section.
Thanks.

romjerome:
> Returning to the dialog you can see that the label now shows that the female is reported as a Man in the 1st participant position 

same as the [previous one](https://github.com/gramps-project/gramps/pull/347#issuecomment-277431657).

romjerome:
> When you enter a Male as the first participant and unknown for gender on the 2nd participant you see the partner label (Should be an uppercase P)

Yes, I did not want to force uppercase() on the first letter because some locales like Japanase, Chinese, Arabic or Hebrew does not use uppercase on the translated string.

Many ways for showing the "uppercase P" (either via Gtk or python), but [none of them sounded good for me](https://github.com/gramps-project/gramps/pull/347#issuecomment-277016877).

romjerome:
> Each participant should have a gender symbol shown (like used in the Relationship view)

:+1: 

It was one idea on feature request [3346](https://gramps-project.org/bugs/view.php?id=3346#c11514).
See also feature requests [5730](https://gramps-project.org/bugs/view.php?id=5730) and
[https://gramps-project.org/bugs/view.php?id=9098](9098).

romjerome:
> labels are currently semi-automatic and only change on the next opening of the dialog

Right, I only looked at monitored fields for the relation type into the family.
As it was handled by relationship calculator, I did not look further.
Yes, could be like parent data once selected.

romjerome:
Note, by cleaning my testing version, where I also looked at feature requests like:
* [9119](https://gramps-project.org/bugs/view.php?id=9119): Additional Uses for Gramps: Botany Genetics - Plants (eg:Grapes) 
* Animals pedigree

there was a fallback based on translated "Father" & "Mother" strings, and a user config key for forcing (or not) customization. It was too much for one Pull Request.

romjerome:
> Did not test 3146: Family Group Report shows unmarried parents as husband and wife

The same proposal on this report can help us to test the code and localization as there is 
a support for localized relationship handlers, a lang selector on report dialog and a specific class for displaying translated strings.

romjerome:
Made some tweak improvements on Family Editor.
Yes, to use [gender symbol](https://github.com/gramps-project/gramps/pull/347#issuecomment-277432141) (like Relationship view) could be a more 
elegant solution.

Still have a problem for finding a nice solution around uppercase for strings returned by the Relationship calculator. To force uppercase is not a good solution, and to call some translated strings for specific cases is no more a generic (common) and global solution as
there is an exception.
Maybe we should return the relationship on tooltip string only?
No more as label (replaced by gender symbols), but into a tooltip: will be translated according to cultural issues.

romjerome:
> [9899: real 'dynamic' change on parent(s) labels](https://github.com/gramps-project/gramps/pull/347/commits/75df9bef2214f9e38a2c3e7a131971ea9ee521b3)

Maybe still need to monitor if people become parents (first baby => father or mother),
and to monitor if relation has changed (married, custom, civil union, etc ...).

sam-m888:
and 'gramps/gui/utils.py'. They should be related. If need, I could
> separate them.
>
>
No need, you are right, please keep together.

romjerome:
> you see the partner label (Should be an uppercase P)

Fixed on last version.
As Chinese, Japanese, Arabic and Hebrew locales do not have a localized relationship handler, they are using the english strings (fallback)!
Still need to check on the related translation files, but last version should already "force"
uppercase for english and some locales.

To monitor some specific data like relation type or number of children could be an extra work as we could look at check_for_family_change(), a callback for 'family-update' signal.
Problem, it seems to ignore change on relation type and it is based on dbstate.

Something like self.data_has_changed() seems more designed for that, but I do not have a proper solution yet.

romjerome:
Once more I cannot properly rebase ...
Fortunately, I disabled my push ability on gramps upstream.
Much more comfortable with 'git merge' command! Sorry.

Nick-Hall:
Rebased.  @romjerome Please check this.

In the future, instead of using `git merge`, use:
```
git checkout feature_3146_9899
git rebase master
```
Then force push to your fork.

Only merge your feature branch into master when you are finally happy with it.

romjerome:
@nick_h

That's the problem ...

I did:
git rebase master
first!

Then, when I saw the mistake on index (history)!
I made a merge after...

Also, see some tips like rather use:
git rebase -i

Sorry, but something is still confusing around
github, track, upstream and my local git repo (fork)
as after my pull request, new commits were checked
by codecov. Sure, I should limit changes after creating
the pull request. This could be the most difficult for me ...

About the code itself, I suppose I am able to play with
current requests and features related to this PR.

Just have a problem for properly monitoring some actions
not stored on database. e.g., on Family Edition, how to use signal
if we change the relation type or add/remove the number of
children. It is possible during DB commit as we check if data changed,
a little bit more difficult on the fly for a real dynamic update.

I have a dirty draft solution which will monitor action on Gtk object
(once active), then rebuild Gtk labels for parents. I will never add
such workaround and it does not work well for items on Gtk Combo Box.

Jérôme

--------------------------------------------
En date de : Mar 7.2.17, Nick Hall <notifications@github.com> a écrit :

 Objet: Re: [gramps-project/gramps] Feature request for 3146 9899 (#347)
 À: "gramps-project/gramps" <gramps@noreply.github.com>
 Cc: "romjerome" <romjerome@yahoo.fr>, "Mention" <mention@noreply.github.com>
 Date: Mardi 7 février 2017, 22h55
 
 Rebased.  @romjerome Please
 check this.
 In the future, instead of using git merge,
 use:
 git checkout feature_3146_9899
 git rebase master
 
 Then force push to your fork.
 Only merge your feature branch into master when you are
 finally happy with it.
 
 —
 You are receiving this because you were mentioned.
 Reply to this email directly, view
 it on GitHub, or mute
 the thread.

romjerome:
> Only merge your feature branch into master when you are finally happy with it.

There is a '_9899_large.patch_' file attached on [feature request 9899](https://gramps-project.org/bugs/view.php?id=9899). It should illustrate what could be possible by going further.

For now, I am stuck with this current feature branch!
I could go a little bit further on polish and improvements.
```diff
diff --git a/gramps/gui/editors/editfamily.py b/gramps/gui/editors/editfamily.py
index f3a6b4b..95b7b51 100644
--- a/gramps/gui/editors/editfamily.py
+++ b/gramps/gui/editors/editfamily.py
@@ -487,8 +487,7 @@ class EditFamily(EditPrimary):
         self.update_mother(mhandle)
 
     def reload_people(self):
-        self.father_label.rebuild()
-        self.mother_label.rebuild()
+        self._rebuild_parents_labels()
         fhandle = self.obj.get_father_handle()
         self.update_father(fhandle)
 
@@ -635,10 +634,26 @@ class EditFamily(EditPrimary):
         # TODO help button (rename glade button name)
         self.define_help_button(self.top.get_object('button119'),
                 WIKI_HELP_PAGE, WIKI_HELP_SEC)
+        self.reltype = self.top.get_object('marriage_type')
+        self.origin = self.data_type.get_val()
+        self.dbstate.db.connect(self.reltype, self._rebuild_parents_labels())
 
     def _can_be_replaced(self):
         pass
 
+    def _rebuild_parents_labels(self):
+        self.father_label.show()
+        self.mother_label.show()
+        trans_rel = [_("Civil Union"), _("Married"), _("Unknown"),
+                     _("Unmarried")]
+        trans_rel.sort()
+        trans_rel.append(_('Custom'))
+        # FamilyRelType(fam_key)
+        fam_key = trans_rel[self.reltype.get_active()]
+        if self.origin != fam_key: # only for custom for now
+            self.father_label.hide()
+            self.mother_label.hide()
+
     def _setup_fields(self):
 
         self.private = PrivacyButton(
```

Just cannot see where is a possible way for advanced monitoring on Family Editor, either for ChildEmbedList class or for checking on 'real time' (not on load only) any signals
generated by events (e.g., active gtk object, or changed value not set on database).

romjerome:
Sure, I could also try to use something like:
`self.reltype.connect("changed", ...)`
but to deal with iter, model, node, path & co
could be too much for a simple event signal, no?

romjerome:
arghh, just see that a **combo.get_active_text()** exists!!!
Not easily visible from https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/ComboBox.html#Gtk.ComboBox

romjerome:
Well, I will not say that I am happy with that, but it already sounds a little bit more logical (_for me_). Note, there is maybe something that we need to check on editors: migrate some 'GtkComboBox' to 'GtkComboBoxText'!
e,g,.```diff
diff --git a/gramps/gui/glade/editfamily.glade b/gramps/gui/glade/editfamily.glade
index 1f5f507..ce727d9 100644
--- a/gramps/gui/glade/editfamily.glade
+++ b/gramps/gui/glade/editfamily.glade
@@ -704,7 +704,7 @@
                       </packing>
                     </child>
                     <child>
-                      <object class="GtkComboBox" id="marriage_type">
+                      <object class="GtkComboBoxText" id="marriage_type">
                         <property name="visible">True</property>
                         <property name="can_focus">False</property>
                         <property name="tooltip_text" translatable="yes">The relationship type, eg 'Married' or 'Unmarried'. 
``` we are in a pure "textual" environment and it seems that latency on this widget could be improved by using GtkComboBoxText.

Anyway, I still need to monitor this somewhere else, but I am already a little bit more happy with:
```diff
index f3a6b4b..0db9dc5 100644
--- a/gramps/gui/editors/editfamily.py
+++ b/gramps/gui/editors/editfamily.py
@@ -487,8 +487,7 @@ class EditFamily(EditPrimary):
         self.update_mother(mhandle)
 
     def reload_people(self):
-        self.father_label.rebuild()
-        self.mother_label.rebuild()
+        self._rebuild_parents_labels()
         fhandle = self.obj.get_father_handle()
         self.update_father(fhandle)
 
@@ -635,10 +634,20 @@ class EditFamily(EditPrimary):
         # TODO help button (rename glade button name)
         self.define_help_button(self.top.get_object('button119'),
                 WIKI_HELP_PAGE, WIKI_HELP_SEC)
+        self.reltype = self.top.get_object('marriage_type')
+        self.origin = self.data_type.get_val()
+        self.reltype.connect('changed', self._rebuild_parents_labels)
 
     def _can_be_replaced(self):
         pass
 
+    def _rebuild_parents_labels(self, combo):
+        self.father_label.show()
+        self.mother_label.show()
+        if self.origin != combo.get_active_text():
+            self.father_label.hide()
+            self.mother_label.hide()
+
     def _setup_fields(self):
 
         self.private = PrivacyButton(
```
I will try a
`git rebase master`
then a push on my local git repo...

romjerome:
Sorry, too many cryptic conflicts.
I cannot properly rebase my local feature branch.

romjerome:
Get it working:
```diff
diff --git a/gramps/gui/editors/editfamily.py b/gramps/gui/editors/editfamily.py
index 0db9dc5..9f573c5 100644
--- a/gramps/gui/editors/editfamily.py
+++ b/gramps/gui/editors/editfamily.py
@@ -576,6 +576,10 @@ class EditFamily(EditPrimary):
         self.contexteventbox.drag_source_set_icon_name('gramps-family')
         self.contexteventbox.connect('drag_data_get', self.on_drag_data_get_family)
 
+        self.reltype = self.top.get_object('marriage_type')
+        self.origin = self.obj.get_relationship()
+        self.reltype.connect('changed', self._rebuild_parents_labels)
+
     def on_drag_data_get_family(self,widget, context, sel_data, info, time):
         if info == DdTargets.FAMILY_LINK.app_id:
             data = (DdTargets.FAMILY_LINK.drag_type, id(self), self.obj.get_handle(), 0)
@@ -634,9 +638,6 @@ class EditFamily(EditPrimary):
         # TODO help button (rename glade button name)
         self.define_help_button(self.top.get_object('button119'),
                 WIKI_HELP_PAGE, WIKI_HELP_SEC)
-        self.reltype = self.top.get_object('marriage_type')
-        self.origin = self.data_type.get_val()
-        self.reltype.connect('changed', self._rebuild_parents_labels)
 
     def _can_be_replaced(self):
         pass
```
Now, still need to monitor childref on Family and find a way for pushing without mixup
on index, maybe the most difficult issue for me!

Nick-Hall:
@romjerome I repaired the branch by doing a `git reset --hard` to the commit just before the merge.  You could either repair your local branch in a similar manner, or rename it and fetch the repaired version from your fork.

romjerome:
Well, I was close to use a dirty alternate solution:
`git diff 4e7eaf016a16c8ec2474febd1c8d1f32af1f02f7 > my_patch`
I do not know why, but somewhere the change around 'dummydb' remains.
Same for commits around tests for reports. They both are listed as conflicts
during my rebase.

OK, I will try a
`git reset --hard`
then maybe will make one commit with the generated 'my_patch' file.

Thanks

romjerome:
Maybe need a review and polish, but globals ideas have been tested.

Will look at len(childref) for generating (or not) parents during edition.
This issue should be more simplier than rebasing...

romjerome:
> Will look at len(childref) for generating (or not) parents during edition.

It seems I am trying to go too far...
In the 'real' world a couple becomes parents on the first child. e.g.,
```diff
diff --git a/gramps/gui/editors/editfamily.py b/gramps/gui/e
index 5a34c47..715ee73 100644
--- a/gramps/gui/editors/editfamily.py
+++ b/gramps/gui/editors/editfamily.py
@@ -206,6 +206,9 @@ class ChildEmbedList(EmbeddedList):
         self.call_edit_childref(ref)
 
     def child_ref_edited(self, person):
+        if len(self.family.get_child_ref_list()) == 1:
+            print("New parents, new to rebuild labels")
+            EditFamily(self.dbstate, self.uistate, self.tra
         self.rebuild()
 
     def share_button_clicked(self, obj=None):
```
Maybe it needs some major design changes on childref tab and Family Editor.
So, I will ignore this specific case, where people into Family Editor become parents
for the first time.

romjerome:
> As Chinese, Japanese, Arabic and Hebrew locales do not have a localized relationship handler, they are using the english strings (fallback)!
> Still need to check on the related translation files, but last version should already "force"
> uppercase for english and some locales.


`
$ git grep -A 1 "Civil Union" po/ar.po po/he.po po/ja.po po/zh*.po
po/ar.po:msgid "Civil Union"
po/ar.po-msgstr "زواج مثلي"
 
po/he.po:msgid "Civil Union"
po/he.po-msgstr "נישואים אזרחיים"

po/ja.po:msgid "Civil Union"
po/ja.po-msgstr "シビルユニオン"

po/zh_CN.po:msgid "Civil Union"
po/zh_CN.po-msgstr "民事结合"

po/zh_HK.po:msgid "Civil Union"
po/zh_HK.po-msgstr "民事結合"

po/zh_TW.po:msgid "Civil Union"
po/zh_TW.po-msgstr "民事結合"
`
so, need a proper solution for uppercase handling.

romjerome:
> or rename it and fetch the repaired version from your fork.

Maybe [all these additional commits](https://github.com/gramps-project/gramps/pull/347/commits) could be one commit, too?

Nick-Hall:
> Maybe all these additional commits could be one commit, too?

Yes.  At some point, it would be a good idea to use an interactive rebase to squash some of these commits.

romjerome:
> use an interactive rebase to squash some of these commits.

done.
 Thanks!

romjerome:
> labels are currently semi-automatic and only change on the next opening of the dialog

One solution could be to hide labels once status changed?
This is the way used, even incomplete (add/remove one parent will still keep "old" relation),
it shows a change.

Maybe need to write something for running this "in memory",
on the fly, without looking at database state. 
Currently these generated strings need an active family object.
I suppose this could be an additional function for the Family Editor?

romjerome:
> I created a JSON patch of the changes made in the editor.  When saving
> the object, I read the object again, apply the patch, and then write the
> object back to the database.  The serialize module contains to_json and
> from_json functions to convert an object to and from JSON.
> 
> See:  http://jsonpatch.com/
> 
> @Nick-Hall 

Oh, maybe this could help for using a nicer "data state monitor" for multiple changes on
Family Editor, either on childref or relationship types. Properly rebuild labels could be easier.
Great! Thanks!

Nick-Hall:
@romjerome I have pushed an experimental branch called [json-patch](https://github.com/Nick-Hall/gramps/tree/json-patch) to my fork.  A print statement lets you see the patch to be applied when "OK" is clicked.

romjerome:
@Nick-Hall 
Thanks!

I now see the [json-patch stuff](https://github.com/gramps-project/gramps/compare/master...Nick-Hall:json-patch#diff-a690f9f22ebc668bcfcbf41c7f9a2979). :+1: 

Just need to install '_python3-json-patch_' and maybe to look at [ChildRef](https://gramps-project.org/docs/gen/gen_lib.html#module-gramps.gen.lib.childref) and [Relationship Types](https://gramps-project.org/docs/gen/gen_lib.html#gramps.gen.lib.family.Family.set_relationship) for Family Edition.

Also, something might be difficult (_e.g., I think on customization around gedcom_) to understand when we use API and relationship types set on Family Object. The mapping on types seems to use a sorted order on Family Editor (drop down list) and localisation. So, related "GUI" integer can be different in english and e.g., in french! True, a specific GUI issue, but gedcom handling or any code using the related API do not really make a difference. Same for custom types added at the end of the list. Anyway, it does not block, just can be confusing into non-english environment.

Lowercase for "Mother/Father" labels in japanese, chinese, arabic and hebrew locales is also planned.

romjerome:
Maybe to also update the docstring about "civil union"?
> In present-day France, only civil marriage has legal validity. A religious ceremony may be performed after or before the civil union, but it has no legal effect. Since 1792, with the French Revolution, religious marriage ceremonies in France were made secondary to civil marriage.

Some countries like Argentina, Belgium, Russia or Turquey, seem to also have a different definition of "Civil Union"!

romjerome:
@Nick-Hall 
Is it not possible to use schema?
`print(self.obj.get_schema())`
Currently, change on relationship types is ignored. 
So, it needs an extra check.
```
if self.update_object() != None:
   ...
```

romjerome:
e.g.,
```
from gramps.gen.lib.serialize import to_json
import json
print(json.loads(to_json(self.obj))["type"]["string"])
```

Nick-Hall:
@romjerome I have updated my _json-patch_ branch to include the family editor.  Changing the relationship type in the editor will generate a patch similar to the following:

`[{"op": "replace", "path": "/type/string", "value": "Civil Union"}]`

I'm not sure why you want to use JSON.  To access the relationship type use `self.obj.get_realationship()` or `self.obj.type`.

romjerome:
@Nick-Hall I just thought that it was possible to monitor relationship type with something like a "has_data_changed". So, JSON could be a raw monitoring alternative. Currently, by "monitoring" changes on primary objects, we do not really get steps before the save on Family Object and it is only useful for debugging!

Made a refactor and added a gender symbol support on Family Editor on last revision.
It can light better where is my problem: by changing the relationship type we also need to rebuild labels. This should not be a problem if there is at least one child, as label will be either "Father" or "Mother". The problem might be for updating one parent or changing the relationship type before the validation.

I tested some workarounds, but I would like to use  one clear and common way for handling
labels and tooltips on Family Editor whatever family status in memory.

The last revision uses gender symbol as label and displays relation on tooltip on Family Editor, but something could be improved around Relationship type and Child tab.
The gender value is related to person, so less changes than with relation into the family.

Paul-Franklin:
Personally, I would prefer to review it based on a
patch attached to 9899, if that is possible.  Thanks.

romjerome:
@Paul-Franklin 
Thank you! 

Yes, it was the expected review: a new vision by someone who does not have the focus on something specific. You are also right, maybe I should keep one function for both (gramps.gen and reports), the "xlocale" argument might be the glocale by default.

True, always having right label via either gender symbol or string is a challenge by looking at some specific cases. The last revision should be more "neutral" (more simple and common) on Family Editor.

Paul-Franklin:
Yes, well, remember that what I really would like is to have
the family editor have two combo widgets for the two people,
primed with whatever you want but with other choices available.

romjerome:
As I did not get something ready and complete for Family Editor, I quickly refactored the two functions as asked by @Paul-Franklin and @sam-m888. Sure, it makes it more clean.

"Moderation" is currently used on Family Editor.

> two combo widgets for the two people

@Paul-Franklin 
Like one Person selector without gender filters?
This will remove one feature, no?

> but with other choices available.

Choices?
Do you mean some config keys, like on the "large" patch for plants, animals & co?

Paul-Franklin:
What I want is to have a combo box on the left, next to the two icons.
Initialize it with "Father" or "Man" or "Male" or "Husband" or anything
else you want, as long as the other choices I mentioned are also there,
and also the corresponding choices for the other gender ("Mother" and
"Woman" and "Female" and "Wife"), but also the neutral ones (maybe
"Partner1" and "Partner2"? I don't know).  And ideally the user should
be able to type in their own "title" (although that won't work in reports
and other places in gramps, maybe).  Then there should be another
combo box by the other two icons, with the same set of identical choices
only it will be initialized with the corresponding one for the other sex.

Those two combo boxes will be what sets the "title" of the two people.
Even though I would guess that 99.9% of our users will leave them
alone (set to "Father" and "Mother" or "Male" and "Female" or
whatever you decide their initial settings should be).

Then there'd be some sort of internal table, saying that if
"Father" was chosen the gramps gender should be set to MALE.
And the sex-specific selector widget would be brought up, even
if the two people were the same sex, if the user chose to pick
an existing person.  Or the person editor seeded with the sex
based on the name.

That's what I meant by "choices" in the combo boxes.

Thanks.

romjerome:
> user should be able to type in their own "title" (although that won't work in reports
and other places in gramps, maybe)

Ah yes, this will be no more dynamic labels...

This is already possible when we set the family.type (family relationship).
`two people + family.type + relationship module = 
   pair of hardcoded (and translated) strings for parent`

Nick-Hall:
Rebased.

gramps/gen/utils/parents.py
```diff
@@ -0,0 +1,88 @@
...
+def parents(db, family, xlocale):
+    """
+    Get the label for parent
+    """
+    if family.get_father_handle(): #[0]
+        father = db.get_person_from_handle(family.get_father_handle())
+        if family.get_mother_handle():
+            mother = db.get_person_from_handle(family.get_mother_handle())
+        else:
+            if father.gender == Person.FEMALE:
```
Nick-Hall: We need to the the case where the father is of unknown gender.

gramps/gen/utils/parents.py
```diff
@@ -0,0 +1,88 @@
...
+                return [_T_("Woman"), _T_("Person")]
+            else:
+                return [_T_("Man"), _T_("Person")]
+    else:
+        if family.get_mother_handle(): #[1]
+            mother = db.get_person_from_handle(family.get_mother_handle())
+            if mother.gender == Person.MALE:
```
Nick-Hall: Again we should add the cases where the mother is female or of unknown gender.

gramps/gen/utils/parents.py
```diff
@@ -0,0 +1,88 @@
...
+                return [_T_("Person"), _T_("Man")]
+        else:
+            return [_T_("Person"), _T_("Person")]
+
+    # need it for above strings and as fallback
+    rel_father = _("Man")
+    rel_mother = _("Woman")
+
+    if len(family.get_child_ref_list()) > 0:
```
Nick-Hall: We only get here if the family has two parents.  Do we want to describe a single parent as "Father" or "Mother"?

gramps/gen/utils/parents.py
```diff
@@ -0,0 +1,88 @@
...
+        if father.gender == Person.FEMALE:
+            rel_father = rel_mother
+        if mother.gender == Person.MALE:
+            rel_mother = _T_("Father") # if both parents switched
+    else:
+        if family.get_father_handle() and family.get_mother_handle():
+            rc = get_relationship_calculator(True, xlocale)
+            rel_father = rc.get_one_relationship(db, mother, father, olocale=xlocale)
+            rel_mother = rc.get_one_relationship(db, father, mother, olocale=xlocale)
+        if xlocale.lang[0:2] == "fr":
```
Nick-Hall: Why do we need language specific cases here?

gramps/gui/editors/editfamily.py
```diff
@@ -486,11 +496,61 @@ def show_buttons(self):
...
+        """
         fhandle = self.obj.get_father_handle()
         self.update_father(fhandle)
+        log.debug('father label: %s' % self.father_label.get_label())
 
         mhandle = self.obj.get_mother_handle()
         self.update_mother(mhandle)
+        log.debug('mother label: %s' % self.mother_label.get_label())
+
+        if fhandle and mhandle:
```
Nick-Hall: Some of this code would be better in the `parents` utility function.

sam-m888:
sorry hit the wrong button!

gramps/gen/utils/parents.py
romjerome: Ok, something like:
`if father.gender == Person.UNKNOWN
    return [_T_("Unknown"), _T_("Person")]`

gramps/gen/utils/parents.py
romjerome: same for mother:
`if mother.gender == Person.UNKNOWN:`

    `return [_T_("Unknown"), _T_("Person")]`

romjerome:
As [gedcom 5.5.5](https://www.gedcom.org/) now supports "same-sex relation",
we do no need anymore to keep a male as FATHER
on family for gedcom 5.5.1 support!

This should make this feature request more simple?
Either remove the labels for parents (we do not need to
force using "father" on the left and "mother" on the right
for the Family Editor), or move to a common parent section
like started here?

romjerome:
I cannot [look at all specifications for gedcom 5.5.5](https://gramps-project.org/bugs/view.php?id=11389). Just supposing that [updating/upgrading/dumping gedcom export from 5.5.1 to 5.5.5](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/export/exportgedcom.py#L345) will cover the [HUSB/WIFE case on FAMily](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/export/exportgedcom.py#L836). So, even with unmarried people (with or without child), adults in the FAMily will be exported whatever gender or status as HUSB or WIFE to gedcom 5.5.5.

romjerome:
default

Select_A

select_B

partners

switch

romjerome:
```
        fhandle = self.obj.get_father_handle()
        self.update_father(fhandle)

        mhandle = self.obj.get_mother_handle()
        self.update_mother(mhandle)

        self.phandles = [mhandle, fhandle]
```

Maybe variable names could be improved?
Currently we can easily make a confusion ...
Does 'mhandle' mean 'male handle' or 'mother handle'?
Does 'fhandle' mean 'female handle' or 'father handle'?

PR #348, opened 2017-02-02
bug 9927; fix UndoableEntry for pygobject bug

prculley:
This is a workaround for https://bugzilla.gnome.org/show_bug.cgi?id=644927
What is happening, is that our UndoableEntry class is connecting to the 'insert-text' event, which uses one of the 'in-out' parameters for the position of the cursor in the buffer. The very fact of the connect to the event causes the issue, not something Gramps is doing wrong.

A workaround is to not hook the event, but to replace the 'insert_text' method that gets called internal to Gtk.Editable.

I should note that this is a nuisance bug, Gramps works correctly even with the bug in place.  But the fix eliminates the "Warning: g_value_get_int: assertion 'G_VALUE_HOLDS_INT (value)' failed" console message.

Tested manually, the undo (ctrl-z) and redo (ctrl-shift-z) still seem to work properly...

### DIFF
--- a/gramps/gui/widgets/undoableentry.py
+++ b/gramps/gui/widgets/undoableentry.py
@@ -51,7 +51,7 @@ from .undoablebuffer import Stack
 
 class UndoableInsertEntry:
     """something that has been inserted into our Gtk.editable"""
-    def __init__(self, text, length, position, editable):
+    def __init__(self, text, length, position):
         self.offset = position
         self.text = text
         #unicode char can have length > 1 as it points in the buffer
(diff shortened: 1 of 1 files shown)


sam-m888:
Worked for me.

Had not tried the, the undo (ctrl-z) and redo (ctrl-shift-z) combination on the field before which without the patch gives the following error:

`/usr/lib/python3/dist-packages/gi/overrides/Gtk.py:196: Warning: g_value_get_int: assertion 'G_VALUE_HOLDS_INT (value)' failed return super(Editable, self).insert_text(text, -1, position)
`

sam-m888:
Thank you

bpisoj:
Please see bug ticket 9960: Error Function edit persons window
your patch break with use of non-ascii characters

PR #349, opened 2017-02-04
Add Progress indicators to filters and reports

prculley:
It started with testing the user progress indicator...

Add progress indication for filters with long 'prepare' times (these had scans of the entire db for persons in the 'prepare' stage).

And fix DeepRelationshipPathBetween for more accurate progress; the original code called step many more times than the 'total' progress count, which made for an apparent freeze.  The fix makes indication more accurate, can still be off a bit if the family tree has a lot of non-relatives (filter ends before progress get to 100%), and can be off the other way if the family tree has cycles.

And fix IsSiblingOfFilterMatch for HandleError, just found this during testing of progress, too lazy to make a separate commit.

Fix TreeBaseModel so that filters have correct transient parent.

Add progress indication to long running reports.


sam-m888:
Thank you

PR #350, opened 2017-02-06
Fix 'Extract Info from names' tool

prculley:
It left an unclosed window, if you cancelled instead of run.
It had a bad transient parent issue, it ignored the ManagedWindow setting because the glade file made it visible too soon.
Fixed Gtk deprecation warnings
Added window size/position config.
Also did a pylint, in separate commit.

-----------
https://gramps-project.org/wiki/index.php?title=Gramps_4.2_Wiki_Manual_-_Tools#Extract_Information_from_Names

### DIFF
--- a/gramps/plugins/tool/patchnames.glade
+++ b/gramps/plugins/tool/patchnames.glade
@@ -3,10 +3,8 @@
 <interface>
   <requires lib="gtk+" version="3.10"/>
   <object class="GtkDialog" id="patchnames">
-    <property name="visible">True</property>
+    <property name="visible">False</property>
     <property name="can_focus">False</property>
-    <property name="default_width">500</property>
-    <property name="default_height">450</property>
     <property name="type_hint">dialog</property>
     <signal name="delete-event" handler="on_delete_event" swapped="no"/>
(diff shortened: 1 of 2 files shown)


Paul-Franklin:
I wonder if you want to get rid of the show_all() also?

Paul-Franklin:
And maybe the default sizes in the Glade file as well
as the resize(700,100) ?

Paul-Franklin:
Or perhaps I am getting confused, with one dialog vs the second.

prculley:
@Nick-Hall The 'show_all' and 'resize(700,100)' are both necessary, as they are for the manually created initial dialog.  I did realize that the resize should have been before the show_all otherwise the dialog would end up not centered on the parent.
I did remove the extraneous size in the glade file, as that was covered by the config option.

Paul-Franklin:
I think what confused me is the fact that both dialogs
are getting a set_window call, but once I'd read as far
as the "display" method it became obvious.  Sorry.

sam-m888:
Worked for me on Example.gramps and my personal Family Tree.

Thank you

PR #352, opened 2017-02-08
Sl gramps42

beernarrd:
Corrected issue with display name editor with colliding substring of given and surname in translated keywords. Seems fine now.
Small other updates.


sam-m888:
Thank you

PR #355, opened 2017-02-12
fix bug 9600

beernarrd:
Typo corrected now also in master.


Paul-Franklin:
@Nick-Hall I'd like to suggest Bernard be added to the
translators' list, if he wants to be, so that he doesn't have
to file a p.r. for a one-line typo fix like this, etc.  Thanks.

Nick-Hall:
@Paul-Franklin Good idea.  I have sent him an invitation.

beernarrd:
Thanks both. That makes the work flow easier. Commit pushed directly.

PR #357, opened 2017-02-21
bug 9954; fix Gedcom import for empty FAMC line

prculley:
An empty FAMC xref is a violation of the Gedcom standard.  One of many created by various source software...
That said, we can deal with it a bit better.  I added a patch to detect the empty xref and add the error to the error list and error notes.

Assuming that there was no valid FAMC with an XREF on the same person, Gramps will generate correct family and person references; this is detected in a post processing step of the Gedcom importer. 

Gramps also puts a message to this effect in the Error list shown at the end of import, although this particular error type is NOT added as a note.

During debugging and validation of this I found an additional issue; the second commit deals with this.
Gedcom import has improper creation of error message notes.  The GIDs were not being checked against others created by the Gedcom importer (the importer assigns GIDs when a note reference is first found, but  delays actual note creation until note text is found in the Gedcom file).

The improper note could overwrite another note, due to overlapping GIDs; an eventual result being that we ended up with a corrupt db with a handle reference error.

### DIFF
--- a/gramps/plugins/lib/libgedcom.py
+++ b/gramps/plugins/lib/libgedcom.py
@@ -4555,6 +4555,9 @@ class GedcomParser(UpdateCallback):
         @type state: CurrentState
         """
 
+        if not line.data:  # handles empty FAMC line
+            self.__not_recognized(line, state)
+            return
         sub_state = CurrentState()
         sub_state.person = state.person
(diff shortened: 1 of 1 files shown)


sam-m888:
Thank you

PR #358, opened 2017-02-21
bug 9920; fix for multiline Titles in Source view

prculley:
Source view is set up for single line rows, multi-line Titles cause problems for Gtk (sometimes many unrelated rows get doubled in height).

Some dbs, originating in Gedcom, have Source titles with one or more line breaks.  This patch fixes the Source view ONLY, so reports, exports etc. still get the original data format.
Despite suggestions to get sophisticated with replacing '\n', I decided to keep it simple and go with ' ' (single space).  This is the same as Gtk itself does for our Source Editor Title.

### DIFF
--- a/gramps/gui/views/treemodels/sourcemodel.py
+++ b/gramps/gui/views/treemodels/sourcemodel.py
@@ -99,7 +99,7 @@ class SourceModel(FlatBaseModel):
         return len(self.fmap)+1
 
     def column_title(self,data):
-        return data[2]
+        return data[2].replace('\n', ' ')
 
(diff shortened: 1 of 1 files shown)


Nick-Hall:
The problem is caused because we use a TreeView with fixed height mode in the [ListView](https://github.com/gramps-project/gramps/blob/master/gramps/gui/views/listview.py#L154) class.

An alternative would be to only display the first line with `data[2].split('\n')[0]`, but I am happy with your solution.

sam-m888:
Thank you

PR #359, opened 2017-02-25
Fix webstuff when user css does not exist

enlite93:
As mentioned in https://github.com/gramps-project/gramps/commit/74ffc76af0d3be2463a8ef918a82a8e1df4d6126#commitcomment-21026601
```

  File "/home/travis/build/gramps-project/gramps/gramps/gen/plug/_manager.py", line 181, in reg_plugins
    results = mod.load_on_reg(dbstate, uistate, plugin)
  File "/home/travis/build/gramps-project/gramps/gramps/plugins/webstuff/webstuff.py", line 181, in load_on_reg
    list_files = os.listdir(USER_CSS)
FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/.gramps/css'
```

### DIFF
--- a/gramps/plugins/webstuff/webstuff.py
+++ b/gramps/plugins/webstuff/webstuff.py
@@ -178,14 +178,14 @@ def load_on_reg(dbstate, uistate, plugin):
          path_img_48x48("gramps-geo-mainmap.png"), None, [], [] ],
         ]
     # If we add css user files, we must restart gramps to use them.
-    try:
+    if os.path.exists(USER_CSS):
         list_files = os.listdir(USER_CSS)
         for cssfile in list_files:
             CSS_FILES.append([cssfile, 1, cssfile.replace('.css', ''),
                               os.path.join(USER_CSS,cssfile),
                               None, [], [] ])
         return CSS_FILES
-    except FileNotFoundError:
+    else:
         return CSS_FILES
 
 def process_list(data):
--- a/gramps/plugins/webstuff/webstuff.py
+++ b/gramps/plugins/webstuff/webstuff.py
@@ -185,7 +185,7 @@ def load_on_reg(dbstate, uistate, plugin):
                               os.path.join(USER_CSS,cssfile),
                               None, [], [] ])
         return CSS_FILES
-    except:
+    except FileNotFoundError:
         return CSS_FILES
 
 def process_list(data):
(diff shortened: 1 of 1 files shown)


gramps/plugins/webstuff/webstuff.py
```diff
@@ -178,12 +178,15 @@ def load_on_reg(dbstate, uistate, plugin):
...
-                          None, [], [] ])
-    return CSS_FILES
+    try:
+        list_files = os.listdir(USER_CSS)
+        for cssfile in list_files:
+            CSS_FILES.append([cssfile, 1, cssfile.replace('.css', ''),
+                              os.path.join(USER_CSS,cssfile),
+                              None, [], [] ])
+        return CSS_FILES
+    except:
```
QuLogic: Better not to use a bare `except`.

enlite93: What do you suggest?

QuLogic: Well, you should catch only what you expect to see; `except FileNotFoundError:` or `except OSError:` or something like that.

sam-m888: I'd suggest you just use

`except FileNotFoundError:`

sam-m888:
Worked for me. 👍

Nick-Hall:
Is this only a problem for the unit tests?  If so, why don't we just create the directory in the Travis environment?

daleathan:
> Is this only a problem for the unit tests? If so, why don't we just create the directory in the Travis environment?

@Nick-Hall You already do that on Travis-CI when testing the addons-source https://github.com/gramps-project/addons-source/blob/master/.travis.yml#L80 but not when testing Gramps source.  But really from my point of view the problem is that when I run the test locally I get the same error and in effect your say that I now have to remember to create that css directory before the test are run isn't that a little backward. Another benefit of adding this is that you can then remove that line from Travis-CI for the addons-sources.

Nick-Hall:
@daleathan OK. I see the problem. Why not just add an `if` statement?

`if os.path.exists(USER_CSS):`

prculley:
Regarding the Travis failure; I had to (painfully) debug an issue just like this a while back. It occurs when the order of Travis tests changes such that a test is run that needs plugins, but the test does not include the creation of user directories or is not preceded by tests with code that does this.

I fixed it at the time for tests that use the test_util 'run()' method. Perhaps the failing test uses some other method to operate the Gramps code. See https://github.com/gramps-project/gramps/pull/336/files, the changes for the test_util file.

enlite93:
> Why not just add an if statement? 
> 
> `if os.path.exists(USER_CSS):`
> 

@Nick-Hall Ok, so instead of try/except I'll use an if statement 👍

Nick-Hall:
Fixed in commit c352fa63783ee6ceae689d4ce0c6ad67e2aa88bc.

enlite93:
Was it something I did? Sure my solution wasn't exactly like yours!

Either way errors gone, so thanks I guess?

PR #360, opened 2017-02-26
Fix unodoableentry for international characters

prculley:
Gtk has some parameters in bytes, some in characters; they cannot
be combined.


Nick-Hall:
PR #222 had a similar issue.  See my [comment](https://github.com/gramps-project/gramps/pull/222#issuecomment-262861239).

PR #361, opened 2017-02-28
 Fix Addon manager to recognize newly installed addons immediately

prculley:
During testing of an improved Addon manager, I discovered that the original manager had been changed such that it would no longer recognize any newly installed addons immediately.  They were properly recognized on restart.

The issue was caused by myself, through two inadvertent changes.  A change in the UpdateAddons class to make it a ManagedWindow (which prevented a re-scan of plugins), and the performance improvements for the plugin manager (which by design stopped multiple scans of plugins).

Rather than give up the performance improvements to plugin operations, I decided to make the re-scan explicit.

I discovered that even after re-enabling a rescan, newly installed import, export and docgen plugins were not getting recognized.  This appears to be an old bug.

Then I discovered that updated addons were not recognized.  This appears to be an old bug.

The changes here get all this functionality working.

I do note that updated Views and Gramplets (those that are actively running) are still not being recognized.  I think that is because the original code, while in use, is not getting removed/replaced by Python, even after asking for a reload.  Probably reasonable in that I can imagine all sorts of issues that could occur...

### DIFF
--- a/gramps/cli/grampscli.py
+++ b/gramps/cli/grampscli.py
@@ -337,6 +337,8 @@ class CLIManager:
         """
         self._pmgr.reg_plugins(PLUGINS_DIR, dbstate, uistate, rescan=rescan)
         self._pmgr.reg_plugins(USER_PLUGINS, dbstate, uistate, load_on_reg=True)
+        if rescan:  # supports updated plugin installs
+            self._pmgr.reload_plugins()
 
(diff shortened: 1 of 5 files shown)


sam-m888:
Thank you

sam-m888:
Related to https://github.com/gramps-project/gramps/pull/336

PR #363, opened 2017-03-02
Fix Merge test for complex families

prculley:
Trying to get the last four of these enabled.  On my system, only one actually failed, so I fixed it; I enabled all four.  See what Travis makes of this.

Hmmm; so it fails under Travis but passes on my system.  Looks like Travis (probably nose) is intercepting the assertion error instead of letting it fall through to the stderr like the test expects.  Since I can only run standard unittest on my system, the different test env may be causing the issue.

I my suspicion is correct, we may have to rethink the use of 'assert' for user errors like merging a child with a parent.

Looks like we deal with this fine in GUI, only an issue for CLI.

Trying to fix it in CliMerge addon...

### DIFF
--- a/gramps/gen/merge/test/merge_ref_test.py
+++ b/gramps/gen/merge/test/merge_ref_test.py
@@ -33,10 +33,10 @@ import lxml.etree as ET
 from gramps.plugins.lib.libgrampsxml import GRAMPS_XML_VERSION
 from gramps.test.test_util import Gramps
 from gramps.cli.user import User
-from ...const import DATA_DIR, USER_PLUGINS, TEMP_DIR
-from ....version import VERSION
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Rebased.

prculley:
Ok I now have success.  But please be careful and don't just merge this PR, I have a testing commit that changes the Travis script that we don't want in master.  I think you should cherry-pick the https://github.com/gramps-project/gramps/pull/363/commits/b865556ddd5349d480eb7bb55ca5f7d838ba2db7 commit instead.

I had to add this because I don't have rights to push to addons to update the tgz and listings files.
If one of you gentlemen will 'make' the necessary 'addons' files and push them I would appreciate it.  The addons-source is updated.

PR #365, opened 2017-03-04
Use assertRaises to check for the NameError

enlite93:
Use [assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) to check for the NameError and only output a test passed dot during unitest. eg removes the following output:
```

Warning: TestSignals: Exception occurred in callback function.
  File "/home/enlite93/gramps/gramps/gen/utils/callback.py", line 406, in emit
    fn(*args)
  File "/home/enlite93/gramps/gramps/gen/utils/test/callback_test.py", line 72, in borked
    rubish.append(i)
NameError: name 'rubish' is not defined

```

Also corrected spelling of a word.

### DIFF
--- a/gramps/gen/utils/test/callback_test.py
+++ b/gramps/gen/utils/test/callback_test.py
@@ -65,11 +65,10 @@ class TestCallback(unittest.TestCase):
 
         def borked(i):
             """
-            this intentionally raises a NameError exception
-            FIXME: could use an explanation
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Thanks.  It's good to get rid of some of the extraneous output.

enlite93:
👍 

Yes, ideally only _test passed dot_'s should be shown.

PR #367, opened 2017-03-24
Update Catalan translation to latest template

arnaullv:
Update Catalan translation to latest template and unified translation criterias


sam-m888:
Thank you

PR #369, opened 2017-03-30
bug 9998 partial fix for Note editor Toolbar overflow icons missing

prculley:
Bug Affects Windows only.
Adds text labels to icons which will only appear in the overflow
dropdown menu when dialog is too narrow.  Icons still don't show
due to Gtk bug on Windows, but at least user sees the text labels.

### DIFF
--- a/gramps/gui/widgets/styledtexteditor.py
+++ b/gramps/gui/widgets/styledtexteditor.py
@@ -81,10 +81,10 @@ FORMAT_TOOLBAR = '''
   <toolitem action="%d"/>
   <toolitem action="%d"/>
   <toolitem action="%d"/>
+  <toolitem action="%d"/>
+  <toolitem action="%d"/>
   <toolitem action="Undo"/>
   <toolitem action="Redo"/>
-  <toolitem action="%d"/>
-  <toolitem action="%d"/>
   <toolitem action="%d"/>
   <toolitem action="%d"/>
   <toolitem action="%d"/>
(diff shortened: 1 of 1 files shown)


eno93:
Tested and works for me on Win 7 (Parallels for Mac) with Gramps 4.2.5 AIO 64bit 👍 
newnote-textdropdown

Would it be possible to move those two boxes to the left? Or increase the minimum size of the **new note** window?
newnote-hiddenboxes-text

prculley:
@sam-m888 as requested...
I moved Undo/Redo icons to the right of the drop downs.  I think this keeps a minimum width dialog from losing the drop down.

eno93:
> @sam-m888 as requested...
> I moved Undo/Redo icons to the right of the drop downs. I think this keeps a minimum width dialog from losing the drop down.

Actually I requested it, anyway works well 👍 
new note fixed bu paul

sam-m888:
Tested on Ubuntu 16.04 and can see no issues
ubuntu-16 04 1-noteeditor-minimised

Note that the **transient parent** messages:
> (xgrampspy3) [grampsdev@lfs101x-VirtualBox gramps (prculley-bug9998)]$ ./Gramps.py 
> Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.
> Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.

are to be handled by: https://github.com/gramps-project/gramps/pull/370

sam-m888:
Thank you

PR #370, opened 2017-03-31
Bug 8128; for color chooser dialogs; fixed transient parent

prculley:
for styledtexteditor and Family Lines Graph, Family Colors option


sam-m888:
Confirming  color chooser dialogs in both:
* styledtexteditor 
* Family Lines Graph > Family Colors option

no longer emit the message in question

> Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.

sam-m888:
Thank you

PR #372, opened 2017-04-07
bug 8128; Fix 'print...' dialogs for missing transient parent

prculley:
Requires adding uistate to the docgen basedoc and all files that
override the __init__


sam-m888:
Thank you

PR #374, opened 2017-04-19
add the date-format option to narrativeweb and webcal.

SNoiraud:
This pull request cover the feature request 3452.

mail received from paul : 
...
I didn't add it to either webcal or the narrative web,
partly because I don't know if that option makes sense
for those reports, but mainly because I think of you as
the expert on those reports.
...


sam-m888:
Tested and worked on both reports.

narweb-date-option-grampsmaster
narweb-date-option-grampsmaster-example-output
webcal-date-option-grampsmaster

sam-m888:
Thank you

PR #377, opened 2017-04-26
bug 9564; fix exception on Event filter editor

prculley:
for 'Events occurring on a particular day of the week' filter


sam-m888:
Thank you

PR #379, opened 2017-04-26
bug 9445; fix XML export of notes so lead/trail spaces are preserved

prculley:
XML export does not currently preserve leading or trailing spaces in notes.  XML import does.  This patch corrects this situation.


sam-m888:
Thank you

PR #380, opened 2017-04-27
bug 6130; fix export filtering

prculley:
Under some conditions, the handle passed in for comparison to the various filter lists (plist, nlist etc.) is a 'bytes', whereas the lists always seem to be 'str' type.  The "if handle in self.plist" fails; that is, handles that are present in the list are not found because they are the wrong type.

There has been a lot of history since the 6130 bug was filed on Gramps version 3.4.1.  I tested the current 4.2 branch and everything seems to work correctly with no patches; the original 6130 bug does not seem to be present.  But I found this bug when testing master branch.

I'm not sure if this is the -best- way to fix this, but it does seem to work.  As I ranted before, it would be nice if we could somehow make all handle related APIs the same type, not sometimes 'bytes' and sometimes 'str'.  But things seem to have been this way since the transition to Python3.

### DIFF
--- a/gramps/gen/proxy/filter.py
+++ b/gramps/gen/proxy/filter.py
@@ -78,6 +78,8 @@ class FilterProxyDb(ProxyDbBase):
         Finds a Person in the database from the passed Gramps ID.
         If no such Person exists, None is returned.
         """
+        if isinstance(handle, bytes):
+            handle = str(handle, 'utf-8')
         if handle in self.plist:
             person = self.db.get_person_from_handle(handle)
             if person is None:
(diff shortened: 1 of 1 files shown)


Paul-Franklin:
If I had a Euro for every instance of things like that I've seen
in gramps, all over, I could buy a nice dinner for quite a few folks.

I see nothing wrong with adding more of them.  I'd say go for it.

Nick-Hall:
> I'm not sure if this is the -best- way to fix this, but it does seem to work. As I ranted before, it would be nice if we could somehow make all handle related APIs the same type, not sometimes 'bytes' and sometimes 'str'. But things seem to have been this way since the transition to Python3.

Yes.  I totally agree.  Having a handle that is sometimes `str` and sometimes `bytes` is asking for bugs.

sam-m888:
Committing as requested as per comment on issue 6130

>  [With two reviewers saying "yes" for p.r. 380, I would imagine it could be committed? ](https://gramps-project.org/bugs/view.php?id=6130#c51743)

Thank you

PR #382, opened 2017-05-01
FR10020 Add a filter "within <n> km/miles/degree" of a given place

SNoiraud:


SNoiraud:
Need to solve the unittest problem.

Nick-Hall:
@SNoiraud If "Find" is clicked when no data has been entered into the sidebar filter, I get an error.

Nick-Hall:
Rebased.

PR #383, opened 2017-05-02
ReorderIDs II

apoettker:
The second try to extend the ReorderIDs functionality. Due some spurious merge conflict's I've close the original PR and opened another one.

Several issues in the bug tracking system (#4101 #7120, #9871) claims the functionality of the ReorderID tool. Previously complete automatic I've extended to an interactive tool with different options. Goal is creating a tool to fulfil the different remarks and requests.

There are nine rows contains the primary objects (Person, Family, Event, Place, Source, Citation, Repository, Media, Note) with a Object button to activate / deactivate. The Actual counter and the Quantity can't be changed. The format out of the global settings is editable. As before the well known logic will be Proceed when entering the button. After hitting the Change button you can determine, which Start and Step sequence will be applied while renumbering the data. The Keep button protects everything with different number format model inside the primary objects. Deactivating this barrier will reorder all IDs as predetermined.

Additional hidden buttons - to streamline the design - in Object, Format, Change, Start, Step and Keep column assist the researcher with logical suggestions.

screenshot reorderid 170320

### DIFF
--- a/gramps/plugins/tool/reorderids.glade
+++ b/gramps/plugins/tool/reorderids.glade
@@ -1,0 +1,1596 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.16.1 -->
+<interface>
+  <requires lib="gtk+" version="3.10"/>
+  <object class="GtkDialog" id="reorder-ids">
+    <property name="can_focus">False</property>
+    <property name="resizable">False</property>
+    <property name="modal">True</property>
+    <property name="window_position">center-on-parent</property>
+    <property name="hide_titlebar_when_maximized">True</property>
+    <property name="type_hint">dialog</property>
+    <property name="has_resize_grip">False</property>
+    <child internal-child="vbox">
+      <object class="GtkBox" id="base-box">
+        <property name="visible">True</property>
+        <property name="can_focus">False</property>
+        <property name="orientation">vertical</property>
+        <child>
+          <object class="GtkLabel" id="title">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="margin_top">10</property>
+            <property name="margin_bottom">5</property>
+            <attributes>
+              <attribute name="weight" value="bold"/>
+            </attributes>
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">True</property>
+            <property name="position">0</property>
+          </packing>
+        </child>
+        <child>
+          <object class="GtkGrid" id="selection_table">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="halign">end</property>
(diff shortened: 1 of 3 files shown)


Nick-Hall:
This should be a BatchTool which will automatically bring up the undo history warning in a separate dialog.

apoettker:
@Nick-Hall Starting the extended Reorder module below window pops up with the same message as before, but more line feeds (to fulfil the 600x800 directive).

screenshot reorderid 170510

If you pressing "Proceed" module works with absolute same behaviour as before. After activation "Detailed Selection" the warning message folds in and you've more options (screen shot above) to adjust the reorder process. 

Typically a batch tool doesn't offer options, but that the reason why I extend the Reorder ID module. So its a little bit unclear for me what can be improved. Just give me a hint and I'll code it in.

Nick-Hall:
Have a look at the "Merge Citations" and "Sort Events" tools as examples of batch tools with options.  They present the warning and options in two separate dialogs.

apoettker:
ReorderIDs module uploaded again. After nomal undo history warning in a separate dialog

screenshot reorderid 170512a

we see a modified window with above described options to adjust reorder process:

screenshot reorderid 170512b

Hope this is what expected and can just even hush into Gramps 5.0.

Nick-Hall:
@apoettker Good work! I really like this.

Only one very minor point.  Should the "Actual" and "Quantity" columns contain Label instead of Entry widgets?

We can certainly get this into v5.0.

apoettker:
@Nick-Hall Both "Actual" and "Quantity" column heads are "GtkLabel" type. For optical reasons (column head line should look homogeneous) I draw the frames from "Object", "Format", "Change", "Start", "Step" and "Keep" off. My first approach was giving "Actual" and "Quantity" a frame but I struggled designing a label with an frame looks as elegant as a "GtkButton".

Nick-Hall:
@apoettker I was referring to the Entry boxes, not the headings.  For example, "I00988" is read-only, but uses a Gtk.Entry rather than a Gtk.Label.

apoettker:
@Nick-Hall The same concept keepingthe look somewhere consistent and homogeneous. In the beginning I've had some "GtkLabels" for the "Actual" and "Quantity" columns. IMHO the entries looking a little bit lost in this environment. Its a little bit confusing that this entries are not editable, but within the context the majority of our user will realize very quickly why this item could not changed.

Nick-Hall:
@apoettker Yes, it has a good look and feel at the moment.  I'll go ahead and merge it.

Thanks for your contribution.

prculley:
@apoettker This GUI is not very obvious in what it can do.  Until I looked at the code I did not even realize that the top row of labels were (in some cases) buttons.  I looked at the Help button site; it looks like it needs to be updated to adequately describe all the new features.

PR #384, opened 2017-05-03
Bug 9738; Fix People selector for speed; option of Tree or List view

prculley:
This PR is another potential fix for https://gramps-project.org/bugs/view.php?id=9738 and https://gramps-project.org/bugs/view.php?id=9880 (https://github.com/gramps-project/gramps/pull/337 also tries to fix this another way, but it is not complete and the author is not responding).  See also the email thread on this topic at http://gramps.1791082.n4.nabble.com/Unable-to-collapse-places-and-people-td4679691.html

Prior to v4.2.4 Person selector (used to select person for family or relationship view etc.) brought up the dialog with a collapsed tree mode.
bug8785 was submitted requesting that selectors were opened with a fully expanded tree view, this was implemented for v4.2.4.  Since that time another patch changed the behavior for place selectors, this PR should not effect this.

Since that time it has become evident that with large trees that the fully expanded tree view has (rather severe) performance problems with large trees.

This PR adds an edit/preference option on the 'Display' tab
   'When choosing from existing people, use List view (Tree view if off)' default off.
It also has a tooltip 'Tree view with large trees will slow down the dialog.'
It also reverts the baseselector to the previous (before 4.2.4) state of collapsed tree as the default.
I added back the code removed in https://github.com/gramps-project/gramps/commit/10629b16b8c3945efb68cd763691b2bef4088180 which allowed a tree to expand on a specific person, although I was unable to find anywhere that used this feature.  Perhaps it was used in an addon.

I like the List view instead of expanded Tree; on my machine with 16000 persons, List view opened in 0.9 sec, the collapsed tree opened in 2.7 sec, the current code expanded tree took 8.5 sec.

If a user never changes the preference, he gets the behavior like before 4.2.4.  If he does change the new preference, his person selectors open in list view, which, in my opinion, are faster, and use less screen real estate than the expanded tree view.

If this is accepted, I will prepare a similar patch for the 4.2 branch...

### DIFF
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -187,6 +187,7 @@ register('geography.use-keypad', True)
 # needed), for instance to four 'interface.clipboard' variables --
 # so do a recursive grep for "setup_configs" to see all the (base) names
 register('interface.dont-ask', False)
+register('interface.people-selector-list', False)
 register('interface.view-categories',
          ["Dashboard", "People", "Relationships", "Families",
           "Ancestry", "Events", "Places", "Geography", "Sources",
--- a/gramps/gui/configure.py
+++ b/gramps/gui/configure.py
@@ -1108,6 +1108,14 @@ class GrampsPreferences(ConfigureDialog):
                           row, 'interface.grampletbar-close', stop=3,
                           extra_callback=self.cb_grampletbar_close)
         row += 1
+        # When using Select People, do the trees expand by default?
+        cbox = self.add_checkbox(grid,
+                                 _("When selecting people, use a List "
+                                   "(default: a Tree)"),
+                                 row, 'interface.people-selector-list', stop=3)
+        cbox.set_tooltip_text(_("Tree view with large trees will slow down the"
+                                " dialog."))
+        row += 1
         return _('Display'), grid
 
     def add_places_panel(self, configdialog):
--- a/gramps/gui/editors/editfamily.py
+++ b/gramps/gui/editors/editfamily.py
@@ -323,6 +323,8 @@ class FastMaleFilter:
         value = self.db.get_raw_person_data(handle)
         return value[2] == Person.MALE
 
+    text = ' '
+
 class FastFemaleFilter:
 
     def __init__(self, db):
@@ -331,6 +333,8 @@ class FastFemaleFilter:
     def match(self, handle, db):
         value = self.db.get_raw_person_data(handle)
         return value[2] == Person.FEMALE
+
+    text = ' '
 
 #-------------------------------------------------------------------------
 #
--- a/gramps/gui/selectors/baseselector.py
+++ b/gramps/gui/selectors/baseselector.py
@@ -21,6 +21,16 @@
 
 #-------------------------------------------------------------------------
 #
+# Standard python modules
+#
+#-------------------------------------------------------------------------
+import time
+import logging
+
+_LOG = logging.getLogger(".gui.baseselector")
+
+#-------------------------------------------------------------------------
+#
 # GTK/Gnome modules
 #
 #-------------------------------------------------------------------------
@@ -53,12 +63,13 @@ class BaseSelector(ManagedWindow):
     IMAGE  =  2
 
     def __init__(self, dbstate, uistate, track=[], filter=None, skip=set(),
-                 show_search_bar = True, default=None, expand=True):
+                 show_search_bar = True, default=None, expand=False):
         """Set up the dialog with the dbstate and uistate, track of parent
             windows for ManagedWindow, initial filter for the model, skip with
             set of handles to skip in the view, and search_bar to show the
             SearchBar at the top or not.
         """
+        self.cput = time.clock()
         self.filter = (2, filter, False)
         self.expand = expand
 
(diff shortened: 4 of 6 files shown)


Paul-Franklin:
Additional imports are usually added within parentheses
(as opposed to after a backslash).


More importantly, I find the suggested text for the new
config option to be somewhat confusing, and not at all
intuitive.  I wonder what a translator will think it means?

Instead of 

"When choosing from existing people, use List view (Tree view if off)"

what about something like

"When selecting people, use a List (default: a Tree)"

or maybe

"Whether selecting people is done via a List (default: a Tree)"

prculley:
@Paul-Franklin Updated per your comments...

Paul-Franklin:
Thank  you.

sam-m888:
New option:

new option 9738

sam-m888:
Found the following issue in testing:

* Clean Gramps profile
* Start Gramps and with no family tree
* Selected Edit>Preferences>Display and Select the checkbox for "When selecting people, use a List(default:Tree)" 
* Create a new family tree and import example.gramps
* Select the families category view and create a new family then attempting to "select a person as the father" or "mother" gives the following errors:

```
102018: ERROR: grampsapp.py: line 143: Unhandled exception
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/editors/editfamily.py", line 881, in sel_father_clicked
    skip=[x.ref for x in self.obj.get_child_ref_list()])
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/selectors/selectperson.py", line 77, in __init__
    skip, show_search_bar, default)
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/selectors/baseselector.py", line 122, in __init__
    self.build_tree()
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/selectors/baseselector.py", line 298, in build_tree
    search=filter_info)
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/views/treemodels/peoplemodel.py", line 580, in __init__
    scol=scol, order=order, sort_map=sort_map)
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/views/treemodels/flatbasemodel.py", line 487, in __init__
    self.rebuild_data()
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/views/treemodels/flatbasemodel.py", line 617, in _rebuild_filter
    dlist = self.search.apply(self.db, allkeys, tupleind=1)
AttributeError: 'FastMaleFilter' object has no attribute 'apply'

```

```
288548: ERROR: grampsapp.py: line 143: Unhandled exception
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/editors/editfamily.py", line 836, in sel_mother_clicked
    skip=[x.ref for x in self.obj.get_child_ref_list()])
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/selectors/selectperson.py", line 77, in __init__
    skip, show_search_bar, default)
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/selectors/baseselector.py", line 122, in __init__
    self.build_tree()
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/selectors/baseselector.py", line 298, in build_tree
    search=filter_info)
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/views/treemodels/peoplemodel.py", line 580, in __init__
    scol=scol, order=order, sort_map=sort_map)
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/views/treemodels/flatbasemodel.py", line 487, in __init__
    self.rebuild_data()
  File "/home/grampsdev/aaSandbox/gramps/gramps/gui/views/treemodels/flatbasemodel.py", line 617, in _rebuild_filter
    dlist = self.search.apply(self.db, allkeys, tupleind=1)
AttributeError: 'FastFemaleFilter' object has no attribute 'apply'


```

prculley:
@sam-m888 ; guess I should have tested with add parents instead of just add children...  Seems this uses something called a 'fast filter' which was not supported by list model.  As near as I can tell this type of filter is only used by the edit family parent selector.  Anyway, I got it added in to the list flatbasemodel and working.

Nick-Hall:
Why not use two flat views next to each other?  The left-hand flat view could contain a single surname column.  When a row is selected the right-hand flat view could be populated.  This should be very fast.

A similar approach could be used to select citations.  However, we would need something different for the place hierarchy.

prculley:
@Nick-Hall I am not all that interested in adding new features here, just clearing a bug and completing a PR.  I don't personally have a very large db, so I don't notice the slowdown, although I don't care for the fully opened people selector.  So I will probably leave my preferences set to collapsed Tree view.

Since we seem to have accepted RonJohn (Ron Johnson's) request to have something like a fully expanded view, and a lot of others don't like it, I think I have found a pretty simple way to satisfy both sets of people.

It this is not satisfactory, you can always submit your own PR :smirk:

Nick-Hall:
I implemented the feature request, and in retrospect, it was a mistake and should be reverted.  At the time, I didn't realise the performance impact it would have.

In general, we should design a single solution that suits everyone rather than adding customisation options.  It has long been our policy to resist adding extra preferences.

Paul-Franklin:
"I am not all that interested in adding new features here,
just clearing a bug and completing a PR."

I agree.

The former will only delay the release of 5.0.0.
The latter will speed up the release of 5.0.0.

Nick-Hall:
@Paul-Franklin This PR is more like a new feature.  I don't like it as a bug fix.  The original collapsed tree is probably what users with large databases want.

prculley:
@Nick-Hall I don't mind adding preferences, I know that not every user is the same and they like to do things in different ways.  We should try to make sure that default settings are reasonable for the majority of users, because the less sophisticated may never look at them.

I'm afraid that no matter what non-customizable choice we put into place here we will have some unhappy users.

This PR basically give users the same choice they have in the normal people view, tree or flat.  Or were you planning to eliminate one of those as well?

Nick-Hall:
@prculley It isn't always a good idea to try to keep every user happy.  If we took that approach we would end up with hundreds of preferences.  In fact, I remember such discussions on the list in the past.  We certainly shouldn't add extra preferences to a maintenance branch.

Just because there are two person views, doesn't mean there have to be two options for the person selector.  If we followed that logic, we would have two options for the citation and place selectors for consistency.

prculley:
> Just because there are two person views, doesn't mean there have to be two options for the person selector. If we followed that logic, we would have two options for the citation and place selectors for consistency.

Darn, You just anticipated my next couple of PRs...

Nick-Hall:
@prculley I will deal with this myself.

PR #386, opened 2017-05-04
bug 6664; fix Detailed Descendent Report, multi paths to descendant

prculley:
where the descendant ends up in different generations at the same time.

An earlier patch for bug https://gramps-project.org/bugs/view.php?id=5259 seems to have eliminated multiple reports for different paths so people who are members of different generations via different paths only get reported in the lowest generation number. I expect that this is reasonable in that if we tried to report people in each generation that they are members of, the report would rapidly get unwieldy for convoluted trees.

The current code that creates the text 'is the same person as' is trying to indicate that a descendant is an already reported member of an earlier generation (which we don't do now) and so ends up redundant the way things are now; it is just preventing the normal report result.

I've completely eliminated the code which makes that check, and the associated edit/preferences option.  So the code now completes the report as expected by the bug submitter.


sam-m888:
Thank you

PR #389, opened 2017-05-05
Narrative web features: 5592  + 5408  + 10031

SNoiraud:
This solves one translation problem with note type.
This solves several structure problems in html pages.
5408 children dates
5592 show more details
10031 : maps not displaying all events


Paul-Franklin:
Line 858 and 861 seem to have different "locale" attributes.

PR #390, opened 2017-05-10
remove unused GUI import from plugins.export.exportxml

prculley:
Caused a GUI warning under CLI

### DIFF
--- a/gramps/plugins/export/exportxml.py
+++ b/gramps/plugins/export/exportxml.py
@@ -62,7 +62,6 @@ from gramps.gen.updatecallback import UpdateCallback
 from gramps.gen.db.exceptions import DbWriteFailure
 from gramps.version import VERSION
 from gramps.gen.constfunc import win
-from gramps.gui.plug.export import WriterOptionBox, WriterOptionBoxWithCompression
 import gramps.plugins.lib.libgrampsxml as libgrampsxml
 
 #-------------------------------------------------------------------------


sam-m888:
This causes the "Export options" dialog to not show the "Export options" tab

Working without patch:
working-without-patch

Blank tab with patch:

with-patch

This line was introduced originally in [PR121: Add compression option on XML export](https://github.com/gramps-project/gramps/commit/af726ec8084a50ae2b7f39d7b7e7f45ae5d51fa3)

prculley:
@sam-m888 thank you for doing the testing.  I admit I did not even try to test thoroughly.  It seemed a simple case of a leftover 'unused import' (according to pylint).  I did check and _there was no usage of the imported names_ when I did a search.

And it seemed in line with our policy to not have gui imports in 'gen' type modules (this is a plugin, but it is used in both CLI and GUI).

Now I have to go and try to understand how an apparently unused import is actually used...  Something new to learn every day.

@Nick-Hall Is this sort of thing common in Gramps?  Seems like it is sufficiently non-obvious that perhaps we should add comments to an import line like this...

Nick-Hall:
@prculley Report, tool, export and docgen plugins can have an associated options class.  This is specified in the _gpr_ file.  The options class is usually defined at the bottom of the plugin module, but in the case of export plugins these can be shared.

So, this is uncommon.  I would probably put the import after the others with a comment.

PR #392, opened 2017-05-12
Relationships view new toolbar icon to [Add a new Person]

sam-m888:
_Made this WIP as it may be confusing when you add a person the displayed active person does not change to the just newly added person._

The idea of this is that the user does not have to change to the people list view to create a new person and they can start creating a new family tree from this view.  Current behavior is on an empty family tree all the icon's are inactive with no hint as to what to do.

* "Add a new Person" Icon on Tool bar .

37b3cb7e-374b-11e7-8f62-8d47d5c0d83b

> Current behavior is on an empty family tree all the icon's are inactive with no hint as to what to do.

currentbehavior

 - [ ] Update [wiki](https://gramps-project.org/wiki/index.php?title=Gramps_5.0_Wiki_Manual_-_Categories#Relationships_Category)

### DIFF
--- a/gramps/plugins/view/relview.py
+++ b/gramps/plugins/view/relview.py
@@ -410,6 +410,10 @@ class RelationshipView(NavigationView):
         """
       <placeholder id='otheredit'>
         <item>
+          <attribute name="action">win.AddNewPerson</attribute>
+          <attribute name="label" translatable="yes">Add new Person...</attribute>
+        </item>
+        <item>
           <attribute name="action">win.Edit</attribute>
           <attribute name="label" translatable="yes">Edit...</attribute>
         </item>
@@ -498,6 +502,18 @@ class RelationshipView(NavigationView):
     <placeholder id='BarCommonEdit'>
     <child groups='RW'>
       <object class="GtkToolButton">
+        <property name="icon-name">gramps-person</property>
+        <property name="action-name">win.AddNewPerson</property>
+        <property name="tooltip_text" translatable="yes">'''
+        '''Add a new Person</property>
+        <property name="label" translatable="yes">Add new Person...</property>
+      </object>
+      <packing>
+        <property name="homogeneous">False</property>
+      </packing>
+    </child>
+    <child groups='RW'>
+      <object class="GtkToolButton">
         <property name="icon-name">gtk-edit</property>
         <property name="action-name">win.Edit</property>
         <property name="tooltip_text" translatable="yes">"""
@@ -570,6 +586,7 @@ class RelationshipView(NavigationView):
         self.family_action = ActionGroup(name=self.title + "/Family")
         self.family_action.add_actions(
             [
+                ("AddNewPerson", self.add_new_person),
                 ("Edit", self.edit_active, "<PRIMARY>Return"),
                 ("AddSpouse", self.add_spouse),
                 ("AddParents", self.add_parents),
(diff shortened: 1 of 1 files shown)


sam-m888:
Due to merging https://github.com/gramps-project/gramps/pull/281 need to make sure no menu entry "Add a new Person" is added to the **Edit** menu.

johnbickmore:
@sam-m888 I really like what you're attempting to do with Relationships view by making things self explaining to user, I'd suggest getting this into Gramps as soon as possible when done along with pr 581 🚀 🎇

PQYPLZXHGF:
@sam-m888  Your PR's ( https://github.com/gramps-project/gramps/pull/393 / https://github.com/gramps-project/gramps/pull/394 / https://github.com/gramps-project/gramps/pull/581 ) on usability look great any chance they will make it into 5.1?

Nick-Hall:
They won't make it into v5.1, but we can discuss user interface changes for v5.2.

prculley:
This will have to be redone to work with the new uimanager.  See other code for toolbars for examples.

sam-m888:
Paul, will do thanks


Raised feature request for this.

11232: [Relationships view]Add new toolbar icon to [Add a new Person]
https://gramps-project.org/bugs/view.php?id=11232

wroldwiedbwe:
Good idea, I'm seeing the following message:

> /usr/lib/python3/dist-packages/gi/overrides/GObject.py:541: Warning: /build/glib2.0-pjKWYQ/glib2.0-2.48.2/./gobject/gsignal.c:2635: instance '0x2c6ec90' has no handler with id '21883'
>   return func(*args, **kwargs)
>   
> /usr/lib/python3/dist-packages/gi/overrides/GObject.py:541: Warning: /build/glib2.0-pjKWYQ/glib2.0-2.48.2/./gobject/gsignal.c:2635: instance '0x2c6ea30' has no handler with id '32862'
>   return func(*args, **kwargs)

but it seems to work.

emyoulation:
How about adding the same New Person button to the Charts view toolbar?  However, the focus would have to auto-navigate to the newly created person. (It has seemed odd that the focus does not track to new object when adding in a view.)

A New Person button would allow a abstract block-building of relationships concept to be exploited in the Charts view.

The Pedigree and H-Tree already have placeholders... where a double-click can create a New Relationship dialog for ancestors.  This allows Parents or Siblings to be added.  There is no facility for create a New Relationship for Spouses ... which would also allow partners and children to be added.  (I thought clicking the Pedigree chart's leftmost button "Jump to child..." might reveal partner & child placeholders.  But it does not.)  

However, also adding the "Add Parents" "Select Family" (although it'd be nice to be able to change their Role between partner OR a child) and "Add Life Partner" would make a graphical tree building a feature.   (Another alternative would be to add a Relationships/Families tab to the Person Editor to allow adding/selecting a new Family reference and select the Role. Expanding the References tab with those capabilities might make GUI metaphor sense without adding another tab.)

dsblank (commented):
**Summary:** Adds an "Add new Person..." button/menu item to the Relationships view toolbar and context menu.

**What it does:**
1. Adds a toolbar `GtkToolButton` with `gramps-person` icon and `win.AddNewPerson` action
2. Adds a menu item "Add new Person..." in the `otheredit` placeholder
3. Registers the `AddNewPerson` action in `define_actions`
4. Implements `add_new_person()` which creates a blank `Person` with a default `Surname` and opens `EditPerson`

---

**Issues / Suggestions:**

1. **`add_new_person` signature is inconsistent** — other callbacks like `add_spouse`, `add_parents` only take `(self, obj)`, but this one takes `(self, obj, event)`. Since it's triggered by a toolbar button (not a mouse event), `event` will likely always be `None`. Should match the pattern of other action callbacks.

2. **Comment style** — `#the editor requires a surname` is missing a space after `#`. Should be `# the editor requires a surname`.

3. **String concatenation in XML** — In the toolbar XML:
   ```python
   <property name="tooltip_text" translatable="yes">'''
   '''Add a new Person</property>
   ```
   This implicit string concatenation mid-XML string looks like a line-length workaround but is fragile and unusual. Better to just keep it on one line or restructure.

4. **No navigation to new person** — After creating a person via `EditPerson`, the view doesn't navigate to the newly created person. Other similar flows (e.g. `add_spouse`) typically connect the result back to the view. This may be intentional for the relationships view context, but worth considering.

5. **`WindowActiveError` handling** — Consistent with the rest of the file, so fine.

Overall it's a straightforward and useful addition. The main concerns are the `event` parameter inconsistency and the tooltip string concatenation.

Nick-Hall:

PR #393, opened 2017-05-12
Add Start here gramplet to sidebar in Relationship view

sam-m888:


eno93:
My suggestion:

> To start entering you family history you will need to create a new Family Tree (sometimes called a \'database\'). From the menu select "Family Trees" >"Manage Family Trees" and then press the '"New" button and then name your Family Tree.

Then provide that exact link you mean to the user manual.

It would be great if each category/view had a short help section like this 👍   Even better if screenshots could be embeded 🥇

sam-m888:
@Nick-Hall Any chance this can make it into Alpha 2?

Nick-Hall:
@sam-m888 Does this make sense without #394?

sam-m888:
> Does this make sense without #394?

@Nick-Hall Your right not with the current wording which assumes it is the first view seen. Should I make a modified version?

Nick-Hall:
@sam-m888 Why not just enhance the Welcome gramplet on the Dashboard?

PR #394, opened 2017-05-12
Change Default Navigator views order

sam-m888:
Reordered the categories in the Navigator so that List are grouped below and non list are at the top. This makes it easier for people to understand at a glance the difference between the top 4 categories [Dashboard] & [Programmed Views/Extensions] and those below eg. List.

* Ideally I would like a separator or the label "List" to appear in-between "Geography" and "People"

Old order on left / New order on right
old left - right-new

### DIFF
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -191,10 +191,11 @@ register('geography.use-keypad', True)
 # needed), for instance to four 'interface.clipboard' variables --
 # so do a recursive grep for "setup_configs" to see all the (base) names
 register('interface.dont-ask', False)
+#Default Navigator views (sets order shown)
 register('interface.view-categories',
-         ["Dashboard", "People", "Relationships", "Families",
-          "Ancestry", "Events", "Places", "Geography", "Sources",
-          "Citations", "Repositories", "Media", "Notes"])
+         ["Relationships", "Dashboard", "Ancestry","Geography",
+          "People", "Families", "Events", "Places", "Notes",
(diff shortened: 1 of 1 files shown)


Paul-Franklin:
I somewhat dislike the idea of making what I consider to be a
"major" change to the UI so late in the cycle, just a short period
of time before 5.0.0 is released.  IMHO people will "expect" the
order to be what it has been.  Certainly I do, when I am running
gramps in a different language than my own.  It really helps (for
instance) when I have memorized the order of the preferences
in the various Edit=>Preferences tabs, although I admit that is
not what you are talking about here -- but it's close.

I would suggest delaying this until 5.1.0 but will wait to see if
there are other misgivings besides mine.

(Too bad this change wasn't suggested a year ago, say.  I
might have gotten used to it by now.)  8-)

sam-m888:
>  (Too bad this change wasn't suggested a year ago, say. I
> might have gotten used to it by now.) 8-)

@Paul-Franklin  Damm time machine keeps shorting out :)

dsblank:
I think we can close this draft PR as the ordering has become entrenched in the use of Gramps now.

daleathan:
Hi @sam-m888 

Seeing as pr 2295 https://github.com/gramps-project/gramps/pull/2295 will add another list view for DNA .  This change would be beneficial, especially from the perspective of new users and old

Also note that the [cardview](https://github.com/cdhorn/CardView/blob/master/images/cardview-view-person.png) addon adds a **Tags** list view.  

I'll wait for the forum discussion you mention.

PR #396, opened 2017-05-18
8200: Person Relatives Tab should use the type from the relationship

SNoiraud:


Paul-Franklin:
Since there are now two variables, flagged for translation, when
a gramps.pot is made a comment will be generated suggesting
that the two be named variables instead.

sam-m888:
Worked for me.  

relatives-gramplet-patched
relatives-gramplet-patched-2

SNoiraud:
> Since there are now two variables, flagged for translation, when
> a gramps.pot is made a comment will be generated suggesting
> that the two be named variables instead.
OK. I'll do that.

Paul-Franklin:
Looks fine to me.  Thank you.

PR #397, opened 2017-05-22
fixes for Media Reference Editor (bug 8429)

prculley:
bug 8429; fix for selection frame outside/behind image

This was two different bugs; 
1)the various objects were not responding to the resize events, 
2)the cairo objects were under some conditions hidden behind the image (worse on the Windows platform).  Even on Linux, a zoom in would hide the cairo objects.

Added signal handler to deal with the resize event.  Unfortunately, it appears that Gtk still is doing something odd when the 'Shared Information' expander is clicked and the 'Preview' window shrinks.  The basic resize handler did not correctly deal with this case.  I found it necessary to add another handler for the expander; in that I had to kill the image and allocate a small size, before letting the resize handler bring it back to the new correct size.  Surprisingly, the same trick did NOT work if put into the resize handler.  There is a lot I don't understand about Gtk...

Fixed the cairo object to use the already defined (by Gtk) object rather than making a new one each time.  I suspect that on Windows, the new cairo object was under some conditions getting drawn behind the image.  The Gtk object, when reused seems to always work correctly.

During debugging I found that the spin buttons did not update the selection frame etc. if the image had already been selected via mouse click (frame showing and background greyed out).  Added a signal emit to deal with this.

During debugging I found that the popup instructions did not work; the user could not just drag to draw the selection.  I added code to detect and do this.

Tested on Windows and Linux (Umbuntu).

Also Tested with the Photo Tagging Gramplet, looks like everything still works (better), although I found a Photo Tagging Gramplet specific bug.

### DIFF
--- a/gramps/gui/editors/editmediaref.py
+++ b/gramps/gui/editors/editmediaref.py
@@ -100,6 +100,7 @@ class EditMediaRef(EditReference):
         tblref =  self.top.get_object('table50')
         self.notebook_ref = self.top.get_object('notebook_ref')
         self.track_ref_for_deletion("notebook_ref")
+        self.expander = self.top.get_object('expander1')
         #recreate start page as GrampsTab
         self.notebook_ref.remove_page(0)
         self.reftab = RefTab(self.dbstate, self.uistate, self.track,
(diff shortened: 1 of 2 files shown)


sam-m888:
Confirming that it worked for me on Linux (xubuntu 16.04.2)

2017-05-27 09_44_26-ubuntu-16 04 1-grampsdev-amd64 running - oracle vm virtualbox

bpisoj:
@Nick-Hall 
It look strange to me that latest commit is 25 days old.
Should'n we update date on commit?

git commit --amend --date="now" --no-edit

PR #398, opened 2017-05-23
fix bug 10052, relationship calculator fails (missing uistate param)

prculley:
bug was a missed edit during a general fixup of treemodels to add uistate...

### DIFF
--- a/gramps/plugins/tool/relcalc.py
+++ b/gramps/plugins/tool/relcalc.py
@@ -118,7 +118,7 @@ class RelCalc(tool.Tool, ManagedWindow):
         self.textbuffer = Gtk.TextBuffer()
         self.text.set_buffer(self.textbuffer)
 
-        self.model = PersonTreeModel(self.db)
+        self.model = PersonTreeModel(self.db, uistate)
         self.tree.set_model(self.model)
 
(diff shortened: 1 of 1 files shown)


Paul-Franklin:
I have duplicated this bug.

I concur this is the solution.  It should be committed.

(If it were me, I'd have just committed it, but you are not me.)

Paul-Franklin:
Remember to commit it only to maintenance/gramps50, since
the transfer to master will be done by a "senior developer."

prculley:
@Nick-Hall Do I have to close and restart this PR, since it was originally aimed at master, and now should be aimed at gramps50?  Or is there some way to shift it to the other branch?

Nick-Hall:
@prculley You can use the "Edit" button next to the title to change the base branch.

Nick-Hall:
Fixed in commit 22fb1ab254819f1a3c7d098246830572c789f540.

Paul-Franklin:
@NIck-Hall ISTM it was fixed in "5.0.0-alpha3" (and not ...2).

https://gramps-project.org/bugs/view.php?id=10052

Nick-Hall:
@Paul-Franklin Thanks.  I have changed it to alpha3.

PR #399, opened 2017-05-25
Fix bug 10054 HandleErrors in Gramplets

prculley:
Error occurred when deleting a Family in Family view.
Don't know why the error occurred in one Gramplet, but not all of them, when they generally use the same code methods for looking at the active handle; sigh...

Checked for similar errors in other Family elements/Gramplets.

The citations Gramplet error was a bit different, it appeared when I added a Citation to the Family and selected the family in the Family view.

### DIFF
--- a/gramps/plugins/gramplet/citations.py
+++ b/gramps/plugins/gramplet/citations.py
@@ -133,9 +133,10 @@ class Citations(Gramplet, DbGUIElement):
         for lds in obj.get_lds_ord_list():
             self.add_citations(lds)
             place_handle = lds.get_place_handle()
-            place = self.dbstate.db.get_place_from_handle(place_handle)
-            if place:
-                self.add_place_citations(place)
+            if place_handle:
+                place = self.dbstate.db.get_place_from_handle(place_handle)
+                if place:
+                    self.add_place_citations(place)
 
     def add_association_citations(self, obj):
         for assoc in obj.get_person_ref_list():


Nick-Hall:
The problem actually seems to be in the `get_active` method which should either return a valid handle or `None`.  The [History](https://github.com/gramps-project/gramps/blob/master/gramps/gui/displaystate.py#L75) object should remove deleted handles from the list and emit an _active-changed_ signal when necessary.

The gramplets may also be connecting to a database signal and updating before history list is updated.  If this is the case, we may need to change the order in which the signals are processed.

prculley:
The sequence I observe is:

1. Delete Family at Gui
2. signal: person-update (with handles that are being removed from family)
3. Children Gramplet runs generic update code, does get_active('Family'), gets the family handle from history, BUT, it is already gone from db. (Thus the original bug)
4. signal: family-delete
5. family handle is removed from history
6. callback listview row-delete etc.
7. Signal active-changed, Gramplet runs generic update, gets active handle (another valid family)

Looks like we are actually doing the operations in the family delete, and then, afterword, emitting the signals for that.  I don't have time now to see if we changed anything in this area recently.

I think I understand why only Children Gramplet is affected; since it registers for both person-update and family-update, but runs a generic update routine it was subject to this bug, whereas since the other Gramplets only registered for family-update they were not.

prculley:
Found a way to deal with the get_active HandleErrors for all the Gramplets, so have just fixed citations in this one.

PR #405, opened 2017-05-29
Convert handle type to str in the database API

Nick-Hall:
The mixed type of handles has been a constant source of bugs.
Converting them all to str in the database API should make
maintenance easier. The key to BSDDB tables must still be bytes
in the database layer.


prculley:
Wow; that must have been a marathon editing session.  Thanks.
I think you messed up set_default_person_handle; in order to open a current bsddb, we still have to expect this in bytes mode, right?  So I think we still have to set it that way too...

Are you planning on a similar session for addons?  Or would you like me to try it to share the work?  I think it is polite to fix for everyone when we change the API.

Nick-Hall:
Upgraded the database in commit 16cd995558eda3305203a9112f21a5a5a8dcc9bd.  Now all handles in the metadata table are str.

I updated the third-party addons in commit gramps-project/addons-source@05be639b72fa9e43fa90b1c39af6097f1d6348b9.  If you have time, perhaps you could check to see if I have missed anything.

Paul-Franklin:
I am getting crashes, using the [16cd995] gramps50.

They happen when I select an old (schema 18) DB to
load, then hit "Cancel" in that big warning dialog when
it ask me if I really want to upgrade it to 19.

Let me know if you need more details and can't reproduce
it.  I'm guessing you'll be able to.  (The "researcher" is None,
so it has no is_empty().)

Paul-Franklin:
I am getting other crashes, still using the [16cd995] gramps50,
but this time in a newly-created (schema 19) DB.

AttributeError: 'bytes' object has no attribute 'encode'

I am saving the crash typeouts, but they are on another
machine, not this one, so it's somewhat of a pain to go
there, put them on a flash disk, then create them here,
then paste them somewhere, so I'd rather not.

So I'm guessing you'll be able to reproduce things like
that yourself, once you start trying, but let me know if not..

Paul-Franklin:
This is just a question, since I've never looked into
details like this, but does the DTD need to go up (to
1.7.2 say) with this schema-19 change?

Nick-Hall:
Fixed the DummyDb errors in commit 512fa615bfc92b6d076c9cf5bd22410f472f3bb2.  This would have been a problem for any cancelled database upgrade.

The XML schema remains unchanged.

PR #406, opened 2017-05-30
10049: tabs reorganization in narrative web.

SNoiraud:


Paul-Franklin:
This looks like a good idea to me.  I think it should be committed.

SNoiraud:
Is it a good idea to backport it to 5.0 ?

prculley:
I would think that this is not in the category of bug fixes; rather it
looks to me like an enhancement.  So by our new rules, it doesn't belong in
Gramps50...

My opinion of course.

Paul C.

Paul-Franklin:
I was assuming that it was going into gramps50.  It never
occurred me to think otherwise.

As far as I'm concerned it's a bug fix, since it came about
after a user didn't see the options in the Privacy tab (since
that tab was not visible in the default report dialog), then
filed a bug report (10049) about people not being there.

So I'd say it belongs in gramps50.

My opinion of course.  8-)

PR #409, opened 2017-06-02
6364: dead people are shown as alive. Fixes #06364

SNoiraud:
In the related bug, I attached a sample database test with 4 people.
2 are dead and 2 are alive. Without this patch, some reports doesn't work.
The LivingProxyDB gives incorrect results when we have a death record with an invalid date or without date.


Nick-Hall:
Rebased.

PR #412, opened 2017-06-04
Bug 10068 & 10070 fix for HandleErrors and dbapi undo delete issue

prculley:
Three commits included.
1) Undo of delete person on dbapi fails https://gramps-project.org/bugs/view.php?id=10070.  Actually applies to all deletes.  This one is because of an error in plugins.db.dbapi _do_remove method, where the code first deletes the object, then tries to get it for the undo db, so the undo transaction doesn't save anything.  I note that Doug Blank fixed this for his GPrime, although he refactored the code so much I decided to just do the minimal patch rather than trying to grab his work.

2) Gramplets and Status bar updates were producing HandleError exceptions (some of which resulted in Error dialog) https://gramps-project.org/bugs/view.php?id=10068.  Fixed the Gramplet class get_active methods to check that handle was valid before passing to the Gramplets.  Also fixed the navigation_label method used by status bar to check for good handle first.

3) The above (2) needed a db has_handle method, which was already present in dbapi, but absent in bsddb.  So added method to bsddb.

**Despite title, most of this applies to both bsddb and dbapi**


gramps/gen/plug/_gramplet.py
```diff
@@ -127,15 +128,20 @@ def get_active(self, nav_type):
         """
         Return the handle of the active object for the given navigation type.
         """
-        return self.uistate.get_active(nav_type, self.nav_group)
+        handle = self.uistate.get_active(nav_type, self.nav_group)
+        if handle and nav_type in self.dbstate.db.get_table_names() and \
```
Nick-Hall: The `has_handle` method is not actually part of the public API and should be made protected.  This is currently under review - see bug [#9541](https://gramps-project.org/bugs/view.php?id=9541).

We need to discuss this further on the list.

prculley: I know it was under review ("to be added if and when needed"); it turns out to be a useful function in this case.  I could code around it with a helper that directs the 'nav_type' to the various 'has_???_handle' methods, but that seems a bit inelegant since both db types then collapse them to another (hidden) single helper method method again.

Nick-Hall: I agree, but additions to the API should really be discussed on the list first.

gramps/plugins/db/bsddb/read.py
```diff
@@ -1632,6 +1635,23 @@ def get_raw_note_data(self, handle):
     def get_raw_tag_data(self, handle):
         return self.__get_raw_data(self.tag_map, handle)
 
+    def has_handle(self, obj_key, handle):
+        """
+        Determine in a valid handle exists for the obj_key type
```
johnbickmore: Do you mean   "Determine ~~in~~**if** a valid handle exists for the obj_key type"?

PR #414, opened 2017-06-07
Citations Gramplet doesn't update in many cases, fixes #10069

prculley:
To avoid issues in testing this, https://github.com/gramps-project/gramps/pull/399 should be merged first.  The bugs were a bit different so different PRs and commits.

I went through every type of data that the citations Gramplet(s) can show, and tested that the Gramplet updated when 1) the Citation data was changed, and 2) the Citation was deleted, and 3) the citation was removed from the object it was attached to, and 4) the citation was restored to a monitored object.

Since the Citations Gramplet 'goes deep' for many types of data, this meant on a Persons citations for example; checking citations on Persons Events, checking Citations on Person Events Places, Person Media citations etc.

The original code used a callman method register_obj, which did not do a very good job of finding all the handles that the citations Gramplet needed to monitor for its signals.  So I found a slightly different way to accomplish the job.  I also had to add additional monitoring of the objects that citations are attached to in order to catch cases where citations were deleted, and added to the objects.

This started as an observation during debug of another issue, and turned into an educational debug process or our callback signalling system in its own right.

### DIFF
--- a/gramps/plugins/gramplet/citations.py
+++ b/gramps/plugins/gramplet/citations.py
@@ -57,8 +57,14 @@ class Citations(Gramplet, DbGUIElement):
         """
         called on init of DbGUIElement, connect to db as required.
         """
-        self.callman.register_callbacks({'citation-update': self.changed})
-        self.callman.connect_all(keys=['citation'])
+        self.callman.register_callbacks({'citation-update': self.changed,
+                                         'person-update': self.changed,
+                                         'family-update': self.changed,
+                                         'event-update': self.changed,
+                                         'media-update': self.changed,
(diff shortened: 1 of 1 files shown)


gramps/plugins/gramplet/citations.py
```diff
@@ -82,6 +89,9 @@ def build_gui(self):
         return top
 
     def add_citations(self, obj):
+        cl_name = CLASS2KEY.get(obj.__class__.__name__)
```
Nick-Hall: I don't like using `CLASS2KEY` here, so I have committed a suggested change.

PR #417, opened 2017-06-12
_errorview dialog mapped without transient parent, Fixes #8128

prculley:
@SNoiraud noticed that the errorview dialog had no transient parent on another bug and added another note to #8128.  I looked over the code and determined that there are some cases where the parent window search we did was not catching a parent, yet there was one available.  It seems that sometimes the gtk.is_active() on a window object is not set.  However the gtk.is_visible() seems to be true in every case I could test for.

@Paul-Franklin since you tested a variety of case the last time you worked on this, I'd appreciate if you tested this.  It seems to work ok for the cases I tried on my Windows and Linux systems.

I also noticed that the error report assistant was not getting destroyed; after several tries with it there were several extra copies of it in memory.

During debug I found it necessary to disable the save of window position for errorview, I was confused at first that it wasn't popping up over the failed window ;-)

### DIFF
--- a/gramps/gui/logger/_errorreportassistant.py
+++ b/gramps/gui/logger/_errorreportassistant.py
@@ -119,7 +119,7 @@ class ErrorReportAssistant(ManagedWindow, Gtk.Assistant):
         if self.parent_window is not None:
             self._save_position(save_config=False) # the next line saves it
             self._save_size()
-        self.hide()
+        self.destroy()
         if self.ownthread:
             Gtk.main_quit()
 
--- a/gramps/gui/logger/_errorview.py
+++ b/gramps/gui/logger/_errorview.py
@@ -106,8 +106,13 @@ class ErrorView(ManagedWindow):
     def draw_window(self):
         title = "%s - Gramps" % _("Error Report")
         self.top = Gtk.Dialog(title)
+        # look over the top level windows, it seems the oldest come first, so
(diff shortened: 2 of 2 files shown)


Paul-Franklin:
" It seems to work ok for the cases I tried on my Windows and Linux systems."

@prculley I am happy to trust your judgement -- and fixes.

Especially since (IMHO) it is "easier" for you (on Windows) to
notice such problems (and thus fix them).  I do not get any typeout
for such things for instance, and have to notice such problems
visually, when a dialog does not appear centered.  Additionally,
I am running gramps with old (3.10/3.12) GUI/Gtk libraries, which
do not manifest some "modern" problems (like the AIO ones).

FWIW I view all of my ManagedWindow (and transient parent)
changes as prototypes, to be verified by use and testing, as you
are now doing.  I will not be surprised if anything needs to be
changed.  Thank you for catching these.  (The dual-window
error reporting stuff was a special pain, IIRC.)

I looked at the code and your changes seem fine to me -- and
even more so when I consider that IMHO you have more GUI
knowledge than I do.  8-)

Thanks for catching these -- and fixing them!

(Or at least for filing this/these pull requests.  I am not
sure of the mechanism for implementing them, if they
require some other developer to commit them.  I still
think we are short of active developers, or their time.
In general, pull requests seems to be piling up.)

PR #418, opened 2017-06-13
Delete Multi Events

apoettker:
After selecting multiple events the 'Delete' dialogue offers only 'Confirm Each Delete' or 'Delete All'.
Therefore a user have no chance to escape if he pressed the button by mistake. Rather he's forced to press n-times the 'Cancel' button in the 'Confirm Each Delete' option. 

To improve this dialogue I've added an button (here:  'Cancel Delete') to abort the operation. Some logic is adjusted, the 'dialog.glade' file aligned.

screenshot delete multi events 2017-06-13

### DIFF
--- a/gramps/gui/dialog.py
+++ b/gramps/gui/dialog.py
@@ -105,6 +105,7 @@ class QuestionDialog:
         label2.set_text(msg2)
         label2.set_use_markup(True)
 
+        self.xml.get_object('cancel').set_visible(False)
         self.xml.get_object('okbutton').set_label(label)
 
         if parent:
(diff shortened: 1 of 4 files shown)


bpisoj:
Wouldn't existing MultiSelectDialog be more appropriate for that instead of changing QuestionDialog2.
Family list already use them, see gramps/plugins/view/familyview.py

apoettker:
Excellent idea, consistency and uniform user interface is always a plus! Found it in libpersonview.py too. I've set dialog.glade, dialog.py and listview.py to the original state. Don't know exactly, if they have to upload all again (done for safety reasons). A screen shot now looks like this:

screenshot delete multi events 2017-06-20

Nick-Hall:
Merged in commits 40f47bc22d246dabdedf9a5c05f214c20b472257 and 5276461239325c3196649a6b379e4b8c95366b67.

@apoettker Thanks.  Code improvements are appreciated, but please put them in a separate commit in future.  This makes it easier for us to review your changes.

PR #419, opened 2017-06-13
10076: Filtering in the sidebar doesn't work in the citation view. Fixes #10076

SNoiraud:
This occurs if you use the regular expressions.


Nick-Hall:
Rebased.

Paul-Franklin:
Should it say "open" when it also says "committed" 13 June?

Nick-Hall:
Fixed in commit adb4883291159298781144706fa5bc6331870ef2.

PR #420, opened 2017-06-14
Removes try/except and recursion in ancestor filters +/- N gens

dsblank:
This fix rewrites the Ancestor-of +/- N filters to remove the try/except
and removes the recursion. The recursion would only be a problem
on very deep trees, but the try/except was hiding a variety of
problems. Most directly, if a person did not have a main family
then it would exit out of the loop.

(I don't have a test database to break the code other than my
real data, so you'll have to do that part.)

I also made sure that the meaning of N is consistent in
both filters:

1 - root person
2 - root person's parents
3 - root person's grandparents


Nick-Hall:
Rebased.

PR #422, opened 2017-06-16
Name Editor crash

bpisoj:
Fixes: #10084, #10085, #10086.


bpisoj:
This is caused by change in commit 352d6fc5582ce86c7760a13587cfc138e30a873e

Nick-Hall:
Fixed using is similar approach to that originally suggested in commit fd329764ed7eaae361c705a84eebe7f2ffc1bc7b.

PR #423, opened 2017-06-16
Guarantee a specific order of db signals; fixes for 10068 & 10070

prculley:
Bug #10068: during a Person delete operation, a large number of HandleError exceptions can occur.  These are somewhat masked by a try/except but we print an exception dump to the terminal.  In some cases the exception can actually go to a Gramps Error View dialog.

It turns out that these are happening because the db signal emits come after the actual commits, and, depending on the order they arrive, our UI objects (Gramplets and status bar), can get HandleErrors.

In particular, during a person delete, if the 'person-update' signal arrived first, UI elements try to update their screens with the (now deleted) person handle.
@Nick-Hall suggested arranging for deletes to be emitted first, this seems to deal with this case (but there are others).

The 'object-delete' and 'object-add' Signals from the db should now come out first, and should be entirely consistent across dbs.

Bug #10070 was found while debugging this case for dbapi; the undo of a delete was not working.  Turns out that when saving transaction data for undo, the data object was being loaded after it had been deleted.

This is a replacement for PR412 which tried to do this by testing for bad handles with has_??_handle calls to the db (now closed).

I have also added a unittest that 
1) checks the signals for primary db operations
2) performs undo and redo and checks actual db results
3) checks the signals for undo/redo db operations
4) checks the callback manager operation
5) does all of the above for both bsddb and dbapi
This should expand our code coverage somewhat and pretty well checks that this PR does what it says.

P.S. I found another case when testing if this worked (family merge), will be another bug/PR.

### DIFF
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -44,7 +44,8 @@ import glob
 from . import (DbReadBase, DbWriteBase, DbUndo, DBLOGNAME, DBUNDOFN,
                KEY_TO_CLASS_MAP, REFERENCE_KEY, PERSON_KEY, FAMILY_KEY,
                CITATION_KEY, SOURCE_KEY, EVENT_KEY, MEDIA_KEY, PLACE_KEY,
-               REPOSITORY_KEY, NOTE_KEY, TAG_KEY, TXNADD, TXNDEL)
+               REPOSITORY_KEY, NOTE_KEY, TAG_KEY, TXNADD, TXNUPD, TXNDEL,
+               KEY_TO_NAME_MAP)
 from ..errors import HandleError
 from ..utils.callback import Callback
 from ..updatecallback import UpdateCallback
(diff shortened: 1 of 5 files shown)


prculley:
@bpisoj; I'll do it myself if it doesn't get done in a few more days,
unless you want to release another drop of gramps50.  Let me know if you
want this done when you get ready to do the release.

Paul C.

Nick-Hall:
The code looks good to me, but I haven't had a chance to test it thoroughly.  Please feel free to merge.

bpisoj:
@prculley
I just release GrampsAIO-5.0.0-alpha2-112-g8f69201a7_win64.exe, when this PR get resolved i will make another one (hope it will be soon)

johnbickmore:
> I just release GrampsAIO-5.0.0-alpha2-112-g8f69201a7_win64.exe,

For those of you looking this can be found on sourceforge and is called [Gtk3.18-GrampsAIO-5.0.0-alpha2-112-g8f69201a7_win64.exe](https://sourceforge.net/projects/gramps/files/Unstable/Gtk3.18-GrampsAIO-5.0.0-alpha2-112-g8f69201a7_win64.exe/download) and not here on github.

bpisoj:
No, that version on sourceforge is different one.
It is made with Gtk-3.14 for testing some bug but is named wrongly with Gtk-3.18

prculley:
Manually merged to preserve commits... (why did we disable normal 'Merge' option for squash and merge?)

PR #424, opened 2017-06-17
Gramplets HandleErrors with Family Merge and issues on close/open

prculley:
This PR was tested on top of https://github.com/gramps-project/gramps/pull/423 to deal with as many HandleErrors as I could.  If you intend to test this, you should accept PR 423 first, I cannot guarantee everything works unless both are present.

This should finish #10068, #10081, #10082, #10089, #10090 and possibly some others.

After PR 423, I discovered that doing a Family merge that also merged at least one parent, would still get HandleErrors.  Root cause, a lot of Gramplets were getting connected to Person history active-changed signal, and thus were doing unnecessary updates.  In this case, the Person update occured before the Family Gramplets got shifted off of the family deleted by the merge, so the Gramplets tried to show the (now deleted) family again.  And got HandleErrors.

I changed the base Gramplet class to only make default connections to the Person History active-change signal when the Gramplet had a navtype of 'Person' from the .gpr file.  So unless a Gramplet has navtype=['Person'] in .gpr it will not get an automatic connection to Person active-changed.

It turns out a few Gramplets knew they were getting the 'extra' Person connection, and had a 'disconnect' which now failed with the prior change.  So these extra disconnects are removed.

I also noticed that on db close/open, a lot of signals were getting connected when they already were connected.  This because the connection was made in a db_changed method, but never disconnected.

To prevent multiple connections after db close/open, I decided to make all connections through Gramplet class 'connect', rather than directly to db, and to implement a disconnect on db_close and on db_changed, prior to making the connections again.

In the sweep though all our Gramplets I spotted a few other smaller issues #10082 had a wrong signal connect ('person-edit', rather than 'person-update').  And a few signals were missing from some of the Gramplets.

I apologize for not separating these issues into separate commits, but I discovered after all the testing that many of the changes were too close together for 'git add -p' to break them up.  And I'm too lazy to do it manually.

I have a PR for the addons-source repo that does the same thing for those addons as well.

### DIFF
--- a/gramps/gen/plug/_gramplet.py
+++ b/gramps/gen/plug/_gramplet.py
@@ -64,12 +64,12 @@ class Gramplet:
         self.init()
         self.on_load()
         self.build_options()
-        self.connect(self.dbstate, "database-changed", self._db_changed)
-        self.connect(self.gui.textview, "button-press-event",
-                     self.gui.on_button_press)
-        self.connect(self.gui.textview, "motion-notify-event",
-                     self.gui.on_motion)
-        self.connect_signal('Person', self._active_changed)
+        self.dbstate.connect("database-changed", self._db_changed)
+        self.dbstate.connect("no-database", self._no_db)
+        self.gui.textview.connect("button-press-event",
+                                  self.gui.on_button_press)
+        self.gui.textview.connect("motion-notify-event",
+                                  self.gui.on_motion)
         self._db_changed(self.dbstate.db)
         active_person = self.get_active('Person')
         if active_person: # already changed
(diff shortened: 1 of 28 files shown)


gramps/gen/plug/_gramplet.py
```diff
@@ -383,11 +384,27 @@ def _db_changed(self, db):
...
+            self.disconnect_all()  # clear the old signals from old db
         self.dbstate.db = db
         self.gui.dbstate.db = db
-        self.db_changed()
+        if db.is_open():
+            # the following prevents connecting to every Gramplet, and still
+            # allows Person Gramplets to be informed of active-changed
+            # Some Gramplets .gpr files don't have navtypes set, thus the
+            # try/except
+            try:
```
Nick-Hall: Use `hasattr` instead of `try`/`except`.

daleathan:
Version for addons https://github.com/gramps-project/addons-source/pull/75

PR #425, opened 2017-06-18
FamilyGroup Report crash

bpisoj:
PyICU Collator.getCollationKey(string).getByteArray()
return bytes

Fixes: #10077

### DIFF
--- a/gramps/gen/utils/grampslocale.py
+++ b/gramps/gen/utils/grampslocale.py
@@ -33,6 +33,7 @@ import codecs
 import locale
 import collections
 import logging
+from binascii import hexlify
 
 LOG = logging.getLogger("." + __name__)
 LOG.propagate = True
@@ -921,8 +922,8 @@ class GrampsLocale:
 
         if HAVE_ICU and self.collator:
             # ICU can digest strings and unicode
-            # Use hex() as a fix for bug #10077
-            return self.collator.getCollationKey(string).getByteArray().hex()
+            # Use hexlify() as to make a consistent string, fixing bug #10077
+            return hexlify(self.collator.getCollationKey(string)).decode()
         else:
             if isinstance(string, bytes):
                 string = string.decode("utf-8", "replace")
--- a/gramps/gen/utils/grampslocale.py
+++ b/gramps/gen/utils/grampslocale.py
@@ -923,7 +923,7 @@ class GrampsLocale:
         if HAVE_ICU and self.collator:
             # ICU can digest strings and unicode
             # Use hexlify() as to make a consistent string, fixing bug #10077
-            return hexlify(self.collator.getCollationKey(string)).decode()
+            return hexlify(self.collator.getCollationKey(string).getByteArray()).decode()
         else:
             if isinstance(string, bytes):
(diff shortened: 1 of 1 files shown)


bpisoj:
O.K.
But i leave few comments there also

PQYPLZXHGF:
Yet travis-ci shows this fails for all the test!

jralls:
Well *I* like it, but Travis-CI doesn't: "AttributeError: 'bytes' object has no attribute 'hex'"
Looks like bytes.hex was introduced in Py3.5. The old way appears to be
``` python
from binascii import hexlify
hexlify(self.collator.getCollationKey(string)).decode()
```
Nowhere near as pretty, but I don't suppose that we want to require Py3.5+ just yet.

Paul-Franklin:
@jralls So why not push those changes, just like Nick pushed
his changes to Josip's original code?  Then see what Travis says...

jralls:
@Paul-Franklin: Because this is @bpisoj's PR andthe actual code is in his github account.

Paul-Franklin:
ISTM that Travis failed:

TypeError: 'icu.CollationKey' does not support the buffer interface

jralls:
That's better. @Nick-Hall: is it safe to push the "Squash and merge" button? I trust it will let me put back @bpisoj's original commit message...

prculley:
@jralls I was curious myself on what 'Squash & Merge' did, so I performed an experiment.  The button brings up a confirm dialog with an edit box for the commit message.  I tried it against my personal copy of the Gramps master repo to see what would happen.  Looks like you would expect.  However, if the commit list includes other, non-related changes, perhaps due to something the PR writer did, it might be wiser to download a copy of the PR and do the appropriate interactive rebase locally before pushing back to the Gramps repo anyway.
 ' github

Nick-Hall:
@prculley Yes, squash and merge does exactly that.  It will squash the three commits in the "commit" tab into the first one.  This is probably the most useful action for simple pull requests.

@jralls In this case, I force pushed my commit to replace @bpisoj 's original commit.  I suggest you create a local branch with a single commit.  Make sure you put "Fixes #10077" in the commit message.  Then force push it to josip's branch before you merge and push it back to the repository manually.  GitHub will automatically mark the PR as merged because the commit hashes will match.  Mantis should assign the bug to you and change the status to resolved.

jralls:
@Nick-Hall if squash and merge does the right thing then why not use it?

Nick-Hall:
@jralls Squash and merge would almost work, but I would be the author and my comment about hex() would be in the commit message.

jralls:
Ah, there's no way to edit the commit message? That's no good then.

jralls:
Rebased, merged, and pushed.

If we're going to do a 4.2.6 then I should backport it as well. Will we?

Nick-Hall:
Actually you do get a chance to edit the commit message.  See step 4 in the [Merging a pull request](https://help.github.com/articles/merging-a-pull-request/#merging-a-pull-request-on-github) help.  The author cannot be changed though.

I think we should release a 4.2.6, so a backport would be useful.

jralls:
OK, backport pushed.

PR #427, opened 2017-06-22
Clone event

apoettker:
Clone event

Bug #9604

Cloning Events (eg. to creating blueprints) can save a lot of time. Therefore I've picked up the issue #9604 and added some logic to the event handling. A screen shot now looks like this:

screenshot clone events 2017-06-22

IMHO the references to persons / families should preserve. Never less, I've coded a remove mechanism for references too. Unfortunately I've got an error out of the depth of GTK:

/usr/lib/python3/dist-packages/gi/overrides/GObject.py:530: 
Warning: /build/buildd/glib2.0-2.40.2/./gobject/gsignal.c:2580: 
instance '0x7efd1800dfa0' has no handler with id '10033'
  return func(*args, **kwargs)

This happen after accepting the clone when references are deleted and the system tries to disconnect the model (eventbackreflist.py, line 57 ff.). Don't know when and where the handler is lost, struggle to find it. A patch can be to supress the disconnection via a flag (here: 'if not self.changed'), but I'm afraid this fills the memory with garbage after a while.


prculley:
Interesting issue.  Because you are deleting a reference, you perform an update.  The update code destroys the model and creates it again.  In the process of destruction, the model.connect('row-inserted', self.update_label) in the BackRefList class __init__ code is disconnected for you.

When you later try to close the dialog, the _cleanup_local_connects runs and Gtk cannot find the already deleted connection.

This has not come up until now, because an reference could not be deleted in the editor.

The model's underlying data is created in the Gtk idle loop, so we really don't know in the backreflist code if we even have data.  So we really do need a connect to model 'row-inserted' so we see data later on.
I think the best answer is to move the model.connect to a new 'post_rebuild' method, which will get called after the model is created, but before it gets filled (because we have not gone idle yet).
<pre>diff --git a/gramps/gui/editors/displaytabs/backreflist.py b/gramps/gui/editors/displaytabs/backreflist.py
index 715fa75..1d0707c 100644
--- a/gramps/gui/editors/displaytabs/backreflist.py
+++ b/gramps/gui/editors/displaytabs/backreflist.py
@@ -66,7 +66,7 @@ class BackRefList(EmbeddedList):
         EmbeddedList.__init__(self, dbstate, uistate, track,
                               _('_References'), refmodel)
         self._callback = callback
-        self.connectid = self.model.connect('row-inserted', self.update_label)
+        #self.connectid = self.model.connect('row-inserted', self.update_label)
         self.track_ref_for_deletion("model")

     def update_label(self, *obj):
@@ -135,3 +135,11 @@ class BackRefList(EmbeddedList):
     def edit_button_clicked(self, obj):
         (reftype, ref) = self.find_node()
         edit_object(self.dbstate, self.uistate, reftype, ref)
+
+    def post_rebuild(self, prebuildpath):
+        """
+        Allow post rebuild embeddedlist specific handling.
+        @param prebuildpath: path selected before rebuild, None if none
+        @type prebuildpath: tree path
+        """
+        self.connectid = self.model.connect('row-inserted', self.update_label)</pre>
This seems to work for me with no ill effects, but you should test carefully since this will effect all the reference tabs, not just yours.

apoettker:
Integrated the diff, cleaned up and tested different (not all possible) scenarios. Work like a charm Paul!

Nick-Hall:
This PR is not a good idea.
* If a referenced person or family is being edited when the event is cloned then event references will be lost.  The same problem exists with the third-party census and form editors.
* Source-based data entry would provide a better soltuion.
* This needs to be discussed in conjunction with [GEPS 034](https://gramps-project.org/wiki/index.php?title=GEPS_034:_Improve_usability): Improve usability.

apoettker:
@Nick-Hall: 

- "... event references will be lost": Clone event duplicates an event with a new handle and a new ID. You've the opportunity to delete the references, but you don't have to - that's the default. So therefore all references are in place, visible and editable in the Person / Family editors. Maybe you can detail the problem you've seen a little further?
- "Source-based data ...": The same, maybe you can explain your thoughts a little bit further too?
- "[GEPS 034](https://gramps-project.org/wiki/index.php?title=GEPS_034:_Improve_usability): Improve usability": This runs in a deeper discussion regarding the future data entry way. 
PR 427 is a nice small extension to improve usability within the actual look and feel of Gramps. It's far away - was never intended - from the suggestions / remarks in [GEPS 034](https://gramps-project.org/wiki/index.php?title=GEPS_034:_Improve_usability).
I'm not sure how we can press all informations needed (see e.g. bug tracker isssue #9553) in a 800x600 resolution w/o additional popup screens - that's the mechanism we've today. I've seen several filed issues and mock ups reg. extended data edition in the bug tracker, but it looks that there is not too much appetite to code this in. Some toughs reg. future will be very helpful, cause I've several minor extensions like this in the queue, but all in the actual style.

Nick-Hall:
@apoettker The problem is that you are modifying a person or family, which may already be in the process of being edited, within the event editor.  Consider the following:
1. Edit a person and make a change without closing the editor.
2. Clone an event where the person is a primary participant.
3. Save the event.  The person has a new reference to the event.
4. Save the person.  The event reference list is overwritten losing the new reference.

apoettker:
A new version with interaction to Person / Family editors and ability to work w/o references. I've to include PRs

https://github.com/gramps-project/gramps/commit/7771fa063a5e9c21e9a54005117529517acb4ac0
https://github.com/gramps-project/gramps/commit/a1205009f4d6450a2b8a3487adf9cc7f2d70f68b
https://github.com/gramps-project/gramps/commit/48eb507f984353044bac8729ca7788bb47fa7d55

from Paul Culley due to former issues in our code (see issue in the bug tracker #10144). Above PRs are not merged into Master right now, therefore a lot of touched code pieces.

Nick-Hall:
Overwriting the event reference list may also cause problems.  What happens if the user had already edited an event reference?  We started to discuss this problem on the list in the [Application Level Locking](https://sourceforge.net/p/gramps/mailman/message/35663849/) thread.  I suggested that editors applied JSON patches as a potential solution.

I still think that source-based data entry would be a better approach.  Feature request [#8686](https://gramps-project.org/bugs/view.php?id=8686) has the right idea.  Other possibilities would be to generate events in the Form editor or from transcriptions with suitable mark-up.

apoettker:
@prculley: Is there a way to detect if the event reference is just being edited?

@Nick-Hall: Let's recapitulate: "Clone Event" is a small tool to generate a twin event w/o references. It was not the intention to change fundamentally the way of data submission ([GEPS 034](https://gramps-project.org/wiki/index.php?title=GEPS_034:_Improve_usability)). After the ReorderID tool trouble (which was thankfully fixed by Paul C.) due of the inadequateness of GTK on different platforms my fun on larger user interface changes is low. Also I see the Form mechanism based mainly on decided sources not as a alternative at this time. Source-based data entry is far bigger (and btw. more complex as well as fault-prone) as this tool is designed for.

Nick-Hall:
@apoettker The problem is that you update people and families from within the event editor which is not allowed because it can cause problems.  The Form addon does the same thing, which is why it has never been integrated into core code.

It may be helpful if you came up with some use cases that illustrate the problem you are trying to solve.  I seem to remember that there have been similar suggestions in the past.

An alternative would be to implement this as a third-party addon.

apoettker:
Use case: Marriage banns in Germany in the 16. - 20. century have been typically announced in two parishes (husbands and wife's parish). While importing data from other genealogists the place name in this event are combined in one string (unacceptable in Gramps). Now I've to correct the one place name, than double the event with all his tags, citations, attributes and notes and reference it to the object - a lot of work.

I'm not so familiar with consequences as a third-party addon implementation. Functionality should be easy to access and located locally in the working area where needed.

As an alternative would the Clone Event functionality in two variants be acceptable as:
a. Implement the functionality in the Event tab of the Person / Family editor, where exactly this reference after doubling an event is needed and all references are immediately visible? In this moment the event editor will not be activated, its a copying process in the background.
b. Implement the functionality in the Event list w/o references (to achieve a quick doubling from an blueprint event)? Its like an event addition, but here with default values.

Nick-Hall:
@apoettker Thanks.  Your use case differs from that described in the feature request.  I can see that cloning would be useful when splitting an imported event.

Both of your alternatives are acceptable because they avoid the locking problem.  I prefer (a) because it would allow you to keep one reference.  Opening the event reference editor with the clone would save time and also be consistent with the interface. You will need to remove your modifications to the editor reference tab.

Nick-Hall:
@apoettker You could just force push a new version to the same branch.

apoettker:
Thanks for the advise Nick.

 

My private GitHub was almost 450 commits behind Gramps/master, so I take the chance (before even your mail reached me) to step up to the actual level. I decided to implement both alternatives. Cause it was almost ready, I'll release alt. B (Clone Event w/o references) first, followed by alt. A (Embedded Clone Event).

 

KR

Alois.

 

Gesendet: Mittwoch, 18. April 2018 um 15:12 Uhr
Von: "Nick Hall" <notifications@github.com>
An: gramps-project/gramps <gramps@noreply.github.com>
Cc: "Alois Poettker" <alois.poettker@gmx.de>, Mention <mention@noreply.github.com>
Betreff: Re: [gramps-project/gramps] Clone event (#427)


@apoettker You could just force push a new version to the same branch.

prculley:
I think you may want to do a "git rebase -i upstream/master" on your local
branch with the PR.  And then use the editor that pops up (assuming you
have told git about your editor) to 'drop' those commits not directly
involved in your PR.  If you keep a local copy of 'master' and it is up to
date with out any commits of your own, then you can use that instead of
'upstream/master'.  I often use 'git gui' and then the
'Repository/visualize all branch history' menu command to get a graphical
idea of what is going on when I mess something up.  If you 'file/reload'
after making changes with git, you can see the results graphically.

Paul C.

emyoulation:
Perhaps this could be done with existing tools and a much simpler interface.

Say that you export a single Primary Object (Person) in an uncompressed .gramps format, then the XML also includes all the secondary objects. 

When you paste that XML into the Import Text gramplet, executing an import will Clone everything. If the Gramplet was a bit smarter and assumed any matching handles and IDs were to be shared not cloned, then new options are opened.

Say that a Context menu was added to all Edit Object dialogs that 'clones' by exporting that primary object to a (undocked) improved Import Text gramplet but also generates a new Handle & ID for the primary object (substituting the new handle throughout the XML), then a 'Import' of the Primary object would clone the primary objects and have shared secondary objects. 

Initially, it could be a Power User feature with no more interface than the Import Text gramplet currently has. The exception being that the Import button should spawn Object Editor dialogs for each cloned object. Then the user could tweak the attachments. It should NOT be a pre-populated new Object Editor but a Object Editor with an gramps_id and handle. This is because a committed object can be dragged from the header of its Object Editor to attach it to other objects. (Alternately, it COULD be a new object editor IFF a context menu allowed the Object to be committed and the dialog refreshed to have the Handle & ID.) 

Later, a list of objects interface might be overlaid on the XML listing, offering a 'clone' and 'Skip' checkbox options for all secondary objects. The clone button would toggle between a new handle/ID for the object and restoring the original handle/ID. The list could be flat, hierarchical by object type, or hierarchical by referenced object. 

A 'clone all'. 'omit all' could be offered at the hierarchical parent level.  

An example of use might be to start with a 'clone all' for the flat list, swap to the hierarchical by type and choose not to clone Places or Repo or Sources or Notes. Then swap to hierarchical by reference object and clone certain Notes.

Further, if the full XML text of the object was left available in the interface as a display option, the clone to the Import Text gramplet would offer a simplified workflow for moving an isolated object & secondary objects (via copying the XML to the OS clipboard) to another Tree.

PR #428, opened 2017-06-23
ReorderIds tool crashes with 'bytes' error on bsddb

prculley:
fixes #10098

Th tool used the old 'map' db interface, which had 'bytes' handles in it.  With recent change to 'str' handles, got a crash.

Changed tool to use `iter_???_handles` method instead.


prculley:
PR https://github.com/gramps-project/gramps/pull/428 should fix.

PR #430, opened 2017-06-27
Use generic open (mac) or xdg-open (everything else) instead of webbrowser...

jralls:
to open links.

Webbrowser is broken on macs with Python < 3.5.

Fixes #10105.

### DIFF
--- a/gramps/gui/display.py
+++ b/gramps/gui/display.py
@@ -25,7 +25,7 @@
 #-------------------------------------------------------------------------
 import os
 import webbrowser
-
+import sys
 #------------------------------------------------------------------------
 #
 # GRAMPS modules
@@ -33,7 +33,7 @@ import webbrowser
 #------------------------------------------------------------------------
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 from gramps.gen.const import URL_MANUAL_PAGE, URL_WIKISTRING
-from gramps.gen.constfunc import is_quartz
+from gramps.gen.constfunc import is_quartz, mac
 from gramps.gen.config import config
 from gramps.gui.utils import open_file_with_default_application as run_file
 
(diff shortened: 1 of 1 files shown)


Paul-Franklin:
open_with_default_applicatoin(norm_path)

Looks misspelled to me.

And as long as I'm commenting, any chance of shortening
the import line, just to make pylint happier?

Nick-Hall:
Is `xdg_open` available on Windows?

jralls:
> Is xdg_open available on Windows?
I have no idea, I relied on [the comment on the bug](https://gramps-project.org/bugs/view.php?id=10105#c52157) where you said that it was.

If it isn't then we're back to webbrowser with a special case for mac and python < 3.5.

Nick-Hall:
Sorry, I didn't mean to imply that either `open` or `xdg_open` are available in Windows.  My guess is that they are not.  This is probably why [os.startfile](https://docs.python.org/3/library/os.html#os.startfile) is provided by python.

jralls:
@Nick-Hall no further comments? Shall I merge & push?

gramps/gui/display.py
```diff
@@ -74,4 +74,9 @@ def display_url(link, uistate=None):
     """
     Open the specified URL in a browser.
     """
-    webbrowser.open_new_tab(link)
+    if (mac and sys.version_info.major == 3 and sys.version_info.minor < 5):
```
Nick-Hall: I think you need `mac()` rather than `mac` here.

Nick-Hall:
I see you decided against a new open_with_default_application function in the end.  Is there any benefit in checking the return value and presenting an error dialog if it fails?

jralls:
Yes, there's no point in ```open_with_default_application``` if Windows doesn't know how to do that. This is a pure workaround for broken Python on Mac.

Yes, quite right about mac(), fixed.

Nick-Hall:
Thanks.  This now looks ready to merge and push,

jralls:
Rebased and merged.  I cherry-picked it into gramps50 as well. It's not needed for Gramps.app that I build but others (MacPorts and Homebrew) might not be using a new enough Python.

PR #434, opened 2017-07-13
Fix collation issues

Nick-Hall:


jralls:
This doesn't allow for queries in different locales which would be necessary for reports, but perhaps it's more important to get everything working with the GUI locale first.

Nick-Hall:
Our existing BSDDB code doesn't use the report locale for sorting either.

dermoth:
There was some issues with this PR: not quoting the collation names in queries (which in PostgreSQL looks like `en_CA.UTF-8`) and the code to create collations was still needed (predefined ones on my database are named like `en_CA` and `en_CA.utf8`, and not all collations are present. Safest option is still to create them as needed.

Nick-Hall:
Updated to remove the character encoding from collation names as @jralls suggested. Now the same names can be used in sqlite and postgres.

@dermoth Do we still need to create extra collations in postgres?

dermoth:
Yes the collation is still needed in postgres. I can merge your changes into my PR and test if you like...

Nick-Hall:
We should create a separate PR for each change.  Please test this PR and then add an extra commit to it if necessary.

dermoth:
I have 3 commits in my PR all PostgreSQL fixes. I think they should go in as one, but I can submit them separately if you want.

From what I see you added one method and changed the double quotes to single... I already merged your previous work in my branch so it's easy for me to update it again.

jralls:
> Our existing BSDDB code doesn't use the report locale for sorting either.

True, but easily fixed. 

Replicating bugs in new code isn't exactly "best practice".

Nick-Hall:
The latest update fixes collation for translated reports.

@dermoth What happens if you try to create a collation for a locale that doesn't exist on the server?

dermoth:
Sorry I haven't had time to test so far...

For your question, my understanding is that you can create any valid libc collation and any accepted syntax works (ex. `en_CA`, `en_CA.utf8`, `en_CA.UTF-8`) and then refer to it by the given name.

You can also use pre-existing locales in the `pg_catalog` schema it you refer to them by name, but they have to be created by an admin beforehand. I'm not sure exactly what is installed automatically; I'm pretty sure I installed Debian with `C` and `en_CA.UTF-8` (the later as default) and installed `fr_FR.UTF-8` before configuring that cluster, and the locales I got in `pg_catalog` are `C`, `C.UTF-8`,  `POSIX`,  `en_CA`,  `en_CA.utf8` (both mapped to utf8 variant),  `fr_FR` and `fr_FR.utf8` (both mapped to utf8 variant) plus `ucs_basic`, an alias for `C`, and `default` which shows as blank but appears to end up being the default I configured, `fr_FR.UTF-8` (It's definitively not `C`/`POSIX`).

Finally, when creating locales, they always get created in the default schema (normally `postgresql` but can be set otherwise) but if the same name exists in pg_catalog only this one is listed (and I assume only that one is effective). OTOH to delete locales in the default schema where the same locale name exists in `pg_catalog`, you have to specify the schema.

Creating a locate that does not exists fails. So does using a local that isn't present either default or pg_catalog schema.

jralls:
Isn't collation creation limited to the locales which are installed on the server rather than any that are supported by libc? AFAIK the necessary information is in /share/locale/xx/LC_COLLATE, not somewhere in libc.

dermoth:
The only `LC_COLLATE` file I have on my system is `/usr/lib/locale/C.UTF-8/LC_COLLATE`.

Maybe the locale needs to be installed, but the point is that the best option appears to have Gramps create (on demand) every local needed by users in PostgreSQL; i.e. not use at all the ones in `pg_catalog`.

PR #435, opened 2017-07-14
Fix creating new PostgreSQL database

dermoth:
1. Use all parameters from settings.ini as keyword arguments for PostgreSQL (port was missing, and other can be useful such as `sslmode='verify-full'`)
2. Fix Collation on PostgreSQL (Name each collation to prevent users form clobbering each other's collation)

### DIFF
--- a/gramps/gen/utils/grampslocale.py
+++ b/gramps/gen/utils/grampslocale.py
@@ -935,6 +935,12 @@ class GrampsLocale:
                 return string
             return key
 
+    def get_collation(self):
+        """
+        Return the collation without any character encoding.
+        """
+        return self.collation.split('.')[0]
+
     def strcoll(self, string1, string2):
         """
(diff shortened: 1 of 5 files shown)


dermoth:
NB: this replaces PR #434

Nick-Hall:
Updated to remove the character encoding from collation names as @jralls suggested.  Now the same names can be used in sqlite and postgres.

@dermoth Do we still need to create extra collations in postgres?

Nick-Hall:
Merged parameter change in commit fd2af705e78d416995740360e4a3ef0db9ee65d7.

Nick-Hall:
Merged fix for creating new PostgreSQL database in commit 8f69201a70a2cad039d01028edd23fd3a739e50f.

Nick-Hall:
@dermoth I have added your latest changes to PR #434

PR #436, opened 2017-07-16
Add 'title' properties to more of schema

prculley:
You can argue that this is an enhancement, but I would really like to get it into Gramps50 branch.

Fixed the 'Source' and 'Citation' "srcattr_list" issues, and added 'title' properties to essentially everything that did not have it yet.  And only need two new translation strings...


Nick-Hall:
Thanks.  This is a bug rather than an enhancement.

I split the commit and changed the formatting to one property per line.  Two titles seemed wrong in the Date object so I changed them.  Note that the `sortval` is actually the serial date number, which can be used as a sort value, rather than a _Sort As_ field.

PR #438, opened 2017-07-19
Tidy up mocking code

Nick-Hall:


prculley:
@Nick-Hall I did not know you could get away with mocking the time.* functions directly.  I was under the impression that if some module imported one of these functions before the import tests were run (and the patch was applied), then the patch would not work, as the 'some module' would have a reference to the original time functions, not the mocked ones.  Which would make running the import tests very dependent on where they ended up in the test sequence.  Guess I did not understand how the mock actually works...

Nick-Hall:
@prculley It depends on the import used:  `import time` will require a different patch to `from time import ...`

Now I'm trying to work out why we don't patch `create_id()`.

prculley:
@Nick-Hall Back when I first started and Doug was pushing for tests to show my patches working, I did not know anything about mock.  So I actually modified the create_id() code to have a 'deterministic' mode, just a counter that always starts at 0.  It is reset with a function call at the beginning of the import tests (and some others where consistent results are important).  I also wrote a tiny plugin to set the mode and reset the count so it could be done via the Gui (detid).

prculley:
@Nick-Hall If you decide to change the create_id to something other than the current incrementing number, someone will have to recreate all the test data.  And will have to figure out a method for doing the initial creation via the Gui, since the DetId plugin will no longer work.  Perhaps DetId could use mock as well...

Nick-Hall:
@prculley The deterministic mode is only used for testing so it doesn't really belong in the core code.  Instead, mock objects can be created to replace real objects during testing.  This avoids the need to modify the core code to provide a test mode.  In this case, we can replace `create_id()` with a function that returns an incremental rather than a random id.

No test data should need to be changed.

Nick-Hall:
The first commit is probably worth merging, but I can't get the mock `create_id` function to work, and I don't want to spend any more time on it.

sam-m888:
> The first commit is probably worth merging

Tidy up mocking code https://github.com/gramps-project/gramps/pull/438/commits/95ff4417fd53baf1fa21b56672f65038d78d5e58

Not sure how to merge this and reject the second patch or should I just commit the first patch manually?

sam-m888:
First commit merged as https://github.com/gramps-project/gramps/commit/8ebb4d35fabe295729a5605677de87e352cee4eb to Gramps master.

Should this PR be left open?

sam-m888:
Looks like I didn't test well enough!  

Issue with the importprogen test file dates!

Nick-Hall:
Closing, because I couldn't get it to work properly.

PR #439, opened 2017-07-23
fix translation typo

schoonc:


Paul-Franklin:
Thank you for wanting to make gramps better!

In general, translation work is being concentrated on the
maintenance/gramps50 branch, as that is what will be released
next.  If this typo is in the ro.po file there, I suggest you email
our Russian translator @vantu5z about it, rather than filing
a pull request.

vantu5z:
@schoonc, thank you!
I fix this on the maintenance/gramps50 branch - 6d4e4ba6cca911f924b1ad619efb5aa951e2965d.

PR #440, opened 2017-07-23
Add `pycharm` dir to .gitignore

schoonc:


sam-m888:
[PyCharm 2017.2 Help/Using Git integration: Exclude files from version control](https://www.jetbrains.com/help/pycharm/using-git-integration.html#ignore-files)

PR #441, opened 2017-07-28
Locale Pro-Gen importer II & ReorderIds II

apoettker:
(Translated locale into German (Import Pro-Gen),
corrected minor translation bugs (ReorderIds)).

### DIFF
--- a/gramps/plugins/importer/importprogen.glade
+++ b/gramps/plugins/importer/importprogen.glade
@@ -122,7 +122,7 @@
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
                     <property name="tooltip_text" translatable="yes">Source reference text
-(Text &amp; import Filename).</property>
+(Text &amp; import Filename)</property>
                     <property name="hexpand">True</property>
                     <signal name="key-release-event" handler="on_import_entry_keyrelease" swapped="no"/>
                   </object>
@@ -152,7 +152,7 @@
                     <property name="can_focus">False</property>
                     <property name="valign">center</property>
                     <property name="margin_bottom">10</property>
-                    <property name="label" translatable="yes">Attribut</property>
+                    <property name="label" translatable="yes">Attribute</property>
                     <attributes>
                       <attribute name="stretch" value="semi-condensed"/>
                     </attributes>
@@ -169,7 +169,7 @@
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
                     <property name="tooltip_text" translatable="yes">Source attribute text
-(Text, import Filename &amp; (System-)Date).</property>
+(Text, import Filename &amp; (System-)Date)</property>
                     <property name="opacity">0.9882352941176471</property>
                     <property name="valign">start</property>
                     <property name="margin_bottom">10</property>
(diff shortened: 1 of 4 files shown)


Paul-Franklin:
Another approach would be to just add your new strings
to the existing files.  As opposed to regenerating them and
thus ending up with many hundreds of changed lines, most
of which (I assume) have nothing to do with your additions.

If you only add or change a few lines, github will show the
changes in a diff, whereas now somebody who wants to see
them has to download your p.r., apply it, then wade through
a massive "git diff" typeout, searching for real changes, the
"wheat" not the "chaff".

apoettker:
With approx. 50+ line translations & changes in between different files I don't like to fiddle around manually, so I trust in the power of Git.

Paul-Franklin:
I'm not talking about what git can do.  I'm talking about how
you have prepared this pull request, how it is structured.

Currently, this changes four files, all at once.  Your changes
to the two glade files do not concern me.  I am talking about
the changes you have made to po/de.po and to po/gramps.pot.

I don't recall you replying to my suggestion (in another place)
that you coordinate any possible changes to the German
translation with our current German translator.  Mirko has been
the German translator for gramps for ten years and I don't
want him taken by surprise.  I don't want him unhappy by any
changes you want to make -- or the fact that you are changing
de.po at all.  In my experience he has always been responsive
and changed things when necessary.  Don't you think that
will happen this time too?  Have you contacted him at all?

You are also making changes in the gramps.pot file which
in my opinion don't belong in the gramps.pot file at all.  For
instance you have an eleven-line comment (in code, in a
method in your Pro-Gen importer,) which starts "Pro-Gen
has a text field for the date" and then goes on for ten more
lines.  That comment doesn't belong in gramps.pot at all,
but it being added since you have placed it immediately
adjacent to the translated word _("Unknown").  Why don't
you place those eleven lines in the docstring, immediately
above them?  That seems to me to be a better place, and
then maybe they won't end up in gramps.pot.

You have flagged things in importprogen.glade which
don't make a lot of sense to me (although I haven't
tried running the Pro-Gen importer).  I don't see why
you have put a newline ("\n") in the middle of six
word strings for instance.  Nor why you think single
words like "REFN" need translating at all.

In fact, the more I think about it, the more I wonder
whether it might not make sense for the Pro-Gen
importer and exporter to be removed from "core"
gramps and placed in the third-party addon repository,
since it seems to me to primarily be a third-party
author-supported piece of software, which is not
logically a piece of "core" gramps, the way the
GEDCOM importer and exporter are.  That way
you could do whatever you want to them, as
often as you wanted.  The gramps50 version
of the addon could have these new features
for instance, and you wouldn't have to wait
until 5.1.0 is released for users to have them.
Among other things, that would allow German
translations to be made as often as you wanted,
in the addon repository.

More importantly, the Pro-Gen strings would not
end up in gramps.pot, where they would need to
be translated into every gramps-supported locale,
even ones highly unlikely to need them.  Arabic,
Chinese, Japanese, Turkish, Hebrew come to mind.

But really, isn't the Pro-Gen importer (and maybe
exporter) only likely to be used in the Netherlands
and in German-speaking countries?  Wouldn't it
make a lot more sense to have the Pro-Gen things
as third-party plugins, where they could have only
Dutch and German translation files?  That seems
to me to be a better solution than forcing all our
translators to do extra work, which in all probability
will never be needed in their language.

Paul-Franklin:
I still am in favor of this p.r. not being committed, and its
code turned into an addon instead -- for the reasons I stated
above as well as on gramps-devel.  It's senseless to make all
our translators translate strings while will probably only be
needed in Germany, Austria or the Netherlands.

apoettker:
As requested.

PR #442, opened 2017-08-07
Show main participants on narrative website

SNoiraud:
Resolves #7740

### DIFF
--- a/gramps/plugins/webreport/narrativeweb.py
+++ b/gramps/plugins/webreport/narrativeweb.py
@@ -12,7 +12,7 @@
 # Copyright (C) 2008-2011  Rob G. Healey <robhealey1@gmail.com>
 # Copyright (C) 2010       Doug Blank <doug.blank@gmail.com>
 # Copyright (C) 2010       Jakim Friant
-# Copyright (C) 2010-2016  Serge Noiraud
+# Copyright (C) 2010-2017  Serge Noiraud
 # Copyright (C) 2011       Tim G L Lyons
 # Copyright (C) 2013       Benny Malengier
 # Copyright (C) 2016       Allen Crider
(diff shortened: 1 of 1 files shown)


sebschub:
Sorry for the silence in the bug report. The links on the Event page are wrong. They all miss the `.html`, some are "more wrong". For example, when looking at the example database from the bug report, the first two people of the marriage event with Gramps ID E0000 have the same link.

Furthermore, I would suggest having an output on the Event page like

> Name (Role) [Gramps ID]

This style resembles the style on the Person page.

sebschub:
Added some non-primary participants to the example data base in PR https://github.com/gramps-project/gramps/pull/443.

SNoiraud:
> Added some non-primary participants to the example data base in PR #443.
thanks for this.

> Furthermore, I would suggest having an output on the Event page like
>    Name (Role) [Gramps ID]
I agree.

> They all miss the .html.
Thanks for that. In reality, this is the page extension (.htm, .html, .php, ...)

sebschub:
The links of the two primary persons of any marriage are still wrong.
If it is not too much work, I would prefer that the primary person(s) are always first in the list. This is for example not the case for the Death of Garner, Anderson.
Thanks a lot!

SNoiraud:
> The links of the two primary persons of any marriage are still wrong.

Done with last commit. I put back the marriage event which was missing.

> I would prefer that the primary person(s) are always first in the list

This will be very difficult to do :
I have only a list of handles. For each handle, I need to find the person, and then found the role for the associated event. If you have some persons in your database, I think this is not a problem, but if you have several thousand of events like me, it will have a high cost (cpu+disk).
For that, I need to modify the logic of how we store all events, persons, ...

SNoiraud:
I'll drop this FR.
I'm actually reorganizing the narrative web. I'll include this FR (see #0010151).
It modify all the structures to include this possibility.
I can also easily sort the references by role an have the Primary roles the first.

SNoiraud:
Follow this with next PR : #445

PR #445, opened 2017-08-10
Reorganization of the narrative web report.

SNoiraud:
#010151 : split the narrativeweb in classes.
#007740 : Show main participants on narrative website

Resolves  #010151, #07740


Paul-Franklin:
I don't claim to know anything abut Travis but I observe that
GitHub is reporting that "The Travis CI build failed" (for this
p.r.), so I wonder if you can duplicate it locally with a "python3
setup.py test" ?

prculley:
That would work, but take a long time.  It would be faster to just run the
reports_test.py directly which is the subset that appears to be failing.

GRAMPS_RESOURCES=. python3 -m nose gramps\plugins\test\test_reports.py

Should do it assuming you have nose installed.

GRAMPS_RESOURCES=. python3 gramps\plugins\test\test_reports.py
might work as well.

Paul C.

Nick-Hall:
@SNoiraud Excellent work! This should make future maintenance much easier.  Merging this will not cause any problems, but I am tempted to make an exception and allow it into _gramps50_.  What do you think?

@prculley Unit tests should be capable of being run with `nose`.  See the [Running Unit Tests](https://gramps-project.org/wiki/index.php?title=Unit_Test_Quickstart#Running_Unit_Tests) section in the wiki.  So you should be able to use the following:

`GRAMPS_RESOURCES=. python -m unittest gramps.plugins.test.test_reports`

SNoiraud:
> @SNoiraud Excellent work! This should make future maintenance much easier. Merging this will not
> cause any problems, but I am tempted to make an exception and allow it into gramps50. What do you think?

This is not a problem for me to add this in 5.0. It will be easier for me if we have a bug in 5.0 : I should patch 5.0 and master if it's applied only in master.

SNoiraud:
I am not sure the branch base move was a good solution.
What is the best way to solves this ?
I get errors when trying to resolves conflicts from command line.

Nick-Hall:
Rebased.

PR #448, opened 2017-08-11
Gramps 'Views' are not named in the window header

SNoiraud:
Resolves #001675


sam-m888:
Works , this change will help with users identifying which view they are using 👍 

titles-for-views-gramps50

sam-m888:
Thank you

PR #449, opened 2017-08-11
gvrelgraph: connect arrows to side of nodes instead of anywhere

azrdev:
This sets the headport and tailport attributes to all edges in the
graphviz file. Their default value makes the arrows between persons
and/or family nodes attach their ends to the respective node at any
position. With this change, the position facing the node on the other
side of the arrow is chosen.

The screenshot shows a portion of a big graph with splines=true before the change (left) and after (right).

arrows

### DIFF
--- a/gramps/gen/plug/docgen/graphdoc.py
+++ b/gramps/gen/plug/docgen/graphdoc.py
@@ -64,12 +64,19 @@ log = logging.getLogger(".graphdoc")
 #-------------------------------------------------------------------------------
 _FONTS = [ { 'name' : _("Default"),                   'value' : ""          },
            { 'name' : _("PostScript / Helvetica"),    'value' : "Helvetica" },
-           { 'name' : _("TrueType / FreeSans"),       'value' : "FreeSans"  }  ]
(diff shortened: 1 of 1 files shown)


Paul-Franklin:
Please separate this into two commits, one with your "pylint"
changes (such as changing the number of spaces a line has),
and one with the "real" changes you are proposing.  Thanks.

Also, your p.r. title says "gvrelgraph" but you are proposing
making a change to the "backend" and so all Graphviz reports
will be affected.

Paul-Franklin:
Also, if you decide it belongs in the "backend," I suggest
you make it an option so users can choose.

Do a recursive grep for "DocOptions" to see what I mean.

The ASCII and SVG backends currently each have one,
but I assume the Graphviz backend could be done also.

Paul-Franklin:
And if you are so motivated, you can think about
https://gramps-project.org/bugs/view.php?id=8319
but certainly that should be an option, not wired-in.

azrdev:
@Paul-Franklin thx for the review, I'll take care of your remarks the next days

PR #450, opened 2017-08-12
Correct issue with exception on database change in Gramplets

prculley:
Fixes #10147 and #10154
Exception was occurring on  'no-database'  and 'database-changed' signals under some conditions.
An earlier fix attempted to disconnect signals from old db during a db change to prevent multiple Gramplet updates after multiple db changes.  Problem was occurring in the disconnect when the db had already wiped out all its signals.

The fix is not very elegant, since the problem is that Gramps is not very consistent in the state at the time the signals arrive.  Sometimes when _gramplet gets the 'no-database' signal, the prior db is open, sometimes closed  The current db as can also be opened or closed.  When the prior one is closed the exception will occur.

### DIFF
--- a/gramps/gen/plug/_gramplet.py
+++ b/gramps/gen/plug/_gramplet.py
@@ -399,8 +399,7 @@ class Gramplet:
         self.update()
 
     def _no_db(self):
-        if self.dbstate.db.is_open():
-            self.disconnect_all()  # clear the old signals
+        self.disconnect_all()  # clear the old signals
 
     def get_option_widget(self, label):
         """
--- a/gramps/gen/utils/callback.py
+++ b/gramps/gen/utils/callback.py
@@ -311,8 +311,7 @@ class Callback:
             for key in keymap:
                 self.__callback_map[signal_name].remove(key)
             self.__callback_map[signal_name] = None
-        self.__callback_map = None
-        del self.__callback_map
+        self.__callback_map = {}
 
     def emit(self, signal_name, args=tuple()):
         """


Paul-Franklin:
It may not be very elegant but it solved my problem.  Since
I started using this patch I have never crashed the way I did
in 10147.  I think this should be committed.  Thank you!

Nick-Hall:
I don't like the bare except.  Can we check if the signal object still exists or perhaps prevent an error when attempting to disconnect a signal that has already been disconnected?

prculley:
I could change the bare except to an AttributeError.  That would not be quite so 'bare'.

The problem is that I cannot figure out how to tell if the error is happening locally (within the _gramplet.py) before it happens.  The object I have (a History or Db or ... ) is still valid, although sometimes the Db is closed.  Even closed, I think the object will be valid until the last reference to it is gone.  And _gramplet is holding some references until I disconnect.

I considered looking into the object to see if it had the hidden '_Callback__callback_map' attribute, but I already know you don't like layering violations like this.

And the signal has not exactly "already been disconnected", it is more that the early db.close destroyed the db's '_Callback__callback_map', killing off his record of it.  

I did look into the morass of code that results in the 'no-database' and 'database-changed' signals.  There are a lot of paths to consider, and I think that the structuring of some of this is not exactly clear and obvious, being tied up with the DummyDb and all.
For example, the particular path that caused the issue being patched has 4 opportunities to close the db (places where code asks is_open, then close if not).  And emits the 'no-database' and 'database-changed' signals twice in succession.

Ideally, we would guarantee that the the 'no-database' signal arrived _before_ closing the db.  And I would personally change the 'database-changed' signal and semantics to 'new_database', and require all users of these to monitor both and do the right thing.

But that is a lot more architectural change than I'm willing to do at the moment.  Especially for a bug that is holding up release 5.0.

Attached are my notes taken when I looked into some of the paths though the code.  I only covered the GUI paths, the CLI paths are not examined, and I'm not sure I got them all.  I was frankly afraid that messing with any of that would create more bugs than it fixed.  And it seemed to me that this was not any kind of high performance or often used path, so any minor delays caused by the 'except' would not hurt.
[Gramps Notes on open&close.txt](https://github.com/gramps-project/gramps/files/1228373/Gramps.Notes.on.open.close.txt)

Nick-Hall:
> And the signal has not exactly "already been disconnected", it is more that the early db.close destroyed the db's '_Callback__callback_map', killing off his record of it.

Closing the database calls the `disconnect_all` method.  This destroys the callback map at [line 315](https://github.com/gramps-project/gramps/blob/master/gramps/gen/utils/callback.py#L315).  A subsequent call to disconnect a signal will result in an error.  Does setting the callback map to an empty dictionary instead of destroying it fix the bug?

prculley:
@Nick-Hall Setting to an empty dict seems to correct the issue.  I have to wonder, @bmcage did a bit of work to clean up memory leaks back in 2010, including this.  Was he being belt & suspenders in deleting the  '_Callback__callback_map'?  Or is it really enough to just set it to an empty dict?
I understand Python is _supposed_ to automatically clean up objects when they are no longer in use, but when I see code like this I have to wonder...

Paul-Franklin:
I did not do any rigorous testing, and of course I don't
know what any other effects of the proposed changed
will be, but I can say that when I changed callback.py's
disconnect_all method's last two lines to be just the
suggested "self.__callback_map = {}" that I no longer
had the typical error typeout I reported in 10147, when
I did the usual things which caused the problem.

Nick-Hall:
@prculley Under normal circumstances, there is no need to delete private instance variables. When implementing data structures with cyclic references, such as doubly-linked lists, you may have to be more careful.  We did have garbage collection problems with Gtk2 objects in the past.  I'm not sure if there is still a problem with Gtk3.

PR #452, opened 2017-08-24
Reorder IDs tool fixes

prculley:
Fixes bugs [#10175](https://gramps-project.org/bugs/view.php?id=10175), [#10176](https://gramps-project.org/bugs/view.php?id=10176), [#10177](https://gramps-project.org/bugs/view.php?id=10177)

Original idea was to have the tool look for GetGov type IDs and give user option of keeping them or renumbering them.  But found a variety of issues while debugging and testing.

In fixing the bugs, I got the enhancement mixed up with the fixes quite a bit; it would be very difficult to separate the two.  I hope this still qualifies for gramps50 status.

In order to support the GOV IDs, I put some restrictions on the ID used by Gramps for automatic numbering:
* Can have a prefix and/or suffix, but must be less than three characters (not digits)
* There must be between 3 and 9 digits of number
* number must use leading zero option
* Must be no leading or trailing spaces  (this has caused trouble in some reports)

An attempt to enter an ID format violating these rules, or not having proper "%" and "d" characters, is ignored, and the standard format is used.  This shows up in the GUI when you leave the format field with 'enter' key or select another field.

When the tool runs via the 'OK' button, if it encounters an ID violating one of the above rules, it asks the user if he wants to replace the ID.  That dialog has a "Use this answer for the rest of the items" checkbox.

I changed the dialog to be narrower, so it would work on smaller screens.
I changed the dialog so tabbing between fields it would not stop in fields that were not editable.

There were a couple of errors inside a try/except loop that prevented the code from working as intended in trying to preserve a previous format when renumbering.  It always took the exception.

On an editorial note; I have to wonder if this tool has more functionality than the normal user needs.  I still haven't figured out what all the top header buttons do.  It certainly needs to have the help wiki updated to explain all the functionality.

### DIFF
--- a/gramps/plugins/tool/reorderids.py
+++ b/gramps/plugins/tool/reorderids.py
@@ -66,27 +66,37 @@ from gramps.gui.widgets import MonitoredCheckbox, MonitoredEntry
 WIKI_HELP_PAGE = '%s_-_Tools' % URL_MANUAL_PAGE
 WIKI_HELP_SEC = _('manual|Reorder_Gramps_ID')
 
+PREFIXES = {'person': 'i', 'family': 'f', 'event': 'e', 'place': 'p',
+            'source': 's', 'citation': 'c', 'repository': 'r',
+            'media': 'o', 'note': 'n'}
 #-------------------------------------------------------------------------
 #
 # Actual tool
 #
 #-------------------------------------------------------------------------
 
-# gets the number specified in a format string, eg: %04d returns '04'
-_parseformat = re.compile('.*%(\d+)[^\d]+')
+# gets the prefix, number, suffix specified in a format string, eg:
+# P%04dX returns 'P', '04', 'X'  It has to have the integer format with at
+# least 3 digits to pass.
+_parseformat = re.compile(r'(^[^\d]*)%(0[3-9])d([^\d]*$)')
+
 
 class ReorderEntry(object):
     """ Class for internal values for every primary object """
 
-    def __init__(self, object_fmt, quant_id, nextgramps_id):
+    def __init__(self, object_fmt, quant_id, nextgramps_id, obj):
+        self.object = obj
         self.width_fmt = 4
-        self.object_fmt, self.object_prefix = '', ''
+        self.object_fmt, self.object_prefix, self.object_suffix = '', '', ''
         self.quant_id, self.actual_id, self.actgramps_id = 0, 0, '0'
         self.calc_id(object_fmt, quant_id)
-        self.stored_fmt = object_fmt
+        self.stored_fmt = self.object_fmt
         self.stored_prefix = self.object_prefix
+        self.stored_suffix = self.object_suffix
 
-        self.number_id = int(nextgramps_id.split(self.object_prefix, 1)[1])
+        self.number_id = int(nextgramps_id[len(self.object_prefix):
+                                           (len(nextgramps_id) -
+                                            len(self.object_suffix))])
         self.step_id = 1
         self.active_obj, self.change_obj, self.keep_obj = True, False, False
 
@@ -101,7 +111,7 @@ class ReorderEntry(object):
     def set_fmt(self, object_fmt):
         """ sets primary object format """
         if object_fmt:
-            self.calc_id(object_fmt, self.quant_id)
+            self.calc_id(object_fmt.strip(), self.quant_id)
 
     def get_fmt(self):
         """ gets primary object format """
@@ -121,19 +131,26 @@ class ReorderEntry(object):
 
     def __ret_gid(self, actual):
         """ return Gramps ID in correct format """
-        return '%s%s' % \
-               (self.object_prefix, str(actual).zfill(self.width_fmt))
+        return '%s%s%s' % \
+               (self.object_prefix, str(actual).zfill(self.width_fmt),
+                self.object_suffix)
 
     def calc_id(self, object_fmt, quant_id):
-        """ calculates identifier prefix, format & actual value """
-        self.object_prefix = object_fmt.split('%', 1)[0] if '%' in object_fmt else ''
+        """ calculates identifier prefix, suffix, format & actual value.
+        Requires a valid format or returns the default instead """
         self.object_fmt, self.quant_id = object_fmt, quant_id
 
         # Default values, ID counting starts with zero!
-        self.width_fmt, self.actual_id = 4, 0
         formatmatch = _parseformat.match(object_fmt)
         if formatmatch:
-            self.width_fmt = int(formatmatch.groups()[0])
+            self.object_prefix = formatmatch.groups()[0]
+            self.width_fmt = int(formatmatch.groups()[1])
+            self.object_suffix = formatmatch.groups()[2]
+        else:  # not a legal format string, use default
+            self.object_prefix = PREFIXES[self.object].upper()
+            self.width_fmt = 4
+            self.object_suffix = ''
+            self.object_fmt = PREFIXES[self.object].upper() + "%04d"
         self.actgramps_id = self.__ret_gid(self.actual_id)
 
     def zero_id(self):
@@ -246,7 +263,7 @@ class ReorderIds(tool.BatchTool, ManagedWindow, UpdateCallback):
 
             object_fmt, quant_id, next_id = self.prim_methods[prim_obj]
 
-            obj_value = ReorderEntry(object_fmt, quant_id, next_id)
+            obj_value = ReorderEntry(object_fmt, quant_id, next_id, prim_obj)
             self.obj_values[prim_obj] = obj_value
 
         if self.uistate:
@@ -315,8 +332,10 @@ class ReorderIds(tool.BatchTool, ManagedWindow, UpdateCallback):
         for obj_entry in ['start', 'step', 'keep']:
             obj = self.top.get_object('%s_%s' % (obj_name, obj_entry))
             if obj_entry == 'keep':
-                if self.obj_values[obj_name].stored_prefix != \
-                   self.obj_values[obj_name].object_prefix:
+                if (self.obj_values[obj_name].stored_prefix !=
+                        self.obj_values[obj_name].object_prefix and
+                        self.obj_values[obj_name].stored_suffix !=
+                        self.obj_values[obj_name].object_suffix):
                     self.keep_entries[obj_name].set_val(False)
                 else:
                     obj.set_active(obj_state)
@@ -383,7 +402,7 @@ class ReorderIds(tool.BatchTool, ManagedWindow, UpdateCallback):
         obj_name = Gtk.Buildable.get_name(widget).split('_', 1)[0]
         obj_fmt = self.format_entries[obj_name].get_val()
 
-        self.format_entries[obj_name].set_val(obj_fmt)
+        self.format_entries[obj_name].set_text(obj_fmt)
         self.start_entries[obj_name].update()
 
         return False
@@ -479,17 +498,17 @@ class ReorderIds(tool.BatchTool, ManagedWindow, UpdateCallback):
         self.menu = self.top.get_object("popup_menu")
 
         # ok, let's see what we've done
+        self.window.resize(700, 410)
         self.show()
 
     def _update(self):
         """ store changed objects formats in DB """
 
         update = False
-        for prim_obj, tmp in self.xobjects:
+        for prim_obj, dummy in self.xobjects:
             obj_value = self.obj_values[prim_obj]
             if obj_value.object_fmt != obj_value.stored_fmt:
-                prefix = obj_value.object_prefix.lower()
-                constant = 'preferences.%sprefix' % prefix
+                constant = 'preferences.%sprefix' % PREFIXES[prim_obj]
                 config.set(constant, obj_value.object_fmt)
                 update = True
 
@@ -542,7 +561,8 @@ class ReorderIds(tool.BatchTool, ManagedWindow, UpdateCallback):
             print('\nDone.')
 
     # finds integer portion in a GrampsID
-    _findint = re.compile('^[^\d]*(\d+)[^\d]*')
+    _findint = re.compile('^[^\d]*(\d+)[^\d]*$')
+    _prob_id = re.compile(' *[^\d]*(\d+){3,9}[^\d]*')
     def _reorder(self, prim_obj):
         """ reorders all selected objects with a (new) style, start & step """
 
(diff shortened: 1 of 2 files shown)


sam-m888:
@Nick-Hall Seeing as PR #445 Reorganization of the narrative web report made it into Gramp 5.0, any opposition to this one?

apoettker:
Works in overall as expected. The new restrictions are excellent! The 'Keep' check box protects - or even not - w/o any user request the existing IDs against renumbering. The new algorithm with dialogue which checks additional if an existing ID should be overwritten is therefore double programming. But it's a really good reminder for user with larger databases (10k+), which typically have some numbering exceptions.

I've noticed that  - sometimes - not all element of an objects are renumbered after running the tool. Unfortunately can't reproduce this in an deterministic way. Seen it on Gramps 4.2.5 and on 5.1.0-e5d5cfb (the actual one).

Editorial comment: The module have more functionality as it looks on the first view. The top header buttons are little helper to specify quick a pattern for renumbering, no additional functionality in the background. No question that the help wiki needs to be updated.

Nick-Hall:
@prculley Yes.  This can go into gramps50.

@apoettker Is PR #441 still valid?

gramps/plugins/tool/reorderids.py
```diff
@@ -572,31 +591,60 @@ def _reorder(self, prim_obj):
...
+                label2 = xml.get_object('mainlabel')
+                label2.set_text(_("Do you want to replace %s?" % act_id))
+                top.set_transient_for(self.progress._ProgressMeter__dialog)
+                self.progress._ProgressMeter__dialog.set_modal(False)
+                top.show()
+                response = top.run()
+                do_same = apply_to_rest.get_active()
+                top.destroy()
+                self.progress._ProgressMeter__dialog.set_modal(True)
+                if response != 3:  # 3 = Yes
```
Nick-Hall: Why not use `Gtk.ResponseType.YES` here?

prculley:
@Nick-Hall I tried to use one of the existing dialogs.  One of them had 3 buttons (I needed 2) and the other had the "Use this answer for the rest of the items" hard coded to save the answer to a config setting, which I did not want to share with whatever used that dialog.

If you would prefer, I could go back and try to generalize the standard dialog some more.  I chose to just add my own rather than spread changes around too much.

As you might have guessed, I copied/edited the standard dialog, which already used the '3' response, so I guess my answer to "Why not use Gtk.ResponseType.YES here?", is expediency.

Let me know if you want me to go a different direction.

Nick-Hall:
@prculley OK.  If using an existing dialog isn't easy, then creating a new one is probably best for now.

Please use a predefined Gtk [ResponseType](https://lazka.github.io/pgi-docs/Gtk-3.0/enums.html#Gtk.ResponseType) where possible though.  This allows Gtk to arrange dialog buttons in an OS preferred sequence.  It may possibly also be helpful for themes and RTL layout.  Unfortunately, the Glade editor didn't provide a list so you have to look up the values yourself (-8 and -9 in this case).

apoettker:
@prculley I've observed that starting Gramps (5.0.0.alpha 2, Ubuntu 14.04 LTS) in  a console  and running the ReorderID Tool creates tons of 
"(Gramps.py:3509): Gtk-WARNING **: Unknown property: GtkEntry.max-width-chars" warnings.

@Nick-Hall PR 411 is a combination of German translation for the ReorderID tool and the Pro-Gen importer (which will be an addon-tool over time)

prculley:
@Nick-Hall Changed to gtk.ResponseType as requested.
@apoettker My Glade did not warn me that set_max_width_chars was not available for old Gtk version (<3.12).  I have no idea how to set the dialog Gtk.Entry to smaller width without that feature, so I imagine that if you have a small screen you might have issues on these older Gtk versions.

I removed the offending code from the glade file, and put it into the python code, with a check for the Gtk version.  That way it won't generate errors on older Gtk versions.

reid

gramps/plugins/tool/reorderids.py
```diff
@@ -475,6 +495,16 @@ def _display(self):
                                     self.obj_values[prim_obj].get_keep,
                                     readonly=True)
 
+            if gtkok:
+                # the max_width_chars function was introduced in Gtk 3.12
+                # for earlier versions the dialog might be too large on small
+                # screens since the default Gtk.Entry seems to be about 20
+                # characters or so.
+                obj_actual.set_max_width_chars(10)
```
Nick-Hall: Use `set_width_chars`, which is available in Gtk 3.10, instead.

prculley: @Nick-Hall Unfortunately, `set_width_chars` doesn't seem to work by itself, at least on my Windows Gtk 3.18.9.  The original Glade file had (and still has) that, and the Gtk.Entry fields were all very wide (20 chars or so despite `set_width_chars` being set to 10, or 6.  Leading to the ~1000px wide dialog.  I tried quite a few combinations before I found the stackexchange reference to doing it this way.

Nick-Hall: @prculley Interesting.  It works in the filter sidebars to keep them small.

prculley: @Nick-Hall Attached is the original, unmodified ReorderIds screen from gramps50, running on my kubuntu Linux VM.  Note that even with the `set_width_chars` values that you want, the fields are large.  I also note that my sidebar is very wide as well (that's as small as it gets), I never realized that it wasn't supposed to be that way until I compared with an image from our wiki (also attached), but that is another bug.
So I don't think the issue is Windows specific, but probably is Gtk specific.

So I think the code should stay as it is; at least for me it works well on both Linux and Windows.

redilinux
gramps_sidebaricons_filter

Nick-Hall: What version of Kubuntu are you using?  I'm using Ubuntu 14.10.

When you just want to set the default width and not the maximum width, using `set_max_width_chars` is not a good idea.  It is probably acceptable here, because all the fields are small.  However, it will cause problems if you try to use the same fix the the gramplet sidebar where users will want to increase the size on large screens.

The window for this tool also cannot be resized, which is inconsistent with other tools in Gramps.

prculley: I definitely understand that setting max_width_chars means that the field (and thus the whole dialog) becomes locked to no larger than that.  With the Gtk (bug/feature) that the Gtk.Entry doesn't go smaller than 20 chars (or so), despite set_width_chars values smaller than that, the dialog size becomes completely locked.  If Gtk ever allows a smaller Gtk.Entry, the dialog will still shrink.
But for this application I cannot see why anyone would want to resize it anyway.  I deliberately made the fields almost twice the size needed so there should not be an issue even for different size default fonts.

I'm open to suggestions on how to shrink these fields without using max_width_chars; I spent 1/2 a day searching the internet and trying various possibilities already, none worked.  And I really think that having a dialog ~1000px wide is not a good idea.

My Linux KUbuntu system is 15.10 with Gtk 3.16.7, my Windows system (where the original dialog looks exactly the same) has Gtk 3.18.9  which is where Josip has frozen the AIO Gtk for now.

Nick-Hall: I don't see this issue in Gtk 3.10 or Gtk 3.22 - so it looks like a bug that occurred and was fixed between these two versions.

Our policy is that no dialog should be larger than 800x600, so you need the fix for broken Gtk versions.  However, I would not apply the fix to versions that don't need it and I would also include a link to the Gtk bug report in the comment.  We tend to remove fixes when the minimum required Gtk version reaches the point that they are no longer required.

Dialog windows should be made resizable for consistency, even if there isn't much point for them to be made bigger than their default size.

Paul-Franklin:
I could point out dialogs which are bigger than 800x600, for
instance the person editor, and that doesn't even include dialogs
which aren't in English but which are in other translations, but
since I can't fix such things and the people who can don't seem
to have enough time to do so, mentioning them is a waste of time.

Nick-Hall:
@Paul-Franklin Opening a bug report with a list of large dialogs would be helpful.

prculley:
@Nick-Hall So your insistence on making it resizable caused me to discover that the original author had set it to NOT be resizable.  Once I fixed that I realized that the dialog could be made smaller, down to the sum of the `set_width_chars` values (and extra margins etc.), if done manually.  Seems that the 'natural' size of these Gtk.Entry things is the maximum size, so the dialog wouldn't shrink without a deliberate size change (manually or otherwise) unless I also used `set_max_width_chars`.  Found out some of this by reading the Gtk source code.
I used a ended up removing the `set_max_width_chars` and adding a window.resize(600, 410) (line 501) in the code to force it smaller than the 'natural' size.
I hope this meets with your approval...

Paul-Franklin:
@Nick-Hall

I am reluctant to spend the time to research things like
that since (as I said) I don't think think they'll be fixed.

How about I wait until the one I mentioned gets fixed,
and then I'll start researching others?

And what specifically do you (or somebody) plan to
do about the ones which are too wide in some other
language?  I can remember mentioning that to some
other senior developer and his response was along
the lines of "the translator should use shorter words."

Nick-Hall:
@prculley Thanks for the effort you are putting into getting this right.  I think we are almost there!  However, I just got an error when using a test database where the id prefix was the object name - Person001, Person002 etc...  I solved this by editing the regular expressions.  Then I had to set the Actual, Format and Start columns to expand horizontally and fill to see the contents easily.  Why do we set a size limit for the prefix and suffix?

Whilst looking at the code, I noticed that `eval` is used where `getattr` would be better.  Also setting `self.dbstate = dbstate.db` is confusing.  However, this can be fixed in a separate PR.

Nick-Hall:
@Paul-Franklin The person editor dialog in _gramps50_ is 797x485.  Translators should provide the best translation they can.  Unfortunately, we cannot assume that our developers speak any language other than English.

Paul-Franklin:
@Nick-Hall

Well, I have no way to know what database you are using,
nor what the Person in your DB has in their dialog.

Let me be more specific, when I load the example.gramps
DB and then open the person editor on the default person,
I0044,  the person-editor dialog I get cannot be shrunk less
than 855 wide (nor less than 518 high, which I admit is fine),
at least on my machine.

Note that that is without showing all the tabs, only with the
indicator to the right of the rightmost tab indicating there
are more, if the user a) notices, and b) knows what it means.

And in fact you can't even fully see the top ("birth") line
in the date below, in the tabbed area, without dragging
that dialog taller/higher.

Note also that once such a too-large dialog is opened,
that when it is closed that too-large size then becomes
the size that the next person-editor dialog opens at.
The user will have to then manually drag it smaller, if
they are able to.

So I stand by my statement.

If you want another, the Display Name Editor, up
in Edit=>Preferences=>Display, cannot be made
narrower than 820, at least for me, on my machine.

Paul-Franklin:
"I think we are almost there."

Maybe this needs a "Work in Progress" flag?

prculley:
@Nick-Hall 

> I just got an error when using a test database where the id prefix was the object name - Person001, Person002 etc... I solved this by editing the regular expressions. Then I had to set the Actual, Format and Start columns to expand horizontally and fill to see the contents easily. Why do we set a size limit for the prefix and suffix?

I'm not sure what 'error' you got, but the changes you made to the regular expressions completely destroy my original purpose in doing this PR.  As I explained in the very first post of this PR, I put restrictions on the length of prefix and suffix in order to make it possible to differentiate between normal Gramps IDs and GOV IDs.  I had looked at several hundred GOV IDs and noted patterns of usage (not guaranteed because I did not look at them all).  I noted that they usually had a group of text characters, then some numbers and finally more characters.  Or something like 'object_123456' or 'adm_123456'.  Excepting the two special cases I spotted, the prefix text group was usually 6 or more characters, the number group was usually 2 digits, and the suffix was usually 2 text chars. 
I chose a 3 char prefix limit for Gramps ID to avoid conflicting with the GOV IDs I found.
Allowing the prefix to be open ended like you did means that GOV IDs stored in the Gramps ID will be overwritten during a renumber.  My original code used the RE to spot potential GOV IDs and then gave the user the option of overwriting.

Nick-Hall:
@prculley Rather than looking for GOV IDs, it would be better to look for the format of the old prefixes.  The user may be using another external data source for places.

Assuming a prefix of three characters is too restrictive.  Users are allowed to define any length of prefix in the preferences.

Users using the GOV database probably won't want to use this tool for places anyway.

prculley:
@Nick-Hall I'm not sure I agree.  If a user has imported an XML from someone else with a different prefix set, or possibly changed his mind and set different prefix values while creating objects, there could be quite a few different 'old' prefixes around in his data.  That's when the Reorder IDs is most useful.

I'd much rather restrict the choices of format in the preferences to match.  Or insist that GOV IDs or similar be marked in some way that makes them recognizable.  Which would not fix already present data, but...

I personally doubt that very many folks ever used anything but the default, and I doubt if the few who did used long prefixes.

Nick-Hall:
@prculley I take your point about imported data.  However, I don't think that someone using GOV IDs is going to want to re-order their place IDs.  They are more likely going to merge imported places with existing places, or assign new GOV IDs.

We shouldn't place new restrictions in the preferences unless we have no other option.  I don't know what prefixes are used, but we must assume that some users prefer prefixes greater than three characters.  If we did introduce new restrictions, then we would want users to be able to use this tool to change their data.

Putting the GOV ID into the Gramps ID field was only ever intended as a temporary measure.  It was convenient for the experimental getGOV gramplet.  For 5.1, I plan better core GOV integration and GEDCOM-L support.

prculley:
The tool is fine for changing the data, it doesn't totally prevent a user from changing IDs that look like GOV IDs or long prefixes, just provides the (new) warning dialog and option to keep them.

In many ways I would prefer to find a way to mark GOV IDs (or other types) in the Gramps ID location.  The current Reorder ID fix was a workaround, intended to deal with the most likely cases.  Maybe we should just put up a warning that non-standard Gramps IDs, like GOV IDs will be overwritten by the ReorderID tool?

As it is, the new dialog and associated checking code is useless, with the changed RE.

FYI, I have pretty much completed code for Geps043, which does a pretty good job of importing/exporting Gedcom L within the limits of Gramps current data model.  And their not quite complete spec.  Mostly working on some refinements and test cases to make sure everything works as expected.

Nick-Hall:
@prculley If the RE to detect GOV IDs is just used to present a warning, then that is acceptable to me.  However, it mustn't fail with an error when long prefixes are encountered.  I used the object type as a prefix for testing when I got the error.  We could give the user a warning about non-standard IDs, but is this necessary if you get the bug fixed?

We may choose to store the GOV ID differently in the future.  Remember that this is still a fairly new third-party plugin.

Excellent news abot GEPS043!  I look forward to seeing it, and I'm sure it will be a very popular addition to Gramps.

prculley:
@Nick-Hall you never mentioned how you created an error.  Were you trying to enter a long prefix into the Reorder ID tool like "PERSON%04d"?  This should have not been possible as the dialog should just reject it and re-display the default.
Or was there something in your data that contained IDs with long prefixes?  I didn't see any errors there in my testing, they were just detected as non-standard.

Nick-Hall:
@prculley My test data contained IDs with the object type as the prefix.  e.g. Person001, Repository123, Citation111.

prculley:
@Nick-Hall I've been unable to recreate any error with objects named as you suggest.  So I am doing something different than you are, as usual.  If you would take the time to try to recreate, and describe it better, I would appreciate it.  I am still using the original PR, before you changed the RE and glade files.

Nick-Hall:
After clicking "proceed" at the initial warning dialog, I immediatley get:
```
line 99, in __init__ len(self.object_suffix))])
ValueError: invalid literal for int() with base 10: 'erson033'
```

PR #454, opened 2017-08-26
fix Undo; crashes due to race in Gtk

prculley:
Some Redraws occur during a model clear when the row changed signal
is emmitted.  Model is only partially cleared at that point.
Under some conditions, some more of the model gets
cleared before the redraw completes, and redraw fails.
fixes #9932

### DIFF
--- a/gramps/gui/undohistory.py
+++ b/gramps/gui/undohistory.py
@@ -120,10 +120,11 @@ class UndoHistory(ManagedWindow):
         scrolled_window.add(self.tree)
         self.window.vbox.pack_start(scrolled_window, True, True, 0)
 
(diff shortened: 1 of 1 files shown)


sam-m888:
@prculley Do you have the "crash-test-2017-02-11.gpkg" file mentioned in comment [51235](https://gramps-project.org/bugs/view.php?id=9932#c51235) if so can it be attached to issue 9932 please?

prculley:
@sam-m888 I never made the test file he described, so don't have anything to attach.  I just did the steps in the GUI a few times until I saw the crash.  The reporter never replied with any feedback, so I don't know for sure that this fixed his issue, but it seemed to fix the crash for me.

PR #455, opened 2017-09-01
fix invalid export (and import) of Gedcom EMAIL records

prculley:
Gedcom spec 5.5 requires '@' in general text to be doubled '@@'
fixes #10167


sam-m888:
Thank you

PR #456, opened 2017-09-02
GraphViz arrow ports

azrdev:
GraphViz reports: connect arrows to side of nodes instead of anywhere

This sets the headport and tailport attributes to all edges in the
graphviz file. Their default value makes the arrows between persons
and/or family nodes attach their ends to the respective node at any
position. With this change, the position facing the node on the other
side of the arrow is chosen.

This PR supersedes #449

### DIFF
--- a/gramps/gen/plug/docgen/graphdoc.py
+++ b/gramps/gen/plug/docgen/graphdoc.py
@@ -191,6 +191,13 @@ class GVOptions:
         spline.set_help(_("How the lines between objects will be drawn."))
         menu.add_option(category, "spline", spline)
 
+        node_ports = BooleanOption(_("Align line ends"), True)
+        node_ports.set_help(_("Force connecting lines to attach to nodes on "
+                              "the side they're coming from. Otherwise, they "
+                              "take the shortest way and may attach anywhere "
+                              "to the node."))
+        menu.add_option(category, "node_ports", node_ports)
+
         # the page direction option only makes sense when the
         # number of horizontal and/or vertical pages is > 1,
         # so we need to remember these 3 controls for later
@@ -415,6 +422,7 @@ class GVDocBase(BaseDoc, GVDoc):
         self.vpages        = get_value('v_pages')
         self.usesubgraphs  = get_value('usesubgraphs')
         self.spline        = get_value('spline')
+        self.node_ports    = get_value('node_ports')
 
         paper_size          = paper_style.get_size()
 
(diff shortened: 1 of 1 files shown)


azrdev:
@Paul-Franklin you wrote in the original PR
> Also, if you decide it belongs in the "backend," I suggest you make it an option so users can choose. Do a recursive grep for "DocOptions" to see what I mean. The ASCII and SVG backends currently each have one, but I assume the Graphviz backend could be done also.

I looked at that, but it seems these (SvgDraw)DocOptions are not used by the GraphViz reports at all: the SVG option (background color) appears in "graphical reports" when selecting SVG, but not in Graph reports. I'm a bit lost how to introduce some `GraphVizDocOptions` here

Paul-Franklin:
Yes, I didn't mean for you to actually use those existing DocOptions
classes.  I meant for you to use them as a model to introduce a new
class.  However I forgot that Graphviz isn't using our DrawDoc and
TextDoc classes (or gpr.py files).

I will think about adding that capability to the Graphviz backend
(docgen options), but it might take a while since I have lots of
things on my list.  But also it might not really be needed.

But I still think it should be optional, so the user can choose.  There
are lots of other Graphviz options however.  Maybe you can find an
appropriate tab and add it that way?  If it really is just a yes/no, then
a BooleanOption would be fine.  If there are four choices, then
put them in an EnumeratedListOption.  Add it to add_menu_options
then use it where it is appropriate.

azrdev:
> Yes, I didn't mean for you to actually use those existing DocOptions
> classes. I meant for you to use them as a model to introduce a new
> class. 

That's how I understood you in the first place, however, I struggled to understand how to integrate my own model based on how they do it.

> I will think about adding that capability to the Graphviz backend
> (docgen options), but it might take a while since I have lots of
> things on my list. But also it might not really be needed.

That would be great, but no hurry.

> But I still think it should be optional, so the user can choose. There
> are lots of other Graphviz options however. Maybe you can find an
> appropriate tab and add it that way? If it really is just a yes/no, then
> a BooleanOption would be fine. […] Add it to add_menu_options
> then use it where it is appropriate.

I'll do that.

Thx for looking after this!

Paul-Franklin:
I have committed a slightly-modified version of your patch to master.

https://github.com/gramps-project/gramps/commit/f31caf1ea4df40d000e2669b2655e5de40a9e495

As far as I am concerned this p.r. may be closed now.

Thank you for helping to make gramps better.

azrdev:
thank you!

PR #457, opened 2017-09-12
StyledText.join method fails if the joint_text has StyledText Tags

prculley:
fixes #10192
StyledText join method is supposed to mimic the string.join method in that
`join_text.join([a_text, b_text])`
should work for all three items containing tags. But if the join_text contains StyledTextTags, they are lost.
Since the StyledText.replace uses the .join method internally, it also fails if the replacement text contains tags.

I include a new unittest that demonstrates the patched code working correctly. The styledtext_test can be used without the patch to show the failure.

### DIFF
--- a/gramps/gen/lib/styledtext.py
+++ b/gramps/gen/lib/styledtext.py
@@ -27,6 +27,7 @@
 # Gramps modules
 #
 #-------------------------------------------------------------------------
+from copy import copy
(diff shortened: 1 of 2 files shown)


sam-m888:
>  I include a new unittest that demonstrates the patched code working correctly. The styledtext_test can be used without the patch to show the failure.

Thank you that helped with testing 👍

PR #459, opened 2017-09-19
Fix make_unknown to create valid Place

prculley:
Fixes #10202
The gen.utils.unknown make_unknown function creates objects for Check and Repair or bad Gedcoms (at least) when one is referenced but doesn't exist.

When making an unknown Place, the function makes a place with a title but no name. This may not be strictly invalid, but does create issues with some other code.


Nick-Hall:
Yes.  All places should have a name.  The legacy title is optional because it can be generated, but there is no harm in leaving it set.

PR #460, opened 2017-09-20
Updated pt_PT translation

pagramps:
Please insult me if I made a mess


Paul-Franklin:
Did you really want this committed to "master" or is it for
"maintenance/gramps50" instead?

pagramps:
Why are you not calling me stupid?
Of course it's for gramps50. I don't know what I was thinking of. How do I "undo" this?

Paul-Franklin:
Well, I've never done it, so this is from memory and thus
could be wrong, but if I remember correctly you can click
on the "Edit" button in the upper right corner. and then
(somehow) change which branch it will be applied to.

Paul-Franklin:
Well, since I don't understand git or GitHub, I can't begin to understand
why there are now these additional comments, some from June, as well
as (possibly) additional commits.  Somebody else will have to tell you
what to do, but I'm guessing your pt_PT.po commit shouldn't be just
committed based on this pull request.  Sorry.

Paul-Franklin:
Maybe the line at the top is giving a clue, when it says "40 commits"
into "gramps-project:maintenance/gramps50" from "pagramps:master"?

Maybe it should somehow be made to say that it's coming from
"pagramps:maintenance/gramps50" instead?  Not that I know how
to do that.

pagramps:
Well, if you don't know, neither do I. To much trouble for a simple translation. Maybe someone can delete this mess, somehow. Sorry.

Paul-Franklin:
As I know how to "close" a pull request, I have done as you asked.

Thanks for trying to use git and GitHub, though.

Nick-Hall:
Create a new branch from _gramps50_, then cherry-pick your commits into it and create a new PR.  It is not a good idea to work in _master_.

PR #463, opened 2017-09-25
Fix ManagedWindow/GrampsWindowManager for subsidiary window close

prculley:
fixes #10194
The 10194 bug has Check&Repair being run with a bunch of Gramps editor windows open, particularly opened each from the previous.
What is happening is that the Check&Repair eventually gets to do *-rebuild signals from the db.  These in turn tell the various open editor windows to close.

Issues occur when the editor windows attempt to close in a different order than they were opened.  In this case the GrampsWindowManager is designed to close subsidiary windows first, on its own.  But the code to do so was very simple and did not actually 'close' the entire editor object, but rather just destroyed the Gtk Window and part of the GrampsWindowManager state.

A bit later during the rebuild, another signal arrives to the (partially closed) editor which then attempts to really close the Gtk window.  This created an error in the GrampsWindowManager which appears on the console ("Missing item from window manager").  A further issue is that the ManagedWindow code attempts to 'present' (make visible) the parent window during the close operation.  Since in several cases the parent window had already been (partially) closed (actually destroyed), the 'present' creates a long list of Gtk-CRITICAL errors and then puts back up an empty window for the user to see.

Note that the improper closing of subsidiary Windows (which can be done by the GUI) also creates Memory leaks as the objects are still present in memory, and since the Window Manager has destroyed is record of them, opening them again creates new copies.  A 'rebuild' signal occurring anytime after that will cause issues.

This patch for the GrampsWindowManager and ManagedWindow code makes the automatic closing of subsidiary windows actually run the object 'close' routine.  It seems to work OK in all the windows I've tested, but there are a LOT of dialogs and windows in all the Gramps code.  So this will require watching to see if it works properly with everything.

### DIFF
--- a/gramps/gui/managedwindow.py
+++ b/gramps/gui/managedwindow.py
@@ -196,10 +196,14 @@ class GrampsWindowManager:
 
     def close_item(self, item, *args):
         # Given an item, close its window and remove it's ID from the dict
+        if item.opened:
+            item.close()
         if item.window_id:
             del self.id2item[item.window_id]
+            item.window_id = None
         if item.get_window():
             item.get_window().destroy()
+            item.window = None
 
     def remove_item(self, track):
         # We need the whole gymnastics below because our item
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Thanks for tracking this down.  Did you intend the fix for the _gramps50_ branch?

prculley:
@Nick-Hall Yes I intended it for gramps50 branch; sigh.  I keep forgetting that even though I'm looking at gramps50 branch when I click the 'make a PR' button, it reverts to 'master' when it opens the PR page.  Usually I get a 'cannot merge' warning to let me know I forgot, but you had just merged master with gramps50 so that did not come up.  I used the 'Edit' button above to change to gramps50.

PR #464, opened 2017-09-25
Add login dialog and username/password command line options

Nick-Hall:


cz172638:
i miss following optional cmdline options / dialog fields:
--hostname (if omitted - localhost, ignored for sqlite)
--port (if omitted - default port for selected dbtype)
--dbtype (if omitted - sqlite)
--dbname (if omitted gramps.db under folder for appropriate gramps database)

Nick-Hall:
In the GUI, the database type, host and port are set in the "Family Tree" tab of the preferences dialog.  These settings are used when a new database is created in the family tree manager.

When creating a new database from the command line the name would be specified using the -C option.  Again, the database type, host and port will be taken from the configuration settings.  What flexibility do we really need when creating a new database from the command line?

PR #466, opened 2017-09-27
Narrative Web Feature requests :

SNoiraud:
Sort "Surname" web page by given name and birth date.
Surname list doesn't use default name format

Fixes #05315, #08067

### DIFF
--- a/gramps/plugins/webreport/basepage.py
+++ b/gramps/plugins/webreport/basepage.py
@@ -163,6 +163,18 @@ class BasePage: # pylint: disable=C1001
         name = _nd.display(person)
         return (name, person.get_gramps_id())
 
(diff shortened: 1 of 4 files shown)


SNoiraud:
@Nick-Hall @prculley 

This PR didn't close the two associated bug report!
Fixes #05315, #08067 doesn't work

Did I make an error somewhere ?

prculley:
Is it possible that the 'automatic close' doesn't work for Feature requests?

Nick-Hall:
We got a webhook timeout. I re-sent the message, but the bug tracker didn't resolve the issue, so I linked the changesets manually.

I'm not sure what the problem is.

PR #469, opened 2017-09-30
Geps043a - Improving GEDCOM support for places

prculley:
This is the preliminary implementation of GEPS043, improved GEDCOM support for places.
It is documented in detail in https://gramps-project.org/wiki/index.php?title=GEPS_043:_Improving_GEDCOM_support_for_Places#Details

Note: this was developed under gramps50 branch as that was generally most up to date.  It can be moved to master branch if desired for release purposes.
Note: there are three commits, the second one is NOT stand-alone, it MUST be used after the first.

I have put in a lot of time testing, first with all our standard test files, then with crafted test files, and with a variety of GEDCOM files found around the internet, in particular the MUSTER_GEDCOM_UTF-8.ged from the [Gedcom L](https://gramps-project.org/bugs/view.php?id=688) group.
In every case, I manually compared the Gedcom data with the Gramps data to make sure that the code performed as expected.
An end result is that using nose, with the coverage option, almost all of the newly written or modified code has been tested by the included tests.  There are a few paths not taken for conditions for which I could not devise tests.  The overall import coverage is above 92% and export coverage is above 93%, in both cases the missing elements are mostly due to code unrelated to this PR.


daleathan:
@prculley Should this be changed to the master branch?

sam-m888:
Included as part of https://github.com/gramps-project/gramps/pull/809

prculley:
Since this is now included in the GEPS045 work, I will close this.  It would not be nearly as elegant without the GEPS045 changes.

PR #471, opened 2017-10-01
Fix 'Welcome' Gramplet and StyledTextEditor for Link handling

prculley:
Issue [#10179](https://gramps-project.org/bugs/view.php?id=10179)

This fixes the StyledTextEditor so that it actually responds only to ctrl-click on a Link to open it.  Before, it would also respond to shift-click, caps-lock-click etc.

To resolve part of #10179 "Welcome Gramplet doesn't respond to links", the styled text editor was also changed to respond to a left-click on a link immediately, if the editor was not editable.

I also added to the tool-tip for the link, when it was editable "Ctrl-Click to follow link" as that is not the usual way to open links.  This is similar to Microsoft Office apps or LibreOffice Writer.

Note, there is a pending discussion on MAC usage of the CONTROL key for this and lots of other places in Gramps; this PR does not address this as I think it should be a separate commit.

### DIFF
--- a/gramps/gui/widgets/styledtexteditor.py
+++ b/gramps/gui/widgets/styledtexteditor.py
@@ -50,6 +50,7 @@ from gi.repository import Pango
 #
 #-------------------------------------------------------------------------
 from gramps.gen.lib import StyledTextTagType
+from gramps.gen.constfunc import mac
 from .styledtextbuffer import (ALLOWED_STYLES,
                                           MATCH_START, MATCH_END,
                                           MATCH_FLAVOR, MATCH_STRING,
@@ -319,8 +320,9 @@ class StyledTextEditor(Gtk.TextView):
             if url.startswith("gramps://"):
                 obj_class, prop, value = url[9:].split("/")
                 display = simple_access.display(obj_class, prop, value) or url
-        return display + \
-            ("\nCtrl-Click to follow link" if self.get_editable() else '')
+        return display + ((_("\nCommand-Click to follow link") if mac() else
+                           _("\nCtrl-Click to follow link"))
+                          if self.get_editable() else '')
 
     def on_button_release_event(self, widget, event):
         """
--- a/gramps/gui/widgets/styledtexteditor.py
+++ b/gramps/gui/widgets/styledtexteditor.py
@@ -319,7 +319,8 @@ class StyledTextEditor(Gtk.TextView):
             if url.startswith("gramps://"):
                 obj_class, prop, value = url[9:].split("/")
                 display = simple_access.display(obj_class, prop, value) or url
-        return display
+        return display + \
+            ("\nCtrl-Click to follow link" if self.get_editable() else '')
 
     def on_button_release_event(self, widget, event):
         """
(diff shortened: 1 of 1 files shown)


Paul-Franklin:
Does "\nCtrl-Click to follow link" get translated in some
non-obvious fashion, elsewhere?

prculley:
@Paul-Franklin Oops; forgot the '_()'.  Will fix shortly.

gramps/gui/widgets/styledtexteditor.py
```diff
@@ -319,7 +319,8 @@ def make_tooltip_from_link(self, link_tag):
             if url.startswith("gramps://"):
                 obj_class, prop, value = url[9:].split("/")
                 display = simple_access.display(obj_class, prop, value) or url
-        return display
+        return display + \
```
Nick-Hall: Please use implied line continuation inside parentheses instead of a backslash.  See the [Maximum Line Length](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) section in the PEP8 style guide.

gramps/gui/widgets/styledtexteditor.py
```diff
@@ -319,7 +319,8 @@ def make_tooltip_from_link(self, link_tag):
...
+            (_("\nCtrl-Click to follow link") if self.get_editable() else '')
```
Nick-Hall: This may be Command-click on a Mac.  I see that @jralls has opened PR #474 to address similar issues.

prculley: I was hoping he would support either control-click or Command-click; that way anyone who was used to the way Gramps is now could continue to do the same.  And I would not have to add an 'if mac()'...

gramps/gui/widgets/styledtexteditor.py
prculley: I guess the Eclipse/Pydev PEP8 checker doesn't check for that.  Usually if I just get rid of the warning flag, it passes pylint etc.

gramps/gui/widgets/styledtexteditor.py
jralls: PR #474 does indeed make it "command-click" on a Mac. The control-click behavior never worked on Macs because is_right_click() processed the event first and opened the context menu.

prculley:
The tooltip is upgraded to say "Command-Click to follow link" if on a Mac.  And the Python now follows the PEP8.
This should be committed with or after PR #474 to avoid confusion to users.

Nick-Hall:
PR #474 was partially reverted in commit 0f0a9c941bb6ad74ae0eee4cd052e0621dacca3d.

prculley:
@Nick-Hall Merge conflict fixed and rebased.

Nick-Hall:
Partial fix in commit 2bc4139c65295810588e2028c8b1a46833c44f16.

lordemannd:
@prculley I think this solves most of the issues in #0010179
My guess is that the strangeness regarding the insert point is beyond Gramp's control.

Nick-Hall:
Added tooltip in commit be11466b35fa759f9a9686ddf783741c0da6fc5e.

PR #472, opened 2017-10-05
 Fix relationship Graph so Unicode chars on Multiple pages works

prculley:
This PR fixes the long standing bug that makes Relationship Graph (and other Graphs) incorrectly show non-Latin-1 characters if run with multiple pages.

Fixes: #9783, #9359, #8080, #6782, #6108 and possibly others and is related to
Bugs: #6431, #4164, #1485, #2964, #2790

Graphvis 'dot' command can properly convert utf-8 non-Latin-1 characters to Postscript if used with the Cairo renderer option.  But it is not capable of putting out multi-page Postscript of PDF with that option enabled.

If more that one page is requested with Cairo enabled, 'dot' puts out a single large page.
The current Gramps code deals with this by disabling Cairo when multiple pages are desired, but then the users see bad utf-8 characters in their PDFs.

There have been various attempts to fix this over the years, the best of which was done by user 'embar' in https://gramps-project.org/bugs/view.php?id=6431.  For whatever reason, his code submission was not committed.

I took his idea, cleaned it up for Python3, and optimized it to run with fewer calls to Ghostscript.

The basic idea was to let 'dot' output the large Postscript page, and then use Ghostscript to slice it into pages.  I spent a bunch of time getting this to work correctly.

I also added a small bit of overlap between pages so that if a user tries to 'cut & paste' the pages together (literally, with scissors and tape), he will be able to do so more easily as his cuts need not be pixel accurate.

Tested on:
  Windows with Graphvis 2.41, Ghostscript 9.16
  Linux with   Graphvis 2.38, Ghostscript 9.16
  Mac Sierra   Graphvis 2.40, Ghostscript 9.21

Also did a pylint in a separate commit, 9.62 rating.

I changed the way Ghostscript is located.  This takes effect for non-Windows systems if Ghostscript is not found on the normal PATH.  It is intended to make it possible for Mac OS X systems locate the Ghostscript executable, where Gramps is passed a PATH that does not include elements added by the terminal shell.  The additional search paths are  /bin, /usr/bin, /usr/local/bin, /opt/bin and /opt/local/bin .
This is implemented in a new function "where_is" which is similar to the Linux "whereis -b file" command.

### DIFF
--- a/gramps/gen/plug/docgen/graphdoc.py
+++ b/gramps/gen/plug/docgen/graphdoc.py
@@ -25,7 +25,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
-
+""" Graphviz adapter for Graphs """
 #-------------------------------------------------------------------------
 #
 # Standard Python modules
@@ -36,19 +36,18 @@ import os
 from io import BytesIO
 import tempfile
 from subprocess import Popen, PIPE
-import sys
 
-#-------------------------------------------------------------------------------
+#-------------------------------------------------------------------------
 #
 # Gramps modules
 #
-#-------------------------------------------------------------------------------
+#-------------------------------------------------------------------------
 from ...const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
 from ...utils.file import search_for
 from . import BaseDoc
 from ..menu import NumberOption, TextOption, EnumeratedListOption, \
-                          BooleanOption
+    BooleanOption
 from ...constfunc import win
 
 #-------------------------------------------------------------------------
@@ -57,41 +56,41 @@ from ...constfunc import win
 #
 #-------------------------------------------------------------------------
 import logging
-log = logging.getLogger(".graphdoc")
+LOG = logging.getLogger(".graphdoc")
 
-#-------------------------------------------------------------------------------
+#-------------------------------------------------------------------------
 #
 # Private Constants
 #
-#-------------------------------------------------------------------------------
-_FONTS = [ { 'name' : _("Default"),                   'value' : ""          },
-           { 'name' : _("PostScript / Helvetica"),    'value' : "Helvetica" },
-           { 'name' : _("TrueType / FreeSans"),       'value' : "FreeSans"  }  ]
+#-------------------------------------------------------------------------
+_FONTS = [{'name' : _("Default"), 'value' : ""},
+          {'name' : _("PostScript / Helvetica"), 'value' : "Helvetica"},
+          {'name' : _("TrueType / FreeSans"), 'value' : "FreeSans"}]
 
-_RANKDIR = [ { 'name' : _("Vertical (↓)"),      'value' : "TB" },
-             { 'name' : _("Vertical (↑)"),      'value' : "BT" },
-             { 'name' : _("Horizontal (→)"),    'value' : "LR" },
-             { 'name' : _("Horizontal (←)"),    'value' : "RL" } ]
+_RANKDIR = [{'name' : _("Vertical (↓)"), 'value'   : "TB"},
+            {'name' : _("Vertical (↑)"), 'value'   : "BT"},
+            {'name' : _("Horizontal (→)"), 'value' : "LR"},
+            {'name' : _("Horizontal (←)"), 'value' : "RL"}]
 
-_PAGEDIR = [ { 'name' : _("Bottom, left"),                  'value' :"BL" },
-             { 'name' : _("Bottom, right"),                 'value' :"BR" },
-             { 'name' : _("Top, left"),                     'value' :"TL" },
-             { 'name' : _("Top, Right"),                    'value' :"TR" },
-             { 'name' : _("Right, bottom"),                 'value' :"RB" },
-             { 'name' : _("Right, top"),                    'value' :"RT" },
-             { 'name' : _("Left, bottom"),                  'value' :"LB" },
-             { 'name' : _("Left, top"),                     'value' :"LT" } ]
+_PAGEDIR = [{'name' : _("Bottom, left"), 'value'  : "BL"},
+            {'name' : _("Bottom, right"), 'value' : "BR"},
+            {'name' : _("Top, left"), 'value'     : "TL"},
+            {'name' : _("Top, Right"), 'value'    : "TR"},
+            {'name' : _("Right, bottom"), 'value' : "RB"},
+            {'name' : _("Right, top"), 'value'    : "RT"},
+            {'name' : _("Left, bottom"), 'value'  : "LB"},
+            {'name' : _("Left, top"), 'value'     : "LT"}]
 
-_RATIO = [ { 'name' : _("Compress to minimal size"),    'value': "compress" },
-           { 'name' : _("Fill the given area"),         'value': "fill"     },
-           { 'name' : _("Expand uniformly"),            'value': "expand"   } ]
+_RATIO = [{'name' : _("Compress to minimal size"), 'value': "compress"},
+          {'name' : _("Fill the given area"), 'value': "fill"},
+          {'name' : _("Expand uniformly"), 'value': "expand"}]
 
-_NOTELOC = [ { 'name' : _("Top"),    'value' : "t" },
-             { 'name' : _("Bottom"), 'value' : "b" }]
+_NOTELOC = [{'name' : _("Top"), 'value' : "t"},
+            {'name' : _("Bottom"), 'value' : "b"}]
 
-_SPLINE = [ { 'name' : _("Straight"), 'value' : "false" },
-            { 'name' : _("Curved"), 'value' : "true", },
-            { 'name' : _("Orthogonal"), 'value' : 'ortho'} ]
+_SPLINE = [{'name' : _("Straight"), 'value' : "false"},
+           {'name' : _("Curved"), 'value' : "true", },
+           {'name' : _("Orthogonal"), 'value' : 'ortho'}]
 
 if win():
     _DOT_FOUND = search_for("dot.exe")
@@ -110,20 +109,21 @@ else:
     else:
         _GS_CMD = ""
 
-#-------------------------------------------------------------------------------
+
+#------------------------------------------------------------------------------
 #
 # GVOptions
 #
-#-------------------------------------------------------------------------------
+#------------------------------------------------------------------------------
 class GVOptions:
     """
     Defines all of the controls necessary
     to configure the graph reports.
     """
     def __init__(self):
-        self.h_pages    = None
-        self.v_pages    = None
-        self.page_dir   = None
+        self.h_pages = None
+        self.v_pages = None
+        self.page_dir = None
         self.dpi = None
 
     def add_menu_options(self, menu):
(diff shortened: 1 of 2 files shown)


Paul-Franklin:
Thank you for doing this pull request!

Some nits.  It's "Graphviz" not "Graphvis".  Also, if we are
going to be using Ghostscript more heavily, I'd suggest that
it be "strongly recommended" in README.md, as Graphviz
presently is.  (I don't see Ghostscript at all, there.)  Speaking
of that, do you or John have a solution for Mac users, so
gramps does not have to be run from the command line?

Also, if possible I'd like to see the copyright section up top
give credit to the user who did the original code, unless you
think you have changed it beyond all recognition by him.  As
well as your name, of course.

Paul-Franklin:
Also, does this affect your Edit=>Preferences fontconfig
choice for Windows users (make it obsolete?)?

prculley:
This doesn't add any additional need for Ghostscript beyond what was done
before.  It just uses it a bit more; instead of just to convert to pdf, it
is used to split and convert.  If Ghostscript is not installed, then users
just don't have multi-page pdf for graphs, they can still do single page.

John R. would have to add Ghostscript to his 'build' for MAC to (like Josip
does) to use it without a user install (and then starting Gramps from
command line).  Would make the '.deb' file that much larger.  If you think
we should make that a standard part of the Gramps build/dependencies, then
lets make it so for Linux as well.  I suggest you take this issue to the
developers list.  If we do this, then the README can be updated accordingly.

I'm not particularly concerned on seeming my name added to every bit of
code I touch (quite a few modules by now).  I did add my name to the
author/contributors list that pops up when you do help/about and push the
button.  Should show up in next build.  As for 'embar' who had the original
idea, I cannot find out his actual name, just his bugtracker id. So it
would be hard to add him as a contributor.

Also, does this affect your Edit=>Preferences fontconfig
> choice for Windows users (make it obsolete?)?
>

That issue is completely different.  It has to do with finding an
appropriate font with the utf-8 chars that cause the trouble.  Most fonts
don't contain all the Unicode characters (over 100,000 at the moment).  The
standard Gtk Windows font adaption code doesn't always do a good job of
finding a font with the character to print; the 'alternate fonthandler'
written for Linux but usable on Windows seems to do a better job although
it also makes Gramps itself look a bit different since it chooses a
different font for the defaults.  It also seems to have some performance
effects during install and if a font is added or changed on the Windows
system.

Paul C.

Paul-Franklin:
If you have installed gramps from its dmg on your Mac VM
(congratulations on being able to do that, by the way,) I
would have guessed you'd be able to see that John has
put Ghostscript into it for quite a few years (IIRC).  I don't
know the details of course, like where he puts it.  Josip
used to (and maybe still does) put all his executables into
the same directory (like Ghostscript), then made sure
that directory was in the Windows PATH.  But you and
John have said that Apple does things differently.

I just think that since the whole purpose of this pull
request is to make something work which has never
worked before, that you (or John) should make sure
it works on all three platforms we support.  That seems
like the approach we should take with every bug we fix.
As I understand it (I could be wrong, and I hope I am,)
this pull request won't help our Mac users, unless they
start gramps on Terminal (or some equivalent), so that
seems wrong to me.

The fact that it uses Ghostscript "a bit more" is exactly
why it should be "strongly recommended" in the README.
IMHO.

As far as "embar" you'll see his full name in the XML
tree he attached to the bug where you found his patch.
If you then search the Internet you'll probably be able
to find his email.  I did, back in 2013 when I wrote him.
If you don't want to do that, you can recursively search
every file in gramps for either his first name or his last
name and you'll see his email where he contributed to
gramps back in 2013.

Yes, I understand that you don't feel like adding your
name up at the copyright area is important.  I think
Nick feels the same way, since he doesn't do it and
doesn't tell new developers to do it either,  But I was
trained by an earlier administrator who felt differently
so I feel differently.  I've seen you add your name if
you add new code, like in new modules, so I think
you'll agree that "embar" should get credit (a 2013
copyright line) for the new code he added.  It does
not seem fair to me to leave him out.

Nick-Hall:
> Yes, I understand that you don't feel like adding your
name up at the copyright area is important. I think
Nick feels the same way, since he doesn't do it and
doesn't tell new developers to do it either

I tend to only add a copyright if I make a significant change to a file.

> As far as "embar" you'll see his full name in the XML

In the bug tracker his details are:  Mindaugas <embar@super.lt>

Paul-Franklin:
@Nick-Hall

"I tend to only add a copyright if I make a significant change to a file."

Yes, that's what we all do.

But everybody makes their own determination as to what
is "significant" and yours is drastically different than mine.

As an example, I remember when you refactored the book
report, splitting its code into two (or three) modules,
and changing things to make it work that way.  Yet you
never added your name (IIRC).  I guess that wasn't a
"significant" change as far as you were concerned.

The only times I can recall you (or Paul) adding your
name is when either of you create a brand-new module,
with totally-new code.  That seems extreme to me.

But I don't want to belabor the point.  Clearly neither of
you is going to change (nor will I).  But I still think that
credit should be given to Mindaugas.

And I still think that Ghostscript should be in README.
The fact that most (all?) linux packagers (as well as our
two) already do it, doesn't take away from the fact that
if we need it then we should say we need it.

Nick-Hall:
> As an example, I remember when you refactored the book
report, splitting its code into two (or three) modules,
and changing things to make it work that way. Yet you
never added your name (IIRC). I guess that wasn't a
"significant" change as far as you were concerned.

I probably just forgot.

> The only times I can recall you (or Paul) adding your
name is when either of you create a brand-new module,
with totally-new code. That seems extreme to me.

I added a copyright line when I tidied up the new database backend code.

Nick-Hall:
> And I still think that Ghostscript should be in README.

Please raise this on the list as Paul suggested.  I have no objections.

prculley:
I've added my name and Mindaugas Baranauskas to the header of the file.  Fixed Graphviz spelling.

I just looked at the Mac release of Gramps-Intel-5.0.0-alpha3-2.dmg that John put together, no Ghostscript.  Maybe he used to do it, but not now.  Have no idea how much that would add to the size of the release.

So Windows is the only platform that comes ready to run that includes Ghostscript.  Mac and Linux users have to add it themselves, and then Mac users need to start from Terminal, as far as I can tell.

Paul-Franklin:
"So Windows is the only platform that comes ready to run that includes Ghostscript. Mac and Linux users have to add it themselves, and then Mac users need to start from Terminal, as far as I can tell."

Yes, I was wrong.  The Mac bundle has never contained Ghostscript.
Sorry.  I made a wrong assumption.

But I believe there are Apple programs which manipulate PostScript,
so if possible I would like this pull request modified to use them for Mac
users (once it's discovered what they are and how to use them).  I do
not think it likely that Ghostscript will be added to the Mac bundle.  At
least in the near future (like before 5.0.0 is released).

I would argue that if Ghostscript is required (for instance to make
multi-page reports using non-English scripts, or whatever), and it
is not present, that the user be told that, somehow, preferably in a
way which works for both GUI and CLI.  Ideally she would be told
what to do to get it, or told about some workaround (like an Apple
PostScript manipulator, if it is not going to get coded into this p.r.,
then massaging the PostScript manually to multi-page it).

The fact that this pull request, as it is at the moment, works for
Windows users (since Josip bundled in both Graphviz and also
Ghostscript into his AIO), but not for linux users or Mac users
unless they happen to have both installed (and installed in the
correct place for Mac users, not in /usr/local), makes me feel
that this p.r. is not ideal yet.  (That may be one reason why
enbar's original patch was never added to gramps, as you
wondered about, but I am only guessing.)

If you don't want to change it to make it better, perhaps even
doing a rewrite so that both Graphviz and Ghostscript were
invoked via their Python-interface (library, or whatever it is),
as opposed to running them in the shell, then at the very least
I'd like to see explicit warning messages given to the user,
saying that what they want to do is not possible at the moment.

prculley:
No warning is necessary.  If Ghostscript is not available, then the PDF (Ghostscript) option is not available.  The PDF (Graphvis) option remains available and works, although the multiple page options are greyed out.  So it is just not possible for a user to make a multi-page PDF out of a relationship (or other graph).

I am not aware of any method of using Ghostscript other than the shell command.  The same is true for Graphvis 'dot' command.  I note that John includes the 'dot' in the Mac .dmg bundle which is why we can make the single page PDF, SVG, JPEG etc..

As I understand it, Apples Applications use some 'Quark' APIs to print things, including printing to PDF.  We would have to create some sort of interface to those APIs (which would be MAC specific code) to fix this 'right'.  Even then, I have no clue how to 'poster print', I'm not sure the APIs would do this for us.

The Apple message boards have a couple of ways to use standard apps (poorly) to print very large images across multiple pages, but they mostly recommend using specific purchased applications to do it well.

As I am a confirmed Windows user, and have always hated Apple and their 'special' ways of doing things, I am very unlikely to undertake any large project like this when there are a couple of easy ways to get past it.  Perhaps we can get John to make two install modules, one with Ghostscript, one without.  Or just put some additional instructions in the wiki on how to deal with it.

prculley:
@Paul-Franklin After some testing and further correspondence with John Ralls I determined that the 'security' issue John was concerned about, doesn't seem to apply in this case.  So I just updated this with a config option that allows access to Ghostscript, even with normal Finder/Launchpad startup of Gramps.
User still will have to install Ghostscript himself.  The default is set so the most likely version of Ghostscript will just work after install.  If this PR is accepted I'll update the wiki with instructions on how to deal with this and the best version of Ghostscript to use.

If you want Ghostscript added to the Mac bundle and the dependencies for Gramps, I suggest you take it to the developers list.

P.S.  The python interface to Graphvis we are already using.  The python interface to Ghostscript is a thin layer that still requires Ghostscript to be installed in the right place to be found.  If I understand the OS X security documents correctly, it would NOT work the way the Ghostscript is currently installed, since it is an unsigned library installed in the wrong place.  So I don't think this would help, might even make the problem harder.

P.P.S.  I'm willing to do minor adjustments to Gramps code to deal with platform issues, what I said was that I'm not willing to make major efforts to deal with a single platform, particularly one I don't understand or use, and have reasons to dislike.  You won't see me doing major work for Windows (or Linux) that isn't useful to the other platforms either.

Nick-Hall:
I don't like this last commit.  I suggest writing a utility similer to the Unix `which` command that returns a path.  This could initially call `search_for` but then check the usual binary paths:  /bin, /usr/bin, /usr/local/bin, /opt/bin and /opt/local/bin.

Then we could simply write:  `_GS_CMD = which('gs')`

Paul-Franklin:
Note that search_for was already modified (by Josip, 15 Jan 2011)
to look in the current folder on Windows, which is not usually in PATH,
since prepending it to the PATH is critical for the AIO's ability to find
the "correct" DLL, and not one (say) in system32 instead.

https://gramps-project.org/bugs/view.php?id=4532
https://github.com/gramps-project/gramps/commit/5c8a3f74889ca8d9dd064f0fd98adaf10b2f17f4

prculley:
@Nick-Hall I'm not quite sure which aspect of the last commit you did not like.  Was it the config option that allows the user to put something in place if nothing else worked?  Or the place the code was located?  Or not having enough possible paths?

I can easily modify the gen.utils.search_for function to test for all the suggested paths, instead of doing it locally for Ghostscript.  But after a quick Google on where Unix style OS puts applications, it seemed that they could be almost anywhere.  Normally this would not be an issue since the system PATH variable would allow finding it.  But OS X is giving a different PATH to its 'Applications' that is severely limited, presumably to improve security.  Thus the config option to allow a user to just tell Gramps the path if nothing else worked.

Nick-Hall:
@prculley I don't like the location of the fix or the extra Ghostscript specific configuration option.

Applications on a Unix-style OS can be installed anywhere, but there is a [conventional directory layout](https://en.wikipedia.org/wiki/Unix_filesystem#Conventional_directory_layout).  It is reasonable that we search in the conventional binary directories and the PATH.

We are likely to meet this problem for other applications, so modifying the utility functions is a better solution.  If you feel that a configuration option is really necessary it should contain a general search path.

prculley:
@Nick-Hall change implemented as requested.  The new "where_is" is located with other file functions in gen.util.file.py.  I dropped the configurable path option.

PR #474, opened 2017-10-10
Replace use of CONTROL_MASK with PRIMARY_INTENT.

jralls:
For better UX on Macs. Mac users expect to use the "command" modifier key pretty much everywhere that users of other systems use the "control" key. Gtk provides for this preference with the  Gdk.ModifierIntent.PRIMARY_ACCELERATOR, which should always be used instead of GDK_CONTROL_MASK. Gtk.Widget uses a private GtkBuilder function to implement the alias "primary" in Builder XML (a.k.a. glade) files.

### DIFF
--- a/gramps/gui/clipboard.py
+++ b/gramps/gui/clipboard.py
@@ -60,7 +60,7 @@ from gramps.gen.constfunc import mac
 from .glade import Glade
 from .ddtargets import DdTargets
 from .makefilter import make_filter
-from .utils import is_right_click
+from .utils import is_right_click, get_primary_mask
 
 #-------------------------------------------------------------------------
 #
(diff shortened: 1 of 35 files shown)


SNoiraud:
We use right click in geography views. Do we need to do something for that.
the impacted file is : gramps/plugins/lib/maps/osmgps.py

jralls:
@SNoiraud yes, line 507 should be either is_right_click() or GDK.Event.triggers_context_menu().

Another thing to consider is how to provide button-2 operation for pointers that don't have one and don't provide middle-button emulation.

SNoiraud:
I tested GDK.Event.triggers_context_menu(). It seems to work for me. You can add it in your PR.
Button 2 in the geography view is only used to zoom in on an area of the map. So, it may not be useful to dive into the code for this replacement : 

@@ -504,7 +504,9 @@ class OsmGps:
         elif event.button == 2 and event.type == Gdk.EventType.BUTTON_RELEASE:
             self.end_selection = current
             self.zone_selection = False
-        elif event.button == 3 and event.type == Gdk.EventType.BUTTON_PRESS:
+        #elif event.button == 3 and event.type == Gdk.EventType.BUTTON_PRESS:
+        #     for MAC users who don't have third button :
+        elif Gdk.Event.triggers_context_menu(event):
             self.build_nav_menu(osm, event, lat, lon)
         else:
             self.save_center(lat, lon)

jralls:
Done.

Zooming also responds to scrollwheel events and the touchpad equivalent; there are also clickable zoom controls in the window so I guess that's covered.

That brings up {{{gui/plug/quick/_quicktable.py}}} and {{{gui/widgets/selectionwidget.py}}} which use button 3 for purposes other than opening a context menu. The former is quite complicated and seems to have a bunch of side effects, though it does sometimes display a context menu. The latter emulates the X11 left-click/right-click selection method which will be rather foreign to most users.

PR #475, opened 2017-10-11
Check&Repair; add checks for 'backlinks' and repair if needed.

prculley:
While working on bug https://gramps-project.org/bugs/view.php?id=10221 I wanted a method of checking and repairing the db to assist in my debugging.  I decided to just add this to the Check & Repair tool.

The 'Repair' part does the same thing as the "Tools/Family Tree Repair/Rebuild reference maps" tool.  The check portion checks all objects for:

1. each reference has a backlink
2. backlinks that don't have a referenced object
3. backlinks that don't have a reference within the object, although referenced object exists

It is remotely possible that the 'references' tables in the db (source of backlinks) could also have orphan entries which are indexed by something missing and no object has a reference to that missing something either, this tool cannot detect this.  But it would not hurt anything in the db other than taking up space.

This can be (partially) tested using a delete operation on 5.0.0-alpha3 (until bug 10221 is fixed) using an sqlite (dbapi) db.  The delete leaves backlinks (if the object had references to something else) that should not be there.

### DIFF
--- a/gramps/plugins/tool/check.py
+++ b/gramps/plugins/tool/check.py
@@ -61,7 +61,7 @@ ngettext = glocale.translation.ngettext  # else "nearby" comments are ignored
 from gramps.gen.lib import (Citation, Event, EventType, Family, Media,
                             Name, Note, Person, Place, Repository, Source,
                             StyledText, Tag)
-from gramps.gen.db import DbTxn
+from gramps.gen.db import DbTxn, CLASS_TO_KEY_MAP
 from gramps.gen.config import config
 from gramps.gen.utils.id import create_id
 from gramps.gen.utils.db import family_name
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Travis is failing on this PR.

prculley:
Interesting.  The Travis tools_test uses the test case generator to make a truly messed up db.  So bad that any attempt to run `find_backlink_handles` on the resulting db fails with a weird bsddb exception (it thinks the references db is closed).

This did not happen on dbapi; there it just had 78 bad backlink issues.

So I just added an overall try/except (Yeah raw except) to the backlink check and forced a reindex.

Paul-Franklin:
I like the idea of this pull request.

Nick-Hall:
The checks are run within a [batch transaction](https://github.com/prculley/gramps/blob/64a62f25a1de18fd71a45cd3fa0a44fb454f6880/gramps/plugins/tool/check.py#L187).  In BSDDB, secondary tables are closed to improve performance.   This means that `find_backlink_handles` cannot be used.  You should run this test in a non-batch transaction.

gramps/plugins/tool/check.py
```diff
@@ -1215,6 +1217,87 @@ def check_events(self):
                 len(self.invalid_events) == 0:
             logging.info('    OK: no event problems found')
 
+    def check_backlinks(self):
+        '''Looking for backlink reference problems'''
+
+        total = (
```
Nick-Hall: Consider using [`get_total`](https://github.com/gramps-project/gramps/blob/master/gramps/gen/db/base.py#L1966) here.

prculley:
@Nick-Hall If the bsddb has the 'references' db closed for batch mode, then that would imply that it must rebuild the 'references' table at the end of every batch transaction.  If that is so, then there is no point to running this 'check backlinks' test at all for bsddb, since all the rest of the checks are run in batch mode.  And the 'references' table must be getting rebuilt anyway.

Do I have this right?  I don't have time to research this at the moment.

Nick-Hall:
@prculley The reference table is still open, but it has two secondary tables associated with it, which act as indexes on the primary and referenced handles.

prculley:
I moved the backlink check out of the db batch transaction into its own transaction, with batch=False.  And removed the try/except block.
It appears to be working with bsddb, tested by making an error by deliberately preventing a reference table update during a delete with my debugger.   Also working with sqlite, tested as per first comment.
And passes the tools_test with both db types.

PR #476, opened 2017-10-12
New convenience function to return database methods

Nick-Hall:
Examples:

db.method('get_%s_from_handle', 'Person')
db.method('get_%s_from_%s', 'Event', 'gramps_id')

Returns None if the method doesn't exist.

Formats 'iter_%s' and 'get_number_of_%s' use the plural forms and
are not yet implemented.

Replaces old get_table_metadata method.

Bug #9541.

### DIFF
--- a/gramps/gen/db/base.py
+++ b/gramps/gen/db/base.py
@@ -1433,6 +1433,34 @@ class DbReadBase:
         """
         return False
 
+    def method(self, fmt, *args):
+        """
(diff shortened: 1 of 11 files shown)


prculley:
You could deal with plural forms by putting aliases in DbReadBase or DbWriteBase __init__.
self.get_number_of_person = self.get_number_of_people
Unfortunately has to be after instantiation or inheritance messes it up.

PR #477, opened 2017-10-13
Not a pull request - Comment only about hardcoded bsddb not my defaul…

fxtmtrsine:
…t database format

Hello, could not find a way to raise an issue here on github, hence the pull reqest 

Unexpected behavior when I've selected SQLite as my default. 
 * Create a new family tree
 * Use the gui manager to "archive" my family tree
 * select the archived tree from the gui and the button changes to "Extract" so press that
 * In the "Database type" column "BSDDB" is listed instead of my default of sqlite

Keep up the great work.

Tested on Arch Linux

### DIFF
--- a/gramps/gui/dbman.py
+++ b/gramps/gui/dbman.py
@@ -665,7 +665,7 @@ class DbManager(CLIDbManager, ManagedWindow):
 
         self.__start_cursor(_("Extracting archive..."))
 
-        dbase = make_database("bsddb")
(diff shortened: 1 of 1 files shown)


neic:
Hi @fxtmtrsine 
The bugtracker lives on https://gramps-project.org/bugs

Nick-Hall:
Fixed in commit 50091d1f86cbce68fdb40a762e528c5df386815f.

fxtmtrsine:
@neic @Nick-Hall Thanks

PR #479, opened 2017-10-13
Dbapi: Fix for broken 'backlinks' after a delete.

prculley:
Fixes #10221

The dbapi was missing a way to update the references table (backlinks) when object are deleted.  Actually the forward references by the deleted object were not removed, but the issue showed up as bad backlinks.

This PR deals with that specific issue.   I looked into refactoring the entire 'delete' code, but decided to make that a separate PR (if I do it at all).

You can test that this works by deleting an object with forward references (contains a Note, Media etc.).  And then run the updated Check & Repair from #475 while on a Dbapi db like Sqlite.  Prior to this fix Check and repair will show an error, after the fix, the same test procedure should not.

### DIFF
--- a/gramps/plugins/db/dbapi/dbapi.py
+++ b/gramps/plugins/db/dbapi/dbapi.py
@@ -35,7 +35,7 @@ import logging
 #
 #------------------------------------------------------------------------
 from gramps.gen.db.dbconst import (DBLOGNAME, DBBACKEND, KEY_TO_NAME_MAP,
-                                   TXNADD, TXNUPD, TXNDEL,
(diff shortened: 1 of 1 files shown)


Nick-Hall:
The code in `_remove_backlinks` looks very similar to some of the code in` _update_backlinks`, which aslo adds new backlinks.  Are you sure we are not just missing a call to ` _update_backlinks`?

prculley:
@Nick-Hall It is similar.  I suppose I could have removed the references from the object to be deleted, via get_referenced_handles_recursively and a bunch of calls to obj.remove_handle_references and then called update_backlinks, which would have detected this change and then removed them from the references table, but this seemed more direct.

PR #480, opened 2017-10-14
Fix Geography view 'Find' when db is closed.

prculley:
Fixes #10229

Crash when clicking 'Find' in Geography view when db is closed.

View attempts to recreate the map with the filter, and the filters don't deal well with closed db.
PR just skips the attempt to put places on the map when db is closed.

### DIFF
--- a/gramps/plugins/view/geoclose.py
+++ b/gramps/plugins/view/geoclose.py
@@ -323,10 +323,12 @@ class GeoClose(GeoGraphyView):
         all handling of visibility is now in rebuild_trees, see that for more
         information.
         """
+        self.lifeway_layer.clear_ways()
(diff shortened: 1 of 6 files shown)


SNoiraud:
We have the same call to the generic filter in geoevents.py

prculley:
@SNoiraud Thanks for the tip.  I did not realize that so much of the code was duplicated across the different types of geo views.  I think I have found all the potential places that cause troubles with closed db now.

The only view that did not need attention was geomoves, because it depended on a single person from get_active, and it is not set when the db is closed.

I also found it useful to clear the 'lifeway_layer' in a couple of cases, because it got left active when the db closed.

PR #487, opened 2017-11-07
Multiple Surname Editing fixes

prculley:
Fixes #10254, #9868, #6828, #6257
Issues fixed:
Typing a surname 'Multiple Surnames' box, followed by mouse click anywhere but in the 'Multiple Surnames' box (like the 'OK' button) lost the typed surname.

Changes made in the person editor 'Multiple Surnames' box did not appear in the 'Preferred name', unlike other changes.  Two cases, name changes, and order of names, needed different fixes.

If shrunk to the minimum size, the 'Primary' field of the Person Editor 'Multiple Surnames' box was partly cut off.

If shrunk to the minimum size, the 'Primary' field of the Name Editor 'Family Names' box was almost completely cut off.

Empty surnames were not very obvious, particularly if the 'Multiple Surnames' box was partly cut off.

Editing directly in the 'Multiple Surnames' box is not very intuitive, not consistent with other edits of lists in Gramps, and for this box in particular was darn ugly.  The attached screen cap shows how the "Origin" field edit originally looked on Windows; the drop down is not clear, partly cut off (box is big enough), how it overlaps the "Connector" and "Primary" fields.
multsurnames before

I decided to create a separate "Surname Editor" popup window, rather than trying to figure out how to completely fix editing in the 'Multiple Surnames' box.  It seemed to me to be a bit more consistent with the rest of Gramps, and was clearer on how to complete the edit.
surnameeditor

I adjusted the 'Multiple Surnames' box for better fitting when the windows are resized.  This was done by making the 'Surname' field in the box the one that changes with overall window size.  The user can still change the other field sizes by adjusting the column breaks in the header.

I added a "Missing Surname" text, shown only in this box (not saved as part of name) if the Surname field is empty.

### DIFF
--- a/gramps/gui/editors/displaytabs/embeddedlist.py
+++ b/gramps/gui/editors/displaytabs/embeddedlist.py
@@ -519,9 +519,12 @@ class EmbeddedList(ButtonTab):
                 # insert the colum into the tree
                 column.set_resizable(True)
             column.set_clickable(True)
-            column.set_sizing(Gtk.TreeViewColumnSizing.FIXED)
-            #column.set_min_width(self._column_names[pair[1]][2])
-            column.set_fixed_width(self._column_names[pair[1]][2])
+            if self._column_names[pair[1]][2] != -1:
+                column.set_sizing(Gtk.TreeViewColumnSizing.FIXED)
+                #column.set_min_width(self._column_names[pair[1]][2])
+                column.set_fixed_width(self._column_names[pair[1]][2])
+            else:
+                column.set_expand(True)
 
             column.set_sort_column_id(self._column_names[pair[1]][1])
             self.columns.append(column)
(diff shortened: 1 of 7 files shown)


Nick-Hall:
The multiple surnames box certainly needs tidying up.  However, I don't like the extra surname editor.  Users already complain about our excessive use of dialogs.

Try replacing the TreeView with a Grid containing dynamically generated widgets, one row per surname.

prculley:
I've already invested several days in this, and I think it is obvious, consistent with the rest of Gramps,  and usable.
If you don't like it, I guess it is your turn to write the code.  Or leave the bugs open.

Nick-Hall:
OK.  Leave it to me.

sam-m888:
See https://github.com/gramps-project/gramps/pull/527 Surname Editor fixes Take 2!

PR #488, opened 2017-11-07
Fix Session Log for exception on closed db reference click

prculley:
The Session Log Gramplet keeps a running record of changes made.  The original version kept this even across db close/open operations.  However, if the user mouse clicked one of the links to a closed db, an exception would occur.

I patched the HandleError in the grampletpane code so clicks on bad links would not get the exception.

During testing I noticed that deleted persons or families were not being identified; turns out that for bsddb the 'interesting' method of looking into the transaction log was getting a 'bytes' handle and trying to compare with str handle.  So put back in code to convert to str if bytes was found.

I then noticed that the Person select code was not getting executed; this was due to an earlier fix that stopped generating active-change for persons for everything.  Added the proper connect for that.  And added a similar connect for Family active-change  while I was at it.

These bugs found while trying to recreate a bug from the bug tracker (not successfully).

### DIFF
--- a/gramps/gui/widgets/grampletpane.py
+++ b/gramps/gui/widgets/grampletpane.py
@@ -638,6 +638,8 @@ class GuiGramplet:
         for (tag, link_type, handle, tooltip) in self._tags:
             if iter.has_tag(tag):
                 if link_type == 'Person':
+                    if not self.dbstate.db.has_person_handle(handle):
+                        return True
                     person = self.dbstate.db.get_person_from_handle(handle)
                     if person is not None:
                         if event.button == 1: # left mouse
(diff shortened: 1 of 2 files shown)


Nick-Hall:
> During testing I noticed that deleted persons or families were not being identified; turns out that for bsddb the 'interesting' method of looking into the transaction log was getting a 'bytes' handle and trying to compare with str handle. So put back in code to convert to str if bytes was found.

Yes,  this is probably the best fix.  We can't issue the 'delete' signal before the transaction, when the handle is still valid, in case it fails and needs to be rolled back.  Changing the signal order may well also cause adverse effects.

The transaction `get_record` method only appears to be used in the session log gramplet, so I suppose we could do the type conversion in the database layer.  Alternatively, we could re-write the gramplet so the _delete_ loop is outside of the _handles_ loop - but is it worth the effort?

prculley:
The 'ideal' fix would be to correct in the db layer so we never see 'bytes' handles.  But I was a bit surprised that anything other than the db itself ever looked into the transaction log; it did not seem to be a reasonable public API.  So I reasoned that this was a 'special' situation and deserved the very minor special (and easy) fix.  If we ever dump bsddb we can drop the check.

Nick-Hall:
I agree.

PR #489, opened 2017-11-18
Mainz css file produces left aligned FamilyMap pages

SNoiraud:
Solving Mainz problem creates new ones with Basic-Blue
and Nebraska stylessheet.

Some cleanup with unused css file (GeoView.css)
Strange effects on the navigation bar for the Mainz stylesheet.
Suppression of the first image in the family because we can't have selection regions in a thumbnail.
Placement problems with all stylesheets for the ancestor tree in the person page.

Fixes #10268


SNoiraud:
Testing these patches, I saw the media file name is not created if we have unused or unreferenced media.
I correct this.

PR #494, opened 2017-11-25
Filter fixes and addtional Travis tests

prculley:
Was looking into filters to see if a users issue with run time could be improved.  But first wanted to make sure tests were complete so that any changes did not cause more issues.

Test developments showed up a couple of issues themselves, some HandleErrors and the RegExpName filter missing from the GUI.
This PR corrects those.  Fixes #10286.
If test are run without the fixes, there should be errors.

### DIFF
--- a/gramps/gen/filters/rules/person/__init__.py
+++ b/gramps/gen/filters/rules/person/__init__.py
@@ -187,6 +187,7 @@ editor_rule_list = [
     HasNote,
     HasNoteRegexp,
     RegExpIdOf,
+    RegExpName,
     Disconnected,
     ChangedSince,
     IsRelatedWith,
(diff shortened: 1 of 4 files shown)


daleathan:
> Nick-Hall merged commit 33ebd5d into gramps-project:maintenance/gramps50
>
>  prculley deleted the prculley:filts branch on Dec 9, 2017

@prculley @Nick-Hall Does this PR's (as well as #496 and #497)  need to be closed or does something further need to be do like merging them? @prculley Great stuff speeding up the filters 👍 

are not going to be merged for gramps 5?

Nick-Hall:
@daleathan I'll have a look at them now.  They were tagged as enhancements rather than bugs.

PR #497, opened 2017-11-28
Filter fixes to prevent MatchesFilter types from running multiple times

prculley:
The filters that include 'OfFilterMatch' in their name run 'on top' of other filters; they execute the other filters first and then use the results as their starting point.
If there are multiple filter rules that include 'OfFilterMatch' and they are executing the same base filter, then the current code runs the same base filter multiple times.  The base filter gives the same result each time, so this represents extra work.

The current code attempted to correct this in the gen.filters.rules._rule.py in the 'requestprepare' method by tracking if a filter had been already run.  The already run indication was cleared by the 'requestreset' method.  

Unfortunately, the filters that include 'OfFilterMatch' were executing the 'requestreset' as soon as they were individually done with the base filter result during their 'prepare' step.  This meant that the base filter 'already run indication' was also cleared.  A subsequent attempt to use the same base filter would have to execute it again.

To avoid the multiple runs of base filters, I've moved the base filter 'requestreset' call from 'OfFilterMatch' filters from the 'prepare step' to their 'requestreset' step, so that all the filters are reset together at the end of operations.  Thus the state of the base filters is retained for use multiple times, if needed.

This PR will increase memory usage when 'OfFilterMatch' rules are used, during the execution of the overall filters.  The time saved will depend greatly on the number of times the base filters are reused, and the time spent in their 'prepare' step.

I recommend that this PR be accepted after PR https://github.com/gramps-project/gramps/pull/494, which includes additional tests to demonstrate the correct execution of the filters.

### DIFF
--- a/gramps/gen/filters/rules/person/_deeprelationshippathbetween.py
+++ b/gramps/gen/filters/rules/person/_deeprelationshippathbetween.py
@@ -43,31 +43,6 @@ _ = glocale.translation.gettext
 #-------------------------------------------------------------------------
 
 
-def filter_database(db, user, filter_name):
-    """Returns a list of person handles"""
-
-    filt = MatchesFilter([filter_name])
-    filt.requestprepare(db, user)
-
-    if user:
-        user.begin_progress(_('Finding relationship paths'),
-                            _('Retrieving all sub-filter matches'),
-                            db.get_number_of_people())
-    matches = []
-    for handle in db.iter_person_handles():
-        person = db.get_person_from_handle(handle)
-        if filt.apply(db, person):
-            matches.append(handle)
-        if user:
-            user.step_progress()
-    if user:
-        user.end_progress()
-
-    filt.requestreset()
-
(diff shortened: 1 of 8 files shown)


Nick-Hall:
This needs a rebase.

prculley:
@Nick-Hall Rebase completed.

daleathan:
@prculley PR https://github.com/gramps-project/gramps/pull/499 also needs a rebase (assuming it's part of all these filter fixes 👍 )

daleathan:
@Nick-Hall Will these two ( PR #497 & #499 ) make it into the 5.0 beta release?

Nick-Hall:
@daleathan I wasn't considering them because they are labelled as enhancements rather then bugs.  I don't mind if Paul wants to include them in the beta release.  However, after the beta release we should be strict about bugs only.

prculley:
Rebased to master.

PR #499, opened 2017-11-29
Filters optimization; Use cache proxy with filters in views

prculley:
One more improvement in use of filters.  This one uses the CacheProxyDb to provide some speedup in the use of filters in views.  This is applied only when a filter is being executed in the view.

Provides a good speedup with complex filters, and larger trees less so with simpler ones.  I expect good improvements until the tree size approaches the size of the cache, (currently 100,000 objects) after which a smaller effect.

The improvement is less noticeable if the other filter optimizations (PR https://github.com/gramps-project/gramps/pull/496 , https://github.com/gramps-project/gramps/pull/497) are also put into effect.


prculley:
Shifted to master branch; I think its worth the effort...

PR #502, opened 2017-12-07
Fix QuestionDialog display for html like characters in title

prculley:
Fixes #10298

QuestionDialog was used to display Family Tree names, which may have characters interpreted as html, ('&') even though they are not.

I scanned all Gramps source and Addons Source, and no-one was putting html in title (msg1), so used the html.escape() call to protect the title, as suggested by user DougReider


prculley:
@Nick-Hall Done, as requested.

PR #506, opened 2017-12-19
Fix interface problems with dark themes

Nick-Hall:
* List views now use the default foreground colour from the theme rather than using black.
* Find the link colour from the theme.  Changing the theme will still require a restart.
* Change the shading in the relationship view to use black for dark themes and white for light themes.  I use the amount of red to distinguish a light theme from a dark theme.  There may be a better way to do this.

### DIFF
--- a/gramps/gui/utils.py
+++ b/gramps/gui/utils.py
@@ -538,6 +538,21 @@ def rgb_to_hex(rgb):
         rgbint = (int(rgb[0] * 255), int(rgb[1] * 255), int(rgb[2] * 255))
         return '#%02x%02x%02x' % rgbint
 
+def get_link_color(context):
+    """
+    Find the link color for the current theme.
+    """
+    from gi.repository import Gtk
+
+    if Gtk.get_minor_version() > 11:
+        col = context.get_color(Gtk.StateFlags.LINK)
+    else:
+        found, col = context.lookup_color('link_color')
+        if not found:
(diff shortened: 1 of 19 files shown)


vantu5z:
Relationship view is not good. Color is very contrast, we should get some color from theme (that used for Gtk.TextView for example).

For information: Vertex-dark theme don't have this bug as Adwaita-dark.

Nick-Hall:
@vantu5z Thanks for the feedback.  Updated to use the background colour of a TextView for shading in the relationship view.  Unfortunately, I can't see an alternative to using the deprecated `get_background_color` method.

I haven't tested a dark theme with Gtk 3.10 yet.

vantu5z:
Now looks good.
Maybe don't look color and replace it. But use some container with native color that make shading effect?

vantu5z:
Read docs tell that `override_background_color` deprecated too. We should use `Gtk.render_background` to change background color theme. Internet give some example:
```
context = my_entry.get_style_context()
context.save()
context.add_class('warning')
Gtk.render_background(context, blah, blah, blah)
context.restore()
```
How I can understand we should found class name for TextView and we can replace rendering of widget background.

Nick-Hall:
@vantu5z I have created a new widget with a shaded background.

vantu5z:
It works for me! Good job 👍

PR #513, opened 2017-12-27
fixing polish places translations

neonkowy:


Nick-Hall:
Thanks for the updated translation.

PR #515, opened 2018-01-02
Geocoding: associate a lat/lon to a place name

SNoiraud:
Fixes #09642
Issue #10361

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,13 @@ The following packages are optional:
 
  More font support in the reports
 
+* **gir1.2-geocodeglib-1.0**
+
+ A library use to associate a geographical position (latitude, longitude)
+ to a place name. This is used if you already have osmgpsmap installed.
+ If installed, when you add or link a place from the map, you have a red line
+ at the end of the table for selection.
+
 Optional packages required by Third-party Addons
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -98,12 +98,16 @@ The following packages are optional:
...
 
-* **gir1.2-geocodeglib-1.0**
+* **geocodeglib**
 
  A library use to associate a geographical position (latitude, longitude)
  to a place name. This is used if you already have osmgpsmap installed.
  If installed, when you add or link a place from the map, you have a red line
  at the end of the table for selection.
+ Debian, Ubuntu, ... : gir1.2-geocodeglib-1.0
+ Fedora, Redhat, ... : ?
```
QuLogic: Probably [geocode-glib](https://apps.fedoraproject.org/packages/geocode-glib) as well.

SNoiraud: I look at in this rpm, but the gir file is missing. So we need another rpm which provides ...

Ah yes, new versions doesn't need the gir file. only the typelib is necessary.
This rpm provides : 
/usr/lib/girepository-1.0/GeocodeGlib-1.0.typelib
So it's good. I modify the readme file.

PR #516, opened 2018-01-03
fix Fan Chart for exception on right-click of person with no parent

prculley:
Fixes #10349

One of many HandleErrors...

### DIFF
--- a/gramps/gui/widgets/fanchart.py
+++ b/gramps/gui/widgets/fanchart.py
@@ -1598,6 +1598,8 @@ class FanChartGrampsGUI:
                 siblings.append(sib_id)
         # Collect a list of per-step-family step-siblings
             for parent_h in [fam.get_father_handle(), fam.get_mother_handle()]:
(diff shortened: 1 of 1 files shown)


gramps/gui/widgets/fanchart.py
```diff
@@ -1595,7 +1595,8 @@ def on_popup(self, obj, event, person_handle, family_handle=None):
                     continue
                 siblings.append(sib_id)
         # Collect a list of per-step-family step-siblings
-            for parent_h in [fam.get_father_handle(), fam.get_mother_handle()]:
+            for parent_h in [hnd for hnd in [fam.get_father_handle(),
```
Nick-Hall: Instead of a list comprehension it maybe clearer to use:  `if not parent_h: continue`

prculley:
@Nick-Hall Fixed as per your wishes...

Nick-Hall:
@prculley Thanks.

PR #517, opened 2018-01-04
Narrative web: multiple problems.

SNoiraud:
1 - Added an option for the statistics page.
2 - We have problems with notes : pages are too large
3 - Difficult to see some page on a mobile phone when you have notes.
4 - Resizing of images on a mobile
5 - Maps are too bigs on a mobile
6 - Split the event line to have a narrower page
7 - The note is spanned on two columns (Place + Description)
7 - In media pages, the progress meter doesn't work well.
8 - Add a progress meter for the medialistpage
9 - Nebraska and Basic-Blue css files : extra space between lines
10 -  some css files contains individuallist instead of IndividualList (case problem)

Issues #10344

### DIFF
--- a/data/css/Web_Basic-Blue.css
+++ b/data/css/Web_Basic-Blue.css
@@ -288,10 +288,6 @@ table.infolist thead tr th {
 table.infolist tr td {
     border-bottom: dashed 1px #000;
     vertical-align: middle;
-    padding: 6px 0 6px 10px;
-}
-table.infolist tr td a {
-    display: block;
 }
 table.infolist tr.BeginLetter td, table.infolist tr.BeginSurname td {
     border-top: solid 1px #000;
@@ -351,15 +347,15 @@ div#Individuals {
     margin: 0;
     padding: 0;
 }
-div#Individuals table.individuallist {
+div#Individuals table.IndividualList {
     border-bottom: solid 1px #000;
 }
-div#Individuals table.individuallist tbody tr td.ColumnSurname a:hover,
-div#Individuals table.individuallist tbody tr td.ColumnSurname a:active {
+div#Individuals table.IndividualList tbody tr td.ColumnSurname a:hover,
+div#Individuals table.IndividualList tbody tr td.ColumnSurname a:active {
     cursor: default;
     background: none;
 }
-div#Individuals table.individuallist tbody tr td.ColumnName a {
+div#Individuals table.IndividualList tbody tr td.ColumnName a {
     vertical-align: middle;
 }
 div#Individuals div table.infolist tr td p {
--- a/data/css/Web_Basic-Blue.css
+++ b/data/css/Web_Basic-Blue.css
@@ -1129,9 +1129,10 @@ div.narrative {
 }
 .narrative p {
     font: normal .9em/1.4em sans-serif;
-    margin-top: .5em;
-    margin-bottom: 0;
-    padding: 0 20px 1em 20px;
+    margin: 0.1em 0 0.2em 0;
+}
+i + div.grampsstylednote p {
+    margin: 0.1em 0 0.2em 0;
 }
 
 /* Subsections : References
--- a/data/css/Web_Nebraska.css
+++ b/data/css/Web_Nebraska.css
@@ -276,7 +276,6 @@ table.infolist tr th a:hover {
 table.infolist tr td {
     font:normal 1.1em/1.4em serif;
     vertical-align:middle;
-    padding:.1em 10px;
 }
 table.infolist tr td a {
     display:block;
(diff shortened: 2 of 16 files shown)


SNoiraud:
@prculley : 
I have a problem with : 
FAIL: test_navwebpage (gramps.plugins.test.reports_test.TestDynamic)
where is this test ?

prculley:
The test found the the message "Failed to write report." in the stderr.  You can look at the complete message if you run the reports_test.py module with a breakpoint set at line 91 (self.assertTrue(test_function(out, err, report_name, **options))) and watch for your report_name to be 'navwebpage' then look at the 'err' variable.   Might take a few hits at the breakpoint to get to your test (on my system it was the 20th test). The 'args' contain the CLI parameters used for the test.

prculley:
I looked at this a bit.  I see no obvious issues with most of your fixes, however the 'cleanup_var' does not do anything useful as is, and I think the concept is too dangerous to use.

As is, you pass in the 'NavWebReport' class object, which doesn't have an '__iter__' attribute, so the only thing the routine does is to try to delete the class object and do a gc.  Which doesn't do anything at that point since the 'NavWebReport' class object still has references to it.

I made a quick try to get it to utilize the class '__dict__' to go recurse down, and it crashed python completely.  I started to follow along as it ran, and found it trying to delete characters from strings, and dive into all the objects owned by Gramps, like the db, the user class etc. and delete their data structures.  Probably not what you intended, so my fault for trying to fix this in this way.

I think we should be a lot more surgical to try to clean up; perhaps del the proxycache (already covered I see, via clear_cache) and self.tab in the top level narrativeweb to release the next set of page class objects.  The latter will break one set of circular references.

If you want to look at circular references in more detail, you might grab the 'leak.py' Gramplet from the master branch; it has been upgraded to make it easier to see the circular references (still not easy, but better).

SNoiraud:
I'll have a look at leak.py
I tried cleanup_var different ways without succes. As the reporter wrote in the bug report, it seems the gc collector is a bit lazy.
I already tried to do : cleanup_var(self.obj_dict) and cleanup_var(self.bkref_dict) but it took a long time
I can remove it.

I found the travis problem.
I use self.user._progress.set_header() in the patch. It works well except in the travis test. I think we have not the graphical environment and self.user._progress is not used.
Perhaps I need to include this statement in a try except.

Nick-Hall:
> I found the travis problem.
I use self.user._progress.set_header() in the patch. It works well except in the travis test. I think we have not the graphical environment and self.user._progress is not used.
Perhaps I need to include this statement in a try except.

Variables that start with an underscore are protected.  They should only be accessed in sub-classes.

In this case, we should consider adding a `set_header` method to the `User` class.  Reports can be run in both GUI and command line environments.

SNoiraud:
> In this case, we should consider adding a set_header method to the User class. Reports can be run in both GUI and command line environments.

I'll remove this function. I'll do that for master in a new PR.

prculley:
I've just spent several days updating the 'leak' Gramplet to make it easier to spot cyclic references.  Also did a lot of research and experimentation.  My conclusions:

The Python garbage collection really does work!  It is possible that earlier versions of Python don't work quite as well, clearly some of the historical v2.x had real issues, but the versions I tested (3.4, 3.6) work fine. 

My earlier difficulty with seeing no evidence that it was working was due to the fact that the "Uncollected Objects" Gramplet (leak.py) was effectively disabling the gc for diagnostic purposes.  Once I figured this out (after doing lots of work updating the leak Gramplet) I realized we can probably mostly ignore manually fixing up reference loops. 

I don't think we need to worry about this much, because Gramps is not exactly a high performance time sensitive application; a few Python garbage collections now and then, usually taking >0.2 sec are unlikely to be noticed in the normal operations.

So the extra work to try to clear and delete objects with NavWeb itself appears to me to be no longer necessary.

I do think we should programmatically run the garbage collection after major operations (like reports) to speed up the process of memory release.  That should stop the growth of memory footprint with multiple runs of reports, like NavWeb.  My inclination is to put "gc.collect(2)" in gui.viewmanager.py run_plugin at the end of the method, after reports/tool is run (about line 1772).

I also learned that although the garbage collection frees up memory from Python's viewpoint, it doesn't always quickly get deallocated at the system level.  Python keeps some of it around for allocating new objects.  So using Windows or Linux memory status tools doesn't always show immediate memory reductions.

If you or anyone should want to look at my updated Leak Gramplet update, it is https://github.com/gramps-project/addons-source/tree/leakv2/Leakv2

Nick-Hall:
> The Python garbage collection really does work! It is possible that earlier versions of Python don't work quite as well, clearly some of the historical v2.x had real issues, but the versions I tested (3.4, 3.6) work fine.

Most of our historic memory leak issues were related to Gtk2.

SNoiraud:
> The Python garbage collection really does work!

I agree

> I don't think we need to worry about this much
+1 

> I do think we should programmatically run the garbage collection after major operations (like reports) to speed up the process of memory release.

I'am OK with that.

Do you think I can remove the self.database.clear_cache() statement ?

prculley:
I'm inclined to leave the  self.database.clear_cache() statement  in place, as it does give the gc less work to do, although I have not run tests to see if the cache clear is faster than the gc.  Either way SHOULD work.

sam-m888:
> 1 - Added an option for the statistics page.

Updated image in user manual to show new option on "Include" tab

https://gramps-project.org/wiki/index.php?title=Gramps_5.0_Wiki_Manual_-_Reports_-_part_7#Include

PR #521, opened 2018-01-09
relview doesn't follow the background theme.

SNoiraud:
Fix #010360

### DIFF
--- a/gramps/plugins/view/relview.py
+++ b/gramps/plugins/view/relview.py
@@ -317,8 +317,7 @@ class RelationshipView(NavigationView):
 
         self.scroll = Gtk.ScrolledWindow()
 
-        st_cont = self.scroll.get_style_context()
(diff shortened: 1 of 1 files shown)


Nick-Hall:
We already have a proposed solution for this in PR #506.

SNoiraud:
Yes, but this is not the same problem. Perhaps we need to merge the two PRs.
The PR #506 solves the foreground colors for tags in listview

Nick-Hall:
PR #506 solves three issues with dark themes:  the background colour in list views, the colour of links and the shading in the relationship view.

The proposed solution of a new ShadeBox widget reacts to a change of theme without the need to restart Gramps.  It also removes some deprecated methods.

SNoiraud:
I remove this PR. It's a duplicate of  PR #506

PR #523, opened 2018-01-14
Color schemes

vantu5z:
- add combobox to select color scheme in config dialog;
- change `add_color` and `update_color` methods to support color schemes;
- change signal `color-set` -> `notify::color` to catch programmatically color changes;
- add restore to defaults button;
- add colors for family node and home person;
- add colors for divorced families;

For backward compatibility colors saved not only to scheme colors but in past preferences too.

ToDo:
- add page for printing colors;
- check for views and addons that this PR don't brake something.

Printscreens:

### DIFF
--- a/gramps/gui/editors/displaytabs/surnametab.py
+++ b/gramps/gui/editors/displaytabs/surnametab.py
@@ -204,6 +204,16 @@ class SurnameTab(EmbeddedList):
         if self.on_change:
             self.on_change()
 
+    def post_rebuild(self, prebuildpath):
+        """
+        Called when data model has changed, in particular necessary when row
+        order is updated.
+        @param prebuildpath: path selected before rebuild, None if none
+        @type prebuildpath: tree path
+        """
+        if self.on_change:
+            self.on_change()
+
     def column_order(self):
         # order of columns for EmbeddedList. Only the text columns here
         return ((1, 0), (1, 1), (1, 2))
--- a/gramps/gui/editors/editperson.py
+++ b/gramps/gui/editors/editperson.py
@@ -197,9 +197,6 @@ class EditPerson(EditPrimary):
         self.singsurnfr = SingSurn(self.top)
         self.multsurnfr = self.top.get_object("hboxmultsurnames")
         self.singlesurn_active = True
-        self.surntab = SurnameTab(self.dbstate, self.uistate, self.track,
-                                  self.obj.get_primary_name(),
-                                  on_change=self._changed_name)
 
         self.set_contexteventbox(self.top.get_object("eventboxtop"))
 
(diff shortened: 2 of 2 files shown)


SNoiraud:
I think we could add two new colors for the family (used in GraphView) : 
family background color and family divorced background color

SNoiraud:
I think there is an error in gramps/gui/configure.py at line 401
We have : if shceme_type: and I think it should be if scheme_type:

vantu5z:
Yes I think to add colors too.
Thanks for your feedback and typo detection.
I shall do this.

sam-m888:
Looking good.

Note that travis-ci fails because of trailing white space see https://travis-ci.org/gramps-project/gramps/builds/331085694#L2807 :

> gramps/gui/configure.py:399:        Add color chooser widget with label to the grid. 
>

SNoiraud:
I think the following should be associated to another PR.
I think colors for divorced family should be a good thing to view divorced families. For the moment, I add a tag to show them.
The following patch add the possibility to show them. The colors in the patch are exemples and could be added to this PR.
[divorced.patch.txt](https://github.com/gramps-project/gramps/files/1655143/divorced.patch.txt)

Nick-Hall:
I like this PR.  We could even consider including it in _gramps50_.

The code was beginning to look a bit messy so I have pushed a commit with a few suggestions.  I created a new _colors_ configuration section and made the naming more consistent.  We need to check how much this would impact existing code.

vantu5z:
Looks good!
I don't have a lot of time and knowledge, but think about this. I like your solution.

- About new section: when I start do the same with new section `colors`, but this broke code that use this preferences, so I had to return to old names.
- About `add_color` and `update_color`: some addons have own config dialogs with `add_color` without scheme support but have to use modified `update_color` method with `self.__config.set('colors.%s' % constant, hexval)`. So `add_color` use only `constant` without `colors` section but update applies another way. I wriggled out from this issue in my solution , so we should think how to solve this in yours.

SNoiraud:
For the standard plugins, only gramps/gui/configure.py is impacted and color_graph_box is used by 
gramps/plugins/view/pedigreeview.py and gramps/gui/widgets/fanchart.py
Geoplaces use its proper color scheme for type of place, but maps are not theme dependant. So there is nothing to do for geoplaces.
For addons, only dynamicweb use these colors.

Nick-Hall:
Changed the colour settings to use either a string or list.  The list consists of two hex colours:  a light theme colour followed by a dark theme colour.

Nick-Hall:
Merged in commit 028f9582f58ca0c8e38986829e0bceab1b3135f5.

PR #524, opened 2018-01-14
Export Web Family Tree; error on file write is now a message, not exception.

prculley:
Fix Export Web Family Tree for errors on file write.    Fixes [#10364](https://gramps-project.org/bugs/view.php?id=10364)

Also did a Pylint editing pass in separate commit.

Tested by exporting example.gramps before and after changes, files match.


sam-m888:
Thank you

PR #525, opened 2018-01-14
Fix Export Web Family Tree for errors on file write

prculley:
Fixes [#10364](https://gramps-project.org/bugs/view.php?id=10365) similar to prior PR  https://github.com/gramps-project/gramps/pull/524 for gramps50 branch.

P.S. As a side effect, this actually also includes changes from prior gramps50 commits https://github.com/gramps-project/gramps/commit/a9685a64ffb223c9e4df9757bf13252ace3bc355#diff-3eb686b3a04410bd10765916abba1ae1 which changed to the "with open" style of file open, and https://github.com/gramps-project/gramps/commit/f7f708225fc0f9553c5cbf3ad57d4bbf69a9f0b2#diff-3eb686b3a04410bd10765916abba1ae1 which opens the file utf-8.

I did not cherry-pick these since the first one included a bunch of other stuff unrelated to Export Web Family Tree.

The new error dialog also doesn't contain a transient parent, because including that would have required a series of other patches to infrastructure to get the parent, all of which were done only to gramps50.

### DIFF
--- a/gramps/plugins/export/exportftree.py
+++ b/gramps/plugins/export/exportftree.py
@@ -48,6 +48,9 @@ log = logging.getLogger(".WriteFtree")
 from gramps.gen.utils.alive import probably_alive
 from gramps.gui.plug.export import WriterOptionBox
 from gramps.gui.glade import Glade
+from gramps.gui.dialog import ErrorDialog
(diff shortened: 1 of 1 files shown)


sam-m888:
Thank you

PR #527, opened 2018-01-17
Surname Editor fixes Take 2!

prculley:
While looking at another problem I ran into a technique that allowed me to fix an annoyance I tried to fix previously, without requiring another dialog, which was considered the wrong way to fix this.  So here is another take at getting this corrected.

Issues fixed:
* Typing a surname 'Multiple Surnames' box, followed by mouse click anywhere but in the 'Multiple Surnames' box (like the 'OK' button) lost the typed surname.  Annoying to mouse users.  Fixes [#9868](https://gramps-project.org/bugs/view.php?id=9868), [#6828](https://gramps-project.org/bugs/view.php?id=6828), [#6257](https://gramps-project.org/bugs/view.php?id=6257)  This was done by creating my own CellRendererTextEdit class, which was the same as the original Gtk class except that it saved the data on focus-out instead of tossing it out.

* Changes made in the person editor 'Multiple Surnames' box did not appear in the 'Preferred name', unlike other changes. Two cases, name changes, and order of names, needed different fixes.  Fixes [#10254](https://gramps-project.org/bugs/view.php?id=10254), 

* If shrunk to the minimum size, the 'Primary' field of the Person Editor 'Multiple Surnames' box was partly cut off.  If shrunk to the minimum size, the 'Primary' field of the Name Editor 'Family Names' box was almost completely cut off.
For that one I adjusted the 'Multiple Surnames' box for better fitting when the windows are resized. This was done by making the 'Surname' field in the box the one that changes with overall window size. The user can still change the other field sizes by adjusting the column breaks in the header.

* While editing a cell, the edit box would expand beyond the cell size, sometimes even going outside of the dialog visible area (Try editing a prefix with a minimum size Person or Surname editor dialog).  Some changes to the underlying Gtk.Entry size (set_width_chars) made the editing stay confined to the cells.

Three separate commits, the first two in their own commit, the sizing changes in a single commit, dependent on the first one.

### DIFF
--- a/gramps/gui/editors/displaytabs/embeddedlist.py
+++ b/gramps/gui/editors/displaytabs/embeddedlist.py
@@ -42,6 +42,7 @@ from gi.repository import Pango
 # Gramps classes
 #
 #-------------------------------------------------------------------------
+from ...widgets.cellrenderertextedit import CellRendererTextEdit
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
 from ...utils import is_right_click
@@ -482,7 +483,10 @@ class EmbeddedList(ButtonTab):
             type_col = self._column_names[pair[1]][3]
 
             if (type_col in [TEXT_COL, MARKUP_COL, TEXT_EDIT_COL]):
-                renderer = Gtk.CellRendererText()
+                if type_col == TEXT_EDIT_COL:
+                    renderer = CellRendererTextEdit()
+                else:
+                    renderer = Gtk.CellRendererText()
                 renderer.set_property('ellipsize', Pango.EllipsizeMode.END)
                 if type_col == TEXT_COL or type_col == TEXT_EDIT_COL:
                     column = Gtk.TreeViewColumn(name, renderer, text=pair[1])
--- a/gramps/gui/editors/displaytabs/embeddedlist.py
+++ b/gramps/gui/editors/displaytabs/embeddedlist.py
@@ -523,9 +523,12 @@ class EmbeddedList(ButtonTab):
                 # insert the colum into the tree
                 column.set_resizable(True)
             column.set_clickable(True)
-            column.set_sizing(Gtk.TreeViewColumnSizing.FIXED)
-            #column.set_min_width(self._column_names[pair[1]][2])
-            column.set_fixed_width(self._column_names[pair[1]][2])
+            if self._column_names[pair[1]][2] != -1:
+                column.set_sizing(Gtk.TreeViewColumnSizing.FIXED)
+                #column.set_min_width(self._column_names[pair[1]][2])
+                column.set_fixed_width(self._column_names[pair[1]][2])
+            else:
+                column.set_expand(True)
 
             column.set_sort_column_id(self._column_names[pair[1]][1])
             self.columns.append(column)
--- a/gramps/gui/editors/displaytabs/surnametab.py
+++ b/gramps/gui/editors/displaytabs/surnametab.py
@@ -45,7 +45,7 @@ _ENTER = Gdk.keyval_from_name("Enter")
 #
 #-------------------------------------------------------------------------
 from .surnamemodel import SurnameModel
-from .embeddedlist import EmbeddedList, TEXT_COL, MARKUP_COL, ICON_COL
+from .embeddedlist import EmbeddedList, TEXT_EDIT_COL
 from ...ddtargets import DdTargets
 from gramps.gen.lib import Surname, NameOriginType
 from ...utils import get_primary_mask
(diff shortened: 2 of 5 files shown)


prculley:
P.S. In case anyoune reading this has any ideas.

I cannot seem to figure out how to shrink the 'Origin' CellRendererCombo in the vertical when it is active.  If the embedded ComboBox is set to something other than ALIGN_FILL, it becomes smaller, but is still offset a bit from the cell boundaries, as if the Edited cell is still larger than the normal cell.  This seems to show that the ComboBox itself is not the limiting vertical size.  The ComboBox has no obvious container, other than CellRendererCombo, which is not really a Widget, but something using the GtkCellLayout interface, so I'm at a loss on how to adjust the vertical size.

A minor annoyance...

lordemannd:
prculley, you are the man! I took your changes on pull 527 against 5 alpha 3 and it looks great to me. 

GRAMPS: GrampsAIO64-5.0.0-alpha3 +  
Python: 3.6.2rc1 (default, Jun 25 2017, 13:29:15...  
BSDDB: 6.1.0 (6, 0, 30)  sqlite: 3.19.3 (2.6.0) 
LANG: en_US.UTF-8 
OS: Windows  

PLUS helper: gspawn-win64-helper-console.exe (#c52804) 
PLUS helper: gspawn-win64-helper.exe 
PLUS directory: lib/lib2to3 (#c53234) 
PLUS gramps50 changes to 2018-01-20 04:17:17 (except translation files) 
PLUS pull 512 (fix sidebar) Commits on Jan 8, 2018 
PLUS pull 528 (memory leak) Commits on Jan 18, 2018 
PLUS pull 527 (name editor) Commits on Jan 17, 2018

my apologies, that i didn't see the 1st version, but I know I like this one. (including the fix for the field geometry/overlapping. )

lordemannd:
I also tried it on debian, it works great.
(can it be adapted for inclusion in 4.2.7?)

GRAMPS: 5.0.0-alpha3 + all current changes on gramps50 branch
Python: 3.6.4 (default, Jan  5 2018, 02:13:53) [... 
BSDDB: 6.1.0 (5, 3, 28) 
sqlite: 3.21.0 (2.6.0)
LANG: en_US.UTF-8
OS: Linux
Distribution: 4.14.0-3-amd64

gramps/gui/editors/displaytabs/embeddedlist.py
```diff
@@ -42,6 +42,7 @@
 # Gramps classes
 #
 #-------------------------------------------------------------------------
+from gramps.gui.widgets.cellrenderertextedit import CellRendererTextEdit
```
Nick-Hall: Use a relative import here.

gramps/gui/widgets/cellrenderertextedit.py
```diff
@@ -0,0 +1,63 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2015-2016 Gramps Development Team
+# Copyright (C) 2016      Nick Hall
```
Nick-Hall: Add your own copyright.  I'm not sure my copyright is needed.  Did you copy this file from somewhere?

gramps/gui/widgets/cellrenderertextedit.py
```diff
@@ -0,0 +1,63 @@
+#
```
Nick-Hall: Include in the POTFILES.skip file.

gramps/gui/editors/displaytabs/surnametab.py
```diff
@@ -122,10 +122,10 @@ def build_columns(self):
         renderer.connect('edited', self.on_orig_edited, self._column_combo[3])
         # add to treeview
         column = Gtk.TreeViewColumn(name, renderer, text=self._column_combo[3])
-        column.set_resizable(True)
+        column.set_resizable(False)
```
Nick-Hall: The name origin column is no longer easily resizable.  This may be annoying for long text in translations or custom entries.

lordemannd:
seeing @Nick-Hall 's comments about esc key, and using the Commits on Jan 17, 2018,
I notice: esc key behavior is different depending on which field is active.  Editing a previously existing name:
esc in Given name field, closes the dialog saving any changes in Given name.
esc in Surname, Prefix, Connector, etc. changes the focus and saves the field contents,
a 2nd esc then closes the dialog, any changes in Surname etc. are saved.
BUT in both cases, that name's entry in the list on the names tab, doesn't get updated. if you re-edit the name you can see the changes that were entered before the escape.
NOW re-edit the name and close using the OK button, now the name list is updated.

Nick-Hall:
@lordemannd The severity of the Esc key problem will depend on whether the user primarily uses the keyboard or mouse.  We should leave the keyboard behaviour unchanged otherwise we will probably get complaints.

lordemannd:
@Nick-Hall, opps I hope I didn't sound argumentative, I agree that the esc key behavior should be made to work right.
If you look at 4.2.6 and use the name editor to make a change to a Given name, then use esc to exit the dialog, you will see the same wrong behavior. The name list shows no change but by editing the name you can see that the changes are preserved. 
is it possible that the meaning of the esc key is to suspend the edit, saving the changes for some later OK that commits them? 

I only meant vote for including the improvements even if they don't change the previous esc key behavior.

gramps/gui/editors/displaytabs/embeddedlist.py
prculley: FWIW I think this policy needs adjustment.  Relative imports seem to me to be fine when you are importing from the same directory (one '.').  But I think they are confusing and hard to read if you are going deeper.  If we ever have to restructure our sources again, the multi-dot syntax will NOT make for an easy search/replace.

But I fixed it.

prculley:
@Nick-Hall All the fixes you requested are as suggested.
I note that I did not do anything about the inconsistency of using 'esc' on the fields that are not part of the Surname Editor; they still update the name even while closing the name editor.  With the noted lack of update of the 'Preferred Name' in the main Person edit dialog.  I regard this as another issue, possibly deserving another PR.

The 'esc' key in the surname editing fields now behaves the same as in prior versions.

I also fixed the POTFILES and copyright issues, You are correct, I copy/pasted from somewhere and forgot to change the names.

PR #529, opened 2018-01-19
Some bug commits originally done on gramps50 branch, for gramps42

prculley:
These have been cherry-picked from my commits to gramps50 where they seemed to be useful for gramp42.  I did not take all of my commits, some were only for bugs against gramps50, and some did not apply cleanly due to other people's changes to gramps50 that were not in gramps42.

As these have already been approved once, I will merge them myself after a few days to make sure they get into the 4.2.7 release.

I keep hoping that we will soon be done with gramps42 maintenance; it is a pain to apply bug fixes to multiple branches...

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -33,11 +33,11 @@ all required and optional dependencies. Missing dependencies will
 result in runtime errors.
 
 To build and install, whether from a tarball or git repo:
-   python setup.py build
-   sudo python setup.py install
+   python3 setup.py build
+   sudo python3 setup.py install
 
 You can avoid using sudo for the install step by specifying a prefix to which you have write priviledge. The default is /usr/local, which is usually owned by root. You can learn of more options with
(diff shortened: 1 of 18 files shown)


Nick-Hall:
@prculley Thanks.

Bugs reported in v4.2 should be fixed in the gramps42 maintenence branch until v5.0 is released.  However, I understand that there will be exceptions to the rule.

prculley:
@Nick-Hall I've been using and doing debug almost exclusively in gramps50 branch, as that is where I hope we end up.  I've also always planned to back-port to gramps42 (or forward port in the few cases where that was necessary), but prefer to wait until the initial commits are accepted (so I don't have to redo things twice).

Now that I know a release is imminent, I'm doing this.

Same thing for addons, except even worse, since there are three branches to update; wish we could do a merge up from gramps50 to master to deal with that...

PR #530, opened 2018-01-21
Main Window menu 'Add' keyboard Acellerators (Gramps50)

prculley:
Two commits, 
* The bug fix changes the keyboard accelerators so <shift>-e ('E') doesn't activate the Event editor, as when typing in Filter or other Gramplet boxes capital 'E' is sometimes used.  Made all the 'Add' accelerators <Alt><Shift>.  https://gramps-project.org/bugs/view.php?id=10379
* Also changed 'Add' menu so the subsidiary 'New' is removed (for now) as it seems redundant uless/until additional key accelerators are defined.


sam-m888:
Thank you

sam-m888:
I will fix the duplicate keybindings (Person and Place) and (Family and Repository) are identical


duplicate-keybinding-add-new-menu

PR #531, opened 2018-01-25
Fix Book XML handler to deal with unusual characters in Book name

prculley:
Fixes #10387
Three issues;
1) Book name could have legal characters that messed up the html ('&')
2) Book could have characters that could not be stored in the system default encoding.  The latter issue is more likely in Windows with its 'code page' default encoding.
3) 'Generate Book' dialog had a bad transient parent so it sometimes ended up under the main book dialog.
4) Book report name could have legal characters that messed up the html ('&')

First commit deals with first two issues.  Changed book.xml file to be stored utf8, and added necessary 'escape'.  To deal with older previously stored non-utf8 book.xml, added a check for UnicodeDecodeError, with a retry using system default encoding.

Second commit deals with bad transient parent.

Third commit fixes bad report name.

### DIFF
--- a/gramps/gen/plug/report/_book.py
+++ b/gramps/gen/plug/report/_book.py
@@ -495,14 +495,14 @@ class BookList:
         """
         Saves the current BookList to the associated file.
         """
-        with open(self.file, "w") as b_f:
+        with open(self.file, "w", encoding="utf-8") as b_f:
             b_f.write("<?xml version=\"1.0\" encoding=\"utf-8\"?>\n")
             b_f.write('<booklist>\n')
             for name in sorted(self.bookmap): # enable a diff of archived copies
                 book = self.get_book(name)
-                dbname = book.get_dbname()
+                dbname = escape(book.get_dbname())
                 b_f.write('  <book name="%s" database="%s">'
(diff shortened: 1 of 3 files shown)


lordemannd:
@prculley, I don't know if you are ready for any testing yet, if it is to soon, just ignore this. I tried the 2nd commit, and on my windows machine, I don't see any change regarding the naked &. When added to the output file name, it doesn't get converted to "&amp;", so when Gramps re-reads the file, the list of books is appears empty.

prculley:
@lordemannd Thanks for noticing this.  I misunderstood your original bug report, thinking that the name of the 'book' was the issue (it had the same html escape problem).  I now realize that it was the name of the 'report' generated that you were talking about.  Hope this is all the spots that this can occur...

lordemannd:
I have taken the Jan 26 commit. It looks good to me. Solves the issues I know about.
If there is a 4.2.7 can it also have these changes?

prculley:
If this is accepted, I will back-port to 4.2 branch, hopefully before 4.2.7 is released.

Nick-Hall:
@prculley Thanks.  Please go ahead and back-port.

PR #533, opened 2018-01-26
Fix view buttons order keeps changing on different startups

prculley:
Fixes #10391

Root cause, the creation of the plugins view (and other) lists had an intermediate 'set' to store the data.  This was a vestige of an old (no longer allowed) capability to have multiple plugins with the same ID, apparently different versions.  See https://github.com/gramps-project/gramps/commit/12918718f9fe702f41e997cdad983613c87ab386 for the original capability addition.

Currently several places in the code prohibit this, so the 'set' is no longer needed, and it does mess up the repeatability of the plugin order.

### DIFF
--- a/gramps/gen/plug/_pluginreg.py
+++ b/gramps/gen/plug/_pluginreg.py
@@ -1245,8 +1245,7 @@ class PluginRegister:
         """
         Return a list of :class:`PluginData` that are of type ptype
         """
-        return [self.get_plugin(id) for id in
-                set([x.id for x in self.__plugindata if x.ptype == ptype])]
+        return [x for x in self.__plugindata if x.ptype == ptype]
 
     def report_plugins(self, gui=True):
         """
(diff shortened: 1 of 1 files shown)


lordemannd:
I checked it on 5alpha3 and it looks good.
Will there be a version for 4.2.6/7?

prculley:
@lordemannd Unfortunately, as the 4.2.x branch does not contain a number of other performance enhancement patches that were done in 5.0.x.  So I cannot apply the same (relatively simple) fix to gramps 4.2.7.  I looked at it for a while and decided not to pursue this.  Sorry...

lordemannd:
Good call, the original report was against 4.2.6, the reporter considered it low priority. 
(I liked the effort he went through to add a pretty complete description, well enough written that google translate could get the point across)

mlitvinenko:
Guys, thank you for fixing it in further versions of your great product. If I will see bugs in future, I will describe them in English. I saw the bugtracking webinterface in Russian and thought that it will be ok to write in the same language.

prculley:
Google Translate is our (Americans at least) friend...

PR #535, opened 2018-01-26
Add support for new genealogy tree report category

Nick-Hall:
Issue #10223.


Nick-Hall:
@wroldwiedbwe It looks like I need to specify utf-8 encoding for Windows.

wroldwiedbwe:
@Nick-Hall Thanks, Works for me and creates a "FamilyTree1.tex" file. How should I view the **tex** file or is something missing in Gramps AIO for these reports?

wroldwiedbwe:
No time to play around with all those; so I'll wait until the AIO has them 👍

PR #536, opened 2018-01-27
Fix Book XML handler to deal with unusual characters in Book name Gramps42

prculley:
Fixes #10387, same as https://github.com/gramps-project/gramps/pull/531 only for gramps42.
Three issues;

1) Book name could have legal characters that messed up the html ('&')
2) Book could have characters that could not be stored in the system default encoding. The latter issue is more likely in Windows with its 'code page' default encoding.
3) 'Generate Book' dialog had a bad transient parent so it sometimes ended up under the main book dialog.
4) Book report name could have legal characters that messed up the html ('&')

First commit deals with first two issues. Changed book.xml file to be stored utf8, and added necessary 'escape'. To deal with older previously stored non-utf8 book.xml, added a check for UnicodeDecodeError, with a retry using system default encoding.

Second commit deals with bad transient parent.

Third commit fixes bad report name.

### DIFF
--- a/gramps/gen/plug/report/_book.py
+++ b/gramps/gen/plug/report/_book.py
@@ -457,79 +457,98 @@ class BookList(object):
         """
         Saves the current BookList to the associated file.
         """
-        f = open(self.file, "w")
-        f.write("<?xml version=\"1.0\" encoding=\"utf-8\"?>\n")
-        f.write('<booklist>\n')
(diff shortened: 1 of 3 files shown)


lordemannd:
@prculley  I have test these changes on on 4.2.6 and it functions well.

sam-m888:
Thank you

PR #538, opened 2018-02-05
Fix several intl date displayers for missing parameter.

prculley:
Fixes #10196

It appears that when the 'inflect' parameter and capability was added to the date displayer, Paul F. did not realize that changes would also have to be made to any _display_gregorian methods as well. I'm not familiar with the purpose of the 'inflect' parameter, but I think we can simply ignore it for those languages that have not already been modified. It appears that we should do like was done for the french version in _date_fr;
change
def _display_gregorian(self, date_val):
to
def _display_gregorian(self, date_val, **kwargs):

This should be forward ported to gramps50 when accepted.

### DIFF
--- a/gramps/gen/datehandler/_date_de.py
+++ b/gramps/gen/datehandler/_date_de.py
@@ -221,7 +221,7 @@ class DateDisplayDE(DateDisplay):
         )
         # this definition must agree with its "_display_gregorian" method
 
-    def _display_gregorian(self, date_val):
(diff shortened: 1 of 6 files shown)


prculley:
Note that this bug only asserts when the language settings are such that the date is translated; on my test system I had to set LC_TIME to 'de' to (I was testing the German) to get the failure.  I had previously set LANGUAGE to 'de.utf-8', but that was apparently insufficient.

PR #540, opened 2018-02-05
Fix Family Lines Graph when 'unknown' places are present

prculley:
Fixes #10402

Current code initializes 'place_text' to None; then later on (after potentially setting value to a string) tries to do a string replace on it.  When the place type is 'Unknown' the 'place_text' never is set.  Looking over the code it appears initialization to empty string works just as well; there are several test of the (potential) str value for any value, they should fail with empty str as well as None.

Works with Example db and bug reporter's source Gedcom file.,


lordemannd:
it does the trick for me.

PR #541, opened 2018-02-05
Accels

jralls:
This is a two-fer: It fixes both [bug 6943](https://gramps-project.org/bugs/view.php?id=6943) and the part of the fix to [10231](https://gramps-project.org/bugs/view.php?id=10231) that had to be reverted.

With the first commit mnemonics are operated with ctrl-alt instead of just alt so that they don't interfere with typing non-ascii characters. The second changes all instances of GDK_CONTROL_MASK to GDK_META_MASK (which is what Gdk maps the command key to in the Quartz backend).  That works on all versions of Gtk3, unlike GDK_MODIFIER_INTENT_PRIMARY_MASK.. and saves a little indirection in the Gtk code in the process. (Part of the motivation is that I've had it shoved home yet again that Gtk+ > 3.14 doesn't seem to get along well with PyGobject on MacOS, see [bug 10185](https://gramps-project.org/bugs/view.php?id=10185) and friends.)

### DIFF
--- a/gramps/gui/dialog.py
+++ b/gramps/gui/dialog.py
@@ -35,6 +35,7 @@ _LOG = logging.getLogger(".dialog")
 #-------------------------------------------------------------------------
 from gi.repository import GObject
 from gi.repository import Gtk
+from gi.repository import Gdk
 from gi.repository import GdkPixbuf
 
 #-------------------------------------------------------------------------
@@ -46,6 +47,7 @@ from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
 from gramps.gen.const import ICON, URL_BUGHOME
 from gramps.gen.config import config
+from gramps.gen.constfunc import is_quartz
 from .glade import Glade
 from .display import display_url
 
@@ -506,6 +508,12 @@ def main(args):
     win = Gtk.Window()
     win.set_title('Dialog test window')
     win.set_position(Gtk.WindowPosition.CENTER)
+    #Set the mnemonic modifier on Macs to alt-ctrl so that it
+    #doesn't interfere with the extended keyboard, see
+    #https://gramps-project.org/bugs/view.php?id=6943
+    if is_quartz():
+        win.set_mnemonic_modifier(
+            Gdk.ModifierType.CONTROL_MASK | Gdk.ModifierType.MOD1_MASK)
     def cb(window, event):
         Gtk.main_quit()
     win.connect('delete-event', cb)
--- a/gramps/gui/glade.py
+++ b/gramps/gui/glade.py
@@ -46,6 +46,7 @@ from gi.repository import Gtk
 #
 #------------------------------------------------------------------------
 from gramps.gen.const import GLADE_DIR, GRAMPS_LOCALE as glocale
+from gramps.gen.constfunc import is_quartz
 
 #------------------------------------------------------------------------
 #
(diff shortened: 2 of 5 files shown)


lordemannd:
i tried these changes on windows and see the following problem on startup:
40845: ERROR: grampsgui.py: line 395: 
Gramps failed to start. Please report a bug about this.
This could be because of an error in a (third party) View on startup.
To use another view, don't load a Family Tree, change view, and then load your Family Tree.
You can also change manually the startup view in the gramps.ini file 
by changing the last-view parameter.
  File "C:\Program Files\GrampsAIO64-5.0.0-alpha3\gramps\gui\grampsgui.py", line 363, in __startgramps
    Gramps(argparser)
  File "C:\Program Files\GrampsAIO64-5.0.0-alpha3\gramps\gui\grampsgui.py", line 285, in __init__
    self._vm.post_init_interface()
  File "C:\Program Files\GrampsAIO64-5.0.0-alpha3\gramps\gui\viewmanager.py", line 779, in post_init_interface
    self.__open_activate(None)
  File "C:\Program Files\GrampsAIO64-5.0.0-alpha3\gramps\gui\viewmanager.py", line 1177, in __open_activate
    dialog = DbManager(self.uistate, self.dbstate, self, self.window)
  File "C:\Program Files\GrampsAIO64-5.0.0-alpha3\gramps\gui\dbman.py", line 166, in __init__
    self.glade = Glade(toplevel='dbmanager')
  File "C:\Program Files\GrampsAIO64-5.0.0-alpha3\gramps\gui\glade.py", line 148, in __init__
    data = builder_file.read().replace('\n', '')
  File "AIO/encodings/cp1252.py", line 23, in decode
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 3475: character maps to <undefined>

it looks like file dbman.glade is being read and the offending character might be in the line:
<property name="invisible_char">●</property>

is there some thing else i need? or it is unfair to test on windows?

jralls:
None of the above and thanks very much for testing on Windows!

I should have set an encoding on the read; without an encoding it assumes the local terminal charset, which Windows persists in setting to the antiquated DOS code pages.

jralls:
I've set encoding to utf-8 and force-pushed. Please try it again.

lordemannd:
now it starts up normally on Windows 10. since most of the issues look like Mac issues, I can't speak to how that part functions. Is there something specific I should check on Windows?

jralls:
As you observed, it's all about Mac, so if keyboard shortcuts all work they way they always have on Windows then it's good.

gramps/gui/dialog.py
```diff
@@ -506,6 +508,11 @@ def main(args):
     win = Gtk.Window()
     win.set_title('Dialog test window')
     win.set_position(Gtk.WindowPosition.CENTER)
+    #Set the mnemonic modifier on Macs to alt-ctrl so that it
+    #doesn't interfere with the extended keyboard, see
+    #https://gramps-project.org/bugs/view.php?id=6943
+    if(is_quartz()):
```
Nick-Hall: C:514, 0: Unnecessary parens after 'if' keyword (superfluous-parens)

gramps/gui/dialog.py
```diff
@@ -506,6 +508,11 @@ def main(args):
...
+        win.set_mnemonic_modifier(Gdk.ModifierType.CONTROL_MASK | Gdk.ModifierType.MOD1_MASK)
```
Nick-Hall: C:515, 0: Line too long (93/80) (line-too-long)

Nick-Hall:
Also the format of your commit messages may mean that the bug report will not be automatically closed.

The regexp is currently set to:

/(?:fixe?d?s?|resolved?s?)+\s*:?\s+(?:#(?:\d+)[,\.\s]*)+/i

gramps/gui/dialog.py
jralls: Heh, my C++ bias is showing again...

SNoiraud:
This PR change the place of the Help button. See #10585

PR #542, opened 2018-02-09
Fix Geography views for bad 'dbstate.is_open()' test

prculley:
Fixes #10417

Apologies; thought I tested that back-port from gramps50...


lordemannd:
I tried on Windows (AIO from 4.2.6, plus 4.2.7 tag from git, plus these changes) seems to be working.
this was reported to the console:
(grampsd.exe:12864): OsmGpsMap-CRITICAL **: Map source setup called twice

PR #543, opened 2018-02-09
Fix Place Tree view for proper display after filter cleared.

prculley:
Fixes #10416
The recent addition of the 'Within' filter for Places caused the get_filter method to not recognize when a filter had been reset.  As a result, once a filter was used, it continued to be used even though all the parameters were cleared.  During tree build, the presence of the filter changes the algorithm so that the tree doesn't get constructed right.  (Which may be a different bug).

This fixes the sidebar filter so that when reset, no filter is used at all.

I also cleared the 'Within' back to an empty string, rather than '0' as the '0' was inconsistent with the default, never used setting, which was an empty string.


lordemannd:
with his db I was able to reproduce what he is seeing with a clean version of 5 alpha 3, Ic then tested with everything on gramps50 and this change, the issue is solved. great job!

PR #544, opened 2018-02-10
Fix character encoding for new tree category reports

Nick-Hall:
Fixes #10418.

### DIFF
--- a/gramps/gen/plug/docgen/treedoc.py
+++ b/gramps/gen/plug/docgen/treedoc.py
@@ -25,7 +25,7 @@
 #-------------------------------------------------------------------------
 from abc import ABCMeta, abstractmethod
 import os
-import shutil
+from subprocess import Popen, PIPE
 from io import StringIO
 import tempfile
 import logging
@@ -97,6 +97,7 @@ _NOTESIZE = [{'name': _("Tiny"), 'value': "tiny"},
 
 if win():
     _LATEX_FOUND = search_for("lualatex.exe")
+    DETACHED_PROCESS = 8
 else:
     _LATEX_FOUND = search_for("lualatex")
 
@@ -582,7 +583,7 @@ class TreeTexDoc(TreeDocBase):
         if self._filename[-4:] != ".tex":
             self._filename += ".tex"
 
-        with open(self._filename, "w") as texfile:
+        with open(self._filename, 'w', encoding='utf-8') as texfile:
             texfile.write(self._tex.getvalue())
 
 
@@ -605,11 +606,14 @@ class TreePdfDoc(TreeDocBase):
             self._filename += ".pdf"
 
         with tempfile.TemporaryDirectory() as tmpdir:
-            with open(os.path.join(tmpdir, 'temp.tex'), "w") as texfile:
-                texfile.write(self._tex.getvalue())
-            os.system('lualatex -output-directory %s temp.tex >/dev/null'
-                      % tmpdir)
-            shutil.copy(os.path.join(tmpdir, 'temp.pdf'), self._filename)
+            args = ['lualatex', '-output-directory', tmpdir,
+                    '-jobname', self._filename[:-4]]
+            if win():
+                proc = Popen(args, stdin=PIPE, stdout=PIPE, stderr=PIPE,
+                             creationflags=DETACHED_PROCESS)
+            else:
+                proc = Popen(args, stdin=PIPE, stdout=PIPE, stderr=PIPE)
+            proc.communicate(input=self._tex.getvalue().encode('utf-8'))
 
 
 #------------------------------------------------------------------------------


Nick-Hall:
Needs testing on Windows.

wroldwiedbwe:
@Nick-Hall Thanks, Works for me and creates a "FamilyTree1.tex" file. How should I view the **tex** file or is something missing in Gramps AIO for these reports?

adamvazquez82:
Same error on 4.2.6 using the the files from [10223#c53662](https://gramps-project.org/bugs/view.php?id=10223#c53662)

```
2018-02-10 13:26:36.113: ERROR: _reportdialog.py: line 739: Failed to run report.
  File "C:\Program Files\GrampsAIO64-4.2.6\gramps\gui\plug\report\_reportdialog.py", line 704, in report
    MyReport.end_report()
  File "C:\Program Files\GrampsAIO64-4.2.6\gramps\gen\plug\report\_reportbase.py", line 87, in end_report
    self.doc.close()
  File "C:\Program Files\GrampsAIO64-4.2.6\gramps\gen\plug\docgen\treedoc.py", line 585, in close
    texfile.write(self._tex.getvalue())
  File "AIO/encodings/cp1252.py", line 19, in encode
UnicodeEncodeError: 'charmap' codec can't encode characters in position 1892-1899: character maps to <undefined>

```

bpisoj:
That fixes unicode errors!

Generated pdf looks like it misses half of page, all tree types looks incomplete.

Can not find anything suspicious in log file except this line
LaTeX Warning: Unused global option(s):
    [portrait].
Note that i am generating report in a4 portrait mode.

Nick-Hall:
@bpisoj The diagram doesn't scale to fit the page.  For a large tree you will have to select a large paper size.

Can you reproduce the problem using the Gramps example database?

I also expect the [command](https://github.com/gramps-project/gramps/blob/maintenance/gramps50/gramps/gen/plug/docgen/treedoc.py#L610) that generates the _pdf_ file directly from Gramps to fail in Windows.

Is LaTeX included in the AIO?

bpisoj:
You are right about "diagram doesn't scale to fit the page", if i choose large inaf page it works but how would i know which page size diagram needs. Without proper scaling i don't see much use of those trees.

Such redirection will produce "The system cannot find the path specified." error.
I just generate tex file and let gramps open it in their default viewer which automaticaly convert it to pdf.

AIO curently not include any of requirements. LaTeX is multiple bigger then hole AIO.

Nick-Hall:
These new _genealogytree_ reports provide an alternative to _graphviz_ reports.  They each have advantages and disadvantages.  I think that the layout and presentation of the new reports is good, but may involve some iteration to get right.

What advice should we give Windows users?  In PR #535, @wroldwiedbwe said he will wait for the AIO to contain LaTeX.  Clearly this is not practical.  Is the installation fairly easy?

I can disable the _pdf_ output option for Windows if it not useful.

bpisoj:
Yes genealogytree is very good piece of software with amazing documentation and yours reports already look nice just needs scaling sorted out. 

Hard to say about easines of installation (it easy for me) but keep in mind that aunt Martha have problem with AIO install also.

Pdf output is useful just need to be portable and dev/null isn't.
os.system('lualatex -output-directory %s temp.tex >os.devnull' % tmpdir)
works for me.

bpisoj:
Opsss it will not work that way.
Windows use NUL so how about using that?

`if win():
  os.system('lualatex -output-directory %s temp.tex >NUL' % tmpdir)`

QuLogic:
The best solution is to use `subprocess` and not `os.system` at all.

Nick-Hall:
Updated to use `subprocess`.  Needs testing on Windows.

bpisoj:
With that Gramps become unrensponsive.
My aproach with "NUL" works in both gui and cli mode but in gui mode console window will pop up briefly. To prevent that we can use Popen with creationflags=8 which stands for "DETACHED_PROCESS" like in GraphView

Popen(['lualatex', '-output-directory', tmpdir, 'temp.tex'],
  creationflags=8, stdin=PIPE, stdout=PIPE,
  stderr=PIPE).communicate()

bpisoj:
Sorry for closing, wrong button pushed!

Nick-Hall:
Updated to use `Popen` and remove the need for a temporary file.  Please test on Windows.

lordemannd:
on windows:
- no option to include private data?
- generating Grandparent report (center person no grandparents)
1087224: ERROR: _reportdialog.py: line 745: Failed to run report.
  File "C:\Program Files\GrampsAIO64-5.0.0-alpha3\gramps\gui\plug\report\_reportdialog.py", line 706, in report
    my_report.write_report()
  File "C:\Users\dave\AppData\Roaming\gramps\gramps50\plugins\GenealogyTree\grandparent.py", line 122, in write_report
    self.write_subgraph(0, 'child', [handle1], self.f1_handle)
UnboundLocalError: local variable 'handle1' referenced before assignment

Nick-Hall:
@lordemannd How did you install LaTeX?  What is the location of your `lualatex.exe`?

I'll add a check for no grandparents and the Grandparent report.

lordemannd:
@Nick-Hall, sorry that i haven't got that far. i have made several attempts ending in too much frustration.
I am going to try again, stepping through the instructions in the message:
[Gramps-devel] LaTeX genealogytree reports on Windows

Nick-Hall:
@lordemannd There is a newer version of MikTeX available.  Try clicking the button on their [download](https://miktex.org/download) page.

lordemannd:
program is located at:
C:\Users\dave\AppData\Local\Programs\MiKTeX 2.9\miktex\bin\x64\lualatex.exe

for 1st sandclock I got this error:
! Illegal parameter number in definition of \gtrDBname.
<to be read again>
1
l.2477 }
?
! You can't use `macro parameter character #' in horizontal mode.
\gtrDBname ->Antoine ##
                       1 \surn {St. Germain}
l.2477 }
?

bpisoj:
@Nick-Hall
Working good on Windows, please commit it.

Other bugs like LaTeX 10 special characters and grandparent report subgraph deserve bug-ticket of thier own.

PR #545, opened 2018-02-12
Some dates are not translated in webcal

SNoiraud:
This also fixes a link problem in the year glance page.

Fixes #10425


SNoiraud:
I think I need to simplify webcal and reuse a maximum of code from the narrative web => 5.1

SNoiraud:
The last patch doesn't solves all problems. Depending on the LANGUAGE when gramps starts, we may have some problems.

SNoiraud:
To solve this problem, all file names for month are set to lower case.
Tested with gramps in French and created the report for japanese, danish, finnish and Italian.
I have no more problems.

PR #546, opened 2018-02-14
Escape special LaTeX characters

Nick-Hall:
Attempt to fix bug [10433](https://gramps-project.org/bugs/view.php?id=10433).

### DIFF
--- a/gramps/gen/plug/docgen/treedoc.py
+++ b/gramps/gen/plug/docgen/treedoc.py
@@ -459,6 +459,8 @@ class TreeDocBase(BaseDoc, TreeDoc):
             media = db.get_media_from_handle(mediaref.ref)
             path = media_path_full(db, media.get_path())
             if os.path.isfile(path):
+                if win():
+                    path = path.replace('\\', '/')
                 self.write(level+1, 'image = {%s},\n' % path)
                 break # first image only
         self.write(level, '}\n')
--- a/gramps/gen/plug/docgen/treedoc.py
+++ b/gramps/gen/plug/docgen/treedoc.py
@@ -26,6 +26,7 @@
 from abc import ABCMeta, abstractmethod
 import os
 import shutil
+import re
 from subprocess import Popen, PIPE
 from io import StringIO
 import tempfile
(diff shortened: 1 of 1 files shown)


lordemannd:
Now I'm getting this warning, on a file names that include these characters.

LaTeX Warning: File `C:/Users/dave/Desktop/- GRAMPS media/Bratlien/Alvin&Annett
e-wedding.jpg' not found on input line 583.

! Misplaced alignment tab character &.
\filename@base ->Alvin&
                       Annette-wedding
l.583 }

lordemannd:
oops, unfair test, I had turned on images. (that maybe a separate issue)

without images, the escaping seems to be working and the escaped characters appear as the normal character in the pdf file.

bpisoj:
For me gramps fail to produce pdf but create valid tex file which can be converted to pdf by hand

Nick-Hall:
The fix in PR #544 doesn't work.  It creates the output file in a temporary directory.  I was just seeing the contents of previous _pdf_ files.

lordemannd:
on windows, the escape part seems to be working and maybe because running lualatex manually, we avoid the temp file problem.

lordemannd:
windows: when images are included, the image file names don't seem to be working, the look like the directory separator "\" needs to be escaped in the file name, also trouble with names which include other character sets like "test image ÄÖÜäöüßŸÿ german text.JPG"

Nick-Hall:
The _pdf_ output should now be fixed, and this PR works for me.  Please test.

I tried escaping the image file name, but it failed.

lordemannd:
Windows 10: if I try to output report to pdf file, Gramps becomes unresponsive, using no CPU, memory use unchanging. had to be killed.

I looked at the image file names, yes, this escape isn't working here either. I commented out the escape for the image path.

Without the escapes on the file path, in the .tex file the path appears to be correct.
"    image = {C:\Users\dave\Desktop\CommonMurres.jpg},"
But when I run the .tex into lualatex, I get this message:
"! Undefined control sequence.
<argument> C:\Users
         \dave \Desktop \CommonMurres
l.98 }"

does this look like lualatex is having the issue?

lordemannd:
regarding the images: it would be good if it worked like Narrated Web Site Report.
- pick the image that is 1st in the gallery for the person, the same one seen on the person editor
- the cropping associated with that image should be applied, the same as the seen on the person editor

Nick-Hall:
Updated to replace backslash in image path for Windows.

@lordemannd The four reports that I have written only have basic functionality.  Please feel free to open feature requests for enhancements.

PR #547, opened 2018-02-15
Apply places changes to the narrativeweb.

SNoiraud:
Also change order between locations and locations alternate names.

Fixes #10427

### DIFF
--- a/gramps/plugins/webreport/basepage.py
+++ b/gramps/plugins/webreport/basepage.py
@@ -2578,7 +2578,7 @@ class BasePage: # pylint: disable=C1001
         if altloc:
             tbody += Html("tr") + Html("td", "&nbsp;", colspan=2)
             trow = Html("tr") + (
-                Html("th", self._("Alternate Locations"), colspan=1,
+                Html("th", self._("Alternate Names"), colspan=1,
                      class_="ColumnAttribute", inline=True),
                 Html("th", self._("Language"), colspan=1,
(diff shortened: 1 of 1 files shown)


SNoiraud:
@Nick-Hall 
get_alternate_locations always return an empty list.
The set_alternate_locations in only used in the sanitize_place method in the gramps/gen/proxy/private.py source file
perhaps we need to remove these two methods : set_alternate_locations and get_alternate_locations

SNoiraud:
Only the narrative web use get_alternate_locations.
For the addons, only DynamicWeb and PersonEverything use it.

Nick-Hall:
@SNoiraud Alternate locations in an existing family tree or imported from a backup are still visible in the [place](https://github.com/gramps-project/gramps/blob/master/gramps/gui/editors/editplace.py#L243) and [place reference](https://github.com/gramps-project/gramps/blob/master/gramps/gui/editors/editplaceref.py#L239) editors.  For new family trees the alternate locations editor tab is hidden.  Sometimes the alternate locations are accessed via `alt_loc` variable or `add_alternate_loaction` method.

I would eventually like to remove the alternate location functionality, but this needs to be discussed on the list first.

SNoiraud:
OK, so it's not the same thing. I put these alternate locations back.
I found an example of this situation. tested with the next commit. it works.

SNoiraud:
This what we get now : 
alternatelocations 3

PR #548, opened 2018-02-16
Add Buchheim tree ancestor trees to Narrated Web Report

papadeltasierra:
The ancestor trees of the narrated web report use a simple binary model and leave large amounts of whitespace. The space and layout also depends on the maximum depth (generations) selected.
The Buchheim/Walker algorithm is slower but creates a better layout that is based on the presence of nodes, is independent of 'maximum allowed depth' (it depends on the actual depth, which may be less than the allowed maximum) and does not leave acres of whitespace when there are no nodes.
I have implemented this as a 'compact' option, leaving the old algorithm unchanged for those who prefer it.  This feature consists of 3 checkins:
- The actual code
- A sample set of trees as a single XML; these demonstrate the algorithm and can be used to compare original and compact output.
- The PO translation file (apologies; missed this from the first checkin).

See the [design notes](https://gramps-project.org/wiki/index.php?title=Narrated_Website_Ancestry_Tree_Design_Notes) in the wiki.


Nick-Hall:
Thanks for you contribution.  The build is failing due to trailing whitespace.

sam-m888:
Excellent, tested works as advertised. Thank you 👍 

narweb-option-box-for-generate-compact-graph-51


sidebyside

 - [ ] Add to documentation (when accepted and new manual rolled.)

gramps/plugins/webreport/person.py
```diff
@@ -585,9 +588,12 @@ def individualpage(self, report, title, person):
             if sect13 is not None:
                 individualdetail += sect13
 
-            # display ancestor tree
+            # display ancestor tree
```
sam-m888: Remove trailing white space from here.

gramps/plugins/webreport/buchheim.py
```diff
@@ -0,0 +1,269 @@
...
+# The following modifications are noted:
+#
+# - The root node is 'west' according to the later nomenclature
+#   employed by Walker with the nodes stretching 'east'
+# - This reverses the X & Y co-originates of the Buchheim paper
+# - The algorithm has been tweaked to track the maximum X and Y
+#   as 'width' and 'height' to aid later layout
+# - The Buchheim examples track a string identifying the actual
+#   node but this implementation tracks the handle of the
+#   DB node identifying the person in the Grams DB.  This is done
```
sam-m888: Should "Grams" be "Gramps" ? or does it have an alternate meaning related to the paper?

adamvazquez82:
From my point of view you have fixed an annoying bug with the graph, who really wants spreadout graphs that make it difficult to have a proper overview?  @papadeltasierra Thank you 🥇 

I'd suggest seeing as the beta has not been released it be added as the default method and remove the extra option.

Also possibly move **buchheim.py** somewhere in Gramps as a library or utility that can be used by other reports if it is generic enough?

SNoiraud:
I have several remarks : 

- If you ask to generate a narrative for the descendants of X, The tree should not show his/her ascendants. I know this problem already existed. 

- Why did you modify media.py. Your modification break the logic. If you have no referenced media and unused media, you never saw them. I patched media.py yesterday in gramps50 for that.

- I think you can remove the old method : display_tree

- Are you sure you want to have 99 levels for the tree? The html page will be too large!

- Did you try all css files ?

lordemannd:
what am i doing wrong?
  File "/usr/lib/python3/dist-packages/gramps/plugins/webreport/narrativeweb.py", line 1729, in __add_more_pages
    default_path_name = config.get('paths.website-extra-page-name')
  File "/usr/lib/python3/dist-packages/gramps/gen/utils/configmanager.py", line 377, in get
    (section, setting))
AttributeError: No such config setting name: 'paths.website-extra-page-name'

sam-m888:
> what am i doing wrong?

@lordemannd Not sure; but none of the code here touches what your error shows, so looks like a separate issue, you will need to see if you can recreate and see which config options make it happen.

lordemannd:
@sam-m888, strange, I see essentially the same issue on windows and linux/debian. when I back out these changes, then I am back to working.

sam-m888:
@lordemannd Yes, the error refers to the new feature
https://github.com/gramps-project/gramps/pull/465 "*Add extra page to
narrativeweb.*" by @SNoiraud that is only available on gramps master; which
needs more testing by the looks of it as you have found out, how did you
trigger the error, I still can't recreate it? (best move this discussion to
its own issue if you can please?)

SNoiraud:
Are you sure the gramps/config.py contains : (lines 231)
register('paths.website-extra-page-uri', '')
register('paths.website-extra-page-name', '')
This was the commit : d677a1a785920fbdb87cfa6cfcdce9839661f3e1
If "Your extra page path" is not none, we add in the navigation tab the "Extra page name"
I tried with the papadeltasierra:buchheim branch and it works correctly for me.

lordemannd:
OOPS, I was too interested in seeing this feature and didn't notice it was on master, I was applying it against the gramps50 branch. Should I delete all my messages related to this issue to clean up the thread?

lordemannd:
@sam-m888, @SNoiraud, I am sorry for consuming your time, I have rewritten my original note (as it referred to in mantis) and deleted my intervening comments, and will delete this comment tomorrow.

papadeltasierra:
Questions from [SNoiraud[(https://github.com/SNoiraud)

Q:/ If you ask to generate a narrative for the descendants of X, The tree should not show his/her ascendants. I know this problem already existed.
A:/ I didn't attempt to fix existing problems such a this.  To be honest, I would not even have realised that people considered it was a problem :-).

Q:/ Why did you modify media.py. Your modification break the logic. If you have no referenced media and unused media, you never saw them. I patched media.py yesterday in gramps50 for that.
A:/ I patched it because I was unable to produce any output until I did.  I'll check the current master and assuming it already has your fix, back out my changes.

Q:/ I think you can remove the old method : display_tree
A:/ Providing people are happy to completely lose the old system, that's fine by me.  Are you 'the man' who can tell me 'go ahead'? 

Q:/ Are you sure you want to have 99 levels for the tree? The html page will be too large!
A:/ Agreed; would 10 be OK?  As far as the algorithm is concerned there is no limit.

Q:/ Did you try all css files ?
A:/ No - I'll try them out.  Is this why there seems to be support for 'shadow' but I'm not seeing any?  Only some CSS support this?

papadeltasierra:
Reponses to [adamvazquez82](https://github.com/adamvazquez82l)

I'll take a look.  I don't believe there is anything that is specific to displaying the ancestry tree in the buchheim and layout code (you need both - build a layout, pass it to the buchheim code).  One thing that someone else might like to add if we ever need it is the ability to draw the tree 'root at top, leaves below'.  This is not difficult but it would add a few if()s, hurting performance a little so I didn't bother doing it here.

papadeltasierra:
I have a question about the CSS.  The 'Visually Impaired' style seem to make the page narrow, chopping off the right hand part of the tree - is this expected?
Also, the 'No style sheet' option simply creates new pages, leaving the CSS unchanged at whatever it was before - is this also expected?

gramps/plugins/webreport/buchheim.py
papadeltasierra: Typo - should be Gramps.

gramps/plugins/webreport/person.py
papadeltasierra: OK, doing that now.  I presume this, and other changes, are still on my fork i.e. papadeltasierra/buchheim.  Let me know if not.

papadeltasierra:
[SNoiraud](https://github.com/SNoiraud), I took a look at the master branch of media.py and I still believe it will fail.  I believe even with zero media objects, execution can reach line 161 where it will hit this:
prev = sorted_media_handles[len(sorted_media_handles)-1]
Since there are zero sorted_media_handles, this creates an index of -1 into an empty array; this is the bug that I hit and which caused me to change media.py to step out if there are no media objects.
Do you want to take a look and let me know whether you agree and if so, do you want to apply and fix and then I will back out the changes that I made?

papadeltasierra:
Serge,

Just to let you now that I added a few comments and questions regarding the Buchheim stuff via GitHub last night.

Regards,
Paul DS.

SNoiraud:
> Q:/ If you ask to generate a narrative for the descendants of X, The tree should not show his/her ascendants. I know this problem already existed.
> A:/ I didn't attempt to fix existing problems such a this. To be honest, I would not even have realised that people considered it was a problem :-).

> Q:/ Why did you modify media.py. Your modification break the logic. If you have no referenced media and unused media, you never saw them. I patched media.py yesterday in gramps50 for that.
> A:/ I patched it because I was unable to produce any output until I did. I'll check the current master and assuming it already has your fix, back out my changes.

These  changes applied to gramps 5.0 and not in master. From time to time, there is a merge from gramps 5.0 into master. This is to avoid to make patches in two repositories. The media.py is the same into gramps 5.0 which is patched and master.

> Q:/ I think you can remove the old method : display_tree
> A:/ Providing people are happy to completely lose the old system, that's fine by me. Are you 'the man' who can tell me 'go ahead'?
I think your method is better than the oldest one, so I think you can do that. This feature is in master and it's too late to add it in 5.0. So we have time to test it.

> Q:/ Are you sure you want to have 99 levels for the tree? The html page will be too large!
> A:/ Agreed; would 10 be OK? As far as the algorithm is concerned there is no limit.

I tested with 12 levels, and I saw the limits seems to be 9. The problem is not the number of level, We can't move the tree to have the highest levels. It depends on the width of your browser and I think this was the reason why we had no more than 5 levels. Gramps we be contained with a maximum width of 800*600 because we have many users with small screens.

> Q:/ Did you try all css files ?
> A:/ No - I'll try them out. Is this why there seems to be support for 'shadow' but I'm not seeing any? Only some CSS support this?

We have another problem here. It already existed before. when you have people with pictures and others without, the lines between people are not placed correctly. I tried to do that, without success. I think we need to resolve that.

papadeltasierra:
Serge,

Thanks for the comments – I’ve added some more notes in-line below.

Paul DS.

SNoiraud:
This is what I see on my firefox browser. The tenth box is partialy visible.
new_tree 1

papadeltasierra:
I see.  I presume that you cannot scroll the browser screen to see the box?  This might imply that there is a bug in the HTML generation.  The graph is drawn in an ‘area’ (existing code but I have had to change the calculation of the area so the new calculation might be wrong.  There was also a bug in the original where the calculation didn’t allow for the _XOFFSET so I’ll see if I can test that further too.

BTW, are you familiar with the Firefox developer tools?  They’re very good at showing the actual HTML and highlighting the bit of the screen generated.  This is often a good way of spotting where, for example, something is drawn correctly but truncate because the parent element is too small.

I can run Firefox at home to I’ll give it a try when I look at the other issues.

Thanks,
Paul DS

SNoiraud:
This is surely a css problem. The css for the tree is : data/css/ancestortree.css

papadeltasierra:
Serge,


It's not the CSS but I think there are two bugs at play here.


  1.  If you look at the 'T' join just below Hilaire Baptiste Auguste (GIRAUD) you will see a 1 (?) pixel gap.  Minor but I'll try and figure out why as that's definitely something I did wrong.
  2.  The person boxes in your display does not have the '*' or '+ ...' cruft that I reported so someone has changed the code to not display them if empty.  However they have not altered the display logic to alter the relative placement of the lines and boxes and the boxes are now too short to reach where the lines are placed.


I can't be sure of the fix to #2 until I've looked at what has happened to the '*/+ ...' code but one possible solution springs to mind.  The lines typically try to place themselves in the 'center of a maximum size box' (this is, and always was, hard-coded not calculated) and this means that the lines are often below the actual center when a box is less than maximum size.  If we bias the line towards the top, the logic will work and won't care if the boxes grow.


Two other alternatives that might be possible, although I do not want to do this unless we really need have to, would be to either calculate the actual, box size (hard because we don't know what's going in it or how the renderer will size the text until later) or bias the boxes upwards as they grow (i.e. fix their bottom left corner instead of the top left corner) - and I'm assuming we can managed to do this in HTML; I'm no expert but I know some people who are 😉.


For now I'll bias the lines to the top of the boxes and we can see what people think.


Paul DS.


Sent from Outlook<http://aka.ms/weboutlook>


________________________________

SNoiraud:
For the tree, If we have no pictures, it works correctly.
If you display a few levels in which all people have a picture, it works well.
The problem occurs when we have some people with a picture and others without.
I already tried to correct this without success.

FYI, when we have a picture, we add the height of the picture which has always the same size.
I think the solution is to calculate the box size for that.

For # 2, I made a mistake : "+ ..." means always alive

SNoiraud:
Yes! it works.

Nick-Hall:
Rebased.

gramps/plugins/webreport/narrativeweb.py
```diff
@@ -1493,7 +1493,8 @@ def set_mtime(tarinfo):
                 try:
                     shutil.copyfile(from_fname, dest)
                     os.utime(dest, (mtime, mtime))
-                except:
+                except Exception as exception: # pylint: disable=broad-except
```
Nick-Hall: Although we don't have a policy regarding this, it is generally not a good idea to disable pylint warnings.  In this case the exceptions that we should probably be catching here are `IOError` and `OSError`.  It is useful to see these warnings when running pylint.  Someone may wish to improve the code at a later date.

It is also useful to put pylint improvements in a separate commit.  This makes the other changes easier to read.  However, if it not easy to separate them out, don't bother to spend time doing so in this PR.

papadeltasierra:
OK, I’ll update these and do the Pylint stuff separately next time.

Sorry,
Paul DS.

papadeltasierra:
Nick Hall, I've removed the pylint disable statements.

sam-m888:
> Nick Hall, I've removed the pylint disable statements.

@papadeltasierra You probably need to push the change :)

papadeltasierra:
Doh – if only you could get the staff.

Paul DS.

Nick-Hall:
This needs a rebase.

Nick-Hall:
Something went wrong in the rebase.  There are some merges we don't need and the build is failing with trailing whitespace again.

papadeltasierra:
Oh joy.  Will try and take a look tonight. It seemed to be a surprisingly difficult merge as if some of the code had been merged into master already – is that the case?

Whilst we’re on the subject, what is the end-game for pull requests and how is the build system using them?


  *   Are they being dynamically applied to the master branch in a special ‘all-pull-requests’ build somehow?
  *   Do they cease to be at the next cut of a release?

Thanks,
Paul DS.

SNoiraud:
There is an issue #0003425 with problem on boxes. Is it solved ?

papadeltasierra:
Serge,

Bottom line, whatever it did before, it will do with the Buchheim stuff but if there is overlap, it will be a little more likely because the nodes closer to the root on a sparse graph will be now closer than in the simple binary graph as they won’t be excessively spaced in anticipation of (great/grand)parental nodes that don’t actually exist.

The layout algorithm allows a fixed size for each box (plus inter-box space both vertically and horizontally) and takes no account of the true size of the boxes are so there are two extremes possible:


  1.  The boxes are not as big as the allowed size (I believe this is often the case) and so there is additional whitespace that a ‘perfect’ algorithm might remove;  this was why I had to tweak the connected lines to be ‘towards the top’ rather than ‘towards the bottom’ when the maximum size of the boxes was increased.
  2.  More data is written to the boxes than the allotted size allows.  It is possible that the HTML/CSS might grow the box to accommodate the text, making the box too large and potentially overlap with the following box (there is some spacing allowed so the box has to grow over that too) but it might also simply crop the text to fit the maximum box size.

I definitely see #1 and I’m most certainly not proposing we try and write a perfect algorithm.

For #2, I haven’t looked at the HTML/CSS to see if the boxes are bounded and would crop.  The problem might happen with a very long name and possible with lots of the extra data that is now possible in a box so if you can remind me what the other data is (dates of both & death?) I might be able to try it out and see whether Gramps truncates grows the box to fit.

Paul DS.

Nick-Hall:
Rebased.

SNoiraud:
I found a problem when I use "show half sibling"
the display_step_family call has one missing argument on line 1873 and 1877
I can add the patch if you want.

Nick-Hall:
@SNoiraud Thanks.  That would be good.

wroldwiedbwe:
Reapplied to my personal Gramps 5.0.1. Will be happy when this merged 👍

SNoiraud:
This will be included in 5.1 and not in 5.0.x

sam-m888:
Issue noticed by user:
[11382: Narrative web: Include ancestor's tree/family tree display looks weird (person boxes overlap)](https://gramps-project.org/bugs/view.php?id=11382)

ex3oct

PR #549, opened 2018-02-16
crash in narrative web if no unused media

SNoiraud:
Fixes #10438


lordemannd:
Looks like it solves the problem.
I tested on windows against GrampsAIO-5.0.0-alpha3-191-gf428069ac_win64.exe + recent changes

PR #552, opened 2018-02-19
[Debian ]Update deb directory

RossGammon:
Big tidy up of the debian/ directory for 5.0.0 beta-1. It is now almost exactly the same as what I would upload to Debian, except a couple of files that only make sense inside the Debian world.

It is exactly what I used to build the .deb package that was just uploaded to to the gramps github pages. Previously, each upload was done out of the git source tree, and had got out of sync (e.g. missing some history in the debian/changelog).

By the way, I had to patch one of the tests which failed to import a module. There is also a warning about a syntax error in the manpage which will need fixing before I can upload in Debian. Some tests working on travis fail for me in my debian/unstable chroot. I will follow up with PRs and bugs for these separately.


Nick-Hall:
Thanks.  Do we need commit deede677be407be41ff96517884b29f2d991a869 now that PR #553 has been merged?

RossGammon:
Correct, it is not required now. But it can safely be left there until the next release. It is standard Debian procedure to refresh any patches before each subsequent upload, and drop any that have been applied in the source since.

I will try and update the instructions on the wiki for building the Debian package in the next few days. I need to practise it in a clean VM to make sure I don't miss any steps (e.g. packages to install that I have forgotten that you need to install).

Nick-Hall:
OK.  Is this ready to merge?  I assume that you don't want to squash the commits.

RossGammon:
@ Nick

I'm sorry - I thought I saw a merge message in my mails.

Squashing is not necessary - there are no mistakes to hide - this time :-)

But if you prefer one simple commit to be merged, that is also fine with me. Actually, I see I have the power to "squash and merge" on the web interface. Let me know if you are happy for me to hit that button.

sam-m888:
Debian page on wiki

https://www.gramps-project.org/wiki/index.php?title=Debian

PR #554, opened 2018-02-23
Fix Report Options when used with empty db

prculley:
Fixes [#10457](https://gramps-project.org/bugs/view.php?id=10457)

Avoids HandleError when looking for a person in report options.


sam-m888:
Thank you.

PR #555, opened 2018-02-23
Privacy problem with the relation to the center person.

SNoiraud:
Bug #12345


PQYPLZXHGF:
Was this meant to be bug [10454](https://gramps-project.org/bugs/view.php?id=10454) and not 12345?

SNoiraud:
Yes, I don't know why there is 12345 here! I don't remember to have put 12345 in the commit.
This PR is really for bug #10454

PR #557, opened 2018-02-23
Fix export gpkg when media files have fractional timestamps.

prculley:
Fixes [#10424](https://gramps-project.org/bugs/view.php?id=10424)
I'm pretty sure this is a Python tarfile error.  Found by analyzing the code. If the media file has one or more timestamps with fractional seconds, it will fail. Seems that the default GNU_FORMAT code doesn't convert the time values to int.

Changed fix from using PAX_FORMAT back to GNU_FORMAT, and adding filter per @Nick-Hall suggestion.

May want to backport to 4.2.x if we are going to have another version.

### DIFF
--- a/gramps/plugins/export/exportpkg.py
+++ b/gramps/plugins/export/exportpkg.py
@@ -86,6 +86,12 @@ def writeData(database, filename, user, option_box=None):
     writer = PackageWriter(database, filename, user)
     return writer.export()
 
+
+def fix_mtime(tarinfo):
+    """ this fixes a bug in the python tarfile GNU_FORMAT where if mtime has
+    a fractional component, it fails."""
+    tarinfo.mtime = int(tarinfo.mtime)
+    return tarinfo
 #-------------------------------------------------------------------------
 #
 # PackageWriter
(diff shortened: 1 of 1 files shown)


prculley:
A user tested my patch:

> I made the patch you suggested and all is well now. 
> 
> I imported the .gpkg into a new family tree and the transfer appears to be perfect.
> 
> Thank you, Elvin Birth

sam-m888:
While it works, on windows and linux, when I use a PAX-format created backup on windows and extracted with 7-zip the archive contains a number of files that don't exist with backups created the previous way named "@PaxHeader" which when opened contain text like the following "27 mtime=1520029609.700665" so I'm not sure what needs to be done here?

bpisoj:
If tool does not support pax headers it will treat them as files.

prculley:
As Josip  says, 7-Zip does not support the Pax headers otherwise known as
Posix.1-2001.
But do we care? The Gramps .gpkg really shouldn't be opened by anything but
Gramps. The fact that it can be opened by a generic unzipper is just a nice
benefit for those who like to poke around.

Paul Culley

Nick-Hall:
An alternative would be to fix _mtime_ with something like:
```
def fix_mtime(tarinfo):
    tarinfo.mtime = int(tarinfo.mtime)
    return tarinfo
```
and:

`archive.add(filename, archname, filter=fix_mtime)`

prculley:
Changed as per @Nick-Hall suggestion.

PR #559, opened 2018-02-23
Changed behavior of "Look up with Map Services"

jgon6:
Removed the section that looked up by city, and country from the Map
Services lookup for Google and Open Street Map.

Resolves [#9486](https://gramps-project.org/bugs/view.php?id=9486)


Nick-Hall:
Thanks.

PR #560, opened 2018-02-26
Fix multi-page Graph output to pdf with filenames containg spaces

prculley:
Fixes [#10470](https://gramps-project.org/bugs/view.php?id=10470)

Should be backported to gramps42 when accepted.


lordemannd:
works for me on windows.

sam-m888:
Works on both Windows and Linux.

sam-m888:
Thank you

Cherry picked to gramps42

PR #562, opened 2018-02-27
Add missing quick view citation filters

Nick-Hall:
Implements feature request [10469](https://gramps-project.org/bugs/view.php?id=10469).

The new "all citations" filter failed during testing with the example database.  I think that is a result of bug [10448](https://gramps-project.org/bugs/view.php?id=10448).


sam-m888:
Waiting on result from https://github.com/gramps-project/gramps/pull/593

PR #565, opened 2018-02-28
Fix Book TOC and Index numbering and placement

prculley:
Fixes [#10452](https://gramps-project.org/bugs/view.php?id=10452)

Initial bug states that the TOC page number for the TOC itself was wrong if the TOC had more than one page.  Was a comparison problem with one based page numbers and internal zero based page index.

During testing I noted that the Index, when used, overwrote part of the report.  This was because the Index's page index was not being adjusted for multi-page TOC (and vise versa).

Both of these worked fine with TOC/Index smaller than one page, it was only when the first encountered was larger than a single page that the bug showed up.

### DIFF
--- a/gramps/plugins/docgen/cairodoc.py
+++ b/gramps/plugins/docgen/cairodoc.py
@@ -146,6 +146,8 @@ class CairoDocgen(libcairodoc.CairoDoc):
                     if offset > 0:
                         self.__increment_pages(toc, index, toc_page, offset)
                         rebuild_required = True
+                    if index_page and toc_page < index_page:
+                        index_page += offset
                 else:
                     toc_pages = []
 
@@ -157,6 +159,8 @@ class CairoDocgen(libcairodoc.CairoDoc):
                     if offset > 0:
                         self.__increment_pages(toc, index, index_page, offset)
                         rebuild_required = True
+                    if toc_page and toc_page > index_page:
+                        toc_page += offset
                 else:
                     index_pages = []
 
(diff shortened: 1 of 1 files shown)


lordemannd:
I was able to reproduce the original problem (on Windows). But noticed on the console several repeats of:
2018-02-28 08:36:06.735: WARNING: _enumeratedlist.py: line 143: Value 'None' not found for option 'Place format'
2018-02-28 08:36:06.735: WARNING: _enumeratedlist.py: line 144: Valid values: [(None, 'Default'), (0, 'Full')]
Which might be related to PR 561 which I have also included.

After including PR 565 (this one) I get the following on the console and the report is aborted.
(gramps.exe:2876): Gtk-WARNING **: Failed to set text '<span weight="bold" size="larger">Could not create C:\Users\dave\Desktop\tstweb\test__testbook_book.pdf
'<' not supported between instances of 'int' and 'NoneType'</span>' from markup due to error parsing markup: Error on line 2 char 10: Odd character 's', expected a '=' after attribute name 'not' of element '''

(gramps.exe:2876): Gtk-WARNING **: Failed to set text '<span weight="bold" size="larger">Could not create C:\Users\dave\Desktop\tstweb\test__testbook_book.pdf
'<' not supported between instances of 'int' and 'NoneType'</span>' from markup due to error parsing markup: Error on line 2 char 10: Odd character 's', expected a '=' after attribute name 'not' of element '''

prculley:
@lordemannd Thanks; a dumb mistake on my part, got to remember to run additional test cases.

sam-m888:
Thank you

PR #566, opened 2018-02-28
Fix Clipboard for right-click on empty clipboard

prculley:
Issues #10473

Bug 10473 contains two problems, this fixes the right-click on an empty clipboard (which popped up a blank menu).

The other bug was for the **addon** Clipboard **Gramplet** and is addressed in a PR https://github.com/gramps-project/addons-source/pull/119 for that addons-source.


sam-m888:
Thank you.

PR #567, opened 2018-03-01
[Narrative Web Report]Incorrect heading for stepmother or stepfather

SNoiraud:
Fixes #[10478](https://gramps-project.org/bugs/view.php?id=10478)

### DIFF
--- a/gramps/plugins/webreport/person.py
+++ b/gramps/plugins/webreport/person.py
@@ -50,7 +50,8 @@ import logging
 # Gramps module
 #------------------------------------------------
 from gramps.gen.const import GRAMPS_LOCALE as glocale
-from gramps.gen.lib import (ChildRefType, Date, Name, Person, EventRoleType)
+from gramps.gen.lib import (ChildRefType, Date, Name, Person, EventRoleType,
+                            EventType)
 from gramps.gen.lib.date import Today
 from gramps.gen.plug.report import Bibliography
 from gramps.gen.plug.report import utils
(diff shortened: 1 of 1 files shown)


SNoiraud:
Is it a bug or an enhancement ?

gramps/plugins/webreport/person.py
```diff
@@ -1635,6 +1638,55 @@ def display_ind_parent_family(self, birthmother, birthfather, family,
...
+        @param: handle The handle of the father or the mother
+        self.person    The child for whom we need this result
+        """
+        bd_date = Today()
+        bd_event = get_birth_or_fallback(self.r_db, self.person)
+        if bd_event:
+            bd_date = bd_event.get_date_object()
+        for event_ref in family.get_event_ref_list():
+            event = self.r_db.get_event_from_handle(event_ref.ref)
+            if event.type == EventType.DIVORCE and \
```
Nick-Hall: PEP8 prefers implied [line continuation](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) inside parentheses.

gramps/plugins/webreport/person.py
```diff
@@ -1635,6 +1638,55 @@ def display_ind_parent_family(self, birthmother, birthfather, family,
...
+                (event_ref.get_role() == EventRoleType.FAMILY or
```
Nick-Hall: It would be more pythonic to write:

`event_ref.get_role() in (EventRoleType.FAMILY, EventRoleType.PRIMARY)`

Nick-Hall:
> Is it a bug or an enhancement?

I think that this is a bug.

gramps/plugins/webreport/person.py
SNoiraud: OK. I suppose you prefer : 
            if (event.type == EventType.MARRIAGE and
                (event_ref.get_role() == EventRoleType.FAMILY or
                 event_ref.get_role() == EventRoleType.PRIMARY )):

gramps/plugins/webreport/person.py
SNoiraud: So the solution is : 
            if (event.type == EventType.MARRIAGE and
                (event_ref.get_role() in (EventRoleType.FAMILY, EventRoleType.PRIMARY ))):

sam-m888:
Thank you

PR #568, opened 2018-03-03
Make relationships optional in narrative web

SNoiraud:
Fixes [#10454](https://gramps-project.org/bugs/view.php?id=10454)

### DIFF
--- a/gramps/plugins/webreport/narrativeweb.py
+++ b/gramps/plugins/webreport/narrativeweb.py
@@ -1566,6 +1566,7 @@ class NavWebOptions(MenuReportOptions):
         self.__securesite = False
         self.__extra_page_name = None
         self.__extra_page = None
+        self.__relation = False
         db_options = name + ' ' + dbase.get_dbname()
         MenuReportOptions.__init__(self, db_options, dbase)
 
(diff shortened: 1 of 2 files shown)


SNoiraud:
May I ask one question ?
You added "Documentation" as a new label. What does that means ?
- I must modify the documentation ?
- This PR or the bug report references the documentation ?
- or something else ?

Is there a place where the rules are written which describes all these ?

sam-m888:
> You added "Documentation" as a new label. What does that means ?

@SNoiraud Helps me keep track of PR's that modify the user interface, in this case you've added an option (see screenshot) , but in this case as the Gramps 5.1 user manual has not been created this will have to wait.  

568narweb-new-option-51

https://gramps-project.org/wiki/index.php?title=Gramps_5.0_Wiki_Manual_-_Reports_-_part_7#Report_Options

> I must modify the documentation ?

No, user manual has not rolled over and the last time you and I tried to create additional documentation for the "Find the database Loop" tool it got lost in the rollover to gramps50!

>  Is there a place where the rules are written which describes all these ?

Not sure, but if any PR or commit changes the user interface, I'd appreciate it being marked with the "[Documentation](https://github.com/gramps-project/gramps/labels/Documentation)" tag 👍

Nick-Hall:
Merged in commit d98a1857dc225c7400076ebd523c532ca0d88403.

PR #569, opened 2018-03-04
Fix up Clipboard for various issues

prculley:
Bug[10475](https://gramps-project.org/bugs/view.php?id=10475) started this by noting HandleErrors when changing db.  During debug I found a lot of other issues.
* Various HandleErrors if trying to edit, set active, or otherwise do 'right-click' actions on clips when the clip was invalid for the current db. 
I changed the action routines to check for valid handle first.  Note that I did these changes before realizing that it would be better to have the invalid clips not show up at all, but decided to leave them in place as defensive programming.

* During a db change, or when going from a closed to an open db, there was a nasty race condition where the Gui class code got the db change first, and asked for the clip classes to validate against the new db.  The clip classes did not get the db change notification until later.  All the classes had independent references to the db.  So the clip classes tried to validate against a closed or previous db.  If it was closed, it got a db error about closed db, if open, the validation passed leaving a clip that could be dropped to create a corrupted db.
I changed the clipboard module to have a module level reference to the db, instead of all the class level references.  I left only the gui to deal with db change signals and update the module level reference.  That way all the clip code used the same db.  Also reduced the number of signals to process, since each clip did not have to register for and deal with change signals.

* Clip objects that had internal references to Note, Citation, and Repository objects were not being validated; this allowed these objects from other dbs to be apparently safe to paste, when they were not.  This would generate a corrupt db.

To avoid replicating some of these fixes multiple times, I used the new 'db.method' to create common code and removed a bunch of replicated code.

I also did a limited pylint in another commit (limited in that it still doesn't go over 9, due to missing docstrings, access to protected members, unused variables etc.).

Note: See https://github.com/gramps-project/addons-source/pull/122 for the Clipboard Gramplet to fix up some more issues.

### DIFF
--- a/gramps/gui/clipboard.py
+++ b/gramps/gui/clipboard.py
@@ -1274,6 +1274,12 @@ class ClipboardListView:
             if dbname:
                 _ob._dbname = dbname
 
+            # If this was a ClipPersonLink with a list for a handle, we don't
+            # want it.  Can happen in People Treeview with attemp to drag last
+            # name group.
+            if (isinstance(_ob, ClipHandleWrapper) and
+                    isinstance(_ob._handle, list)):
+                return None
             # If the wrapper object is a subclass of ClipDropList then
             # the drag data was a list of objects and we need to decode
             # all of them.
--- a/gramps/gui/clipboard.py
+++ b/gramps/gui/clipboard.py
@@ -69,6 +69,7 @@ from .utils import is_right_click, get_primary_mask
 #-------------------------------------------------------------------------
 WIKI_HELP_PAGE = '%s_-_Navigation' % URL_MANUAL_PAGE
 WIKI_HELP_SEC = _('manual|Using_the_Clipboard')
+clipdb = None  # current db to avoid different transient dbs during db change
 
 #-------------------------------------------------------------------------
 #
@@ -178,9 +179,7 @@ def model_contains(model, data):
 class ClipWrapper:
     UNAVAILABLE_ICON = 'dialog-error'
 
-    def __init__(self, dbstate, obj):
-        dbstate.connect('database-changed', self.database_changed)
-        self.database_changed(dbstate.db)
+    def __init__(self, obj):
         self._obj = obj
         self._pickle = obj
         self._type  = _("Unknown")
@@ -188,11 +187,8 @@ class ClipWrapper:
         self._handle = None
         self._title = _('Unavailable')
         self._value = _('Unavailable')
-        self._dbid = self._db.get_dbid()
-        self._dbname = self._db.get_dbname()
-
-    def database_changed(self,db):
-        self._db = db
+        self._dbid = clipdb.get_dbid()
+        self._dbname = clipdb.get_dbname()
 
     def get_type(self):
         return self._type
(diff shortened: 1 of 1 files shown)


fxtmtrsine:
>  if open, the validation passed leaving a clip that could be dropped to create a corrupted db.

and

> This would generate a corrupt db.

Well that explains why I need to keep going back to a backup which I do frequently! Time to stop using the clipboard until it's fixed!

Downloaded your fixed file and it crashed for me, does it rely on something else that I need to install?   What I did was open the clipboard and drag across the collapsed top level surname entry from the list of grouped people, then double click the entry in the clipboard.

> 294601: ERROR: grampsapp.py: line 143: Unhandled exception
>   File "C:\Program Files (x86)\GrampsAIO32-5.0.0-beta1\gramps\gui\clipboard.py", line 1575, in on_button_press
>     self.edit_obj(objclass, handle)
>   File "C:\Program Files (x86)\GrampsAIO32-5.0.0-beta1\gramps\gui\clipboard.py", line 1611, in edit_obj
>     if self.dbstate.db.method('has_%s_handle', objclass)(handle):
>   File "C:\Program Files (x86)\GrampsAIO32-5.0.0-beta1\gramps\plugins\db\bsddb\read.py", line 1588, in has_person_handle
>     return self.__has_handle(self.person_map, handle)
>   File "C:\Program Files (x86)\GrampsAIO32-5.0.0-beta1\gramps\plugins\db\bsddb\read.py", line 1579, in __has_handle
>     return table.get(handle.encode('utf-8'), txn=self.txn) is not None
> AttributeError: 'list' object has no attribute 'encode'
>

prculley:
@fxtmtrsine I suspect that the issue that caused your exception is a brand new bug, not addressed by these fixes.  I'll see if I can track it down.  Did the 'clip' say 'Person Unavailable Unavailable'?

prculley:
I fixed @fxtmtrsine issue, and removed my original idea to synchronize the various clipboards; turns out the different clipboard data was considered a 'feature'.

gramps/gui/clipboard.py
```diff
@@ -69,6 +69,7 @@
 #-------------------------------------------------------------------------
 WIKI_HELP_PAGE = '%s_-_Navigation' % URL_MANUAL_PAGE
 WIKI_HELP_SEC = _('manual|Using_the_Clipboard')
+CLIPDB = None  # current db to avoid different transient dbs during db change
```
Nick-Hall: This isn't a constant.

gramps/gui/clipboard.py
```diff
@@ -1627,75 +1573,13 @@ def edit_obj(self, objclass, handle):
...
-        elif objclass == 'Citation':
-            ref = self.dbstate.db.get_citation_from_handle(handle)
-            if ref:
+        if obj2class(objclass):  # make sure it is an editable object
+            if self.dbstate.db.method('has_%s_handle', objclass)(handle):
+                g_object = self.dbstate.db.method(
+                    'get_%s_from_handle', objclass)(handle)
                 try:
-                    EditCitation(self.dbstate, self.uistate, [], ref)
+                    locals()['Edit' + objclass](
```
Nick-Hall: There is already an [EditObject](https://github.com/gramps-project/gramps/blob/master/gramps/gui/editors/__init__.py#L77) utility.  Could it be used here?

prculley: @Nick-Hall I looked into using EditObject, but am not comfortable with the fact that it logs warnings when attempting to edit non-primary objects, or objects with invalid handles.  The original code was silent in these cases.  By the time I add in code to deal with these before calling EditObject, I have almost everything but the two line call you are commenting on.  I'd prefer to leave it as is.

BTW, an enhancement would be to allow editing of non-primary types, either here in the clipboard or elsewhere (like EditObject).  Then such a change would be more interesting.  But that is not the subject of this PR.

Nick-Hall: OK.  I just thought that you may have been unaware of EditObject.

sam-m888:
Thank you

PR #570, opened 2018-03-04
Fix Organize Tags editor for exception on 'x' close

prculley:
Fixes [#10479](https://gramps-project.org/bugs/view.php?id=10479)


lordemannd:
looks good to me on Win 10 64bit

PR #572, opened 2018-03-06
Fix Fan charts for scrolling/resizing of window; bad rendering

prculley:
Fixes [#10381](https://gramps-project.org/bugs/view.php?id=10381)
During fan chart scrolling/resizing when less than 1/2 of the figure was visible, Gtk was having a lot of trouble getting the whole chart rendered.  Mostly text was missing, although sometimes stripes of the chart were slightly offset from each other.  Did not seem to be related to Gtk version (Issue on v3.18.9 and 3.22.28).  It really seemed to be a Gtk bug though.  This following figure shows up the problem pretty well.

peek 202018-01-21 2014-28

I think what is happening is that the charts take too long to draw, and somehow get interrupted by Gtk, or perhaps there is a problem with their caching or double-buffering algorithms.

I spent a day trying various fixes but ended up by changing our drawing method.  Instead of drawing in detail when the 'draw' signal arrives, I performed the draw to an ImageSurface when something changed on the Gramps side.  When the 'draw' signal arrived (as it does when scrolling etc.), I copied the surface to the cairo context.  This seems to have stopped the issue, with the added benefit that scrolling or resizing the window is much smoother now.

During testing I found another minor issue; the legend was not getting located properly when the chart was printed if the chart had been re-centered by the user, because the centering data was not getting reset during printing.

### DIFF
--- a/gramps/gui/widgets/fanchart.py
+++ b/gramps/gui/widgets/fanchart.py
@@ -120,6 +120,7 @@ class FanChartBaseWidget(Gtk.DrawingArea):
         self.dbstate = dbstate
         self.uistate = uistate
         self.translating = False
+        self.surface = None
         self.goto = None
         self.on_popup = callback_popup
         self.last_x, self.last_y = None, None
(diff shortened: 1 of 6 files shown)


fxtmtrsine:
>  This figure shows up the problem pretty well.

The figure link only shows this:

> Access Denied.
> 
> Login
> 
> Proceed

Does this mean I have to get an account, which I'd rather not.  Can you you paste it here please?

fxtmtrsine:
Thanks 👍

sam-m888:
Thank you

PR #573, opened 2018-03-07
Enhancements to birthday and anniversary report

jgon6:
These commits implement Issues #3540, #5948. 

New report features:

- An option to indicate that a person in the report is dead
- Symbols to indicate the type of event being listed in the report
- An option to print the year of each event at the beginning of each entry
- Death anniversaries. This includes a switch to disable, an event type indicator, and support for the year option.

I've included a pdf of the report with these changes using the sample database.
[birthday_report.pdf](https://github.com/gramps-project/gramps/files/1787212/birthday_report.pdf)

Resolves #3540, #5948


Nick-Hall:
The build is failing due to trailing whitespace.

PR #574, opened 2018-03-07
Move dbapi-specific code out of DbGeneric

Nick-Hall:
These changes will make it easier to write third-party database addons.

### DIFF
--- a/gramps/gui/configure.py
+++ b/gramps/gui/configure.py
@@ -1496,13 +1496,13 @@ class GrampsPreferences(ConfigureDialog):
 
     def set_connection_widgets(self, db_choice):
         """
-        Sets the connection widgets sensitive for PostgreSQL.
+        Sets the connection widgets insensitive for embedded databases.
         """
         for widget in self.connection_widgets:
-            if db_choice == 'postgresql':
-                widget.set_sensitive(True)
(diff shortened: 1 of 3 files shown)


gramps/gui/configure.py
```diff
@@ -1496,13 +1496,13 @@ def database_backend_changed(self, obj):
...
         """
-        Sets the connection widgets sensitive for PostgreSQL.
+        Sets the connection widgets insensitive for embedded databases.
         """
         for widget in self.connection_widgets:
-            if db_choice == 'postgresql':
-                widget.set_sensitive(True)
-            else:
+            if db_choice in ('bsddb', 'sqlite'):
                 widget.set_sensitive(False)
```
prculley: I have to wonder if the unsophisticated user really needs to see these fields at all when running on the standard db.  I would think it would only cause confusion when there are fields present that never seem to do anything.  Perhaps make them not visible instead?

FuruFam61:
"Perhaps make them not visible instead?"

I agree.

gramps/gui/configure.py
Nick-Hall: This PR doesn't make any changes to the user interface.

PR #575, opened 2018-03-07
Allow Copying of text fields for Details tabs

sam-m888:
Allow selection of text fields from details tab in selected views with "Details" gramplets.
* Person Details
* Place Details
* Repository Details

Resolves [#7874](https://gramps-project.org/bugs/view.php?id=7874)


prculley:
Seems to work with the clipboard (results as text).

PR #576, opened 2018-03-08
Fix Relationship Graph for extraneous families when using filter

prculley:
Fixes [#10492](https://gramps-project.org/bugs/view.php?id=10492)

An earlier fix to allow showing families with no parents resulted in extraneous families when using filters.
This corrects that.  I also shifted a 'set' used for inclusion testing to a different spot, since I needed it in a second place, and then used it everywhere for inclusion testing.


sam-m888:
Thank you

PR #577, opened 2018-03-08
Geography: KML media objects not shown on map.

SNoiraud:
Geography doesn't conform to the relative paths
If a place has no coordinates, KML files are not displayed.

Fixes [#10487](https://gramps-project.org/bugs/view.php?id=10487)


sam-m888:
Thank you

PR #578, opened 2018-03-08
Fix places in example.gramps

sam-m888:
Was annoying me that the places for Greece did not actually group under that country.

- Moved Greek places into top level Greece entry & added English names.

- Removed Puerto Rico as a country, as part of USA

before and after example gramps


emyoulation:
This commit should show in the history

https://github.com/gramps-project/gramps/commits/ebc213f0accf12392925b7197d67e6e42b03bd78/example/gramps/example.gramps

prculley:
Looks ok to me. https://github.com/gramps-project/gramps/commits/maintenance/gramps52/example/gramps/example.gramps

PR #583, opened 2018-03-18
List sort by number for citation confidence level

sam-m888:
Fixes [#10506](https://gramps-project.org/bugs/view.php?id=10506)

Citation confidence levels are now sorted numerically and not alphabetically.

Fix tooltip on Confidence level entry for the Citation Editor, so text does not run on to each other.

### DIFF
--- a/gramps/gui/glade/editcitation.glade
+++ b/gramps/gui/glade/editcitation.glade
@@ -199,10 +199,10 @@
                     <property name="visible">True</property>
                     <property name="can_focus">False</property>
                     <property name="tooltip_text" translatable="yes">Conveys the submitter's quantitative evaluation of the credibility of a piece of information, based upon its supporting evidence. It is not intended to eliminate the receiver's need to evaluate the evidence for themselves.
-Very Low =Unreliable evidence or estimated data
-Low =Questionable reliability of evidence (interviews, census, oral genealogies, or potential for bias for example, an autobiography)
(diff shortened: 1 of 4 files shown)


prculley:
@sam-m888 The '/' you added to the tooltip in the Glade file seems to be causing problems.  On my system that seems to be preventing the line breaks in the text from working.  Also see https://gramps-project.org/bugs/view.php?id=10514 where it may be preventing the translations from working.  In his bug he has a photo of the tooltip showing the line break issue as well.  I don't know why you did this, but it appears to be a problem.

sam-m888:
Tooltip commit https://github.com/gramps-project/gramps/pull/583/commits/038e191f6dab1fe6b1a1f1270db2c4db91e58f6e reversed in https://github.com/gramps-project/gramps/commit/89ad092ac12e7269bc4405b71a8ead9b60080c32

sam-m888:
@prculley Looks like I solved the wrong issue, I felt it was just a formatting problem with the [tooltip](https://gramps-project.org/bugs/view.php?id=10506#c54178) that I noticed when I fixed issue [10506](https://gramps-project.org/bugs/view.php?id=10506) which turned out to be an issue introduced by commit https://github.com/gramps-project/gramps/commit/f688d376f54273c9deb82377bfd2ea8a1a8f137e#diff-b5a9b56444769115f8016d643f71a8e3

PR #584, opened 2018-03-22
Fix 'References' Gramplet for issue when activated during an import

prculley:
Fixes [#10507](https://gramps-project.org/bugs/view.php?id=10507) ... Maybe

I noticed that the active_handle was being returned as empty string, not None so changed test for that.
Based on Nick's observation that you cannot do a backlinks lookup and other things during an import, I disabled input to the main window during the import.

The popup from Gedcom import for Ansel character sets still works.  Other main window input seems to be locked out.  The progress bar still updates.

### DIFF
--- a/gramps/gui/dbloader.py
+++ b/gramps/gui/dbloader.py
@@ -506,10 +506,9 @@ class GrampsImportFileDialog(ManagedWindow):
 
                 for plugin in pmgr.get_import_plugins():
                     if extension == plugin.get_extension():
-                        self.do_import(import_dialog,
-                                       plugin.get_import_function(),
-                                       filename)
                         self.close()
+                        self.do_import(plugin.get_import_function(),
+                                       filename)
                         if callback is not None:
                             callback(self.import_info)
                         return
@@ -565,12 +564,10 @@ class GrampsImportFileDialog(ManagedWindow):
 
         return False
 
-    def do_import(self, dialog, importer, filename):
+    def do_import(self, importer, filename):
         self.import_info = None
-        position = self.window.get_position() # crock
-        dialog.hide()
-        self.window.move(position[0], position[1])
         self._begin_progress()
+        self.uistate.window.set_sensitive(False)
 
         try:
             #an importer can return an object with info, object.info_text()
@@ -590,6 +587,7 @@ class GrampsImportFileDialog(ManagedWindow):
                 parent=self.uistate.window)
         except Exception:
             _LOG.error("Failed to import database.", exc_info=True)
+        self.uistate.window.set_sensitive(True)
         self._end_progress()
 
     def build_menu_names(self, obj): # this is meaningless since it's modal
--- a/gramps/gui/dbloader.py
+++ b/gramps/gui/dbloader.py
@@ -567,7 +567,8 @@ class GrampsImportFileDialog(ManagedWindow):
     def do_import(self, importer, filename):
         self.import_info = None
         self._begin_progress()
-        self.uistate.window.set_sensitive(False)
+        self.uistate.set_sensitive(False)
+        self.uistate.viewmanager.enable_menu(False)
 
         try:
             #an importer can return an object with info, object.info_text()
@@ -587,7 +588,8 @@ class GrampsImportFileDialog(ManagedWindow):
                 parent=self.uistate.window)
         except Exception:
             _LOG.error("Failed to import database.", exc_info=True)
-        self.uistate.window.set_sensitive(True)
+        self.uistate.set_sensitive(True)
+        self.uistate.viewmanager.enable_menu(True)
         self._end_progress()
 
     def build_menu_names(self, obj): # this is meaningless since it's modal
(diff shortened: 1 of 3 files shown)


Nick-Hall:
Certainly a possibility.  The import will run in a batch transaction so a call to `find_backlink_handles` will fail because the _references_ table will be closed.

Nick-Hall:
> I do have to wonder if allowing GUI activities during import is a really good idea... Didn't test since I did not want to find more bugs ;-)

Allowing GUI activities during any batch transaction is not a good idea with our current BSDDB backend.  So this will also be a problem when running certain tools.

During a batch transaction, the reference table is closed so the `find_backlinks_handle` method will be unavailable.  This will affect three gramplets and also the participants information in the event view.

The secondary tables, which are like indexes in BSDDB, will also be closed.  So I expect problems with anything that attempts to get an object from a Gramps ID, such as certain filters and reports.

I seem to remember discussing this with @ennoborg in the past.  We thought about making all batch tool dialogs modal, but that obviously doesn't provide a complete solution.  The new backends which don't have the same design constraint are another consideration.

prculley:
@Nick-Hall @sam-m888 So instead of patching the various elements to deal with non-functional backlinks, I stopped allowing input during the import.  I think this should finally fix this issue...

sam-m888:
@prculley Much better 👍  Unfortunately I worked out a way to crash Gramps while importing from the menus :(

Nick-Hall:
@prculley Yes, that it a better approach.

prculley:
@sam-m888 I still don't understand how you were able to access the menus when I cannot.  However, I looked at the code again and realized that the original code attempted to solve this problem by keeping the filechooserdialog (a modal dialog) around, but hiding it during the import.  In theory, this should have prevented access to the main window as well.  But something must have changed in Gtk since that was coded, since it did not work.  So, on the off chance that this was interfering with my patch to set the main window insensitive, I eliminate that bit of original code.  This also works fine for me, I hope it works for your system.

sam-m888:
@prculley Must be something to do with either Gtk or the desktop
environments for Ubuntu I'm testing against  as when I test on Win 10 with
the AIO I get the results you see and on ubuntu I get a different!  I've
added my latest test result to the bug report.  Thanks for persevering.

prculley:
@Nick-Hall @sam-m888 I understand now.  I am using KDE on Ubuntu, and you are using Gnome.  I did not realize until now that the menus under Gnome are not part of the Gramps main window, they are somewhere else, so setting the main window insensitive is not enough.  I've been looking around for a while now and have not yet found how to disable these; I suspect I have to somehow disable the specific Application menu widget.  But I have not yet found where it is.  Nick, if you can give advice I would appreciate it.

prculley:
I thought I figured out how to disable the menus, but it did not work...

Nick-Hall:
@prculley You need to use the `set_sensitive` method of either a Gtk [Action](https://developer.gnome.org/gtk3/unstable/GtkAction.html#gtk-action-set-sensitive) or [ActionGroup](https://developer.gnome.org/gtk3/stable/GtkActionGroup.html#gtk-action-group-set-sensitive).

This reminds me that the `UIManager` code needs to be re-written to avoid using deprecated methods.

prculley:
@sam-m888 @Nick-Hall So I tried several other approaches to getting the menus disabled.  They all worked on my Windows installation and under KDE but failed under Gnome.  So I used Nick's suggestion to set the 'ActionGroup's insensitive.  It is not as clean or surgical as I would like since it leaves the menus visible, under Gnome but it does seem to prevent their activation while an import is in progress.

I looked into updating UIManager, Actions, ActionGroup etc. a while ago, and concluded that this would be a pretty big job, with changes all over the place.  I think the modern idea is to use Gtk.Application.  Maybe we need to make this a GEPS to keep it on our radar?

Nick-Hall:
> I looked into updating UIManager, Actions, ActionGroup etc. a while ago, and concluded that this would be a pretty big job, with changes all over the place. I think the modern idea is to use Gtk.Application. Maybe we need to make this a GEPS to keep it on our radar?

Yes, it is a big job.  I have added it to the v5.1 roadmap.

sam-m888:
Tested and works perfectly.

On ubuntu 16.04 with gnome/unity all menus that can create an action are greyed out.

menus-greyedout

Nick-Hall:
@prculley Did you consider making the import progress dialog modal?  That might also work.

prculley:
I tried a modal "please wait" dialog; which failed because the importers pop up other modal dialogs and they conflict.  On Gnome, it actually locked up Gramps, since I could not dismiss the "please wait" and that was the only one which had focus, so the other ones could not be dismissed.  We could probably redesign each importer to use its own progress dialog, rather than the callback dialog used now; that would avoid the conflicts.  But, as usual, I was trying to keep the changes more limited.

Nick-Hall:
@prculley Yes.  Keep the changes limited.  I only mentioned the modal dialog because I thought that it might have offered a simpler solution.

sam-m888:
Thank you

PR #586, opened 2018-03-24
Fix Relationship path between filter for silent fail

prculley:
Fixes [#10509](https://gramps-project.org/bugs/view.php?id=10509)

Filter failed with HandleError but was silent because it had a try/except pass around it.


sam-m888:
Thank you

PR #587, opened 2018-03-24
Fix FTM for delete tree, followed by close; Title etc. shows old tree

prculley:
Fixes [#10513](https://gramps-project.org/bugs/view.php?id=10513)


sam-m888:
Thank you

PR #588, opened 2018-03-24
Add build option to not compress manpages

herbmillerjr:
The package managers of some distros, in my case Portage on Gentoo, handle compression of manpages themselves. This PR adds a flag that allows those distros to build the manpages without compression so they can take care of that as they wish.

Python isn't my strongest language so this likely isn't perfect, but I wanted to get the discussion started since 5.0.0 is coming up and I'd like to get it packaged for Gentoo.


Nick-Hall:
Thanks.  I haven't tested this yet, but it looks good and see no reason why it shouldn't be included in the next release.

herbmillerjr:
Thank you much!

PR #590, opened 2018-03-28
Crash when selecting an old note, event, media, ...

SNoiraud:
Try to sort on another column.

Fixes [#10519](https://gramps-project.org/bugs/view.php?id=10519)

### DIFF
--- a/gramps/gui/selectors/baseselector.py
+++ b/gramps/gui/selectors/baseselector.py
@@ -318,12 +318,6 @@ class BaseSelector(ManagedWindow):
             self.model.reverse_order()
         self.build_tree()
 
-        handle = self.first_selected()
-        if handle:
-            path = self.model.on_get_path(handle)
-            self.selection.select_path(path)
-            self.tree.scroll_to_cell(path, None, 1, 0.5, 0)
-
         return True
 
     def show_toggle(self, obj):


gramps/gui/selectors/baseselector.py
```diff
@@ -318,12 +318,6 @@ def column_clicked(self, obj, data):
             self.model.reverse_order()
         self.build_tree()
 
-        handle = self.first_selected()
-        if handle:
-            path = self.model.on_get_path(handle)
```
Nick-Hall: You could keep the current handle selected and visible with:
```
iter_ = self.model.get_iter_from_handle(handle)
path = self.model.get_path(iter_)
```

SNoiraud: Removing all these lines change nothing.
The selected lines is always selected after sorting on another column.
As I said in the bug report, I think it was a workaround for an old Gtk problem.

I'll try what you say and see if something changes.

Nick-Hall: OK.  I it works anyway it is better to remove the code.

SNoiraud: Tried and the result is the same. I think we can remove these lines.

Nick-Hall: Thanks for checking.  I'll merge this now.

PR #594, opened 2018-04-04
Crash when looking for a place within an area

SNoiraud:
Fixes [#10531](https://gramps-project.org/bugs/view.php?id=10531)

The within filter needs the two fields. The prior patch remove the second field.


SNoiraud:
The better way should certainly to have the unit selected in the preferences.
But this is not the case for the moment.

SNoiraud:
I'll try your suggestion.

For geoplaces, If you click on reset filter then click on find, you'll get all places. After that, moving backward or forward in the history always show all places.
To test this, you must go to the place view, select several places with coordinates to have an history. Go back to the geoplaces view and test.

SNoiraud:
I think the problem is like in geoplaces. I corrected the problem clearing self.generic_filter.

This is another problem.
Perhaps the best way to do that is to set self.generic_filter to None when resetting the filter ?

SNoiraud:
OK Found it. within[1] is not necessary for the empty value.

PR #596, opened 2018-04-09
Bugfix: use same box margin for SVG and PDF

jkuebart:
When drawing text boxes in SVG, the text is often placed too far to the right and touches the right frame. For an example, generate a »Descendent tree« in SVG format. In the PDF version, the text does not touch the frame of any boxes.

This PR fixes the problem by using the same text displacement (margin) inside the box as the PDF driver. The code is taken from gramps/plugins/lib/libcairodoc.py.


Nick-Hall:
Rebased.

Nick-Hall:
Thanks for the fix.

jkuebart:
that was quick – thank you too!

PR #597, opened 2018-04-11
Fix Person views to update on changes in birth/death event/place

prculley:
Fixes [#10532](https://gramps-project.org/bugs/view.php?id=10532)

User noticed that changes to person birth/death event did not get updated on people list/tree view screen.


sam-m888:
Thank you

adamvazquez82:
further update in https://github.com/gramps-project/gramps/pull/603

PR #598, opened 2018-04-19
Geps/gep 039 genealogical symbols

SNoiraud:
Issue [#9098](https://gramps-project.org/bugs/view.php?id=9098)

Gramps should be able to use genealogy symbols everywhere.

Several kinds of use for the genealogical symbols :
- The graphical interface (GUI, editors, graphical views, ...)
- The Web reports (webcal, narrativeweb, ...)
- The Text reports (odf, rtf, pdf, ...)
- The graphical reports (odf, rtf, svg, ...)

Hope I don't forget something.

### DIFF
--- a/gramps/gen/utils/symbols.py
+++ b/gramps/gen/utils/symbols.py
@@ -1,0 +1,189 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2015-      Serge Noiraud
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# https://en.wikipedia.org/wiki/Miscellaneous_Symbols
+# http://www.w3schools.com/charsets/ref_utf_symbols.asp
+#
+
+#-------------------------------------------------------------------------
+#
+# Standard python modules
+#
+#-------------------------------------------------------------------------
+from gramps.gen.const import GRAMPS_LOCALE as glocale
+_ = glocale.translation.sgettext
+
+# pylint: disable=superfluous-parens
+# pylint: disable=anomalous-unicode-escape-in-string
+
+class Symbols(object):
+    # genealogical symbols
+    SYMBOL_FEMALE                    = 0
+    SYMBOL_MALE                      = 1
+    SYMBOL_ASEXUAL_SEXLESS           = 2 # Unknown
+    SYMBOL_LESBIAN                   = 3
+    SYMBOL_MALE_HOMOSEXUAL           = 4
+    SYMBOL_HETEROSEXUAL              = 5
+    SYMBOL_HERMAPHRODITE             = 6
+    SYMBOL_TRANSGENDER               = 7
+    SYMBOL_NEUTER                    = 8
+
+    SYMBOL_ILLEGITIM                 = 9
+    SYMBOL_BIRTH                     = 10
+    SYMBOL_BAPTISM                   = 11 # CHRISTENING
+    SYMBOL_ENGAGED                   = 12
+    SYMBOL_MARRIAGE                  = 13
+    SYMBOL_DIVORCE                   = 14
+    SYMBOL_UNMARRIED_PARTNERSHIP     = 15
+    SYMBOL_BURIED                    = 16
+    SYMBOL_CREMATED                  = 17 # Funeral urn
+    SYMBOL_KILLED_IN_ACTION          = 18
+    SYMBOL_EXTINCT                   = 19
+
+    all_symbols = [
+               # Name                     UNICODE       SUBSTITUTION
+               (_("Female"),              '\u2640',     ""),
+               (_("Male"),                '\u2642',     ""),
+               (_("Asexuality, sexless, genderless"), '\u26aa', ""),
+               (_("Lesbianism"),          '\u26a2',     "oo"),
+               (_("Male homosexuality"),  '\u26a3',     "oo"),
+               (_("Heterosexuality"),     '\u26a4',     "oo"),
+               (_("Transgender, hermaphrodite (in entomology)"), '\u26a5', ""),
+               (_("Transgender"),         '\u26a6',     ""),
+               (_("Neuter"),              '\u26b2',     ""),
+
+               (_("Illegitimate"),        '\u229b',     ""),
+               (_("Birth"),               '\u002a',     "*"),
+               (_("Baptism/Christening"), '\u007e',     "~"),
+               (_("Engaged"),             '\u26ac',     "o"),
+               (_("Marriage"),            '\u26ad',     "oo"),
+               (_("Divorce"),             '\u26ae',     "o|o"),
+               (_("Unmarried partnership"), '\u26af',   "o-o"),
+               (_("Buried"),              '\u26b0',     "d"),
+               (_("Cremated/Funeral urn"), '\u26b1',    "d"),
+               (_("Killed in action"),    '\u2694',     "d"),
+               (_("Extinct"),             '\u2021',     ""),
+              ]
+
+    # genealogical death symbols
+    DEATH_SYMBOL_NONE                      = 0
+    DEATH_SYMBOL_X                         = 1
+    DEATH_SYMBOL_SKULL                     = 2
+    DEATH_SYMBOL_ANKH                      = 3
+    DEATH_SYMBOL_ORTHODOX_CROSS            = 4
+    DEATH_SYMBOL_CHI_RHO                   = 5
+    DEATH_SYMBOL_LORRAINE_CROSS            = 6
+    DEATH_SYMBOL_JERUSALEM_CROSS           = 7
+    DEATH_SYMBOL_STAR_CRESCENT             = 8
+    DEATH_SYMBOL_WEST_SYRIAC_CROSS         = 9
+    DEATH_SYMBOL_EAST_SYRIAC_CROSS         = 10
+    DEATH_SYMBOL_HEAVY_GREEK_CROSS         = 11
+    DEATH_SYMBOL_LATIN_CROSS               = 12
+    DEATH_SYMBOL_SHADOWED_LATIN_CROSS      = 13
+    DEATH_SYMBOL_MALTESE_CROSS             = 14
+    DEATH_SYMBOL_STAR_OF_DAVID             = 15
+    DEATH_SYMBOL_DEAD                      = 16
+
+    # The following is used in the global preferences in the display tab.
+    #                Name
+    #                                                  UNICODE    SUBSTITUTION
+    death_symbols = [(_("Nothing"),                    "",        ""),
+                     ("x",                             "x",       "x"),
+                     (_("Skull and crossbones"),       "\u2620",  "+"),
+                     (_("Ankh"),                       "\u2625",  "+"),
+                     (_("Orthodox cross"),             "\u2626",  "+"),
+                     (_("Chi rho"),                    "\u2627",  "+"),
+                     (_("Cross of Lorraine"),          "\u2628",  "+"),
+                     (_("Cross of Jerusalem"),         "\u2629",  "+"),
+                     (_("Star and crescent"),          "\u262a",  "+"),
+                     (_("West Syriac cross"),          "\u2670",  "+"),
+                     (_("East Syriac cross"),          "\u2671",  "+"),
+                     (_("Heavy Greek cross"),          "\u271a",  "+"),
+                     (_("Latin cross"),                "\u271d",  "+"),
+                     (_("Shadowed White Latin cross"), "\u271e",  "+"),
+                     (_("Maltese cross"),              "\u2720",  "+"),
+                     (_("Star of David"),              "\u2721",  "+"),
+                     (_("Dead"),                       _("Dead"), _("Dead"))
+                ]
+
+    def __init__(self):
+        self.symbols = None
+    #
+    # functions for general symbols
+    #
+    def get_symbol_for_html(self, symbol):
+        """ return the html string like '&#9898;' """
+        return '&#%d;' % ord(self.all_symbols[symbol][1])
+
+    def get_symbol_name(self, symbol):
+        """
+        Return the name of the symbol.
+        """
+        return self.all_symbols[symbol][0]
+
+    def get_symbol_for_string(self, symbol):
+        """ return the utf-8 character like '\u2670' """
+        return self.all_symbols[symbol][1]
+
+    def get_symbol_fallback(self, symbol):
+        """
+        Return the replacement string.
+        This is used if the utf-8 symbol in not present within a font.
+        """
+        return self.all_symbols[symbol][2]
+
+    #
+    # functions for death symbols
+    #
+    def get_death_symbols(self):
+        """
+        Return the list of death symbols.
+        This is used in the global preference to choose which symbol we'll use.
+        """
+        return self.death_symbols
+
+    def get_death_symbol_name(self, symbol):
+        """
+        Return the name of the symbol.
+        """
+        return self.death_symbols[symbol][0]
+
+    def get_death_symbol_for_html(self, symbol):
+        """
+        return the html string like '&#9898;'.
+        """
+        return '&#%d;' % ord(self.death_symbols[symbol][1])
+
+    def get_death_symbol_for_char(self, symbol):
+        """
+        Return the utf-8 character for the symbol.
+        """
+        return self.death_symbols[symbol][1]
+
+    def get_death_symbol_fallback(self, symbol):
+        """
+        Return the string replacement for the symbol.
+        """
+        return self.death_symbols[symbol][2]
+
+    #
+    # functions for all symbols
+    #
+    def get_how_many_symbols(self):
+        return len(self.death_symbols) + len(self.all_symbols) - 4
--- a/gramps/gui/configure.py
+++ b/gramps/gui/configure.py
@@ -6,6 +6,7 @@
 # Copyright (C) 2010       Benny Malengier
 # Copyright (C) 2010       Nick Hall
 # Copyright (C) 2012       Doug Blank <doug.blank@gmail.com>
+# Copyright (C) 2015-      Serge Noiraud
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -72,6 +73,8 @@ from .plug import PluginWindows
 from .spell import HAVE_GTKSPELL
 from gramps.gen.constfunc import win
 _ = glocale.translation.gettext
+from gramps.gen.utils.symbols import Symbols
+from gramps.gen.constfunc import get_env_var
 
 #-------------------------------------------------------------------------
 #
@@ -363,6 +366,14 @@ class ConfigureDialog(ManagedWindow):
         grid.attach(text, start, index, stop - start, 1)
         return text
 
+    def add_button(self, grid, label, index, constant, extra_callback=None, config=None):
+        if not config:
+            config = self.__config
+        button = Gtk.Button(label=label)
+        button.connect('clicked', extra_callback)
+        grid.attach(button, 1, index, 1, 1)
+        return button
+
     def add_path_box(self, grid, label, index, entry, path, callback_label,
                      callback_sel, config=None):
         """
@@ -555,7 +566,8 @@ class GrampsPreferences(ConfigureDialog):
             self.add_date_panel,
             self.add_researcher_panel,
             self.add_advanced_panel,
-            self.add_color_panel
+            self.add_color_panel,
+            self.add_symbols_panel
             )
         ConfigureDialog.__init__(self, uistate, dbstate, page_funcs,
                                  GrampsPreferences, config,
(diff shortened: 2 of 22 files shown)


gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+    SYMBOL_LESBIAN                   = 2
+    SYMBOL_MALE_HOMOSEXUAL           = 3
+    SYMBOL_HETEROSEXUAL              = 4
+    SYMBOL_HERMAPHRODITE             = 5
+    SYMBOL_TRANSGENDER               = 6
+    SYMBOL_ASEXUAL_SEXLESS           = 7
+    SYMBOL_NEUTER                    = 8
+    SYMBOL_ILLEGITIM                 = 9
+    SYMBOL_BIRTH                     = 10
+    SYMBOL_BAPTISATION               = 11 # CHRISTENING
```
Nick-Hall: Use _SYMBOL_BAPTISM_ rather than _SYMBOL_BAPTISATION_.

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+    SYMBOL_UNMARRIED_PARTNERSHIP     = 15
+    SYMBOL_BURIED                    = 16
+    SYMBOL_CREMATED                  = 17 # Funeral urn
+    SYMBOL_KILLED_IN_ACTION          = 18
+    SYMBOL_EXTINCT                   = 19
+
+    all_symbols = [
+               # Name                                            (UNICODE char)
+               #                      (HTML char) SUBSTITUTION string
+               (_("Male"),                                       '\u2642',
```
Nick-Hall: It is not necessary to store both the unicode character and decimal html.  Use:

`html_string = '&#%d;' % ord(c)`

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+    DEATH_SYMBOL_DEAD                      = 16
+
+    # The following is used in the global preferences in the display tab.
+    #                Name
+    #                              (UNICODE char) (HTML char) SUBSTITUTED string
+    death_symbols = [(_("Nothing"),
+                                    "",         "",        ""),
+                 ("x",
+                                    "x",        "x",       "x"),
+                 (_("Skull and crossbones") +' : \u2620',
```
Nick-Hall: I suggest that you don't add the character to the name.  In English, we wouldn't put a space before the colon.  It would be better to use a translated string format to add the character.

`_('%s: %s)`

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+    #
+    def get_symbol_for_html(self, symbol):
+        """ retun the html string like '&#9898;' """
+        return self.all_symbols[symbol][2]
+
+    def get_symbol_for_string(self, symbol):
+        """ retun the utf-8 character like '\u2670' """
+        return self.all_symbols[symbol][1]
+
+    def get_symbol_replacement(self, symbol):
```
Nick-Hall: Use fallback rather than replacement?

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+                return element[3]
+        return ""
+
+    #
+    # functions for all symbols
+    #
+    def get_how_many_symbols(self):
+        return len(self.death_symbols) + len(self.all_symbols) - 3
+
+if __name__ == '__main__':
```
Nick-Hall: Is it easy to write a unit test instead?

gramps/gui/configure.py
```diff
@@ -1713,3 +1724,251 @@ def update_surn_height(self, obj, constant):
...
+        try:
+            # remove the old messages with old font
+            self.grid.remove_row(8)
+            self.grid.remove_row(7)
+        except:
+            pass
+        font = config.get('utf8.selected-font')
+        symbols = Symbols()
+        my_characters = _("What you will see") + " :\n"
+        my_characters += "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz "
```
Nick-Hall: The `string` class can sometimes be useful, but maybe not worth the effort here:

`string.ascii_letters`

gramps/gui/configure.py
```diff
@@ -1713,3 +1724,251 @@ def update_surn_height(self, obj, constant):
...
+        font_description = Pango.font_description_from_string(font)
+        text.modify_font(font_description)
+        self.activate_change_font()
+        text.set_halign(Gtk.Align.START)
+        text.set_text(my_characters)
+        scrollw.add(text)
+        scrollw.set_policy(Gtk.PolicyType.AUTOMATIC, Gtk.PolicyType.AUTOMATIC)
+        self.grid.attach(scrollw, 1, 7, 8, 1)
+
+        my_characters = symbols.get_symbol_for_string(Symbols.SYMBOL_MALE) + " "
```
Nick-Hall: Use a loop?

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+               #                      (HTML char) SUBSTITUTION string
+               (_("Male"),                                       '\u2642',
+                                      "&#9794;",  ""),
+               (_("Female"),                                     '\u2640',
+                                      "&#9792;",  ""),
+               (_("Lesbianism"),                                 '\u26a2',
+                                      "&#9890;",  ""),
+               (_("Male homosexuality"),                         '\u26a3',
+                                      "&#9891;",  ""),
+               (_("Heterosexality"),                             '\u26a4',
```
QuLogic: Missing u

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+                                      "&#9892;",  ""),
+               (_("Transgender, hermaphrodite (in entomology)"), '\u26a5',
+                                      "&#9893;",  ""),
+               (_("Transgender"),                                '\u26a6',
+                                      "&#9894;",  ""),
+               (_("Asexuality, sexless, genderless"),            '\u26aa',
+                                      "&#9898;",  ""),
+               (_("Neuter"),                                     '\u26b2',
+                                      "&#9906;",  ""),
+               (_("Illegitim"),                                  '\u229b',
```
QuLogic: Illegitimate?

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+                                    "x",        "x",       "x"),
+                 (_("Skull and crossbones") +' : \u2620',
+                                    "&#9760;",  "\u2620",  "+"),
+                 (_("Ankh")                 +' : \u2625',
+                                    "&#9765;",  "\u2625",  "+"),
+                 (_("Orthodox cross")       +' : \u2626',
+                                    "&#9766;",  "\u2626",  "+"),
+                 (_("Chi rho")              +' : \u2627',
+                                    "&#9767;",  "\u2627",  "+"),
+                 (_("Cross of lorraine")    +' : \u2628',
```
QuLogic: Lorraine

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+                                    "&#9768;",  "\u2628",  "+"),
+                 (_("Cross of jerusalem")   +' : \u2629',
```
QuLogic: Jerusalem

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+                                    "&#9769;",  "\u2629",  "+"),
+                 (_("Star and crescent")    +' : \u262a',
+                                    "&#9770;",  "\u262a",  "+"),
+                 (_("West syriac cross")    +' : \u2670',
```
QuLogic: Syriac

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+                                    "&#9840;",  "\u2670",  "+"),
+                 (_("East syriac cross")    +' : \u2671',
```
QuLogic: Syriac

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+                                    "&#9841;",  "\u2671",  "+"),
+                 (_("Heavy greek cross")    +' : \u271a',
```
QuLogic: Greek

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+                                    "&#10010;", "\u271a",  "+"),
+                 (_("Latin cross")          +' : \u271d',
+                                    "&#10013;", "\u271d",  "+"),
+                 (_("Shadowed White Latin cross")+' : \u271e',
+                                    "&#10014;", "\u271e",  "+"),
+                 (_("Maltese cross")        +' : \u2720',
+                                    "&#10016;", "\u2720",  "+"),
+                 (_("Star of david")        +' : \u2721',
```
QuLogic: David

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+                                   _("Dead"),  _("Dead"), _("Dead"))
+                ]
+
+    def __init__(self):
+        self.symbols = None
+    #
+    # functions for general symbols
+    #
+    def get_symbol_for_html(self, symbol):
+        """ retun the html string like '&#9898;' """
```
QuLogic: return

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,409 @@
...
+        return self.all_symbols[symbol][2]
+
+    def get_symbol_for_string(self, symbol):
+        """ retun the utf-8 character like '\u2670' """
```
QuLogic: return

gramps/gen/utils/symbols.py
SNoiraud: Done

gramps/gen/utils/symbols.py
SNoiraud: Done

gramps/gen/utils/symbols.py
SNoiraud: Done

gramps/gen/utils/symbols.py
SNoiraud: Done

gramps/gen/utils/symbols.py
SNoiraud: When I started to write the module, the tests was included.
I removed this code. I'll write a test for that. Can it be a standalone test or it will be included with all other tests.

gramps/gui/configure.py
SNoiraud: Done

gramps/gui/configure.py
SNoiraud: Done

gramps/gen/utils/symbols.py
SNoiraud: Done

gramps/gen/utils/symbols.py
SNoiraud: Done

gramps/gen/utils/symbols.py
SNoiraud: Done

gramps/gen/utils/symbols.py
SNoiraud: Done

gramps/gen/utils/symbols.py
SNoiraud: Done

gramps/gen/utils/symbols.py
SNoiraud: Done

gramps/gen/utils/symbols.py
SNoiraud: Done

gramps/gen/utils/symbols.py
SNoiraud: Done for all Syriac

gramps/gui/configure.py
```diff
@@ -1713,3 +1724,227 @@ def update_surn_height(self, obj, constant):
...
+                value = symbols.get_symbol_for_string(rand)
+                font = fonts[idx]
+                fontname = font.family[0][1]
+                try:
+                    vals = all_fonts[fontname]
+                except:
+                    all_fonts[fontname] = []
+                if font.has_char(value):
+                    if value not in all_fonts[fontname]:
+                        all_fonts[fontname].append(value)
```
QuLogic: Since you don't use the order of the results here, this seems like a good place to use a `set` instead of a list.

gramps/gui/configure.py
```diff
@@ -1713,3 +1724,227 @@ def update_surn_height(self, obj, constant):
...
+                self.progress.step()
+        self.progress.close()
+        nb1 = 0
+        available_fonts = []
+        for font in all_fonts.keys():
```
QuLogic: `for font, font_usage in all_fonts.items():`

gramps/gui/configure.py
```diff
@@ -1713,3 +1724,227 @@ def update_surn_height(self, obj, constant):
...
+                try:
+                    vals = all_fonts[fontname]
+                except:
+                    all_fonts[fontname] = []
+                if font.has_char(value):
+                    if value not in all_fonts[fontname]:
+                        all_fonts[fontname].append(value)
+                self.progress.step()
+        self.progress.close()
+        nb1 = 0
```
QuLogic: This variable is unused.

gramps/gen/utils/symbols.py
```diff
@@ -0,0 +1,188 @@
...
+        return '&#%d;' % ord(self.all_symbols[symbol][1])
+
+    def get_symbol_name(self, symbol):
+        """
+        Return the name of the symbol.
+        """
+        return self.all_symbols[symbol][0]
+
+    def get_symbol_for_string(self, symbol):
+        """ retun the utf-8 character like '\u2670' """
```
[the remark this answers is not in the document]
SNoiraud: Done

gramps/gui/configure.py
```diff
@@ -1713,3 +1724,227 @@ def update_surn_height(self, obj, constant):
...
+                3,
+                'utf8.in-use',
+                extra_callback=self.can_we_use_genealogical_fonts)
+        sel_font = config.get('utf8.selected-font')
+        available_fonts = config.get('utf8.available-fonts')
+        active_val = 0
+        self.all_avail_fonts = [x for x in enumerate(available_fonts)]
+        if len(available_fonts) > 0:
+            self.add_text(self.grid,
+                _('You already run the tools to search genealogy fonts.'
```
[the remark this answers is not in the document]
SNoiraud: Done

gramps/gui/configure.py
```diff
@@ -1713,3 +1724,227 @@ def update_surn_height(self, obj, constant):
...
+            self.add_combo(self.grid,
+                _('Choose font'),
+                5, 'utf8.selected-font',
+                self.all_avail_fonts, callback=self.utf8_update_font, valueactive=True, setactive=active_val)
+            symbols = Symbols()
+            all_sbls = symbols.get_death_symbols()
+            all_symbols = []
+            for symbol in all_sbls:
+                all_symbols.append(symbol[1] + " " + symbol[0])
+            self.all_death_symbols = [x for x in enumerate(all_symbols)]
```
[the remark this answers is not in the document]
SNoiraud: Yes, better

gramps/gui/configure.py
```diff
@@ -1713,3 +1724,227 @@ def update_surn_height(self, obj, constant):
...
+        fonts = fontconfig.query()
+        all_fonts = {}
+        symbols = Symbols()
+        nb_symbols = symbols.get_how_many_symbols()
+        self.in_progress = True
+        self.progress = ProgressMeter(_('Checking available genealogical fonts'),
+                                       can_cancel=True,
+                                       cancel_callback=self.stop_looking_for_font, parent=self.uistate.window)
+        self.progress.set_pass(_('Looking for all fonts with genealogical symbols.'), nb_symbols*len(fonts))
+        for idx in range(0, len(fonts)):
```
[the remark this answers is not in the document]
SNoiraud: It's used latter with : font = fonts[idx] into the two following for statement.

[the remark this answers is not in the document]
QuLogic: But you don't need that line if you change the loop.

gramps/gui/configure.py
```diff
@@ -1713,3 +1724,227 @@ def update_surn_height(self, obj, constant):
...
+            if not self.in_progress:
+                return
+            for rand in range(symbols.SYMBOL_MALE, symbols.SYMBOL_EXTINCT+1):
+                string = symbols.get_symbol_for_html(rand)
+                value = symbols.get_symbol_for_string(rand)
+                font = fonts[idx]
+                fontname = font.family[0][1]
+                try:
```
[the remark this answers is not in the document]
SNoiraud: I'll look at this

gramps/gui/configure.py
SNoiraud: I think it was for an old verification test. removed.

SNoiraud:
To be conform to existing genders (Male, Female, Unknown), I move "Asexuality, sexless, genderless" just after Male, Female.

gramps/gui/configure.py
SNoiraud: If I use a set instead of a list, I get : 
    if value not in all_fonts[fontname]:
TypeError: 'set' object is not subscriptable
perhaps there is a better way to do this.

prculley: I think he meant to use a set instead of a list as the stored value in all_fonts[fontname].  If you use the defaultdict of his earlier suggestion, all_fonts gets initialized like
    all_fonts = defaultdict(set)  # you must use "from collections import defaultdict" at top of file
and then your try and 'if' gets replaced with
    if font.has_char(value):
         if value not in all_fonts[fontname]:
                all_fonts[fontname].add(value)

gramps/gui/configure.py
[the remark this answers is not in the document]
SNoiraud: OK. Done

gramps/gui/configure.py
SNoiraud: I didn't use defaultdict. This was the problem.
Done

SNoiraud:
The first time we click on "Try to find", only the font list was visible. We couldn't see the death symbol list.
This has been solved with the last patch.

SNoiraud:
What could be the first view or report we'll work on ?

gramps/gui/configure.py
QuLogic: Yes, that's what I meant. In fact, you don't need the `if` and can do `all_fonts[fontname].add(value)` unilaterally because the set will keep things unique. You should also remove [this try](https://github.com/gramps-project/gramps/pull/598/files#diff-8c90436ad9f31535f4cc5d1cd05e1a87R1832) because it doesn't do anything with `defaultdict`.

SNoiraud:
Perhaps we could emit a signal when we change the font to force redraw of some views.

SNoiraud:
In person, FEMALE is 0 and MALE = 1, so I need to do the same thing for symbols.
This will be in the next patch.
I added the emit signal.
I tried to work on QuiltView which already use symbols.

SNoiraud:
If someone want to test the QuiltView third party addon, I attached the patch above version 1.0.5 :
[QuiltView.symbols.patch.txt](https://github.com/gramps-project/gramps/files/1934472/QuiltView.symbols.patch.txt)

SNoiraud:
Here is the first try with the pedigree view : 

pedigree with symbols 2
pedigree with symbols

SNoiraud:
@Nick-Hall There is a problem in master with the following: 
commit a216441029159b0cebb0f34f31029b4929c25bc0
Author: Nick Hall <nick-h@gramps-project.org>
Date:   Fri Dec 22 18:31:25 2017 +0000

    Clear old custom name formats when database loaded

With this patch applied, fanchart doesn't work anymore. When the fanchart view is loaded, it creates two custom names format. When all plugins are loaded, we clear the old format so fanchart crash.

prculley:
Regarding the fan chart issue; see https://github.com/gramps-project/gramps/commit/51df2c188db6c6e27cce447759bc8b167e7c5814 which resolved that issue.
Looks like we should do another merge up from gramps50 to master...

SNoiraud:
> Looks like we should do another merge up from gramps50 to master...
I only signaled this problem. We can wait for the merge.

SNoiraud:
timeline pedigree 1
fanchart 2way 1
fanchart 1
fanchart desc 1

SNoiraud:
The Graph View
graphview 1

SNoiraud:
I'll make a new PR for addons.

gramps/gui/configure.py
SNoiraud: Done

prculley:
Thought I'd try out what you have on Windows; got the following.  Looks like your test for fontconfig needs to be moved to configure.py.
  File "C:\Users\prc\workspace\GrampsM\gramps\gramps\gui\grampsgui.py", line 363, in __startgramps
    Gramps(argparser)
  File "C:\Users\prc\workspace\GrampsM\gramps\gramps\gui\grampsgui.py", line 240, in __init__
    from . import viewmanager
  File "C:\Users\prc\workspace\GrampsM\gramps\gramps\gui\viewmanager.py", line 78, in <module>
    from .displaystate import DisplayState, RecentDocsMenu
  File "C:\Users\prc\workspace\GrampsM\gramps\gramps\gui\displaystate.py", line 59, in <module>
    from .views.navigationview import NavigationView
  File "C:\Users\prc\workspace\GrampsM\gramps\gramps\gui\views\navigationview.py", line 51, in <module>
    from .pageview import PageView
  File "C:\Users\prc\workspace\GrampsM\gramps\gramps\gui\views\pageview.py", line 52, in <module>
    from ..configure import ConfigureDialog
  File "C:\Users\prc\workspace\GrampsM\gramps\gramps\gui\configure.py", line 75, in <module>
    from gramps.gen.utils.symbols import Symbols
  File "C:\Users\prc\workspace\GrampsM\gramps\gramps\gen\utils\symbols.py", line 31, in <module>
    import fontconfig
ModuleNotFoundError: No module named 'fontconfig'
Also appears we are going to have to add fontconfig Python adapter to Windows distro...

SNoiraud:
Yes, fontconfig is a prerequisite: python3-fontconfig on linux.
I'll add a test for fontconfig.

SNoiraud:
OK. I have a useless import of fontconfig in symbols.py

prculley:
Finally found a way to get the Python-fontconfig installed under Mingw32.  Next issue is that I got a crash at 
  File "C:\Users\prc\workspace\GrampsM\gramps\gramps\gui\configure.py", line 1828, in can_we_use_genealogical_fonts
    fontname = font.family[0][1]
KeyError: 0

The value of 'font' at that point was FcFont: <FcFont: Corbel>
its path was str: C:/WINDOWS/fonts/corbelb.ttf
 and font.family was dict: {'en': 'Corbel'}
It appears that font.family is always a dict on my system with various keys, I saw 'en', and the font name as keys in my short debug.  Never saw more than one entry in the dict.

SNoiraud:
Strange. for me, I get for font.family:
[('en', 'Tuffy')]

What is your fontconfig version ?
mine is 0.5.1

prculley:
Mine is 0.6.0; they changed that return to a dict with https://github.com/Vayn/python-fontconfig/commit/aaffbdc3c90b19e4ce031525542321be9c0186b5  it is even showing in the readme...

SNoiraud:
Perhaps we need to test if this return a dict or a list ?

prculley:
Sometimes the font.family dict has multiple values; below are some examples.  Interestingly, on my system no font had all the genealogical symbols by default.  The one you popped up as the 'Selected' was {'en': 'Segoe UI Symbol'}

{'en': 'Kozuka Gothic Pro M', 'ja': '小塚ゴシック Pro M'}
{'ar': 'Myriad عربي', 'en': 'Myriad Arabic'}
{'en': 'Myriad Hebrew', 'he': 'Myriad עברית'}
{'Small Fonts': 'Small Fonts'}
{'Terminal': 'Terminal'}
{'MS Sans Serif': 'MS Sans Serif'}

vantu5z:
@SNoiraud When choose font it apply to all app including menu and other text.
Is some way to use these symbols, but stay system fonts for app and views?
I mean just insert symbol and not change font for text around it (maybe use markup for this or something else).

SNoiraud:
> @SNoiraud When choose font it apply to all app including menu and other text. Is some way to use these
> symbols, but stay system fonts for app and views? I mean just insert symbol and not change font for text
> around it (maybe use markup for this or something else).

Much more complicated.
When you use a good OS like linux, The standard font is already utf8. If you use "DejaVu", You have no changes. The problem is for crappy OSes which doesn't use utf8 fonts.

You may have symbols in menus (not for the moment)
Someone could use symbols in new views's name.
Normally, We should not use our own fonts like in some views, but use the theme  font.

I can look at this,

prculley:
I think that your suggestion on how to hadle the differing fontconfig family (list/dict) looks workable.

Another issue; My system happened to have only a single font that had all the symbols.  I found that I could not select the 'Use symbols' button.  Turned out that if the user doesn't select a font from the 'Choose font' combo, or there is only a single font (where choosing it does not trigger the callback because nothing changes) then the 'utf8-selected-font' config setting was never set.
I added the following after the choose font add_combo in two places to correct this.:
            config.set('utf8.selected-font',
                       self.all_avail_fonts[choosefont.get_active()][1])

A second issue, probably gtk related; even when I selected the font that should have had all the symbols, one symbol (second one shown) was still showing as the hex_in_a_box thingie.  This went away when I used the special Windows only setting that changed the gtk font selection code from 'Win32' to 'fontconfig'.  That setting (Use alternate font handler for GUI and reports) was implemented a while ago when the Gramps GUI and reports were not displaying some of the utf-8 characters; switching to the Linux derived fontconfig code seemed to mostly work better even on the Windows systems.

prculley:
I wanted to see the genealogical symbols a bit better than the standard size you initially showed, so I changed the set_text line in configure.py utf8_show_example around 1950 to
        text.set_markup("<span size='x-large'>" + my_characters + "</span>")
You might want to consider this.

SNoiraud:
> @SNoiraud When choose font it apply to all app including menu and other text.
> Is some way to use these symbols, but stay system fonts for app and views?
> I mean just insert symbol and not change font for text around it (maybe use markup for this or something else).

I found one solution for this. Do you think it must be applied ?

SNoiraud:
We are now able to see homosexual relationship.
relationships 1

Nick-Hall:
@SNoiraud There are three conflicts that need to be resolved.  Otherwise is this ready to merge?

SNoiraud:
> Otherwise is this ready to merge?

Yes.

Nick-Hall:
@SNoiraud You will have to resolve the conflicts. There are more than three files and I was unsure what to do in some places.

fxtmtrsine:
@SNoiraud Hmm tried master and tried a to use a Chart view (Pedigree) but only get a blank box with an error message, looking at git blame brings me to this pr, does it work for you?

> 
> ./Gramps.py 
> 2019-06-11 16:50:56.309: WARNING: viewmanager.py: line 812: View 'pedigreeview' failed to load.
>   File "/home/fxtmtrsine/gramps/gramps/gui/viewmanager.py", line 809, in __create_page
>     page = page_def(pdata, self.dbstate, self.uistate)
>   File "/home/fxtmtrsine/gramps/gramps/plugins/view/pedigreeview.py", line 574, in __init__
>     self.func_list.update({
> AttributeError: 'PedigreeView' object has no attribute 'func_list'
> 

https://github.com/gramps-project/gramps/blob/cd092897bcc93bba29673356918ba7579023a796/gramps/plugins/view/pedigreeview.py#L574

SNoiraud:
This problem is not related to this PR. several plugins have this problem too.
Please make a bug report for this.

fxtmtrsine:
done.
https://github.com/gramps-project/gramps/pull/825

SNoiraud:
This problem is related to GEPS044 ( @prculley )
The plugins which have problems are:
gramps/plugins/view/fanchart2wayview.py
gramps/plugins/view/fanchartdescview.py
gramps/plugins/view/fanchartview.py
gramps/plugins/view/pedigreeview.py

prculley:
Actually the issue is (arguably) due to this PR.  Or rather an incompatibility between PRs that was not properly dealt with in the merge.  GEPS044 eliminated the self.func_list functionality.  When this was merged, the self.func_list code snippets got put back in when they should not have been.

I'm a bit busy at the moment, so feel free to fix this; if it doesn't get done in a couple of weeks (or at least before 5.1 release) I will try to find time to do.

SNoiraud:
Done with commit 1ef467c

fxtmtrsine:
fyi still not working :(

sam-m888:
> fyi still not working :(

@fxtmtrsine should be working now.

prculley:
@SNoiraud I just was testing the gramps51 branch prior to release and noticed that the Preferences dialog tab for this said I needed python-fontconfig to work.  The MSYS2/MING64 system did not have that available as a prebuilt binary, so I had to figure out how to make it myself.
I eventually found instructions at https://pypi.org/project/Python-fontconfig/ and of course they did not work since I foolishly decided to use the github source and "python3 setup.py install" method.  This apparently installs version 0.6 which seems incompatible with your code.

Note: I used that method to install since I have in the past had issues with pip binary 'wheels' built with the Microsoft tools which are incompatible with the MSYS MINGW tools the AIO uses.

I figured out that I needed the older 0.5.1 version and was able to build that to make it work.

The reason for the note is I am a little concerned that we are introducing another dependency for Gramps and this appears to be a pretty obscure one, with little activity.

If we really need this, I think we need to update our prerequisites documentation @sam-m888 and to be prepared for others to end up with the wrong version as well.

Nick-Hall:
Would it be possible to use the _fonttools_ package instead?

prculley:
I was kind of hoping for something that did not introduce new dependencies, like the Pango Font parts of Gtk...

SNoiraud:
> Would it be possible to use the _fonttools_ package instead?

I tried but we can't use it.

SNoiraud:
@prculley I aready tested with version 0.6 (on 24 Apr 2018)
This was a question following your request.

In this PR I also added the fact fontconfig is a prerequisite (on 24 Apr 2018)

I didn't find any functions with PANGO to do this job.

prculley:
I just upgraded one of my VMs to the latest Python 3.7.  The python-fontconfig prerequisite will not install at either v0.5.1 or v0.6  as it is referring to internal Python structures that are not part of the public API and the Python guys changed them.

I patched up the python-fontconfig for my AIO to refer to the updated python 'c' code structure and got it to compile.

Then I found that Gramps had an exception when actually using the 'Find Fonts'; it seems that your code line
<pre>                    for lang, name in font.family:     # version 0.6.0 use dict
</pre>
should be
<pre>                    # version 0.6.0 use dict
                    for lang, name in font.family.items():
</pre>

I will create a patch for the python-fontconfig guys and see if they are still around to accept it.  If not, we may need to rethink this again.

The funny thing is that Pango has the exact function we need to find out if a character is available in a font, https://developer.gnome.org/pango/stable/pango-Coverage-Maps.html and https://developer.gnome.org/pango/stable/pango-Fonts.html#pango-font-get-coverage but for some reason they are not introspectable, so no Python equivalents.

In the meantime I think we should correct the Gramps code here.

PR #599, opened 2018-04-20
Clone event IIB

apoettker:
Clone Event w/o references (2nd pass, part B)

Feature request [#9604](https://gramps-project.org/bugs/view.php?id=9604)

This functionality in the Event List doubles an event with all the related tags, citations, media, attributes and notes, but without the references. Its like an event addition, but with default values.

In former PR #427 was closed and substituted by two new ones. This PR (Clone Event 2nd pass, part B) deals with Clone Events out of the Event List with above described functionality. An upcoming PR will handle a clone functionality embedded in the Person / Family editor (Clone Event 2nd pass, part A).


gramps/gui/editors/displaytabs/eventbackreflist.py
```diff
@@ -27,10 +28,24 @@
...
+        self.option = option
 
-    def __init__(self, dbstate, uistate, track, obj, callback=None):
         BackRefList.__init__(self, dbstate, uistate, track, obj,
                              BackRefModel, callback)
 
     def get_icon_name(self):
         return 'gramps-event'
+
+    def get_data(self):
```
Nick-Hall: Why is this needed?  A new event will have no references.

gramps/gui/editors/editevent.py
```diff
@@ -71,6 +72,9 @@
 class EditEvent(EditPrimary):
 
     def __init__(self, dbstate, uistate, track, event, callback=None):
+        """"""
+        self.callback = callback
```
Nick-Hall: Redundant.  This is already set in the parent class.

gramps/gui/editors/editevent.py
```diff
@@ -71,6 +72,9 @@
...
+        self.action = uistate.action.split('-')[1]
```
Nick-Hall: We should be able to determine the action from the handle and gramps_id.

gramps/gui/editors/editeventref.py
```diff
@@ -241,7 +241,7 @@ def _create_tabbed_pages(self):
                              self.uistate,
                              self.track,
                              self.db.find_backlink_handles(self.source.handle),
-                             self.enable_warnbox)
+                             callback=self.enable_warnbox)
```
Nick-Hall: Not part of this change.  Please put pylint improvements in a separate commit.

gramps/plugins/view/eventview.py
```diff
@@ -290,16 +304,38 @@ def merge(self, obj):
...
+        """
+        event_list = self.selected_handles()
+
+        if len(event_list) != 1:
+            msg = _("Cannot clone event object.")
+            msg2 = _("Exactly one event must be selected to perform a clone.")
+            ErrorDialog(msg, msg2, parent=self.uistate.window)
+        else:
+            event = Event()
+            event = copy.deepcopy(self.dbstate.db.get_event_from_handle(event_list[0]))
```
Nick-Hall: Alternatively, you could use:  `event = Event(source=old_event)`

apoettker:
This is how it works:
In eventlist.py we copy an event with everything (ID and handle too), and then delete the ID. The references of Person and Family objects not shown in eventbackreflist.py. If the user save this temporary event, handle will be deleted, it will get a new ID and handle, and have no references anymore .

An alternative will be to compile an temporary event in eventview.py  with all the related tags, citations, media, attributes and notes, but without ID and handle (like the Add method, but here with default variables). This is a much better design, but we need a clear identifier to differentiate between the Add and the Clone action (both cases will have ID and handle equal None).

apoettker:
The Event list get a new button: Clone. Its created within the style of the Merge button, but visualize the two events after this operation. 

clone event i

After pressing the button, the know edit event dialogue pops up, but without ID and handle. 

clone event dialogue

After hitting OK the new created event is sorted into the event list (w/o main participant).

clone event ii

Nick-Hall:
I don't think it is really necessary to modify the event editor.  It should be possible to implement this as a third-party addon.

Nick-Hall:
This is better than the related PR #605.  A "Clone" button next to the existing "Merge" button in the event view toolbar is more consistent with the Gramps user interface.

However, I am not convinced that this functionality is really needed and am concerned that it may confuse users.  Moving towards source-based data entry is the way forward.

An alternative approach would be to write a third-party clone tool, or an enhanced event view.  This could be done without modifying the existing event editor.

In future, this type of feature should be discussed on the mailing list first.  I am going to close it now.

PR #601, opened 2018-04-20
Fix Tag editor for multiple tag removes

prculley:
Fixes [#10552](https://gramps-project.org/bugs/view.php?id=10552)
User attempted to remove multiple tags without closing dialog.  Seems that original code did not update the model correctly during delete.

Should back port to gramps42...


sam-m888:
Thank you

PR #603, opened 2018-04-22
Fix Person/Family/Event view updates on various associated changes

prculley:
Fixes [#10532](https://gramps-project.org/bugs/view.php?id=10532)

This attempts to correct some of the views for updates that effect the 'other' columns in the view.  For example the original bug report noted that the Person view birth/death event date/place were not updated when updating those events.

The 'fix' was to find backlinks to affected objects and do the row updates on those objects.

During testing I noticed that Event view Main Participants also needed work; but it needed a different method to find/update.  I added a second commit to deal with add/update of a person or family, however I could not figure out an elegant (fast) method of dealing with person/family delete.  At the moment, I just have it redoing the whole view on a person/family delete (slow).  The problem is I don't have a good way to figure out what rows to update, based on the person-delete signal.  It includes person handles, but the person objects are already gone from the db, and the event model doesn't contain Main Participant handles to compare against.  Maybe something from the undo db????

Another category that was not covered well was Place updates.  If the update is to an enclosing object (example, change country name, most places in views will not be updated).  Also for Person view, if the birth/death place changed it was not getting updated.  The third commit deals with these issues.

### DIFF
--- a/gramps/gui/views/listview.py
+++ b/gramps/gui/views/listview.py
@@ -33,6 +33,7 @@ from abc import abstractmethod
 import pickle
 import time
 import logging
+from collections import deque
 
 LOG = logging.getLogger('.gui.listview')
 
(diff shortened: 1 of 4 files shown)


sam-m888:
Thank you

PR #605, opened 2018-04-23
Clone Event IIA

apoettker:
Embedded Clone Event w. references (2nd pass, part A)

Feature request [#9604](https://gramps-project.org/bugs/view.php?id=9604)

This functionality in the Person / Family area doubles an event with all related tags, citations, media, attributes, notes and references in the embedded Event Reference editor. Its like an event addition, but with default values.


Nick-Hall:
Closed for the reasons mentioned in PR #599.

PR #607, opened 2018-05-01
Merge embedded events

apoettker:
Merge embedded events.

Feature requests [#8249](https://gramps-project.org/bugs/view.php?id=8249), [#10556](https://gramps-project.org/bugs/view.php?id=10556).

Merging of events inside the related objects Person / Family save's a lot of time while streaming up your database after data import. This PR uses the established method of merging events out of the embedded events tabs in the person / family editors. Afterwards the model will be rebuild, the event list will be refreshed.


apoettker:
The embedded event tab in Person menue get the standard merge button. 

merge person embedded event

After pressing the merge button, next known dialogue pops up:

merge person embedded event dialogue

Now merging can be proceed as usual. Note: Merge button is greyed out until two events are selected. This avoid unnecessary warning pop-ups.
Merging for family events follows the same path.

apoettker:
This PR is independent, but should merge into master after #599 and #605 due to minor code improvements.

Nick-Hall:
@apoettker I probably didn't explain myself clearly enough.  The _Associations_ tab contains a list of person references (`PersonRef` objects) and the _Events_ tab contains a list of event references (`EventRef` objects).  They are not lists of people or events.  In one PR you merge the referenced objects and in the other you don't.

Allowing an editor to modify referenced objects is a fundamental change in design philosophy, and hence must be agreed on the mailing list first.

apoettker:
@Nick-Hall: I'm aware - and so is the enhancement intended - that in the _Association_ tab references  (PersonRef objects) will be merged. In the _Events_ tab I grep through event references  (EventRef objects) down to the events and let the Event Merge Dialogue popup, which at the end merge events. 

All my programming here is making Gramps a little bit more: user friendly. We've heard this sound over and over again from external people. In my understanding the Person / Family dialogue is the right area to double (PR #605) or to merge (PR #607) events, cause here the events belongs to the person / family. 
As an alternative way remembering the key data, going to the Event list, searching this in a big event base, then cloning or merging, then going back, picking the next person / family, is totally boring and a waste of time, it needs far to much clicks. 
If the above requested working style is a fundamental change in design philosophy maybe we can enhance our design philosophy a little in the proposed direction? 

My request for approval is out in the developer list since three weeks (Apr. 24th).

Nick-Hall:
> My request for approval is out in the developer list since three weeks (Apr. 24th).

@apoettker Thanks.  I have replied to the thread with my comment.

Nick-Hall:
I am rejecting this and also PR #616.

The events and associations listed in the person editor are both reference objects, but in one case the references are merged and in the other the referenced objects are merged.  This is not consistent.

The need for this functionality seems to arise from the person merge not working as expected.  Perhaps it would should improve the person merge instead.  An alternative would be to write a tool to locate and fix this type of problem.

PR #609, opened 2018-05-02
Stop malformed Gramps XML creatation due to missing blank [surname] tags

sam-m888:
Fixes [#10078](https://gramps-project.org/bugs/view.php?id=10078)

### DIFF
--- a/gramps/plugins/export/exportxml.py
+++ b/gramps/plugins/export/exportxml.py
@@ -1011,6 +1011,8 @@ class GrampsXmlWriter(UpdateCallback):
         self.g.write('>\n')
         self.write_line("first", firstname, index+1)
         self.write_line("call", name.get_call_name(), index+1)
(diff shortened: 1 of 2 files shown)


prculley:
Regarding the vcard test error; it appears that the test was creating a person with a Nickname, but no surname.  Your change seems to be creating Gramps XML with the <surname /> attribute added; presumably an empty surname as opposed to NO surname.  It is probably easiest to modify the test to add an empty surname to the nickname tests.
`self.vcard[2] = "FN:Lastname"`
I don't have time to try and debug this at the moment, maybe in a couple of days.

sam-m888:
Thanks for the hint eg: Lastname == surname

Hopefully I'm fixing the correct issue and not hiding something!

prculley:
See my latest note on https://gramps-project.org/bugs/view.php?id=10078

johnbickmore:
https://github.com/gramps-project/gramps/pull/613

PR #612, opened 2018-05-06
Export options > Gui alignment issue

SNoiraud:
Fixes [#10576](https://gramps-project.org/bugs/view.php?id=10576)


sam-m888:
While it fixes the issue of the box alignment issue, it introduces an issue where attempting to move up or down a line does not work (some other random line moves instead)

PR #613, opened 2018-05-09
Fix bsddb for person sort with empty Surname list

prculley:
Fixes [#10078](https://gramps-project.org/bugs/view.php?id=10078), [#10577](https://gramps-project.org/bugs/view.php?id=10577)

First found with an attempted export of a tree with persons having no Surnames at all (not just a list with an Empty Surname).  Some importers have been creating dbs this way for a long time, so the case is present in a lot of user trees.

Issue also occurs with some reports.

A gramps50 change added the ability to sort returned lists of people; this was not present in gramps42, the assembly of a fullname from a name that did not take into account the possibility of no surname was the issue.

### DIFF
--- a/gramps/plugins/db/bsddb/read.py
+++ b/gramps/plugins/db/bsddb/read.py
@@ -129,9 +129,12 @@ def find_fullname(key, data):
     #   surname primary,
     #   surname origin type,
     #   surname connector)]
-    fullname_data = [(data[3][5][0][0] + ' ' + data[3][4], # surname givenname
-                      data[3][5][0][1], data[3][5][0][2],
-                      data[3][5][0][3], data[3][5][0][4])]
+    if data[3][5]:  # if Surname available
+        fullname_data = [(data[3][5][0][0] + ' ' + data[3][4],  # combined
+                          data[3][5][0][1], data[3][5][0][2],
+                          data[3][5][0][3], data[3][5][0][4])]
+    else:  # Some importers don't add any Surname at all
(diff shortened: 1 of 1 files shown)


johnbickmore:
Travis-ci [failed](https://travis-ci.org/gramps-project/gramps/builds/376846279) not sure if related to this Pr seems more an issue with travis

Paul-Franklin:
For the record, I still feel this approach should not
be committed, for the reasons I gave in this thread:

http://gramps.1791082.n4.nabble.com/request-for-review-for-5-0-0-td4682172.html

I feel that the patch I attached to that thread should
be committed instead of this pull request, as I said.

However I am willing to discuss it with you, if you want.

Ideally leaving any timing issues out of the discussion.  8-)

If you want to continue the discussion, I suggest that
it be continued on that thread in gramps-devel, since as
I have commented several times, github comments are neither
archived (by SourceForge or Nabble) nor indexed by Google.

However, if you still feel as you said earlier, that this
p.r. is preferable and my patch is "not really necessary,"
then that is fine with me, and as I said on that thread in
May, in that case we both agree that Nick should be the one
who decides which approach is preferable.  Just say so.

Respectfully.

wroldwiedbwe:
@Paul-Franklin Last [message](http://gramps.1791082.n4.nabble.com/request-for-review-for-5-0-0-tp4682172p4682295.html), from @Nick-Hall mentions 

> 
> Your suggestion seems reasonable.
> 
> The next step is to submit this change as a pull request.  Our new
> contribution policy requires a PR for all such changes.
> 
> Regards,
> 
> 
> Nick.

PR #618, opened 2018-05-15
Fix dialog button order on non-Mac systems

Nick-Hall:
Resolves [#10585.](https://gramps-project.org/bugs/view.php?id=10585)

### DIFF
--- a/gramps/gui/glade/editplace.glade
+++ b/gramps/gui/glade/editplace.glade
@@ -60,6 +60,7 @@
                 <property name="expand">False</property>
                 <property name="fill">False</property>
                 <property name="position">2</property>
+                <property name="secondary">True</property>
               </packing>
             </child>
           </object>
--- a/gramps/gui/glade/editplaceref.glade
+++ b/gramps/gui/glade/editplaceref.glade
@@ -31,6 +31,7 @@
                 <property name="expand">False</property>
                 <property name="fill">False</property>
                 <property name="position">0</property>
+                <property name="secondary">True</property>
               </packing>
             </child>
             <child>


SNoiraud:
It's now correct. The help button is on the left side for me (ubuntu 18.04)

jralls:
Doesn't fix the problem on Mac.
```xml
diff --git a/gramps/gui/glade/editplace.glade b/gramps/gui/glade/editplace.glade
index a3dacc324..aab71e3f4 100644
--- a/gramps/gui/glade/editplace.glade
+++ b/gramps/gui/glade/editplace.glade
@@ -60,6 +60,7 @@
                 <property name="expand">False</property>
                 <property name="fill">False</property>
                 <property name="position">2</property>
+                <property name="secondary">True</property>
               </packing>
             </child>
           </object>
```
on the other hand does.
It doesn't explain why none of the other help buttons in Gramps need that property set. It's also odd that  _gtk_builder_parser_parse_buffer cares that the file is read and made into a string by Python instead of g_file_get_contents, but that's not our concern here.

SNoiraud:
For linux, only adding the property solves the problem. no need to repatch glade.py

Nick-Hall:
> Doesn't fix the problem on Mac.

@jralls I didn't expect it to, which is why I asked for your review.

The same problem occurs in the place reference editor.  Are there any other occurrences?

jralls:
No others that I found, including perhaps oddly editplacename. You can make the same addition to editplaceref.glade.

sam-m888:
Confirming that I can not find any others also.

PR #620, opened 2018-05-26
Fix dialogs for crash when canceling via 'x'

prculley:
Fixes #10607
crash on cancelling import dialog with 'x'
crash on cancelling backup dialog with 'x'
crash on cancelling Reports/Graphs/Family Lines/Family Colors/add: select surname dialog with 'x'


sam-m888:
Thank you

PR #622, opened 2018-05-29
Fix menus when operating with read-only db

prculley:
Fixes [#10614](https://gramps-project.org/bugs/view.php?id=10614)

An earlier attempt to deal with closed dbs messed up the read-only ActionGroup settings.  This corrects that situation.


sam-m888:
Works for me.

sam-m888:
Thank you

PR #623, opened 2018-05-29
Fix dbapi dbs for closeing read-only db crash

prculley:
Fixes [#10613](https://gramps-project.org/bugs/view.php?id=10613)

The sqlite (and other dbapi dbs) tried to save their metadata on a close operation, which doesn't work for read-only dbs.


sam-m888:
Looks like Travis [failed](https://travis-ci.org/gramps-project/gramps/builds/385265071?utm_source=github_status&utm_medium=notification) for some unrelated reason on this one!

sam-m888:
Thank you

PR #624, opened 2018-05-29
Fix error when opening bsddb db in read-only mode

prculley:
Fixes [#10615](https://gramps-project.org/bugs/view.php?id=10615)

BSDDB had errors when trying to open in read-only mode.

This should probably be back-ported to gramps42 (if we decide to release more 4.2.x versions).


sam-m888:
Travis failed due to [trailing white space](https://travis-ci.org/gramps-project/gramps/builds/385699182#L2816)

> gramps/plugins/db/bsddb/write.py:622:                    db.DB_INIT_MPOOL 

sam-m888:
Thank you

>> This should probably be back-ported to gramps42 (if we decide to release more 4.2.x versions).

@prculley Yes, can you provide please, even if we don't release a new 4.2.x version

PR #625, opened 2018-06-02
Fix Family Lines/Family Colors picker for bad trasient parent

prculley:
Fixes [#10623](https://gramps-project.org/bugs/view.php?id=10623)


sam-m888:
Thank you

PR #626, opened 2018-06-03
Fix import_as_dict to utilize user gramps_id prefixes

prculley:
Issue [#10622](https://gramps-project.org/bugs/view.php?id=10622)

Found when comparing dbs with large prefixes 'I%05d'; using database differences or import_merge addons.

https://github.com/gramps-project/addons-source/pull/152#pullrequestreview-125397953


sam-m888:
Thank you

PR #627, opened 2018-06-08
Statistics Charts crashes with IndexError

SNoiraud:
Fixes [#010626](https://gramps-project.org/bugs/view.php?id=10626)

In statistics charts, get_month is only localized and the calendar is unused.
If we have an Extra (Sansculottides) month in the french republican calendar
(index 13), we have this IndexError because the gregorian calendar is always
used and contains only 12 values.


sam-m888:
Works for me.

sam-m888:
Thank you

PR #628, opened 2018-06-10
Fix error when opening bsddb db in read-only mode for gramps42

prculley:
Fixes [#10615](https://gramps-project.org/bugs/view.php?id=10615)
BSDDB had errors when trying to open in read-only mode.


sam-m888:
Thank you

PR #629, opened 2018-06-11
Fix Gramplet configure (View/Configure) for large options

prculley:
When a Gramplet uses a BooleanListOption with a lot of entries, along with other options, all the options take on the size of the large BooleanListOption, which doesn't work well at all.
viewconfigbefore
This PR corrects this.
viewconfigafter
I've looked over the various Gramplets that use config options, and the PR doesn't seem to affect any others.


sam-m888:
> looked over the various Gramplets that use config options, and the PR doesn't seem to affect any others.

Confirming this works and that I checked gramplet addons that allow configuration and could not find an issue.

sam-m888:
Thank you

ThariusDawnbreaker:
Thank you Sam

sam-m888:
@ThariusDawnbreaker Your welcome, but the thanks is all to @prculley

PR #630, opened 2018-06-13
Fix Custom filter update when created via sidebar

prculley:
Fixes [#10620](https://gramps-project.org/bugs/view.php?id=10620)

Adding a filter via the sidebar editor originally only updated the current view, other associated views did not see the new filter.  First noticed in People views.  This causes the 'filters-changed' to be emitted which updates all associated view.


sam-m888:
Been using this fix for the past week with no issue that I can find 👍

sam-m888:
Thank you

PR #631, opened 2018-06-21
Fix BaseSelector to avoid long delay before display on large trees

prculley:
Fixes [#10634](https://gramps-project.org/bugs/view.php?id=10634)

The bug report indicates it is possible for the user to create two (or more) modal windows if he started by utilizing a Selector (PersonSelector was demonstrated) that took a long time to appear.  The problem was also caused by the 'loading items' progress indicator which allowed additional Gtk events to be executed during the wait.  The user started a second modal window (a Person Editor) during the wait.  The multiple modal windows confuses Gtk and can result in apparent Gramps freezes.  In this case, under Windows, the Selector dialog was hidden behind the Edit dialog, even though it was active and focused.  If the user shifted the dialog windows to access the Selector dialog, and closed it, both dialogs disappeared and Gramps appeared unresponsive to mouse (a keyboard 'ESC' would still close the hidden Person Editor).

I fixed this by causing the Selector Dialog to show before building the model it is showing.  For large models the Dialog can now appear empty until the model completes.  Previously the dialog did not appear at all until it was fully built.  The fix prevents the user from starting a second dialog.

It may be we should add a 'Loading Items..." into the Selector Dialog during the delay before showing the model; this would be a bit more user friendly, this PR doesn't do that.

### DIFF
--- a/gramps/gui/glade/baseselector.glade
+++ b/gramps/gui/glade/baseselector.glade
@@ -121,6 +121,7 @@
                 <property name="use_action_appearance">False</property>
                 <property name="can_focus">True</property>
                 <property name="receives_default">False</property>
+                <property name="no_show_all">True</property>
                 <property name="halign">center</property>
                 <property name="use_underline">True</property>
                 <property name="xalign">0.5</property>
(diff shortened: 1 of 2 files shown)


lordemannd:
bad things happen after taking this change, Tools->Family Tree Repair->Rebuild Reference Maps.
Gramps seems to working very hard on something (fan speeds up) but makes no progress.
with plain 5.0rc1, this tools completes in a few seconds, after this changes it runs on for a long time, had to kill it.
on Win 10, 5.0rc1 + plus current changes on gramps50 PLUS this PR #631.
(I will now try with 5.0rc1 plus only PR 631.)

prculley:
@Nick-Hall Some of the models seem to use the status progress bars, some do not.  The Busy cursor is already present.  I have added a 'Loading items...' (already in gramps.pot) to the dialog and made sure that it fully appears (previously it was a blank box for some data types) that goes away when the model is built.

@lordemannd I'm pretty sure that this PR has nothing to do with the long Rebuild Reference Maps time; that operation is very dependent on the size of the db, if you were testing with a large db, it can take a while, and the progress bar is frozen as well.

lordemannd:
I have been using these changes for many weeks. I do see the "Loading Items..." message and have not been able to click the add new button to get to the problem that was reported. My DB is only 8000 or so people. But since I do see the "Loading Items..." message I think am testing the initial problem. Win 10, Gramps 5.0 RC1. I don't see any unwanted result to these changes.

PR #633, opened 2018-07-01
Correct binary test logic for primary mask.

jralls:
state & get_primary_mask(another) tested (state & (primary | other)),
which will be true if state matches *either* primary *or* other, but
what is wanted in a not-negated test is state matching all bits of
(primary | other). match_primary_mask does that.

On the other hand there are also cases of "not state & (primary | other)".
no_match_primary_mask handles that, returning true if state matches none
 of the bits in (primary | other).

Fixes [#10646](https://gramps-project.org/bugs/view.php?id=10646).


prculley:
I think you have the wrong bug tracker number on your commit...

jralls:
Thanks, fixed.

PR #635, opened 2018-07-13
Merge Gramps50 with Master

prculley:
As the title says.  During the merge, I encountered several thousand conflicts with the PO files, and saw that some of the automatic merge steps in the po files were not being done properly.  It seems that some po files had been updated on both master and grampss50 branches, other master branch files appeared to be even older than the original 'base' of the merge.

So I decided to use a slightly different process there.  I initially just completed the merge, just selecting everything from gramps50 for po files, and a careful conflict resolution for everything else, and then used the gettext msgcat utility to merge the po files from Gramps50 and original master branch (before merge).  The msgcat properly merges po files, not losing ANY strings, so there might be a few extras when complete, though these will be ignored during the next template merge.  Except for the nn.po, I let the msgcat use the gramps50 version of strings that were translated differently in both (nn.po was clearly just recently redone on master, but not on gramps50).
The results of this were added as a last commit.


jralls:
> I don't think this can be done via a PR because we will merge a merge, which is not what we want.

Agreed. Someone with push should just do it at least every 6-8 weeks, 50->51->master. Run tests at each step and raise a bug (and don't push, of course) or fix it if you can if anything breaks. Anyone who pushes a major change should go ahead and merge up because the details of the change will be fresh in his mind and that will make any needed reconciliation much easier.

Nick-Hall:
@prculley It looks good.  Thank you for volunteering to do this on a regular basis.

How exactly are you using `msgcat` to merge the translation files?  Some translators seem to work in _master_.  Will their recent translation get overwritten?

prculley:
As I mentioned in the first post, msgcat preserves all unique strings (even obsolete ones) from all sources, in my case master and gramps50.  In the case of duplicated msgid with same translations, the source file header comments are preserved from both  (not that it matters much).

In the case of duplicated msgid with different translations (translator did it differently on master vs gramps50) I preserved the translation from the Gramps50 branch, since for all but one of the translations, this was the more recent.  In the one case where the master branch was more recent, I preserved its translations for the duplicated msgids.

Because the one translator did his work in master, we are missing a few translations (recent string changes around 'dbapi' vs. 'sqlite' for example).  I also have not considered trying to backport or redo his translation to gramps50; the release will end up with a pretty old version as it is.

Nick-Hall:
> Because the one translator did his work in master, we are missing a few translations (recent string changes around 'dbapi' vs. 'sqlite' for example). I also have not considered trying to backport or redo his translation to gramps50; the release will end up with a pretty old version as it is.

Only one.  I was concerned there may be more.

It should be easy to use `msgmerge` to merge the po file from _master_ into the pot file from _gramps50_.

We just need to make it clear to the translators which branch to use.  Many of them are non-technical.

PR #637, opened 2018-07-19
Geps044 replace deprecated Gtk.UIManager and associated APIs with newer versions.

prculley:
The PR is about replacing the deprecated Gtk UIManager and associated Actions and ActionGroups.
A discussion about this so far is available at [https://gramps-project.org/wiki/index.php/GEPS_044:_Replace_Deprecated_Gtk.UIManager#Issues](url).

Current status is that it is pretty much complete, unless additional issues come up or comments are raised.  The one area currently not complete is MAC OSX integration; it works on a MAC but certain menu items are duplicated in the 'App menu' and their original location in the Gramps menubar.  Also certain items added by the OSX code (emoji and dictation menu items, at least) are currently missing from the updated code version of Gramps, they were present with the old APIs.

I've started the PR now so that if there are comments about anything, this is a place to let me know, so I can get them addressed.


prculley:
John Ralls has looked at the code as well;

> Looks pretty good.  I think you can go ahead and submit this as a PR for Nick’s review whenever you think you’re ready.

johnbickmore:
Tested ok, only thing of note that I don't like is that the shortcut hints are missing from drop down [menu's](https://gramps-project.org/wiki/index.php?title=Gramps_5.0_Wiki_Manual_-_Navigation#Main_Menus)  righthand side eg: if you select the "Family Trees" menu and look at the entry for "Manage Family Trees..."  the shortcut hint "Ctrl + O" is no longer shown; can they be turned back on?

prculley:
@johnbickmore Could you tell us what system/OS version etc. you are using?  On the 5 systems I tested (Windows, 3 Linux variations, and OSX), the 'accelerator' shortcuts appear as they always have, so if you are not seeing them there must be something different about your system.  Perhaps paste in the printout of "Gramps -v" or the output of the "Prerequisites Checker" addon, so I can figure out what might be different.

gramps/gui/grampsgui.py
```diff
@@ -50,7 +51,365 @@
 
 MIN_PYGOBJECT_VERSION = (3, 12, 0)
 PYGOBJ_ERR = False
-MIN_GTK_VERSION = (3, 10)
+MIN_GTK_VERSION = (3, 12)
```
Nick-Hall: Need to also update the README file.

gramps/gui/uimanager.py
```diff
@@ -0,0 +1,507 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""
+A replacement UIManager and ActionGroup.
+"""
+
+import xml.etree.ElementTree as ET
+from gi.repository import GLib, Gio, Gtk
```
Nick-Hall: Incorrect import order.

gramps/gui/uimanager.py
```diff
@@ -0,0 +1,507 @@
...
+                            parent.insert(indx, update)
+                else:
+                    # updated item not present in original, just add it
+                    # This allow addition of popups etc.
+                    self.et_xml.append(update)
+            #results = ET.tostring(self.et_xml, encoding="unicode")
+            #print(results)
+            #print ('*** Add ui')
+            return changexml
+        except:
```
Nick-Hall: What exception is being caught?  Consider using `Logger.debug`.

There is also some commented out debug code.

gramps/gui/viewmanager.py
```diff
@@ -107,148 +108,20 @@
...
-        import gi
-        gi.require_version('GtkosxApplication', '1.0')
-        from gi.repository import GtkosxApplication as QuartzApp
-        _GTKOSXAPPLICATION = True
-    except:
-        print("Failed to import gtk_osxapplication")
-        _GTKOSXAPPLICATION = False
-else:
-    _GTKOSXAPPLICATION = False
+# if is_quartz():
```
Nick-Hall: Just remove this?  Are there still Mac issues?

gramps/gui/uimanager.py
prculley: The except is to allow diagnostic printout on a crash, with several pages of printout, I did not think using the logging mechanism was really appropriate.  I have added the 'assert False' to complete the crash, as continuing after one of these errors is quite problematic.  It was distressingly easy to cause errors during my development with small syntax issues in the XML etc. so having the extra printout was quite useful.

The extra commented out debug code is something I kept having to put back in to find my issues, I don't think I want that in place and spewing output (even to a logger) unless specifically examining a problem.  I guess I could make it live code with an 'if MYDEBUG' around it, and the MYDEBUG defined at top if you think that is a better way to go.

gramps/gui/viewmanager.py
prculley: I did quite a bit of testing on my OSX VM, I think we are in pretty good shape.  John Ralls also looked at it and seems to concur.  The Gtk.Application and related APIs have covered most of the bases for OSX that GtkosxApplication used to deal with, there were only a few items that had not been fully covered as of the Gtk version that we are packaging for MAC usage.  These items are taken care of with a few scattered is_quartz() calls.

There is one remaining mystery, as far as I am concerned.  OSX automagically adds a few menu items for things like insert emojii, or speech recognition.  For some reason I have not been able to determine, these are not working for this PR code.  On the bright side, even though they were added with our older code, they crashed Gramps if you tried to use them.  On the whole I think this is a win.  If I ever have too much time on my hands, I might try to set up a full MAC debug environment so I can get a better look inside their API code and try to resolve this.  Or perhaps John might give it a go...

gramps/gui/uimanager.py
prculley: Fixed

prculley:
Also fixed up the potfiles.skip, I just don't seem to think about these unless reminded during release...

gramps/gui/uimanager.py
QuLogic: You should still be more specific about the exception; at least use `Exception`, this would not catch things like `SystemInterrupt` (aka Control-C on the terminal.)

PR #638, opened 2018-07-25
Update the Debian changelog after the 5.0.0 release

RossGammon:
Yay - Gramps 5.0 is out there! Well done all.

Just updating the Debian changelog to match the *.deb file just uploaded for the release. No other changes to packaging since rc1.


Nick-Hall:
Thanks for creating the Debian package.

PR #640, opened 2018-07-31
Changed layout of places and dates in familylines plugin

mjbmz:
[#10699](https://gramps-project.org/bugs/view.php?id=10699); [#7788](https://gramps-project.org/bugs/view.php?id=7788l)
In the current layout of the familylines plugin, there is no possibility to differ between birthplace and deathplace when there is only one defined. This changes the layout to looks much sleeker and helps to differ between the place types.

### DIFF
--- a/gramps/plugins/graph/gvfamilylines.py
+++ b/gramps/plugins/graph/gvfamilylines.py
@@ -874,22 +874,20 @@ class FamilyLinesReport(Report):
             elif self.includeid == 2: # own line
                 label += "%s(%s)" % (line_delimiter, p_id)
 
-            if birth_str or death_str:
-                label += '%s(' % line_delimiter
+            if birth_str or birthplace:
+                label += '%s* ' % line_delimiter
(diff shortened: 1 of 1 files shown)


FuruFam61:
Please remember that not every gramps user comes from the same religious
tradition as you do.  In particular, I strongly suggest that having a cross-shaped
symbol for the death of every person is not appropriate for everybody.

If you really think it won't be obvious (which I doubt), why not just use the
standard gramps abbreviations "b." and "d." (which I think already have
translations in other languages, even)?

Nick-Hall:
Yes.  Perhaps we should use the standard abbreviations for now.

Genealogical symbols will be added in PR #598 .  See also [GEPS 039](https://gramps-project.org/wiki/index.php/GEPS_039:_Genealogical_symbols_in_gramps).

Nick-Hall:
I am closing this now.  Please feel free to resubmit if you can fix the conflicts.

PR #643, opened 2018-08-04
Fix Find Database Loop Tool (bad import of _collections)

prculley:
Fixes [#10722](https://gramps-project.org/bugs/view.php?id=10722)

Inappropriate cut/paste from web error; apparently in some configurations there IS an '_collections' Python module, but not in all configurations.  Which is why I did not spot this the first time, sigh...


sam-m888:
Thank you

PR #644, opened 2018-08-04
Fix place reference editor for bad cut/paste on set_latlongitude

prculley:
Fixes [#10719](https://gramps-project.org/bugs/view.php?id=10719)

Bad cut/paste masked by raw except...


sam-m888:
Thank you

PR #646, opened 2018-08-06
Fix validatedmaskedentry Deprecated GObject.PARAM_READWRITE

prculley:
use GObject.ParamFlags.READWRITE (glib 2.42+) instead


Nick-Hall:
Gramps v5.0 requires Gtk 3.10 and above, which in turn requires GLib 2.37.5 and above.

FuruFam61:
If you want to test things like that, you could put Ubuntu 14.04 into a VM, since it has Gtk 3.10 and is LTS so will be supported until 2019.

PR #648, opened 2018-08-09
Fix strings containing deprecated illegal escape sequences

prculley:
Python 3.6 and above has deprecated illegal string escape sequences.  Escape sequences are preceded by a `'\'` and valid ones are "\n\t\r" etc.  Illegal ones are not in the list https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals.

Previous to Python 3.6 these illegal sequences were ignored and the `'\'` was left in place.  Pylint has been noting these for a while now.

This PR corrects these sequences in Gramps.  I used
`find . -name "*.py" | xargs -t -n 1 python3 -Wd -m py_compile 2>&1 | grep Depre`
to locate the failing strings.

I've done this work in gramps50 branch, as I expect it to live for a while.


sam-m888:
Thank you

PR #649, opened 2018-08-09
Fix generate_checksum routine to avoid MemoryError crash

prculley:
with very large files and 32-bit OS

Issue [#10690](https://gramps-project.org/bugs/view.php?id=10690)

Original method tried to load entire file into memory for processing at once; crashed on 32-bit systems.  This processes file in 64k chunks.

Note that I also am fixing the MediaVerify Addon to use the now fixed gen/utils/file generate_checksum routine instead of rolling its own.  So it will also start working when this PR is accepted.


sam-m888:
Thank you

PR #651, opened 2018-08-13
[WhatsNext gramplet]fix warning

vantu5z:
check if db is open before requesting: get_tag_handles

Fixes bug [#10732](https://gramps-project.org/bugs/view.php?id=10732).


prculley:
Does this have an associated bug?  If so, your commit message should include a blank line followed by
`Fixes #123456`

sam-m888:
Thank you

PR #652, opened 2018-08-13
Fix <ctrl>c in view to get selected item to clipboard

prculley:
Fixes [#10682](https://gramps-project.org/bugs/view.php?id=10682)

Not sure what happened the last time I tried this PR, but it seems to have lost the change commit and thus got closed.  Going through old emails from when I was on vacation so found this.  So lets try again.


sam-m888:
Thank you

PR #653, opened 2018-08-17
Fix and restore Statistics Gramplet to 4.2.x status

prculley:
Fixes [#10756](https://gramps-project.org/bugs/view.php?id=10756)

A bit of history.  bug [2060](https://gramps-project.org/bugs/view.php?id=2060) was filed a long time ago (2008), and a patch was added to close it in 2016.  I believe that the bug had already been patched by then, although I cannot be sure.  The patch caused the Gramplet to update anytime the active-changed signal occurred on a person view.  I suspect that this caused a fair amount of overhead on very large dbs as the entire person list was rescanned.

In any event, Doug Blank recently removed the original scan code, as well as some useful functionality.  And left behind a bug where a value was always zero.  The users email list had some complaints about the lost functionality, and I also saw some recent complaints that the gender statistics were incorrect.  Turns out that the db's genderstats data was incorrect, probably due to crashes after db updates.  The Genderstats don't get saved until db closes normally.  In addition, running the Genderstats rebuild tool, did not immediately appear to correct the situation, Gramps had to be restarted to show the updated results.

I added the original statistics code back into the Gramplet, suitably updated for 5.0.x.  So the Gramplet no longer depends on dbs genderstats, it scans for it now on its own.

I removed the update on active-changed function, so overhead is only present on actual db updates (which have their own signals).  I tested that the add, edit, and delete person changes do properly cause the statistics to update now.  So the active-changed signal is not necessary.

### DIFF
--- a/gramps/plugins/gramplet/statsgramplet.py
+++ b/gramps/plugins/gramplet/statsgramplet.py
@@ -48,13 +48,15 @@ _YIELD_INTERVAL = 200
 # StatsGramplet class
 #
 #------------------------------------------------------------------------
+
+
 class StatsGramplet(Gramplet):
+    """ Display some statistics about the tree, with links to filtered quick
+    reports
+    """
     def init(self):
         self.set_text(_("No Family Tree loaded."))
         self.set_tooltip(_("Double-click item to see matches"))
-
-    def active_changed(self, handle):
-        self.update()
 
     def db_changed(self):
         self.connect(self.dbstate.db, 'person-add', self.update)
@@ -68,12 +70,17 @@ class StatsGramplet(Gramplet):
     def main(self):
         self.set_text(_("Processing..."))
         database = self.dbstate.db
+        personList = database.iter_people()
 
+        with_media = 0
         total_media = 0
+        incomp_names = 0
+        disconnected = 0
+        missing_bday = 0
         males = 0
         females = 0
         unknowns = 0
-        bytes = 0
+        bytes_cnt = 0
         notfound = []
 
         mobjects = database.get_number_of_media()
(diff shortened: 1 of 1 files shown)


sam-m888:
@prculley Thanks for restoring original functionality.

> This branch has conflicts that must be resolved

If you have a moment can you please rebase this PR.

I'll take care of updating the user manual (text & screenshots).

prculley:
@Sam888 Rebased...  and thank you for merging a lot of my PRs; I was beginning to think I would have to do them myself.

sam-m888:
Thanks for re-basing 👍 

Further testing shows that on Ubuntu 16.04 using the example.gramps family tree I double clicked on each of the links and all worked except for "Number of individuals: 2157" which quickly flashed up the quickview box (it was all black) then Gramps crashes back to command line with error below, when I restart Gramps the family tree is locked. Also tested that link on Win 10 (AIO32/AIO64) and both worked, so maybe only my environment or GTK version in ubuntu 16.04, can you check also please ?


> (xgrampspy3) [grampsdev@lfs101x-VirtualBox gramps (prculley-bug10756)]$ ./Gramps.py 
> 
> (Gramps.py:3937): Gdk-ERROR **: The program 'Gramps.py' received an X Window System error.
> This probably reflects a bug in the program.
> The error was 'BadAlloc (insufficient resources for operation)'.
>   (Details: serial 271156 error_code 11 request_code 130 (MIT-SHM) minor_code 5)
>   (Note to programmers: normally, X errors are reported asynchronously;
>    that is, you will receive the error a while after causing it.
>    To debug your program, run it with the GDK_SYNCHRONIZE environment
>    variable to change this behavior. You can then get a meaningful
>    backtrace from your debugger if you break on the gdk_x_error() function.)
> Trace/breakpoint trap (core dumped)
>

prculley:
@sam-m888 I got the same error when I tried that on my KUbuntu machine.  I also tried a much older version of Gramps on the same machine (4.2.2) and got exactly the same crash.  I think we really have some sort of bug or limitation on the number of things displayed.  `The error was 'BadAlloc (insufficient resources for operation)'.`
I'm haven't yet figured out how to debug Gtk/Gdk issues; I think you have to build a special version of Gtk/Gdk for debug, and then learn how to us an appropriate debugger.  Not in my skill set (yet).

sam-m888:
@prculley Thanks for checking. Tested further on Gramps 4.1 & 4.0 and can confirm the issue has been around for while and looks to not be anything introduced by this PR so I'll merge soon after raising a new issue as I get a different error message if I attach the Statistis Gramplet to the sidebar in any of the category views eg:

> 
> 74133: ERROR: grampsapp.py: line 143: Unhandled exception
>   File "/home/grampsdev/aaSandbox/gramps/gramps/gui/widgets/grampletbar.py", line 429, in __switch_page
>     gramplet.pui.update()
>   File "/home/grampsdev/aaSandbox/gramps/gramps/gen/plug/_gramplet.py", line 309, in update
>     self._generator = self.main()
>   File "/home/grampsdev/aaSandbox/gramps/gramps/plugins/gramplet/statsgramplet.py", line 81, in main
>     for media in database.iter_media():
>   File "/home/grampsdev/aaSandbox/gramps/gramps/plugins/db/bsddb/read.py", line 1154, in g
>     obj.unserialize(data)
>   File "/home/grampsdev/aaSandbox/gramps/gramps/gen/lib/media.py", line 183, in unserialize
>     date, tag_list, self.private) = data
> ValueError: not enough values to unpack (expected 13, got 12)

but it may just be a different error as it mention media.py!

sam-m888:
Thank you

prculley:
@sam-m888 I did a quick check with the statistics gramplet installed in a couple of different sidebars; it worked fine.  The nature of the error traceback leads me to suspect that the db you were using might have had some corrupted data, as the statistics gramplet really doesn't do anything (related to this bug) other than scan media.  It would be a decidedly odd form of corruption; almost like some failure of a db upgrade.

I'd expect a similar exception if you tried to perform a backup, or maybe ran check/repair on the same db.

PR #654, opened 2018-08-19
Fix Merge Family when same parent is missing from both families

prculley:
Fixes #[10760](https://gramps-project.org/bugs/view.php?id=10760)

Another HandleError... where code assumed no handle meant None person when read from db.


sam-m888:
Thank you

PR #655, opened 2018-08-20
Fix corrupted Bookmarks that can happen after Gramps crash

prculley:
Fixes #[10759](https://gramps-project.org/bugs/view.php?id=10759)

I suspect that this is another example of metadata not being saved on a crash.
If you have an object bookmarked when Gramps is closed normally the db contains a copy of bookmarks.
On next session you delete or merge out the bookmarked object, the internal representation of bookmark data is updated, but the db is not, until a normal close.
If Gramps crashes or is terminated with a task manager etc. then the original (now outdated) bookmark data remains in the db.
When Gramps is next opened, the now bad bookmark causes a crash like the bug above.


sam-m888:
Thank you

PR #656, opened 2018-08-31
Speed up Check and Repair, backlinks check stage.

prculley:
There are complaints that the 'backlinks' check portion of Check and repair tool takes too long for large databases.  So I've marked this as a bug...  ( [10618](https://gramps-project.org/bugs/view.php?id=10618) )

This PR optimizes this process by scanning the db once to build backlink tables, and doing the checks against the in memory tables, rather than the database.

I have a db with 15900 generated individuals, 156000 total objects.  The current code backlink scan took 9.5 hours.  An actual (forced) backlink rebuild took 18 min.  With this optimization, the backlink scan drops to 3.2 minutes for the same db.  No changes to backlink rebuild, as that is in the db itself.
The current code working set on my Win64 machine was 249,860k.
The optimized code working set was 528,228k.

I've based this on Gramps50 branch, even though it might be considered an enhancement, as it seems to me to be worthwhile to get to users faster, to reduce the complaints about execution time.  It uses no new translation strings.

### DIFF
--- a/gramps/plugins/tool/check.py
+++ b/gramps/plugins/tool/check.py
@@ -1237,52 +1237,80 @@ class CheckIntegrity:
 
         total = self.db.get_total()
 
-        self.progress.set_pass(_('Looking for backlink reference problems'),
-                               total)
+        self.progress.set_pass(_('Looking for backlink reference problems') +
+                               ' (1)', total)
         logging.info('Looking for backlink reference problems')
 
+        # dict of object handles indexed by forward link created here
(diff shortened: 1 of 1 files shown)


prculley:
See https://gramps-project.org/bugs/view.php?id=10783 for why I had to fix the Travis failure.

sam-m888:
Tested and works for my family tree run time drops from 7.5 hours to 17minutes.

b1
b2
b3



integrity check results

sam-m888:
Thank you

PR #657, opened 2018-09-01
Fix Check and Repair to deal with bad references empty handle string

prculley:
Fixes #[10783](https://gramps-project.org/bugs/view.php?id=10783)
Check and Repair tool doesn't fix bad references when the reference handle is an empty string.
 Found when rewriting the backlinks scan on same tool. I assumed that the prior checks had already corrected any bad forward references, and did not put in a test for them in my new code. The Travis test failed and when I debugged, I found that the tool had an exception on a handle that consisted of an empty string.

Its interesting to note that when I changed the code, a lot more citation and source references were fixed in the test.


prculley:
See https://gramps-project.org/bugs/view.php?id=10783 for why I had to fix the Travis failure.

johnbickmore:
related to https://github.com/gramps-project/gramps/pull/656

sam-m888:
Thank you

PR #661, opened 2018-09-13
Webcal: link problems in some cases

SNoiraud:
Year 2016 is highlighted by default instead of current year.

Fixes #10801


lordemannd:
using pr661 I can still see the original issue by clearing the check mark for "Create multiple year calenders"

lordemannd:
The updated PR looks like it solves the reported issue.

PR #664, opened 2018-09-25
Fix french translation

khrys63:
Fix french translation

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -21,6 +21,7 @@
 # Frederic Chateaux, 2007-2008.
 # Jérôme Rapinat <romjerome@yahoo.fr>, 2005-2017.
 # Mathieu MD <mathieu.md@gmail.com>, 2012.
+# Christophe <khrys63@gmail.com>, 2018.
 msgid ""
 msgstr ""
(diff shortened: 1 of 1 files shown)


romjerome:
@sam-m888 
It sounds good.

Thank you, @khrys63 !

Nick-Hall:
Is this really intended for _master_ instead of _gramps50_?

romjerome:
@Nick-Hall is right.

Note, I made a pull request (quickly closed ...) which is showing a possible issue between 
_master_ and _gramps50_!

Content of '_fr.po_' file should be very close between both branches,
but the pull request #680 is maybe pointing out a possible [missing reference](https://github.com/gramps-project/gramps/pull/680/files#diff-42793ab00dcd448b33776fc86e5eaaf4R39198).

romjerome:
Maybe we could also still try commands like :
`$ cd po`
`$ intltool-update -m`
inherited from previous GNOME ecosystem and still conform with current gramps' template for translation strings.

Nick-Hall:
There were only 4 strings changed in the _gramps50_ branch, so I merged them manually.

PR #665, opened 2018-09-27
Better typography in graph reports

milotype:
Change for better typography
Line 881:
Replaced the 'hyphen' with 'en-dash'


Nick-Hall:
Thanks for your contribution.

PR #671, opened 2018-10-02
Add Ahnentafel option on hourglass

khrys63:
Issue [10826](https://gramps-project.org/bugs/view.php?id=10826)

new_option_ahnentafel v2

Without option :
hourglass_nonordered

With option :
hourglass_ordered

PR #672, opened 2018-10-03
Add hidden nodes for hourglass

rtclay:
Use hidden nodes to fill the space where missing ancestors would go on the chart

### DIFF
--- a/gramps/plugins/graph/gvhourglass.py
+++ b/gramps/plugins/graph/gvhourglass.py
@@ -112,6 +112,7 @@ class HourGlassReport(Report):
             'family': menu.get_option_by_name('colorfamilies').get_value()
         }
         self.roundcorners = menu.get_option_by_name('roundcorners').get_value()
+        self.hiddennodes = menu.get_option_by_name('hiddennodes').get_value()
 
         self.includeid = menu.get_option_by_name('inc_id').get_value()
 
@@ -126,6 +127,7 @@ class HourGlassReport(Report):
             self.arrowtailstyle = 'none'
 
         stdoptions.run_name_format_option(self, menu)
+        self.hiddennode_count = 0
 
     def write_report(self):
         """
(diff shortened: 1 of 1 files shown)


rtclay:
hourglass001

Without:
hourglass002
With:
hourglass003

prculley:
You charts are so large with low resolution I cannot really see what is different, but it appears that some of the parents in the 'with' chart are missing; is it possible you have the images swapped, or do I not understand what you are doing?

prculley:
After reading the code I understand now; I think your tooltip at least should have some additional explanation:
"This option causes space to reserved on the chart where missing ancestors might eventually be found."
Most users son't have any idea what 'nodes' means, hidden or otherwise.

rtclay:
Good idea. I'll try to make clearer.

gramps/plugins/graph/gvhourglass.py
```diff
@@ -158,22 +162,32 @@ def traverse_down(self, person, gen):
...
 
     def traverse_up(self, person, gen):
         """
         Recursively find the ancestors of the given person.
         """
         if gen > self.max_ascend:
             return
-        family_handle = person.get_main_parents_family_handle()
+
+        try:
```
Nick-Hall: The following might be slightly clearer:
```
if isinstance(person, int):
    person_id = person
    family_handle = None
else:
    person_id = person.get_gramps_id()
    family_handle = person.get_main_parents_family_handle()
```

Nick-Hall:
Rebased.

Nick-Hall:
@rtclay Is this ready to merge?

Nick-Hall:
After merging #671 this PR now has conflicts.  Unfortunately, I don't know the code well enough to resolve them.

Nick-Hall:
I am closing this now.  Please feel free to resubmit if you can fix the conflicts.

PR #673, opened 2018-10-03
Restrict Place view Name col to the primary name

prculley:
while allowing searchbar to find alt and primary names.

Currently, the Place View and Place Selector name column shows a comma separated list of the primary and alt names for a place.  With the advent of GeoNames and Gov data, the alt names lists are getting populated with (sometimes long) lists of (sometimes duplicated) names.  This doesn't look very good in the view.  I've gotten a couple of email complaints about this.

The Name column was originally extended (for Gramps50) to include alt names to allow the use of the top search bar (enabled when sidebar is turned off via 'View/Sidebar'), to perform a search through all the alt names as well as primary name.  https://gramps-project.org/bugs/view.php?id=9276

This PR creates a new (hidden) column in the place model containing the full name list, and reverts the 'Name' column to the primary name.  It also modifies the PlaceSelector and PlaceView classes to cause the search bar to use the new (hidden) column.


lordemannd:
nice job, i like it.

romjerome:
@prculley maybe still need to keep a translated string for the separator!
`-return ','.join([data[6][0]] + [name[0] for name in data[7]])`
`+return _(',').join([data[6][0]] + [name[0] for name in data[7]])`

That's minor, but some locales need a space after (or maybe before) a comma.
https://gramps-project.org/bugs/view.php?id=9276#c49975
https://traduc.org/Guides_pratiques/Petit-guide-du-traducteur.html#ponctuation

So, this might be still handled via translation files.

sam-m888:
Needs a rebase please.

Nick-Hall:
@sam-m888 A maintenance release is overdue.  We have more than enough fixes.  Please post to the list and arrange a convenient date.

romjerome:
> @romjerome The string doesn't need to be translated here because no one ever see it, it is used only for searching where the separator is ignored.

@prculley  oh yes, hidden colomn means not visible ...
I just supposed that spaces around separator could generate additionnal 
issues with regex and related searches, but like LTR/RTL for set of strings (textual values) that's something else not related to this feature and handled by gtk (widgets).
Sorry. Thanks.

PR #674, opened 2018-10-03
In personsidebarfilter, search on each part of name

rtclay:
Fixes mantis bug [#7950](https://gramps-project.org/bugs/view.php?id=7950)
Fixes mantis bug [#7952](https://gramps-project.org/bugs/view.php?id=7952)
Instead of requiring that the entire search string matches a single one of the Person's names, the function now requires that each word in the search string matches any of the Person's name fields.


prculley:
Trailing Whitespace error...

PR #679, opened 2018-10-08
NarrativeWeb: add enclosed by and encloses (place)

SNoiraud:
Fixes #10743

### DIFF
--- a/gramps/plugins/webreport/basepage.py
+++ b/gramps/plugins/webreport/basepage.py
@@ -58,7 +58,8 @@ import logging
 #------------------------------------------------
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 from gramps.gen.lib import (FamilyRelType, NoteType, NameType, Person, UrlType,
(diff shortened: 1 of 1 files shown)


lordemannd:
I can reproduce the failure with this DB
[test-PR 679.zip](https://github.com/gramps-project/gramps/files/2469854/test-PR.679.zip)
 . It runs without errors under a clean version of 5.0.0, but with PR 679 and the current changes on the gramps50 branch. I get the above failure.

prculley:
The particular place that is crashing the Narrative Web report is marked 'Private'.  So when it is loaded from the db, it returns 'None'.
Need to test for 'None' on every db get operation; to deal with filtered objects like this.
The report runs if you check off "Include objects marked Private'.

lordemannd:
Yes, I see it now. Always more conditions to test. 
(now I wish I had left myself a note as to why I had that marked as private)

SNoiraud:
Must be applied on master and not on gramps 5.0
I'll add a test for the Private place.

PR #685, opened 2018-10-16
Fix Adding "ToDo" note crash when no active object

SNoiraud:
Fixes [#10795](https://gramps-project.org/bugs/view.php?id=10795)


gramps/plugins/gramplet/todo.py
```diff
@@ -184,13 +185,20 @@ def new_clicked(self, obj):
...
+            from gramps.gui.editors import EditNote
+            note = Note()
+            note.set_type(NoteType.TODO)
+            try:
+                EditNote(self.gui.dbstate, self.gui.uistate, [], note, self.created)
+            except AttributeError:
+                pass
+        else:
+            WarningDialog(_("No active object"),
+                          _("No active object") + _(":") + _(nav_type),
```
Nick-Hall: The string "No active object" appears in the title bar, primary message and secondary message.  Can we think of something more helpful?  For example:  "First select the object to which you want to attach a note".

SNoiraud: I changed the message as you said.

sam-m888: New Example message

example message

sam-m888: @Nick-Hall Is this message ok to you?

sam-m888:
@SNoiraud Please rebase.

PR #688, opened 2018-10-19
Bug10802  Better default directories for file operations.

prculley:
Use recent-export-dir as default for Export View instead of the console default directory, which can be where program resides (on Windows), or other inappropriate place.

Use USER_HOME for default directory on 'Export' (recent-export-dir) if first run (default settings) rather than default directory, which can be where program resides (on Windows), or other inappropriate
place.

Use USER_HOME for default directory on 'Import' (recent-import-dir) if first run (default settings) rather than recent file which is generally set to the grampsdb path.

Fixes [#10802](https://gramps-project.org/bugs/view.php?id=10802)


gramps/gui/views/listview.py
```diff
@@ -1071,6 +1071,10 @@ def export(self, obj):
         combobox.set_active(0)
         box.show_all()
         chooser.set_extra_widget(box)
+        default_dir = config.get('paths.report-directory')
+        if not default_dir:
```
Nick-Hall: Is this necessary?  The ['paths.report-directory'](https://github.com/gramps-project/gramps/blob/master/gramps/gen/config.py#L228) defaults to `USER_HOME` anyway.

prculley:
During my testing I noticed that some of the 'paths' in the gramps.ini were empty strings, so I added the test for empty, just in case.  The choice of reports vs export paths depends on point of view; I think of a CSV of a view as a report, I suppose others might think of it as an export.  It is not in either category, really.  If you would rather have it set to recent-export-dir, I'll change it.
P.S. if I change it, looks like I need to keep the test for empty.

Nick-Hall:
> During my testing I noticed that some of the 'paths' in the gramps.ini were empty strings, so I added the test for empty, just in case.

Yes.  An empty string refers to the working directory, not the home directory.  The `set_current_folder` method should interpret it as such.

> If you would rather have it set to recent-export-dir, I'll change it.

I think that would be better.  The menu item reads "Export View...".

It would probably be neater to change the defaults in the config.  The default import directory can then just be the value of _recent-import-dir_.

PR #690, opened 2018-10-24
Fix Russian date handler crash when Russian .po (actually .mo) isn't present

prculley:
Fixes [#10855](https://gramps-project.org/bugs/view.php?id=10855)

Original code assumed that the Lexeme format of the translated months is available, not true when the language translation is not installed (Window AIO is not installing languages by default anymore).

### DIFF
--- a/gramps/gen/datehandler/_date_ru.py
+++ b/gramps/gen/datehandler/_date_ru.py
@@ -132,6 +132,11 @@ class DateDisplayRU(DateDisplay):
                                                    inflect, long_months)
         elif date_val[1] == 0: # month is zero but day is not (see 8477)
             return self.display_iso(date_val)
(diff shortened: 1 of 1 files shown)


vantu5z:
I don't know a lot of date handlers, but I think Gramps should use default date handler if translation files is not installed, and it will no need to include fix to each date handler separately.
Anyway this fix should work.

sam-m888:
@prculley What do think about @vantu5z suggestion that the issue be caught in the main date handler?

prculley:
I'm not sure.  The translation files (*.mo) handling is currently quite independent from the selection of the date handler.  I'm not sure mixing the two is a good idea. 

It might make more sense to put up a warning if you are running in a language for which there is no .mo file (caused by not installing the language in the AIO).  Or just changing the AIO back to installing all .mo files by default (increases the installation size about 20MB).

PR #691, opened 2018-10-25
Narrative Web : thumbnails bad alignment.

SNoiraud:
Fixes [#10854](https://gramps-project.org/bugs/view.php?id=10854)

- also solves the following problems in the Nebraska style sheet :
 1 - float parameter has no value "center"
 2 - "[" instead of ";" at end of line 1100

- wrapping of long titles in the thumbnails div for all css

### DIFF
--- a/data/css/Web_Basic-Ash.css
+++ b/data/css/Web_Basic-Ash.css
@@ -450,8 +450,8 @@ div#families table.attrlist td.ColumnType {
 ----------------------------------------------------- */
 #indivgallery .thumbnail {
     float: left;
-    max-width: 130px;
-    max-height: 150px;
+    width: 130px;
+    height: 150px;
     font-size: smaller;
     text-align: center;
     margin: 0.8em 0.5em;
--- a/data/css/Web_Basic-Ash.css
+++ b/data/css/Web_Basic-Ash.css
@@ -109,6 +109,7 @@ p#description {
 a {
     color: black;
     text-decoration: none;
+    word-wrap: break-word;
 }
 a[href]:hover, a[href]:active {
     text-decoration: underline;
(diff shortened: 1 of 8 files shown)


lordemannd:
with pr 691, comparing just basic-ash and nebraska, see the attached. the issue is with space for long titles.
they both have trouble, but basic ash is better. (firefox, win10, minimum width)
side by side

SNoiraud:
Could you try to remove the height directive at line 454 in data/css/Web_Basic-Ash.css. Is it better ?

SNoiraud:
You need to recreate the narrative web after this change. it seems to be better for me.
perhaps we 'll need to do the same modification for all css.
The real problem is when we have a very long title without space. In this case, I don't know how to solves this. Each browser manage this case differently

The problem is : I am not a css specialist and I don't know all subtilities and inheritence of each directive.

lordemannd:
i removed line 454 in data/css/Web_Basic-Ash.css, for this test case it looks great.
I removed similar line at 936 in Web-Nebraska.css, it also looks great.

lordemannd:
I got the new changes, compared basic_ash with nebraska, it looks good to me. I tried it from:
windows with firefox and chrome, 
debian/linux with firefox and chrome
small screen android with chrome, 
ipad with safari, 
all looked good. 

(I will add some of this to #10854, maybe we could get more people to try it out)

lordemannd:
I think the PR is an improvement, maybe more can be done, (re conversation on  #10854)
but as it stands it should b candidate for inclusion.

SNoiraud:
I think effectively this is an improvement. We can leave this PR as it is now for 5.0 and try to make a new PR for 5.1 with some enhancement.

sam-m888:
>  further changes to address #10854 can be worked on in future under a new PR

Sounds good 👍 

Thank you

PR #694, opened 2018-10-26
Add drag Icon to drags from DisplayTabs Gramplet lists

prculley:
Fixes [#10859](https://gramps-project.org/bugs/view.php?id=10859)

Drag from an editor DisplayTab list uses a very long 'row' instead of an icon, like everywhere else.


sam-m888:
Thank you

PR #696, opened 2018-10-27
Simplify the drag/drop code (master)

prculley:
Two commits that try to simplify the drag/drop code a bit.  Not really an enhancement, but more of a code cleanup.  Everything in the original code still works, but seems more complex than necessary.

First commit simplifies the drop code a bit where there were originally two types of source data; the source from the Listview ctrl-c code, and the normal drag sources.

I think this originated from a too quick correction of a bug that occurred during gtk2 to gtk3 conversion.

In any event, I modified the Listview ctrl-c drag source to look like any other drag source, and removed the drop adaptations.

The second commit reverts a bug fix that was originally made back in very early gtk3 timeframe when gtk3 itself had a bug.  That bug was fixed somewhere before gtk3.4.  So the additional complexity is no longer needed.


Nick-Hall:
@prculley Looks good, but needs a rebase.

PR #697, opened 2018-10-29
Fix HandleErrors related to LDS

prculley:
Two commits;
If LDS has no Family handle the ReferencedBySelection proxy failed
Fixes [#10865](https://gramps-project.org/bugs/view.php?id=10865)

Second commit
And while debugging that, I found another HandleError if the LDS did not have a place.
Fix HandleError in Citations Gramplet for lds place missing


sam-m888:
Thank you

PR #698, opened 2018-10-31
FR10850 V1: Refresh button in the gramplet bar.

SNoiraud:
Fixes #10850

This is the first method to avoid reloading gramplet automaticaly.

We need to choose between this PR and the next PR.


prculley:
Not at all certain I like this; it seems like a confusing concept (to an unsophisticated user).  I think users are trained to expect that Gramplets generally update when changing active, having one frozen until a user pressed a little 'Refresh' icon seems likely to confuse users.

I'd be more likely to see what Gramplets have the immediate click to set_active and set those up for right click or some other method to set_active.

PR #699, opened 2018-10-31
FR10850 V2: freeze/unfreeze button per gramplet

SNoiraud:
Fixes #10850

This is the second method to avoid reloading gramplet automaticaly.

We need to choose between this PR and the PR698

### DIFF
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -196,6 +196,7 @@ register('interface.view-categories',
 register('interface.filter', False)
 register('interface.fullscreen', False)
 register('interface.grampletbar-close', False)
+register('interface.grampletbar-freeze', False)
 register('interface.ignore-gexiv2', False)
 register('interface.ignore-pil', False)
 register('interface.ignore-osmgpsmap', False)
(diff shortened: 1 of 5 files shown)


prculley:
Not at all certain I like this; it seems like a confusing concept (to an unsophisticated user).  I think users are trained to expect that Gramplets generally update when changing active, having one frozen if he accidentally pressed a little 'Freeze/Unfreeze' icon seems likely to confuse users.  Particularly if he forgets that he did this in an earlier session.

I'd be more likely to see what Gramplets have the immediate click to set_active and set those up for right click or some other method to set_active.

emyoulation:
Attempting to apply these patches to the 5.1.3-2 (Windows) release makes initiating the Preferences fail.  This must leverage some changes in the 5.2 branch.

dsblank:
@Nick-Hall please choose between:

> I agree with @prculley that this could be potentially confusing to users. 

We rejected another PR in favor of this one. Please close this one if it will not be accepted. See ticket for motivation.

> Also, in Gramps we use the lock icon to indicate privacy which could also be confusing.

OR, give a preference for what kind of icon would be appropriate and we will continue iterating.

Nick-Hall:
I agree with @prculley that this could be potentially confusing to users, so I am closing this PR.

PR #700, opened 2018-11-07
Allow filters/rules via plugin without fail on restart

prculley:
In attempting to add a new filter or rule via an addon, I discovered that the added filter/rule would not work after the first restart after its first use.  In addition I would get a console error:
ERROR: Filter rule 'HasSourceParameter' in filter 'testit' not found!
In this case I had added the new rule 'HasSourceParameter' in the addon, and tried to use it in a custom source filter called 'testit'.

It turned out that the 'custom_filters.xml' file was being processed on startup before the plugins/addons were scanned, so the addon filter was not found.

- First commit deals with this issue.
- Second commit allows a new addon category 'Rule'.
- Third commit allows an addon to provide its own rule GUI elements, for when the already defined rule parameters in filtereditor are not right for the addon.

### DIFF
--- a/gramps/gen/plug/_pluginreg.py
+++ b/gramps/gen/plug/_pluginreg.py
@@ -72,8 +72,9 @@ RELCALC = 9
 GRAMPLET = 10
 SIDEBAR = 11
 DATABASE = 12
-PTYPE = [REPORT , QUICKREPORT, TOOL, IMPORT, EXPORT, DOCGEN, GENERAL,
-               MAPSERVICE, VIEW, RELCALC, GRAMPLET, SIDEBAR, DATABASE]
+RULE = 13
+PTYPE = [REPORT, QUICKREPORT, TOOL, IMPORT, EXPORT, DOCGEN, GENERAL,
+         MAPSERVICE, VIEW, RELCALC, GRAMPLET, SIDEBAR, DATABASE, RULE]
 PTYPE_STR = {
         REPORT: _('Report') ,
         QUICKREPORT: _('Quickreport'),
@@ -88,6 +89,7 @@ PTYPE_STR = {
         GRAMPLET: _('Gramplet'),
         SIDEBAR: _('Sidebar'),
         DATABASE: _('Database'),
+        RULE: _('Rule')
         }
 
 #possible report categories
@@ -211,7 +213,7 @@ class PluginData:
        The python path where the plugin implementation can be found
     .. attribute:: ptype
        The plugin type. One of REPORT , QUICKREPORT, TOOL, IMPORT,
-        EXPORT, DOCGEN, GENERAL, MAPSERVICE, VIEW, GRAMPLET, DATABASE
+        EXPORT, DOCGEN, GENERAL, MAPSERVICE, VIEW, GRAMPLET, DATABASE, RULE
     .. attribute:: authors
        List of authors of the plugin, default=[]
     .. attribute:: authors_email
@@ -362,6 +364,13 @@ class PluginData:
     .. attribute:: reset_system
        Boolean to indicate that the system (sys.modules) should
        be reset.
+
+    Attributes for RULE plugins
+
+    .. attribute:: namespace
+       The class (Person, Event, Media, etc.) the rule applies to.
+    .. attribute:: ruleclass
+       The exact class name of the rule; ex: HasSourceParameter
     """
 
     def __init__(self):
@@ -440,6 +449,9 @@ class PluginData:
         #GENERAL attr
         self._data = []
         self._process = None
+        #RULE attr
+        self._ruleclass = None
+        self._namespace = None
 
     def _set_id(self, id):
        self._id = id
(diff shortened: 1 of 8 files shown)


Nick-Hall:
> I'm calling this a bug because Gramps has long had the theoretical ability to add filters via addons, it was apparently never used before this.

This is not really a bug.  Please move to the _master_ branch.

Nick-Hall:
@prculley I would be interested in an example of your use case here.

prculley:
Some of our users have asked for filters that we don't already have.  I have created a few and privately emailed instructions on how to patch into the Gramps core code, but this is not really sustainable.  I was thinking about creating a 'filter pack' that I would maintain with these as an addon, but when I tried to test a filter rule as an addon, it only worked when first defined, and then failed the next time Gramps restarted, due to the issue noted in the PR.

I'm attaching a very simple example addon that demonstrates the problem.
[HasSourceFilter.zip](https://github.com/gramps-project/gramps/files/2593147/HasSourceFilter.zip)

Nick-Hall:
I really like this idea, but a neater approach would be to actually create a new RULE plugin type.  The GENERAL type is not really intended for this purpose.

prculley:
@Nick-Hall I've added code to put filter rules in their own category and a mechanism to let them define their own GUI parameters in the filter editor.
Attached are a couple of filters that work in the new features and take advantage of them.  Can be used to test.
[plugins.zip](https://github.com/gramps-project/gramps/files/2637334/plugins.zip)

Nick-Hall:
@prculley  The next step is to move the `load_on_reg` code into core Gramps.  Our goal should be to make all rules plugins without needing to modify them.

prculley:
I'm not too clear on how much you want me to do.  I think you are suggesting that I add code to the plugin scanner to recognize 'Rule' type addons and to do the import dictionary adjustments and editor rules lists additions rather than using 'load_on_reg'.  I'd need to add or re-purpose a gpr parameter to specify the object type class for each rule as well. 

Do you also want me to convert all the rules to plugins, moving them to the plugin directory, creating the .gpr files etc.?  A number of the rules use base classes which seem like they would need some sort of library load function if converted to plugins.  Or leave the base classes where they are.

I'm not too clear what any/all of this does to improve the Gramps experience, or even the maintainability.  In fact I could argue that the plugin manager will have an even longer list of 'standard' plugins for users to skip over.

Nick-Hall:
This enhancement should conform to the [ecosystem definition](https://gramps-project.org/wiki/index.php/GEPS_012:_Ecosystem_definition).  We can use views as a template.  I am proposing the following:

* Move the code required to load the rule plugins into core.  This will avoid the need for boilerplate code in every rule addon.
* Add new plugin attributes as required.  You will probably need something like _ruleclass_ and _primaryobj_.
* Keep the base classes where they are as we do for views.
* Convert the existing rules into plugins.  This should simply involve moving them into the plugins directory and creating gpr files.  I suggest keeping the sub-directories and creating one gpr file with multiple entries for each object type.

The idea is that addons should load in exactly the same way as the core functionality.  This was the solution we chose for the other plugin types.  I expect to see the rules in the plugin manager.

Nick-Hall:
Our [programming guidelines](https://gramps-project.org/wiki/index.php/Programming_guidelines#Imports) require that imports within a module should be relative.  I'm not sure if this is regarded as best practice, but it is our current policy.  When moving the rules we will have to use absolute imports as you point out, but this is acceptable.

I just use `grep` to find out where modules are imported.  Apart from in the rules themselves and one unit test, I can't find any other imports that need changing.

It is certainly worth the effort to come up with a good design that is consistent with our ecosystem definition.

prculley:
Its not clear to me where in the ecosystem doc there is any hint that rules should not be part of core Gramps, but Ill leave that argument to the developers list.  I think the movement of all filter rules should be discussed by all developers, perhaps as part of the roadmap.

I'll be happy to get the addon scan code to do the rule boilerplate as you suggested.  But if this PR is going to be held hostage to the overall plugin movement, I think I will decline.

I estimate I would have to edit 280+ files and then test a lot of things to make sure nothing was broken in our tools, reports, exports etc. when that gets done.

prculley:
@Nick-Hall I've updated the PR to have the addon code properly register the rules based on the gpr files.
Attached are a couple of filters that work in the new features and take advantage of them.  Can be used to test.
[plugins.zip](https://github.com/gramps-project/gramps/files/2668325/plugins.zip)

Nick-Hall:
@prculley Thanks.  I'll find time to review it shortly.

gramps/gen/plug/_pluginreg.py
```diff
@@ -362,6 +364,14 @@ class PluginData:
     .. attribute:: reset_system
        Boolean to indicate that the system (sys.modules) should
        be reset.
+
+    Attributes for RULE plugins
+
+    .. attribute:: ruleclass
+       The class (Person, Event, Media, etc.) the rule applies to.
+
+       Note that the rule plugin id MUST be the exact class name of the rule;
```
Nick-Hall: Use the `ruleclass` attribute to hold the class of the rule.  Don't use the `id` for this.

Add a `namespace` attribute to hold the class that the rule applies to.  I see that the term namespace is used for this elsewhere in the rules.

prculley: Changes done as you suggest...

prculley:
Attached are a couple of filters that work in the new features and take advantage of them.  Can be used to test.
[plugins.zip](https://github.com/gramps-project/gramps/files/2691160/plugins.zip)

Nick-Hall:
The unit tests also need fixing.

Nick-Hall:
@prculley I am inclined to drop my last commit moving existing rules, and then merge this PR. Perhaps we should just get an opinion from @bmatherly first.

PR #701, opened 2018-11-07
Filter Rule editor, save pane position

prculley:
This fix saves the Edit/Filter edit rule dialog 'pane' position between the rule tree and the remainder of the dialog.

A long standing annoyance of mine is that the default position doesn't let me see much of the rule tree lines without scrolling or repositioning of the pane edge.  But the pane position doesn't get saved, so I have to do it every time I use the editor.

While it might be considered an enhancement, it doesn't have any compatibility issues or new translatable strings that would require delaying to 5.1.


Nick-Hall:
We should refrain from adding new functionality in a maintenance branch.

prculley:
Sigh; change base to master branch, now I just have to wait for another year or so for 5.1 to use it...

PR #702, opened 2018-11-07
Gedcom export, upgrade OBJE handling to Gedcom 5.5.1 style

prculley:
Fixes #10797

Gramps used to use the older Gedcom 5.5 style of OBJE (media) export record, which could leave some of the database data (notably media source/citation data) un-exported.  The 5.5.1 Gedcom spec allows media objects to use the XREF style of exporting, and the cross referenced OBJE record does allow for a SOUR record.

As usual, when I ran the export tests, I found a bug in the export test routine (that only turns up on a test failure).  That is also fixed here.

### DIFF
--- a/data/tests/exp_sample_ged.ged
+++ b/data/tests/exp_sample_ged.ged
@@ -1,12 +1,12 @@
 0 HEAD
 1 SOUR Gramps
-2 VERS 5.0.0-alpha2
+2 VERS 5.0.1
 2 NAME Gramps
-1 DATE 1 SEP 2017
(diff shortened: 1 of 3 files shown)


sam-m888:
Testing shows the resulting GEDCOM could be imported into Gramps 4.2.x & 5.0.0.

Checking with a few other programs before accepting.

sam-m888:
@prculley If you have a moment can you have a look at https://www.gramps-project.org/wiki/index.php/Gramps_and_GEDCOM and suggest updates based on what you know please?

PR #704, opened 2018-11-10
New CLI commands; Safe mode and Defaults

prculley:
Also a second commit that improves safety in loading an older or corrupted config file.

'**Safe mode**'  ("gramps -S" or "gramps --safe").  This CLI command starts Gramps as if it had never been installed before.  In this mode, any previous family trees can still be loaded, as long as they were stored in the default folder.  All other settings, filters, books, addons etc. are either cleared or returned to their default values.  Other CLI commands can be used, or, if none, Gramps will start the GUI.  Nothing except the actual family tree data is saved.

This actually works by setting the folder that Gramps uses to store its user data (except for family trees) to a temporary directory, which is deleted when Gramps closes.

'**Defaults**' ("gramps -D E" or "gramps --default=E").  This CLI command causes Gramps to clear out or return to defaults the desired settings.  The family tree databases are NOT cleared out or removed.  The sub-commands (replace the 'E' from the example command line above) are:

- 'A' Addons are cleared.  Any installed addons are removed, along with their settings.
- 'F' Filters are cleared.  Any custom filters are removed.
- 'P' Preferences are returned to their default values.
- 'X' Books are cleared, Reports and Tools settings are returned to their default values.
- 'Z' Old '.zip' files from family tree version upgrades are deleted.
- 'E' Everything except the actual family tree data is returned to default settings.  This does all of the above as well as a few more items; deletes thumbnails, maps and the user CSS (used in web reports).

If this PR is accepted, I will create and reference a wiki entry containing the previous information.

Note: I had to shift the code for 'import config'  and its usage a bit to prevent it from getting done before the new commands had a chance to take effect.

### DIFF
--- a/gramps/cli/argparser.py
+++ b/gramps/cli/argparser.py
@@ -39,15 +39,16 @@ import sys
 import os
 import getopt
 import logging
+import shutil
+from glob import glob
 
 #-------------------------------------------------------------------------
 #
 # gramps modules
 #
 #-------------------------------------------------------------------------
-from gramps.gen.const import LONGOPTS, SHORTOPTS, PLUGINS_DIR, USER_PLUGINS
-from gramps.gen.plug import BasePluginManager
-from gramps.gen.config import config
+from gramps.gen.const import (LONGOPTS, SHORTOPTS, USER_PLUGINS, VERSION_DIR,
+                              HOME_DIR, TEMP_DIR, THUMB_DIR, ENV_DIR, USER_CSS)
 from gramps.gen.utils.cast import get_type_converter
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
@@ -81,6 +82,14 @@ Application options
   -y, --yes                              Don't ask to confirm dangerous actions (non-GUI mode only)
   -q, --quiet                            Suppress progress indication output (non-GUI mode only)
   -v, --version                          Show versions
+  -S, --safe                             Start Gramps in 'Safe mode'
+                                          (temporarily use default settings)
+  -D, --default=[APXFE]                  Reset settings to default;
+                 A - addons are cleared
+                 P - Preferences to default
+                 X - Books are cleared, reports and tool settings to default
+                 F - filters are cleared
+                 E - Everything is set to default or cleared
 """)
 
 _USAGE = _("""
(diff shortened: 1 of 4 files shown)


fxtmtrsine:
Normal people will be reluctant to use the command line, how about a menu entry like used in libreoffice or firefox?

prculley:
@fxtmtrsine I considered using menu entries.  But for most people that might need these commands, Gramps is already pretty broken, and in many cases won't start.

The 'Defaults' options may make sense in a menu that appears when running in safe mode...  After all, if it runs in safe mode but not otherwise, then you need to default something or other...

I've already submitted another PR for a Check & Repair after a crash which may help some folks.  I'll keep watching to see if we need other recovery options. Gramps has survived for a very long time without any of these, but of course the goal is to make life easier for non-techno users.

johnbickmore:
@prculley I believe this may pr may have inadvertently caused my issue the other day. Checked and I don't get the error below with Gramps 4.2.8 when attempting to run gramps from command line to list the version details, but not:

 **gramps -v**  
or
 **gramps --version**

Note that  **gramps -h** does work 

Yes I deliberately don't have the gui packages (gtk etc) installed ( I only want package_core & not package_gui as mentioned setup.py )



> 
> john@DESKTOP-JHFWIH38:~/gramps$ ./Gramps.py -v
> 2019-09-23 19:06:38.416: ERROR: grampsapp.py: line 157: Unhandled exception
>   File "./Gramps.py", line 29, in <module>
>     app.main()
>   File "/home/john/gramps/gramps/grampsapp.py", line 503, in main
>     errors = run()
>   File "/home/john/gramps/gramps/grampsapp.py", line 446, in run
>     show_settings()
>   File "/home/john/gramps/gramps/grampsapp.py", line 208, in show_settings
>     gi.require_version('Gtk', '3.0')
>   File "/usr/lib/python3/dist-packages/gi/__init__.py", line 130, in require_version
>     raise ValueError('Namespace %s not available' % namespace)
> ValueError: Namespace Gtk not available

prculley:
John; you guessed wrong 😕.
The issue was created when the Gtk folks started complaining when
"gi.require_version('Gtk', '3.0')" was not present.  So for Gramps 5.0.0 we
added that line.
It now appears that we don't have the correct 'exception' handler, before
we just got an "ImportError", which we handled, but now we are getting a
"ValueError" for the earlier gi.require_version('Gtk', '3.0') line, which
we don't.

I will submit a fix.

Paul C.

johnbickmore:
Those lovely Gtk folk again, thanks for the [fix](https://github.com/gramps-project/gramps/pull/923) 👍

PR #705, opened 2018-11-11
Fix Find Duplicate People tool

prculley:
Throws an exception when deleting someone shown in the tool's list outside of the tool while the tool is displaying the list of potential merges.

Fixes [#10875](https://gramps-project.org/bugs/view.php?id=10875)

P.S. I changed the dellist dict to a set, it appears that nothing was using the data stored in the dict, only the keys were being used.  And added a signal connect to watch for person deletes.


sam-m888:
Thank you

sam-m888:
Should this be backported to Gramps50?

prculley:
@sam-m888 Yes it should be backported.  I submitted a new PR #728 for that.

PR #707, opened 2018-11-12
Fix crash when a filter with loop in definition is defined

prculley:
Fixes [#10621](https://gramps-project.org/bugs/view.php?id=10621)

A user created a custom filter that had a set of rules that eventually referenced each other in a loop.  When trying to 'Find' with that filter, Gramps crashes hard with a
RecursionError: maximum recursion depth exceeded while calling a Python object.

I added a test to watch for too many calls to the filter; this catches the issue.  I also added a new error message.

### DIFF
--- a/gramps/gen/filters/rules/_rule.py
+++ b/gramps/gen/filters/rules/_rule.py
@@ -28,9 +28,11 @@ Base class for filter rules.
 # Standard Python modules
 #
 #-------------------------------------------------------------------------
+import re
+
+from ...errors import FilterError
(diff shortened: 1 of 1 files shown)


Nick-Hall:
@prculley This can go into the _gramps50_ branch.  Sometimes introducing new translatable strings is unavoidable.

prculley:
I need to figure out a better way to switch base branches; but I've got it switched to gramps50 now.

sam-m888:
Thank you

PR #708, opened 2018-11-13
Fix AgeOnDate and some reports using SimpleAccess for missing surname

prculley:
Fixes [#9958](https://gramps-project.org/bugs/view.php?id=9958)

Another issue with completely missing surname  similar to https://github.com/gramps-project/gramps/pull/613.

This was found with the Age on date Gramplet, but would effect any report using SimpleAccess.


sam-m888:
Thank you

PR #710, opened 2018-11-13
Update gramps bugtracker URL in all po files

prculley:
Fixes [#10543](https://gramps-project.org/bugs/view.php?id=10543)

Some of the strings in the po file still point to http://bugs.gramps-project.org/ (which is broken) instead of https://gramps-project.org/bugs/


sam-m888:
Needs a rebase please.

prculley:
@sam-m888 rebased as requested

sam-m888:
Thank you

PR #712, opened 2018-11-15
Fix dbapi set_name_group_mapping to properly close transaction

prculley:
Fixes [#10730](https://gramps-project.org/bugs/view.php?id=10730)

It seems that you have to properly end a transaction (at least) when updating or inserting data into your db.  If you don't, then attempting another transaction gives an error (the bug report [#10730](https://gramps-project.org/bugs/view.php?id=10730) ).

### DIFF
--- a/gramps/plugins/db/dbapi/dbapi.py
+++ b/gramps/plugins/db/dbapi/dbapi.py
@@ -572,6 +572,7 @@ class DBAPI(DbGeneric):
         """
         Set the default grouping name for a surname.
         """
+        self._txn_begin()
         self.dbapi.execute("SELECT 1 FROM name_group WHERE name = ?",
                            [name])
         row = self.dbapi.fetchone()
@@ -582,6 +583,7 @@ class DBAPI(DbGeneric):
             self.dbapi.execute(
                 "INSERT INTO name_group (name, grouping) VALUES (?, ?)",
                 [name, grouping])
+        self._txn_commit()
 
     def _commit_base(self, obj, obj_key, trans, change_time):
         """


gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -575,13 +575,15 @@ def set_name_group_mapping(self, name, grouping):
         self.dbapi.execute("SELECT 1 FROM name_group WHERE name = ?",
                            [name])
         row = self.dbapi.fetchone()
+        self._txn_begin()
```
Nick-Hall: Perhaps move this line to the start of the method.

Nick-Hall:
Do we know why the [unit test](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/db/dbapi/test/db_test.py#L777) was not failing?

prculley:
Unit test did not fail because it runs each subtest independently, so there were no other transactions after this to cause a failure.
I don't fully understand these dbs (yet), but is it necessary to enclose a db 'select' within a transaction?  I though these we basically reads, so that was not needed.

Nick-Hall:
This is a bug in certain versions of Sqlite.  The [documentation](https://sqlite.org/lockingv3.html) describes how autocommit mode can be mixed with BEGIN and COMMIT statements. I have fixed similar bugs before, but obviously missed this one.

Putting the SELECT in the transaction will cause it to take out a lock, which is normally sensible for concurrent transactions - useful if we decide we want a multi-user version in the future.  Unfortunately, in this case, a FOR UPDATE clause would be needed which is not available in Sqlite.  Also two concurrent inserts would cause a primary key constraint violation.

I don't see any harm in putting the SELECT in the transaction, and I have done so in at least one similar place.

gramps/plugins/db/dbapi/dbapi.py
prculley: Done...

sam-m888:
Works for me 👍 

@Nick-Hall I'll leave this one for you to commit.

PR #713, opened 2018-11-15
Narrative web : use latest version from openlayers.

SNoiraud:
Fixes [#10884](https://gramps-project.org/bugs/view.php?id=10884)


sam-m888:
Thank you

PR #714, opened 2018-11-15
Fix IsDuplicatedAncestorOf filter rule to avoid crash on tree loop

prculley:
Fixes [#10685](https://gramps-project.org/bugs/view.php?id=10685)

The Rule (and probably others) crashes if the user has a 'loop' in his family tree.  It was easy to prevent the crash, and slightly speed up the filter in cases where the tree has inbreeding.


sam-m888:
Thank you

PR #717, opened 2018-11-16
Fix Person Sidebarfilter when using 'Event' and Reg expressions

prculley:
Fixes [#10659](https://gramps-project.org/bugs/view.php?id=10659)

When using the 'Event' type in the Person Sidebar Filter, and checking 'Use regular expressions' the person HasEvent rule includes the following parameters:
'Personal event:', 'Date:', 'Place:', 'Description:', 'Main Participants:', 'Primary Role:'
All of these except for 'Primary Role' are strings, and can be used as the source string for re.compile.
The latter causes the exception.
When using the same rule created by the Filter Editor, the last parameter is a string '1' or '0'.  Either works correctly, since the 'apply' converts the parameter to an int for testing, so I changed the Person Sidebar Filter to also start this rule with a string parameter.

### DIFF
--- a/gramps/gui/filters/sidebar/_personsidebarfilter.py
+++ b/gramps/gui/filters/sidebar/_personsidebarfilter.py
@@ -206,7 +206,7 @@ class PersonSidebarFilter(SidebarFilter):
 
             # Build an event filter if needed
             if etype:
-                rule = HasEvent([etype, '', '', '', '', True], use_regex=regex)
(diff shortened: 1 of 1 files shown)


sam-m888:
Thank you

sam-m888:
Should this be backported to Gramps50?

prculley:
@sam-m888 Yes to backport; another error on may part, new PR #729 for gramps50.

PR #718, opened 2018-11-17
Fix Dashboard for multiple copies of a Gramplet

prculley:
Fixes [#10650](https://gramps-project.org/bugs/view.php?id=10650)

If two or more copies of a Gramplet are installed in the Dashboard (don't know why this was supported, but it was, in part), the current code produces a warning similar to the following:
 WARNING: grampletpane.py: line 1163: Failed to load gramplets from /home/dave/.gramps/gramps50/Gramplets_dashboardview_gramplets.ini because While reading from '/home/dave/.gramps/gramps50/Gramplets_dashboardview_gramplets.ini' [line 65]: section 'To Do' already exists 

And then the dashboard is blank.  This does not fix itself on restart.

This PR corrects the situation by properly setting the .ini file section heading with a unique name for each copy of a gramplet.  It also sets the .ini parser to non-strict mode, so if a user already had this issue, the dashboard will at least load (although with only a single copy of the Gramplet).

### DIFF
--- a/gramps/gui/widgets/grampletpane.py
+++ b/gramps/gui/widgets/grampletpane.py
@@ -1152,7 +1152,7 @@ class GrampletPane(Gtk.ScrolledWindow):
         retval = []
         filename = self.configfile
         if filename and os.path.exists(filename):
-            cp = configparser.ConfigParser()
+            cp = configparser.ConfigParser(strict=False)
             try:
                 cp.read(filename, encoding='utf-8')
             except Exception as err:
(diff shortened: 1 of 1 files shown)


Nick-Hall:
> If two or more copies of a Gramplet are installed in the Dashboard (don't know why this was supported, but it was, in part)

It was useful in a few limited cases:  the old ToDo gramplet and the QuickView gramplet are two examples that come to mind.  The title of each gramplet could be changed.

When I created the gramplet bars, I decided not to support this.  I thought that allowing users to change gramplet titles could be confusing, especially when they reported bugs.

Having said that, I didn't know that the title functionality was broken in the Dashboard. As far as I am concerned it should still work.

sam-m888:
Thank you

PR #719, opened 2018-11-17
Fix ToDo Gramplet for multiple attempts to edit a note

prculley:
Fixes [#10645](https://gramps-project.org/bugs/view.php?id=10645)

ToDo Gramplet used wrong exception type for starting the EditNote.


sam-m888:
Thank you

PR #720, opened 2018-11-19
Fix startup messages with bad filename

prculley:
Fixes [#10885](https://gramps-project.org/bugs/view.php?id=10885)

User added a bad file name to command line for GUI start and got a pretty poor crash message.

I reused an already present error message (no new translation) to support a better startup error.

Pretty minor, but tries to make it a bit more friendly.


sam-m888:
Thank you

PR #721, opened 2018-11-19
Updated swedish translation

pehlm:
I'm having an updated swedish translation for Gramps, that I hoping you have some interest in. It's reworked but mostly based on the earlier translator, which so sadly died last year. Over 500 untranslated strings is now translated though, some touches, spellings and grammar errors is also fixed. I've done it with the Gramps 5.0 pot-file, and I've tested it with that version, I don't see anything wrong. Hope it is useful.

### DIFF
--- a/po/sv.po
+++ b/po/sv.po
@@ -20,20 +20,21 @@
 # Jens Arvidsson <jya@sverige.nu>, 2002-2005.
 # Stefan Björk <betula@users.sourceforge.net>, 2005-2006.
 # Peter Landgren <peter.talken@telia.com>, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017.
+# Pär Ekholm <par.ekholm@pekholm.se>, 2018.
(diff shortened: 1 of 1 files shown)


pehlm:
Yes I am, the translation for 5.0 is 100% and I do not think there are many errors either now, if any at all. I would actually be very pleased to be the new translator. Thanks!

Nick-Hall:
Thanks for volunteering.  Welcome to the [Team](https://gramps-project.org/wiki/index.php/Team)!

pehlm:
Many thanks for that. What is the preferred way to commit my changes? Should I create pull requests each time?

Nick-Hall:
@pehlm I suggest you start by submitting pull requests.  There were a couple of unwanted merges in this PR that I had to remove.  Make sure you use the `git pull --rebase`, and check for merges using `gitk` or `git log`.  However, translators are allowed to push translation updates directly to the repository.

pehlm:
Ok, thanks for that!

PR #722, opened 2018-11-21
Narrative Web: thumbnails problems in some cases.

SNoiraud:
Fixes #10890


lordemannd:
I still see problem after this change

PR #723, opened 2018-11-24
Fix Citation List view Source Last Changed Column to sort properly

prculley:
Fixes [#10893](https://gramps-project.org/bugs/view.php?id=10893)

Was using the displayer for sort (so alpha on whatever user prefs were set).  Changed to strict date/time sort like all other 'change' columns.


sam-m888:
Thank you

PR #725, opened 2018-11-25
Fix CSV import to set marriage event role to family

prculley:
Fixes [#9952](https://gramps-project.org/bugs/view.php?id=9952)


sam-m888:
@Nick-Hall @prculley Something strange going on with Travis-ci as even when I restarted the travis run it insist that the pr is against master and not gramps50 and [fails](https://travis-ci.org/gramps-project/gramps/builds/459422134)!

PR works for me by the way 👍

prculley:
@sam-m888 The travis failure was my fault, I neglected to check the test after the change, it correctly caught the change in functionality, so I had to adjust the test file to match the new functionality.  Should be good now.

sam-m888:
Thank you 👍

PR #726, opened 2018-11-25
Fix 'Generate Book' dialog for crash on 'x' close

prculley:
Fixes [#10901](https://gramps-project.org/bugs/view.php?id=10901)

Close got called twice when using the 'x' from titlebar...


sam-m888:
Thank you

PR #727, opened 2018-11-25
Fix dialogs for 'Help' button closes the dialog, and non-functional 'Help'

prculley:
Four commits deal with fixing the 'Help' button closes dialog as described in [Gramps issue#10589](https://gramps-project.org/bugs/view.php?id=10589)

In each case the dialog response was terminating the dialog unnecessarily.  For several of the cases, I could just eliminate the response via the glade file (although I had to define the 'Help' button as secondary to keep it to the left side of the dialog box).

For the other cases, (with no glade file), I re-ran the dialog when the Gtk ResponseCode was 'Help'.

While testing these, I noticed that the Selectors help buttons did not actually do anything, no web site popup.  So the remaining commit corrects that situation.

### DIFF
--- a/gramps/gui/glade/baseselector.glade
+++ b/gramps/gui/glade/baseselector.glade
@@ -65,6 +65,7 @@
                 <property name="expand">False</property>
                 <property name="fill">False</property>
                 <property name="position">2</property>
+                <property name="secondary">True</property>
               </packing>
             </child>
           </object>
@@ -161,7 +162,6 @@
     <action-widgets>
       <action-widget response="-6">cancelbutton1</action-widget>
       <action-widget response="-5">okbutton1</action-widget>
-      <action-widget response="-11">help</action-widget>
     </action-widgets>
   </object>
 </interface>
--- a/gramps/gui/glade/dbman.glade
+++ b/gramps/gui/glade/dbman.glade
@@ -159,6 +159,7 @@
                 <property name="expand">False</property>
                 <property name="fill">False</property>
                 <property name="position">2</property>
+                <property name="secondary">True</property>
               </packing>
             </child>
           </object>
@@ -375,7 +376,6 @@
     <action-widgets>
       <action-widget response="-7">cancel_btn</action-widget>
       <action-widget response="-5">connect_btn</action-widget>
-      <action-widget response="-11">help_btn</action-widget>
     </action-widgets>
   </object>
 </interface>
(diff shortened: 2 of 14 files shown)


sam-m888:
Works for me on each of the mentioned dialogs, thank you 👍 

Only issue with me is I've forgotten where to find the "Select a different person" selector, any hints please?

[x] "Family tree"(manager) dialog

This also occurs on the following dialogs:

The following "selectors"
[x]"Select Father"
[x]"Select Mother"
[x]"Select Child"
[x]"Select Events"
[x]"Select Note"
[x]"Select Media Object"
[ ] "Select a different person"
[x]"Select Source"
[x]"Select Source or Citation"
[x]"Select Repository" ([ ] goes to wrong url should be #Select_Repository_selector )

[x] Also "Books" > "Generate Book" dialog ([] Goes to wrong url 
Gramps_5.0_Wiki_Manual_-_Reports#Generate_Book
instead of
Gramps_5.0_Wiki_Manual_-_Reports_-_part_3#Generate_Book_dialog )

[x] "Generate Testcases for Persons and Families" Tool dialog

prculley:
Oops, I did not test that one either and of course it was broken; I fixed it up so you will have to reload the branch.  The "Select a different person" can be found from a person editor, association tab, add new person dialog.  Thanks for testing. 

I noticed that you have another bug with a LOT of help issues that I will get to as I work backward through the bug tracker; If you would prefer I leave these for you (you created that bug) let me know.

sam-m888:
Thanks for the hint, I don't use associations very often. confirming it works and and I had record the tool tip as the name and not the dialogs name!

[x]"Select a different person"  >> "Select Person" ([ ] goes to wrong url
Gramps_5.0_Wiki_Manual_-_Entering_and_editing_data:_detailed_-_part_1#Person_Reference_Editor
instead of
Gramps_5.0_Wiki_Manual_-_Entering_and_editing_data:_detailed_-_part_1#Select_Person_selector

I've updated and added a section to the the user manual for this selector.

> I noticed that you have another bug with a LOT of help issues that I will get to as I work backward through the bug tracker; If you would prefer I leave these for you (you created that bug) let me know.

You mean issue [#9042](https://gramps-project.org/bugs/view.php?id=9042)  if so yes feel free as always 👍  I'll help as I have time.

sam-m888:
Thank you

PR #728, opened 2018-11-28
Fix Find Duplicate People; exception when deleting someone shown

prculley:
outside of the tool
Fixes [#10875](https://gramps-project.org/bugs/view.php?id=10875)

Original PR #705 was incorrectly based on master; this backports it.

P.S. when master is merged up to date with gramps50, it's easy to forget to reset the base at the top of this screen, since they both work fine for automatic merges...


sam-m888:
Thank you for the backport 👍

PR #729, opened 2018-11-28
Fix Person Sidebarfilter when using 'Event' and Reg expressions

prculley:
Fixes [#10659](https://gramps-project.org/bugs/view.php?id=10659)
Backported from master #717


sam-m888:
Thank you

PR #731, opened 2018-12-01
Fix Rebuild Secondary Indexes tool for dbapi backends

Nick-Hall:
For dbapi backends, this tool will update the secondary columns
that are used for indexing.

Resolves [#10757](https://gramps-project.org/bugs/view.php?id=10757).


Nick-Hall:
@prculley Quite right.  The BSDDB version provides progress so the DBAPI version should also do so.  I have updated the code.

prculley:
Looks good ;-)

PR #733, opened 2018-12-05
Relationship Graph: Father and Mother are connected by an invisible line.

Keepun:
Connecting with an invisible line allows you to generate a more compact and beautiful graph.

### DIFF
--- a/gramps/plugins/graph/gvrelgraph.py
+++ b/gramps/plugins/graph/gvrelgraph.py
@@ -494,6 +494,31 @@ class RelGraphReport(Report):
             self.doc.start_subgraph(fam_id)
             f_handle = fam.get_father_handle()
             m_handle = fam.get_mother_handle()
+            if f_handle and m_handle:
+                father = self._db.get_person_from_handle(f_handle)
+                mother = self._db.get_person_from_handle(m_handle)
+                fcount = 0
+                mcount = 0
+                family_list = father.get_parent_family_handle_list()
+                for fam_handle in family_list:
+                    family = self._db.get_family_from_handle(fam_handle)
+                    if family is not None:
+                        fcount = fcount + 1
+                family_list = mother.get_parent_family_handle_list()
+                for fam_handle in family_list:
+                    family = self._db.get_family_from_handle(fam_handle)
+                    if family is not None:
+                        mcount = mcount + 1
+                first = father
+                second = mother
+                if fcount < mcount:
+                    first = mother
+                    second = father
+                self.doc.add_link(first.get_gramps_id(),
+                                  second.get_gramps_id(),
+                                  "invis",
+                                  "none",
+                                  "none")
             if f_handle in self.persons:
                 father = self._db.get_person_from_handle(f_handle)
                 self.doc.add_link(father.get_gramps_id(),


prculley:
I have not extensively tested this option, however, the 'Use subgraphs' option is supposed to do the same thing.  It also tells Graphvis to group the father and mother together if possible.  So in your testing does this option work better than the 'Use subgraphs'?

Nick-Hall:
I don't use this report myself, so it is probably better if I post to the list to get opinions.

Keepun:
This change complements the 'Use subgraphs' effect.

Many advanced options need to be added to generate a graph, but inexperienced users will get confused.

Any change in the code changes the graph. I made a copy of the "Relationship Graph" for myself, because the changes made since version 4.5 have greatly changed the location of people on the graph.

This change makes the graph more compact with 90 people.

Nick-Hall:
@Keepun There are some conflicts in this PR, otherwise we can merge.

edbennett:
The graph is certainly more compact and beautiful, but it's also now almost impossible to read—different generations appear at the same level, husband and wife appear at different levels… While the previous behaviour made for very wide graphs, the logic was at least clear and easy to follow (the same generation appears at the same level).

Keepun:
> The graph is certainly more compact and beautiful, but it's also now almost impossible to read—different generations appear at the same level, husband and wife appear at different levels… While the previous behaviour made for very wide graphs, the logic was at least clear and easy to follow (the same generation appears at the same level).

Checkbox "Use subgraphs" take off.

edbennett:
> Checkbox "Use subgraphs" take off.

"Use subgraphs" controls an orthogonal functionality. I rely on using subgraphs to group parents together—otherwise they appear very distant on the graph, generating a lot of crossings. I want subgraphs, but not the distortion to the graph that this produces.

I don't think that what I need from the relationship graph is that uncommon (beyond perhaps the size of the graph I'm producing):
- Where families divide discretely into generations, each generation should be aligned.
- Marriages appear between generations
- Parents appear side-by-side, within these constraints
- Children should be as close to beneath their parents as possible within these constraints
- There should be as few line crossings as possible within these constraints

Gramps 5.0.1 as released mostly achieves this (with the caveat that this may be because I provided some hinting in the ordering by transcribing handwritten trees to input the data). The sorting in the current 5.1 branch breaks this somewhat, as it introduces significant amounts of line crossing and horizontal shift between generations, making the graph difficult to read. This change in master completely destroys the first constraint. I'm not sure I see a use case where the new behaviour is more preferable—the aim of a report is to generate an informative and useful representation for reference, while the more compact and beautiful representation does not meet that goal—it is a visually appealing demonstration, but once it comes to use the report in any capacity it is no longer useful.

prculley:
I recently had a chance to try out the new functionality, and I have to say I don't care for it.  I think @edbennett  comments are valid.  In particular I find that the "stacking" of parents on top of each other to be annoying.

I made some tests to see if we could avoid the stacking but still keep parents close together via the invisible line; this is possible and generally seems to do a bit better job than just subgraphs by itself.

I'd like to hear from @Keepun on his original goal for this; if the stacking effect was the primary desire, then I think we should make it another option.  Otherwise I'm inclined to put the parents side by side again, although with the invisible line to hold them together a bit better.  This can be done by using the add_samerank method with the two parents after adding the invisible link.

Keepun:
Used Graphviz to generate a tree. There are no normal settings for creating a graph, because everything translates into Graphviz syntax. The generator interface is terrible.

I used the existing checkbox "Use subgraphs", but you can add an additional checkbox to the GUI for such a tree construction. Because "Use subgraphs" had almost no effect on my tree and parents were in different places.

Any change in this tree generator can coordinately change the appearance of the graph, because everyone solves only their task. I recommend everyone to keep a copy of this module when updating Gramps.

hewettp:
Where can I see the justification for this change? What information is there demonstrating that this change is an improvement? As far as I can see, this change makes the graphs more complicated, confusing and difficult to follow. The previous arrangement (where spouses are shown side by side) makes for much cleaner graphs IMO. Three example comparisons are attached.
file:///home/peter/projects/gramps/733_Subgraph/Example1_subgraph.png
file:///home/peter/projects/gramps/733_Subgraph/Example2_subgraph.png
file:///home/peter/projects/gramps/733_Subgraph/Example3_subgraph.png

hewettp:
Oops, that attempt to include images in the post did not work. Try this one.
https://imgur.com/CYmC3cp
https://imgur.com/oStuEP3
https://imgur.com/nXaPMDF

Nick-Hall:
@hewettp The justification presented by the author was that this PR made the graph more compact.  The PR was left open for 5 months before merging and did not receive any adverse comments.  Anyone could have stopped the merge by giving a "thumbs down" preferably accompanied by an argument against it.

I would suggest discussing this on the gramps-users list to get some feedback.  Adding the change as an option as I suggested on 14th Jan may be a suitable compromise.

prculley:
I've been looking at this again, and I tried a fix https://github.com/gramps-project/gramps/pull/981, which is not working very well. It doesn't do multiple families for a parent on the same rank.

Additionally I found a bug with this PR that causes incomplete nodes to be drawn for individuals that are filtered out, but are part of a family that is filtered in.  When the bug shows up you get a node disconnected from anything with just the Gramps ID in it.

I'm beginning to think we should just revert it.

Nick-Hall:
What does @Keepun think about this?

prculley:
Just for fun I also tested the newrank="true" Graphviz option, as it is supposed to use a different algorithm for dealing with subgraphs and other constraints.  I tested without this PR code (original layout) to see if it did a better job keeping parents together.  It did a bit better in some areas, and worse in others.  I'm not sure it makes enough difference to be worth another option.

Nick-Hall:
We should see if we get a reply from the author of this PR and then make a decision.  Adding too many options will just confuse users, so we should only do this if it gives a real benefit.  Unfortunately, @hewettp's post on the list didn't get any replies.

hewettp:
@Keepun Thank you for contributing to Gramps. I am struggling to use this change in the relationship graph. I am familiar with relationship graphs in Gramps 5.0 and earlier versions. My needs are similar to those described above by @edbennett . On large charts, I use a script to reorder persons to reduce the number of line crossings.

It would help to have some information on how to use the Gramps 5.1 relationship graph to produce useful charts. Can you please provide some info on this, with examples to illustrate the improvement that this latest change brings?
I have put some examples of the difficulty that I am having with this latest version on the bug tracker site 11494 https://gramps-project.org/bugs/view.php?id=11494 . These examples compare the same tree done with Gramps 5.0 and 5.1.

Keepun:
In GraphViz, the result can also change from version to version with the same script.
I also suggested adding a separate parameter:
https://github.com/gramps-project/gramps/pull/733#issuecomment-517501340

Any change in the parameters of the canvas size, orientation, DPI, font size can significantly change the appearance of the graph.

In general, you need to collect sample graphs to verify the code. UnitTest is needed to develop this module.

Without sub-graphs:
[nograph.gv](https://github.com/gramps-project/gramps/files/4008590/nograph.txt)
nograph

With sub-graphs without a patch:
[nopatch.gv](https://github.com/gramps-project/gramps/files/4008591/nopatch.txt)
nopatch

With patch:
[normal.gv](https://github.com/gramps-project/gramps/files/4008592/normal.txt)
normal

hewettp:
In my experience, the layout that Graphviz produces can be very sensitive to the order of the persons in the .gv file. The order of the families doesn't seem to matter. With simple family structures (such as the example presented by @Keepun), it should be possible to have a graph with no crossings. I don't know how to get Gramps to do this, so I get Gramps to output a gv file, then reorder persons to produce a better layout.
For example, the nopatch example above can be redone with zero crossings (by reordering the persons in the dot file). 
https://imgur.com/ya81By3

I use the relationship graph to show family relationships, for up to several hundred persons on one page. The graph is very useful to show relationships that are not simple, such as when spouses have a common ancestor, or when two brothers marry two sisters. Graphviz seems to struggle to achieve a clear layout, and reordering persons can make a big difference.
However, subgraphs are needed to keep spouses together. Gramps 5.0 did a good job of this, but Gramps 5.1 doesn't work as well, particularly where spouses are not shown together.

PR #734, opened 2018-12-06
Allow addon Reports to specify a help button URL for options dialog

prculley:
Fixes [#10910](https://gramps-project.org/bugs/view.php?id=10910)

At the moment, using the 'Help' button on an addon report causes the standard 'Reports' page to pop up.
For example the Help for DescendLines pops up https://gramps-project.org/wiki/index.php?title=Gramps_5.0_Wiki_Manual_-_Reports#Descendants_Lines
Which has nothing to do with the addon.

This PR allows an addon to set the URL for the help button.  In the report's ReportOptions class, the coder would add a line like:
`help_url = "https://www.gramps-project.org/wiki/index.php/DescendantsLines"`
or `help_url = URL_WIKISTRING + "DescendantsLines"`
where help_url is a class attribute.

### DIFF
--- a/gramps/gui/plug/report/_reportdialog.py
+++ b/gramps/gui/plug/report/_reportdialog.py
@@ -581,7 +581,10 @@ class ReportDialog(ManagedWindow):
         pass
 
     def on_help_clicked(self, *obj):
-        from ...display import display_help
+        from ...display import display_help, display_url
+        if hasattr(self.options, 'help_url'):
+            display_url(self.options.help_url)
+            return
         display_help(URL_REPORT_PAGE, self.report_name.replace(" ", "_"))
 
     def on_style_edit_clicked(self, *obj):


prculley:
As usual, I  looked at that.  It would have required changes to viewmanager, _reportdialog, _textreportdialog, graphreportdialog,  _webreportdialog, _drawreportdialog, _docreportdialog, _bookdialog, and _pluginreg to support the URL in the gpr file.

I would have to add plumbing to get the plugindata or the help_url from where it was available in viewmanager, through the 'report' function, all of the classes that use ReportDialog and ReportDialog itself.  Quite a few places that need a new parameter.

I deemed the four line one file change a bit easier.  Changes to each addon would be a wash either way, either modifying a line in the main file or the gpr.  I don't see where adding an optional parmeter to the ReportOptions (derived) classes is any better or more Pythonic than an optional attribute.

Nick-Hall:
Thanks for looking into this.  Bug fixes in a maintenance branch should generally modify the least amount of code.

We can always consider making a change in _master_ to make the code more consistent.

PR #737, opened 2018-12-09
vCalendar: Convenient display on mobile devices.

Keepun:
1809 Abraham Lincoln - Birth
1865 Abraham Lincoln - Death


sam-m888:
What program are you using to display your vCalendar file?  

> # feature requests [2356](https://gramps-project.org/bugs/view.php?id=2356), [1657](https://gramps-project.org/bugs/view.php?id=1657): avoid genitive form

Haven't checked but it looks like you are removing the genitive form feature?

PR #739, opened 2018-12-13
To transfer the mediapath verification code from importgpkg.py to the importxml.py file.

Keepun:
I want to use Git to control changes. After exporting the package, I do the unpacking. Relative paths to files in <file> tags. The <mediapath> tag is missing.

Import problem from such xml file.

https://gramps-project.org/bugs/view.php?id=10915

### DIFF
--- a/gramps/plugins/importer/importgpkg.py
+++ b/gramps/plugins/importer/importgpkg.py
@@ -60,7 +60,6 @@ def impData(database, name, user):
     # Create tempdir, if it does not exist, then check for writability
     #     THE TEMP DIR is named as the filname.gpkg.media and is created
     #     in the mediapath dir of the family tree we import to
-    oldmediapath = database.get_mediapath()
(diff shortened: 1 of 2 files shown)


sam-m888:
Note that travis-ci test broke with your change see
https://travis-ci.org/gramps-project/gramps/builds/467685185#L2014

Nick-Hall:
This needs to be discussed on the _gramps-devel_ list.

I have never fully understood the logic behind importing the base media path, and I'm not sure that we want to set it when importing XML without media.

Keepun:
Now the rules for importing the base media path are the same for packages and XML.

If the base media path was set prior to import, then a warning is displayed.

Nothing new.

sam-m888:
As mentioned by @Nick-Hall 

>> This needs to be discussed on the gramps-devel list.

Please resubmit if discussion agrees to this.

Thank you

PR #741, opened 2018-12-16
Croatian localisation updates

milotype:
please see the two commits ...


Nick-Hall:
@milotype Did you intend these commits to be merged into the _gramps50_ branch?

Nick-Hall:
@milotype Rebased on the _gramps50_ branch.  I used `msgmerge` to merge the po file into the new template.  Please test the result.

Nick-Hall:
> I read through many docs, but don't quite grasp it ... sorry!

I think we need to improve the documentation for translators.  You are not the only one that doesn't fully understand the new workflow.

milotype:
Maybe i misunderstand, but in the folders of the "maintenance/gramps50" i did not find my two updated files.

Btw, while checking, i've noticed, that for the .po files there was an "Update gramps bugtracker URL in all po files (#710)". Please note, that i DO NOT have these changes in my new .po file.

prculley:
It looks like Nick force-pushed to your own personal milotype:master branch so that it will now look like a Gramps50 branch with your changes.  So you need to checkout your master branch and see if it looks ok.  When Gramps is started it should say 5.0.1
If you can do this soon, it may make it into the upcoming 5.0.1 release which I am starting work on tomorrow....

milotype:
Sorry, I don't know how to "checkout".  :-(
I'm a simple user, installing revisions as .dmg files from Gramps download page at:
https://www.gramps-project.org/wiki/index.php?title=Download

prculley:
Ok.  I thought that since you had a Github account and a git master branch that was used for this PR that you were a bit more experienced.  I'll just accept this as is and we will hope for the best.

Nick-Hall:
@prculley Thanks. I don't expect any problems.

PR #742, opened 2018-12-18
Fix contents of enclosed_by secondary dbapi column

Nick-Hall:
The enclosed_by column should contain the first value in the list
instead of the last.

Resolves [#10907](https://gramps-project.org/bugs/view.php?id=10907).


prculley:
Works for newly imported trees and new trees.  Now for something to fix up corrupted trees...

Nick-Hall:
Running the "Rebuild secondary Indexes" tool should fix existing problems.

PR #743, opened 2018-12-20
Maintenance/gramps50

enel53:
Revised and updated Hungarian translation for maintenance/gramps50. Fixed typos and web links.
[hu.po.zip](https://github.com/gramps-project/gramps/files/2698942/hu.po.zip)


prculley:
I've taken your latest hu.po.zip file and used that to update the gramps50 branch.  I'm going to close this now, so you should open a new PR if you decide more updates are needed.

P.S. when this page shows more commits than the specific one you created, that means you did something wrong, typically that the source commit base and the branch you are trying to update are not the same.  I recommend that you make your changes to your own Github branch and use that to make a PR, its usually easier to get things right if you start with that.  Also make sure you have rebased any work on the most current source branch (master or maintenance/gramps50) before pushing.

PR #745, opened 2018-12-23
Set correct font resolution for Mac when drawing text directly to cairo.

jralls:
Fixes [#0010924](https://gramps-project.org/bugs/view.php?id=10924)

If accepted will cherry-pick to master.


Nick-Hall:
> If accepted will cherry-pick to master.

The _gramps50_ branch is now merged into _master_ on a regular basis.

sam-m888:
Thank you

PR #747, opened 2018-12-28
Avoid leading and trailing spaces when copy/paste

SNoiraud:
coordinates from a map provider.

Fixes [#10937](https://gramps-project.org/bugs/view.php?id=10937)


sam-m888:
Thank you

PR #752, opened 2019-01-02
Fix Media editor when using double click the preview of added media

prculley:
Fixes [#10923](https://gramps-project.org/bugs/view.php?id=10923)

Got a Nonetype error because the media object wasn't in database yet.


sam-m888:
Thank you

Nick-Hall:
@sam-m888 Thanks for merging, although we should probably leave a PR open for a week to allow for comments.

sam-m888:
Ok, sorry, forgot and didn't check how long the PR had been active!

PR #753, opened 2019-01-03
Corrections in 'Tip of Day'; changed 'Schlussnotizen' to 'Quellennach…

caigner:
I found some minor mistakes in the Tip of Day section and corrected them.

I also changed the word "Schlussnotizen" to a more meaningful "Quellennachweis".

@prculley  @Nick-Hall @Leonhaeuser  I hope it works this time without you having to correct my git mistakes. :-)

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -656,7 +656,7 @@ msgstr ""
 "<b>Eltern-Kind-Beziehung bearbeiten</b><br/> Du kannst die Beziehung eines "
 "Kindes zu seinen Eltern durch Doppelklicken des Kindes im Familieneditor "
 "bearbeiten. Beziehung kann jede Art sein von Adoptiert, Geburt, Pflege, "
-"Keine, Gefördert, Stief und Unbekannt."
+"Keine, Patenschaft, Stief und Unbekannt."
 
 #: ../data/tips.xml.in.h:28
 msgid ""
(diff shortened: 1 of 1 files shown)


Nick-Hall:
@caigner Looks good this time.

po/de.po
```diff
@@ -5125,7 +5125,7 @@ msgid ""
 "Matches people that have no family relationships to any other person in the "
 "database"
 msgstr ""
-"Liefert Personen, die keine familiären Beziehungen zu irgend jemandem in der "
+"Liefert Personen, die keine familiären Beziehungen zu irgendjemandem in der "
```
sebschub:
```suggestion
"Liefert Personen, die keine familiären Beziehungen zu jemandem in der "
```
sounds even better and also gets the right point, I think.

caigner: You are quite right, but in this PR I focused on other things.

There are more corrections to come in the future. :-)

po/de.po
```diff
@@ -19474,7 +19473,7 @@ msgstr "Keine Hauptperson"
 msgid ""
 "You need to set a 'default person' to go to. Select the People View, select "
 "the person you want as 'Home Person', then confirm your choice via the menu "
-"Edit ->Set Home Person."
+"Edit -> Set Home Person."
```
Leonhaeuser: correcting the original text makes no sence is generated from the sorce code

caigner: Thanks for the info. In this case I have to correct that in the source code.

prculley: You did the same thing at line 1708.  I think it would be best to NOT change this in the source code, as that would invalidate all other translations.  (Unless you want to change the pot and all the translations, in which case you should submit a separate PR for that.)

caigner: I did change it in the source code AND all translations in this PR:
https://github.com/gramps-project/gramps/pull/756

caigner:
@Nick-Hall 
I was wondering whether my changes are already merged into gramps50.
Do I have to do something?

prculley:
PR are usually left open for a week or so to allow time for people to review and comment on them.  Some take MUCH more time than that...

caigner:
@prculley Ah, ok. That's fine. Thanks.

sam-m888:
Thank you

PR #756, opened 2019-01-05
changed 'Edit ->Set Home Person' to 'Edit -> Set Home Person'

caigner:
I like consistency, so I fixed the missing space in the **No Home Person** dialog for the text
`Edit ->Set Home Person`
to
`Edit -> Set Home Person`
in all files.

https://gramps-project.org/wiki/index.php/Gramps_5.0_Wiki_Manual_-_Navigation#Setting_the_Home_Person


sam-m888:
Current **No Home Person** dialog
current-no home person - dialog-50

Updated **No Home Person** dialog

Nick-Hall:
I would have probably fixed this in the _master_ branch, but I suppose we can classify it as a bug.  It should be safe to merge.

Nick-Hall:
This should be rebased onto the _master_ branch and then merged.

PR #763, opened 2019-01-10
Not all place types appears on family lines Graph

SNoiraud:
Fixes [#10968](https://gramps-project.org/bugs/view.php?id=10968)


sam-m888:
Thank you

PR #764, opened 2019-01-10
Fix Place Tree view when using enclosed by sidebar filter

prculley:
the filter was incorrectly using a standard place cursor instead of the tree cursor.

Fixes [#10948](https://gramps-project.org/bugs/view.php?id=10948)

The bug points out that when using a sidebar enclosed by filter (and other filters) the resulting tree view stops displaying correctly.  This is easily shown when picking a high level place (like country) and using its id for 'enclosed by'.  The resulting tree view should still show subsidiary trees for states.  It doesn't.

The bug only appears to effect place trees, as that is the only tree view type that depends on a different cursor type.

The bug came about as a side effect of the optimization from https://github.com/gramps-project/gramps/pull/316.  I think this is the correct way to fix this, but welcome comments.

I was searching for filters that might need modification based on this, and came across ParamFilter which doesn't seem to be used.  Is this correct?

### DIFF
--- a/gramps/gen/filters/_genericfilter.py
+++ b/gramps/gen/filters/_genericfilter.py
@@ -124,6 +124,9 @@ class GenericFilter:
     def get_cursor(self, db):
         return db.get_person_cursor()
 
+    def get_tree_cursor(self, db):
+        return db.get_person_cursor()
+
     def make_obj(self):
         return Person()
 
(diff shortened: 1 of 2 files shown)


sam-m888:
> I was searching for filters that might need modification based on this, and came across ParamFilter which doesn't seem to be used. Is this correct?

Not sure? @Nick-Hall Any idea about this one?

sam-m888:
Thank you

PR #765, opened 2019-01-10
Add help menu entries for Gramplet Bar

sam-m888:
~To improve discoverability of the~ Gramplet Bar Menu ~(Currently a nameless down arrown at end of each Gramplet bar title tab) add a tooltip and~ add a menu entry that goes to the help page.

When the sidebar and bottombar are empty a message appears that calls them both the Gramplet Bar  with the messsage [Select the down arrow on the right corner for adding, removing or restoring gramplets.] eg: on the Relationships category view

Issue [#9042](https://gramps-project.org/bugs/view.php?id=9042)

gramplet-bar-new-menu-entries50

### DIFF
--- a/gramps/gui/widgets/grampletbar.py
+++ b/gramps/gui/widgets/grampletbar.py
@@ -77,6 +77,8 @@ from ..dialog import QuestionDialog
 #
 #-------------------------------------------------------------------------
 WIKI_HELP_PAGE = URL_MANUAL_PAGE + '_-_Gramplets'
+WIKI_HELP_GRAMPLETBAR = 'https://gramps-project.org/wiki/index.php/Gramps_5.0_Wiki_Manual_-_Main_Window#Bottombar_and_Sidebar' # rename section to Gramplet_bar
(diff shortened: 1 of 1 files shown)


fxtmtrsine:
Thanks that tooltip is so sensible, stop me having to guess what the down arrow does!

One issue I have is with this menu being different to the one on the dashboard, any plans to fix that also?

Nick-Hall:
Good idea.  I'm not sure if it should be classified as a bug though.

sam-m888:
Tooltip split to separate PR #768

sam-m888:
> Good idea. I'm not sure if it should be classified as a bug though.

@Nick-Hall if you mean the tooltip then if @fxtmtrsine  's comment is not a good enough example that this is more of a bug than an enhancement then ok sure ! If you mean adding Help links to the submenu  well it benefits the users and is one of the biggest complaints. Either way I've split the PR to be considered separately.  I strongly suggest both go into gramps50 and not master, hopefully you agree?

> One issue I have is with this menu being different to the one on the dashboard, any plans to fix that also?

@fxtmtrsine  No one has raised it as an issue, but I can see the difference in behavior. Please open a issue/feature request on Gramps bugs ( https://gramps-project.org/bugs/ ) on how it should be changed, if at all?

PR #766, opened 2019-01-10
fix Place Format Editor file save/load for difficult names

prculley:
Fixes [#10949](https://gramps-project.org/bugs/view.php?id=10949)

Bug Submitter noticed that when the place format editor name field contained international characters or non-safe for XML characters that Gramps would save it to place_formats.xml but fail (crash) when restarting Gramps.

The place format load routine was not robust against bad XML (so Gramps crashed) and the save routine would create bad XML.

I decided to use the minidom to create the XML file instead of doing manually (the standard Python library dealt with the name escaping).  I also wrote the file with 'utf-8' encoding (not standard on all systems).

P.s. the new error string is already in the .pot file.

### DIFF
--- a/gramps/gen/display/place.py
+++ b/gramps/gen/display/place.py
@@ -54,11 +54,6 @@ class PlaceFormat:
         self.street = street
         self.reverse = reverse
 
-    def to_xml(self):
-        return ('  <format name="%s" levels="%s" language="%s" '
-                'street="%s" reverse="%s"/>\n' %
-                (self.name, self.levels, self.language,
-                 self.street, self.reverse))
 
 #-------------------------------------------------------------------------
 #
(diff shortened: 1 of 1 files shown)


lordemannd:
looks good to me, tested on Windows with many of the problem characters. No start up failures!
good job!

Nick-Hall:
@prculley Two pylint issues, but otherwise it looks good.  Using minidom provides a neat solution for small xml files.

prculley:
Grrr...  The one thing I didn't change was the variable name for the file handle, and of course it is too short.
What do I use for an exception when I don't have any idea what error might be generated?  I tried a couple of ways to make the XML invalid, in addition to the original failure on bad encoding, and got wildly different exceptions.

sam-m888:
Thank you

PR #769, opened 2019-01-12
Fix Relationship Graph; extra people when using filters & subgraphs

prculley:
subgraph option wasn't checking parents against filtered results

Fixes [#10947](https://gramps-project.org/bugs/view.php?id=10947)

Submitter noticed extra people in his graph when using the subgraph option and a filter which should have prevented their view; the nodes were present, although the names and info was not (only the GrampsID).
Turns out they were parents of already existing persons and Family nodes was on.

The add_link method added a node if one was not already present.

### DIFF
--- a/gramps/plugins/graph/gvrelgraph.py
+++ b/gramps/plugins/graph/gvrelgraph.py
@@ -488,13 +488,13 @@ class RelGraphReport(Report):
             self.doc.start_subgraph(fam_id)
             f_handle = fam.get_father_handle()
             m_handle = fam.get_mother_handle()
-            if f_handle:
+            if f_handle in self.persons:
                 father = self._db.get_person_from_handle(f_handle)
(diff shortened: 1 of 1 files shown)


FuruFam61:
Can f_handle or m_handle ever be None?

That's certainly what I would imagine by looking at the original code.

Don't you want something like "if f_handle and f_handle in self.persons"?

prculley:
self.persons is a set of handles, or an empty set.  It will never contain the value 'None' so it is not necessary to do a separate test for a 'None' handle.

FuruFam61:
The question was "Can f_handle or m_handle ever be None?".

If so, then you think "if None in self.persons" won't crash?

prculley:
Yes, f_handle or m_handle can be None.  And None in self.persons will not crash or ever be True.

sam-m888:
Thank you

PR #770, opened 2019-01-12
We should stay on the selected item in list views:

SNoiraud:
Fixes [#10966](https://gramps-project.org/bugs/view.php?id=10966)

When we try to merge two objects, we should stay on the selected row
in list views.
This problem occurs for persons, families, events, places, sources,
citations, repositories, media and notes.

Another problem:
In case we select the first family and select the gramps_id of the
second family, the new gramps_id is ignored: problem in MergeFamilyQuery

### DIFF
--- a/gramps/gen/merge/mergefamilyquery.py
+++ b/gramps/gen/merge/mergefamilyquery.py
@@ -133,7 +133,8 @@ class MergeFamilyQuery:
         old_handle = self.titanic.get_handle()
 
         with DbTxn(_('Merge Family'), self.database) as trans:
-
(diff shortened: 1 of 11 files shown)


Nick-Hall:
A family merge unit test is failing.

prculley:
Yes, I just saw that and am looking into it now.

prculley:
Looks like Travis is failing again, this time because different signals are occurring with the extra commit.  Out of time now, so will deal with it later.  Just have to check/change the test I think...

sam-m888:
Thank you

PR #771, opened 2019-01-13
Fix Name editor crash after clearing a group_as name on dbapi dbs

prculley:
Fixes [#10981](https://gramps-project.org/bugs/view.php?id=10981)

Looks like the dbapi code for set_name_group_mapping is not properly removing the name_group when a 'None' grouping value is provided.
Also the method was not emitting a signal for the GUI to update.


sam-m888:
Thank you

PR #773, opened 2019-01-15
Fix Descendent Tree report for crash when person has multiple

prculley:
families and one of them doesn't have a spouse.

Fixes ([#10661](https://gramps-project.org/bugs/view.php?id=10661))  ([#10971](https://gramps-project.org/bugs/view.php?id=10971))  ([#10983](https://gramps-project.org/bugs/view.php?id=10983))

This seems to fix the specific issue with the reporters crash, and I have not found any obvious issues with the half dozen other trees I tried.
The problem was created with an attempt to fix HandleErrors, a later commit did a more thorough job of dealing with them so the specific line deleted in this PR is no longer needed.


sam-m888:
Thank you

PR #774, opened 2019-01-18
Have a better narrative web site

SNoiraud:
Fixes #10341, #10962, #11018, #11029

10341:
When we are on a mobile phone or a small device, we suppress the navigation tab.
In place, we have a new icon on the upper left which is used to show the dropdown menu.
Thanks to Theo van Rijn for showing me the way to do that.

10962:
For Home, Introduction and Contact, If we have an image and this image contains regions,
show the regions. We can go directly to the person page associated to this region.
If we click on the image,  we go directly to the associated media page. This will be true only if we selected "include images and media objects" and "create and only use thumbnail" is unselected

11018:
The first line identifying a family will be more legible.
The link is not useful in the parents and pedigree section for the current person.
Adapt some css files.

11029:
sort the place references either by date or by name.

### DIFF
--- a/gramps/plugins/webreport/addressbook.py
+++ b/gramps/plugins/webreport/addressbook.py
@@ -12,7 +12,7 @@
 # Copyright (C) 2008-2011  Rob G. Healey <robhealey1@gmail.com>
 # Copyright (C) 2010       Doug Blank <doug.blank@gmail.com>
 # Copyright (C) 2010       Jakim Friant
-# Copyright (C) 2010-2017  Serge Noiraud
+# Copyright (C) 2010-      Serge Noiraud
 # Copyright (C) 2011       Tim G L Lyons
 # Copyright (C) 2013       Benny Malengier
 # Copyright (C) 2016       Allen Crider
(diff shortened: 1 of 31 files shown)


SNoiraud:
The following snapshot if for the image regions in the Introduction page. This functionality is avalaible for Home, Introduction and Contact pages:
new_navweb regions

lordemannd:
can i use these changes with 5.0.1 or does it need to be the master branch?

SNoiraud:
I didn't try this on 5.0.1 perhaps it should work, but there are many files to change (at least 23)

gramps/plugins/webreport/addressbook.py
```diff
@@ -84,12 +84,13 @@ def __init__(self, report, title, person_handle, has_add, has_res, has_url):
 
         # set the file name and open file
         output_file, sio = self.report.create_file(person_handle, "addr")
-        addressbookpage, head, body = self.write_header(_("Address Book"))
+        result = self.write_header(_("Address Book"))
+        addressbookpage, head, body, outerwrapper = result
```
Nick-Hall: You can indicate that a variable is expectedly unused by adding an underscore.  See the regexp in the pylint [configuration](https://github.com/gramps-project/gramps/blob/master/.pylintrc#L290) file.

Consider using `body_` instead of `body`.

SNoiraud: I don't think that works. I replaced body by body_
worst, I have new errors.

SNoiraud:
fr11029 2

Nick-Hall:
@SNoiraud I am happy for this to be merged.  Do you want to keep the full commit history?

SNoiraud:
>  Do you want to keep the full commit history?

No.

PR #775, opened 2019-01-21
Config dialog [master]

vantu5z:
Some changes to config dialog:
- add labels to panels (tabs) and align them to center
- add tooltips to checkboxes
- code cleanup and pylint fixes
- at `General` tab separate addons section by adding label
- set default value for `Add tag on import` option when clear the Entry
- add other colors for family at `Color` tab (fixes 10995)
- split colors to sections

After(top) and Before (bottom), changes are marked:

ToDo:
- [ ] Fix text of tooltips and labels
- [ ] Add tooltips for other elements (if needed)
- [ ] Choose colors for added family colors


vantu5z:
@prculley thanks for your feedback, I just supposed as we use gtk and Gnome use switches in config dialogs, so it is logical to use them (to uniform look).
I agree with you that switches useful only for Mobile applications.
So should I rewrite PR and exclude change checkbox to switch?
What about other changes?

fooltech:
Whilst I am not precious about borders others might be I would be happy with the same border colour for all Family nodes.
I wonder now if this creating too many colours?
Default I consider to be Not Known: (ie the start point when building a tree)
Married:         Religious based legally binding arrangement between two people
Civil Union:     Non religous based legally binding arrangement between two people
Divorced:       Ending of a legally binding arrangement between two people
Unmarried:     Woman who gives birth not married or in civil union, or two people living together with                         no legally binding arrangement.
Are the ones I would use others with a semantically different viewpoint may wish to use Unknown.
Civil Union and Divorce are still relatively rare so for me generally any view would in the main consist of 3 colours for famliy nodes which I feel is more than acceptable.
Thank you for your efforts on this.

SNoiraud:
May I add the patch for bug #11046 ?
I need to add a date error color depending on theme.
I attach a screenshot which show the result. Do you agree ?
bug11046 1

Nick-Hall:
This PR generally looks good.  The only problem that I can see is that centred headings dd not conform to the Gnome HIG specified in our UI guidelines.

Nick-Hall:
> I need to add a date error color depending on theme.

There is actually an "entry.error" node in the Gtk CSS file.  Perhaps we could use the background colour from this instead?

SNoiraud:
> > I need to add a date error color depending on theme.
> 
> The Gtk Entry widget actually has an "error" and "warning" class in the CSS style definition. Perhaps we should use this instead?
I don't understand that: there is only one cssprovider in gramps and it use only the error_color. I don't see warnig for that. The part of code is in gramps/gui/widgets/validatedmaskedentry.py

SNoiraud:
> This PR generally looks good. The only problem that I can see is that centred headings dd not conform to the Gnome HIG specified in our UI guidelines.

Just one question: Where is the UI guidelines? I saw nothing in the documentation.

Nick-Hall:
> Where is the UI guidelines?

There is a [UI style](https://gramps-project.org/wiki/index.php/UI_style) page in the wiki.

vantu5z:
@Nick-Hall Do you mean this guidelines?
https://www.gramps-project.org/wiki/index.php/UI_style
https://developer.gnome.org/hig/stable/

But I can't find about center align.

Nick-Hall:
> Do you mean this guidelines?

Yes.  You should also refer to the older [HIG 2.2.3](https://developer.gnome.org/hig-book/3.12/index.html.en).  In this case, the [Window Layout](https://developer.gnome.org/hig-book/3.12/design-window.html.en) section is relevant.

Note that the headings in configuration dialogs are bold and on the left, and that the use of frames is deprecated.

Nick-Hall:
> The part of code is in gramps/gui/widgets/validatedmaskedentry.py

Yes.  I actually re-wrote that code in a [commit](https://github.com/gramps-project/gramps/commit/e2a5fe8e1243afdc58cdd2d81cdb7dc093f78ee0) about three years ago.  In it I created two new CSS classes, "fade" and "bg", in order to replicate the existing functionality.  However, Gtk themes already provide two classes, "warning" and "error" for this purpose.  Using these classes would provide a look & feel consistent with other applications.  For example, using the "error" class in my theme, I get red foreground text and a red border.

SNoiraud:
As this entry validation is used in several modules and not only date, I changed the name 'Date' to 'Entry'

vantu5z:
Use one setting to line and scrolled window for colors settings.
Снимок экрана от 2019-03-13 16-20-51

vantu5z:
file:///home/vantu5z/%D0%98%D0%B7%D0%BE%D0%B1%D1%80%D0%B0%D0%B6%D0%B5%D0%BD%D0%B8%D1%8F/%D0%A1%D0%BD%D0%B8%D0%BC%D0%BE%D0%BA%20%D1%8D%D0%BA%D1%80%D0%B0%D0%BD%D0%B0%20%D0%BE%D1%82%202019-03-15%2010-54-50.png

Nick-Hall:
Much better!  It's looking good now.

fooltech:
Obviously far more work gone into this than I thought but hopefully it will be of use to more than just me, it appears to fulfill the initial request
Thank you

Nick-Hall:
@vantu5z I am happy for this to be merged. Do you want to keep the full commit history?

vantu5z:
@Nick-Hall I don't need all commits, so you can merge it by one commit.

PR #776, opened 2019-01-22
Properly detect missing birth and death years.

jelmer:
Without this change, birth and death dates get set to "0" in the [fan chart.](https://gramps-project.org/wiki/index.php/Gramps_5.0_Wiki_Manual_-_Reports_-_part_4#Fan_Chart) if missing.


sam-m888:
Thank you. 

Does this references an existing issue on the [bug tracker](https://gramps-project.org/bugs/)? What steps to recreate this do you suggest?

As a bug this should be raised as a PR against the maintenance/gramps50 branch.

PR #777, opened 2019-01-23
Fix EditFamily for adding a child to single parent family with Surname guessing set to combination.

prculley:
Fixes [#10997](https://gramps-project.org/bugs/view.php?id=10997)


lordemannd:
solves the exception for me. (win64)

sam-m888:
Thank you

PR #778, opened 2019-01-25
On restart after crash, offer to run Check & Repair

prculley:
While helping users I've lately noticed a trend where they get db corruptions and then have a difficult time getting Gramps running again; often with a crash when they start up again.  Gramps issue [#10989](https://gramps-project.org/bugs/view.php?id=10989) While it is possible to use the CLI to start Gramps and run Check & Repair, or to start Gramps on another tree and get the view off of the one that is crashing, most users may not be able to figure this out on their own.

This PR tests for HandleError with 'not found' when there is a crash with Error dialog.  If this is the case (a type of error that Check&Repair can deal with), then the new code sets 'behavior.runcheck' in gramps.ini.

The next time Gramps runs, before opening any views on an open db, it sees the config setting and offers to run Check & Repair.

Unfortunately to test this requires a corrupted db, which seems to be harder than the users have found it to be.  I did it by setting a birth event on a person and using my debugger, setting a breakpoint on the Event editor DeleteEventQuery.query_response just after the db transaction was started.  Then I tried to delete the event from event view, when I got the breakpoint, I used the debugger to skip over the handle reference checks and continued execution at the actual remove_event.

This leaves the db in a corrupted state, where references exist to an event that is missing.  When you go to person view you should get crash report as it tries to put up the birth data on the person.  I usually have to kill the Gramps task to get out of this.  If my preferences were set to open previous view and previous db, restarting Gramps and breaking the lock will also (normally) result in the crash again.  But with this PR you should get a chance to run Check & Repair before a crash could occur.

This is an enhancement, but I am optimistically posting it to gramps50 branch, in the hope that it gets to users sooner...


prculley:
I've just found a way to corrupt the db easier.  At least until a PR for that gets accepted.
Add a new Citation, but don't hit 'OK' on the citation editor.  Select or add a Source.  Go to Source view and delete the newly used source.  Return to the Citation Editor and hit 'OK'.  You have just created a Citation with a bad reference to a Source; a corrupted db.  If you return to Citation view, it should bring up the Error dialog.

lordemannd:
Good idea, I like it. How does it relate to the Repair button on the Manage Family Trees dialog?
If it was connected to that button it could allow user to fix the db manually in addition to fixing when the db is loaded.

Although I haven't seen the Repair button "light up" and be active recently (using MS-Windows), in the long distant past it tried to give the non-technical user the ability to fix-up DB problems. My memory is that it tried to fix problems were the DB couldn't be opened because of missing or corrupted files.
(the component files that store a Berkley DB)

sam-m888:
- [ ] Add new dialog to user manual. (TBD by @sam-m888  )

Nick-Hall:
@prculley Needs to be rebased on _master_ and then merged.

PR #779, opened 2019-01-27
Fix for delete of a ref'd primary obj while editing an added obj.

prculley:
Also fixed to update the ref'd obj on changes from outside the editor.

Fixes [#10999](https://gramps-project.org/bugs/view.php?id=10999), [#11000](https://gramps-project.org/bugs/view.php?id=11000),[ #11001](https://gramps-project.org/bugs/view.php?id=11001), [#11002](https://gramps-project.org/bugs/view.php?id=11002)

In all cases we are adding a new primary object and in the object edit dialog.  Then the user uses a view to delete an object referenced in the open dialog.

I initially found one example of this bug, then decided to look around for other examples, after a few bug reports, I decided to switch this to a class of bugs and deal with them in one PR.

added Primary obj | the referenced deleted obj | affected files
------------------- | ---------------------------- | -------------
Person Add | Event Delete | eventembeddedlist.py, editperson.py
Family Add | Event Delete | eventembeddedlist.py
Person Add | Associated Person delete | personrefembedlist.py
Family Add | Mother/Father delete | editfamily.py
Family add | Child delete | editfamily.py
Citation Add | Source delete | editcitation.py
Event add | Place delete | editevent.py
Event Reference add | Place delete | editeventref.py
Place add | enclosed place delete | placerefembedlist.py 

It looks like @bmcage attempted to fix some of this some years ago, I followed his basic ideas in my fixes.
https://github.com/gramps-project/gramps/commit/ee69317b6243d532a336544560945c43b251aec5

### DIFF
--- a/gramps/gui/editors/displaytabs/eventembedlist.py
+++ b/gramps/gui/editors/displaytabs/eventembedlist.py
@@ -131,15 +131,15 @@ class EventEmbedList(DbGUIElement, GroupEmbeddedList):
             refs = self.get_data()[self._WORKGROUP]
             ref_list = [eref.ref for eref in refs]
             indexlist = []
-            last = 0
+            last = -1
             while True:
                 try:
-                    last = ref_list.index(handle)
+                    last = ref_list.index(handle, last + 1)
                     indexlist.append(last)
                 except ValueError:
                     break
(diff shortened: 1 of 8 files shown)


sam-m888:
Thank you

PR #780, opened 2019-01-29
Fix XML import to add tags to Events, Sources, Places, Repos, Cits

prculley:
Fixes [#11004](https://gramps-project.org/bugs/view.php?id=11004)

Apparently the feature to add a tags was originally not going to work with all object types.  https://github.com/gramps-project/gramps/commit/771ae761e28d166b5c8d93eaa0f28cf65602b4b4
added support for the original planned tagged objects, but when all objects became taggable, the XML import default tag stuff did not get updated.  https://github.com/gramps-project/gramps/commit/9c66c62d5c6e1b245c055b6a21566cc6c627bb0f


sam-m888:
Thank you

PR #782, opened 2019-01-31
Fix Gedcom import for multiple notes on OBJE (MULTIMEDIA_LINK)

prculley:
Fixes [#10277](https://gramps-project.org/bugs/view.php?id=10277)

A user pointed out that we are not consistent in our support for multiple notes attached to a media.  In older versions of Gramps we could export them, but not import the exported Gedcom file without losing a note.
This PR corrects this.

This really only applies to a non standard method of attaching notes to a OBJE in Gedcom, if the user's program had supported Gedcom v5.5.1 properly, it would have imported fine.

A recent Gedcom export patch upgraded Gramps OBJE export to v5.5.1 as well, so we can now fully support notes on media objects both with the non-standard Gedcom syntax and the standard one.


sam-m888:
Thank you

PR #783, opened 2019-02-03
New tool to suppress leading and trailing spaces.

SNoiraud:
Fixes [#10824](https://gramps-project.org/bugs/view.php?id=10824)

This tool is looking for place names with leading or/and trailing spaces.
It also looks for in the first name and surname.

For each entry which contains leading or trailing spaces,
a row is added in a treeview.
You can see where the spaces are for each row as the name is underlined.

If you double click on the row, you can edit the Place or the Person.
For places, I search for comma. They are forbidden for lat/lon (related to (https://gramps-project.org/bugs/view.php?id=11114)

### DIFF
--- a/gramps/plugins/tool/removespaces.glade
+++ b/gramps/plugins/tool/removespaces.glade
@@ -1,0 +1,235 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.22.1 -->
+<interface>
+  <requires lib="gtk+" version="3.10"/>
+  <object class="GtkDialog" id="removespaces">
+    <property name="can_focus">False</property>
+    <property name="default_width">650</property>
+    <property name="default_height">400</property>
+    <property name="type_hint">dialog</property>
+    <signal name="delete-event" handler="on_delete_event" swapped="no"/>
+    <child>
+      <placeholder/>
+    </child>
+    <child internal-child="vbox">
+      <object class="GtkBox" id="dialog-vbox1">
+        <property name="visible">True</property>
+        <property name="can_focus">False</property>
+        <property name="orientation">vertical</property>
+        <child internal-child="action_area">
+          <object class="GtkButtonBox" id="dialog-action_area1">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="layout_style">end</property>
+            <child>
+              <object class="GtkButton" id="helpbutton1">
+                <property name="label">gtk-help</property>
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="can_default">True</property>
+                <property name="receives_default">False</property>
+                <property name="use_stock">True</property>
+                <property name="always_show_image">True</property>
+                <signal name="clicked" handler="on_help_clicked" swapped="no"/>
+              </object>
+              <packing>
+                <property name="expand">False</property>
+                <property name="fill">False</property>
+                <property name="position">1</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkButton" id="close">
+                <property name="label">gtk-close</property>
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="can_default">True</property>
+                <property name="receives_default">False</property>
+                <property name="use_stock">True</property>
+                <property name="image_position">right</property>
+                <property name="always_show_image">True</property>
+                <signal name="clicked" handler="destroy_passed_object" swapped="yes"/>
+              </object>
+              <packing>
+                <property name="expand">False</property>
+                <property name="fill">False</property>
+                <property name="pack_type">end</property>
+                <property name="position">1</property>
+              </packing>
+            </child>
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">True</property>
+            <property name="pack_type">end</property>
+            <property name="position">0</property>
+          </packing>
+        </child>
+        <child>
+          <object class="GtkBox" id="vbox1">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="border_width">6</property>
+            <property name="orientation">vertical</property>
+            <property name="spacing">6</property>
+            <child>
+              <object class="GtkLabel" id="title">
+                <property name="height_request">1</property>
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="double_buffered">False</property>
+                <property name="justify">center</property>
+                <property name="ellipsize">start</property>
+              </object>
+              <packing>
+                <property name="expand">False</property>
+                <property name="fill">False</property>
+                <property name="padding">5</property>
+                <property name="position">0</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkBox" id="vbox2">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="margin_top">2</property>
+                <property name="margin_bottom">2</property>
+                <property name="orientation">vertical</property>
+                <property name="spacing">5</property>
+                <property name="homogeneous">True</property>
+                <property name="baseline_position">top</property>
+                <child>
+                  <object class="GtkBox">
+                    <property name="visible">True</property>
+                    <property name="can_focus">False</property>
+                    <property name="orientation">vertical</property>
+                    <child>
+                      <object class="GtkBox">
+                        <property name="visible">True</property>
+                        <property name="can_focus">False</property>
+                        <property name="orientation">vertical</property>
+                        <child>
+                          <object class="GtkLabel" id="label_1">
+                            <property name="visible">True</property>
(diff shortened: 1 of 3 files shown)


SNoiraud:
fr10824 1

Nick-Hall:
Do you plan to keep the handle column?

SNoiraud:
I could hide it. I need it to call the editor.

prculley:
If you want to get this out to users anytime soon, you might consider packaging it as an addon.  The way thing are going, I don't think we will see Gramps5.1 anytime this year.

Nick-Hall:
> The way thing are going, I don't think we will see Gramps5.1 anytime this year.

We will probably release v5.1 in July or August.

lordemannd:
Can it also look at Alternate Names, Longitude, Latitude? (really any user enter-able text field)

lordemannd:
I added the newest changes to the 5.0.1 version on windows and get the error reporting wizard.
Is this a GTK version issue? Should I just wait for 5.1?

8609344: ERROR: tool.py: line 257: Failed to start tool.
  File "C:\Program Files\GrampsAIO64-5.0.1-WORK\gramps\gui\plug\tool.py", line 253, in gui_tool
    callback = callback)
  File "C:\Program Files\GrampsAIO64-5.0.1-WORK\gramps\plugins\tool\removespaces.py", line 77, in __init__
    top_dialog = Glade()
  File "C:\Program Files\GrampsAIO64-5.0.1-WORK\gramps\gui\glade.py", line 158, in __init__
    self.add_from_string(data)
  File "AIO/gi/overrides/Gtk.py", line 464, in add_from_string
GLib.GError: gtk-builder-error-quark: .:125:58 Invalid property: GtkScrolledWindow.max_content_height (11)

* Python 3.6.4 (Success version 3.2.0 or greater installed.)
* GTK+ 3.18.9 (Success version 3.10.0 or greater installed.)
* PyGObject 3.26.1 (Success version 3.12.0 or greater installed.)

Nick-Hall:
The [max_content_height](https://developer.gnome.org/gtk3/stable/GtkScrolledWindow.html#GtkScrolledWindow--max-content-height) property requires Gtk 3.22 and above.  We decided on Gtk v3.12 for Gramps v5.1.

lordemannd:
After "Come back to Gtk 3.12 : remove max_content_height"
I get this error. (I will now go back and try to confirm I have all the changes)

279416: ERROR: tool.py: line 257: Failed to start tool.
  File "C:\Program Files\GrampsAIO64-5.0.1-WORK\gramps\gui\plug\tool.py", line 253, in gui_tool
    callback = callback)
  File "C:\Program Files\GrampsAIO64-5.0.1-WORK\gramps\plugins\tool\removespaces.py", line 77, in __init__
    top_dialog = Glade()
  File "C:\Program Files\GrampsAIO64-5.0.1-WORK\gramps\gui\glade.py", line 158, in __init__
    self.add_from_string(data)
  File "AIO/gi/overrides/Gtk.py", line 464, in add_from_string
GLib.GError: gtk-builder-error-quark: .:136:67 Invalid property: GtkScrolledWindow.propagate_natural_width (11)

SNoiraud:
It's difficult to know all properties glade uses. I am on Ubuntu with Gtk 3.22.
I hope I forgot nothing else.

lordemannd:
Thanks, its working now. I will do some testing.

lordemannd:
It doesn't seem to be checking the alternate surname or given names.

SNoiraud:
Can you attach a database example (.gramps) for that.

lordemannd:
Using current Gramps, the process of backup (.gramp) and then import fixes many of these issues.
I tried using other export methods, they also result in the data being corrected.

The specific issue in my test was caused originally by cutting/pasting names from web sources. When doing this, sometimes the copied text includes control characters (carriage return, line feed, tab, vtab, bell, etc.) that are considered "white space". 

In older versions of Gramps these control characters where not displayed, making it nearly impossible to find them. (but causing problems in many generated reports )
Current Gramps makes them visible.

On Windows, I can recreate the problem in alternate names easily. 
(my apology for the painful detail, just trying to be clear)
- in Gramps, add a new name to a person to open the name editor.
- Windows, open the notepad application
- add 2 lines of text.
- move cursor to the end of the 1st line
- left click and hold button
- drag cursor to end of 2nd line
- release button
you should now have the 2nd line of text highlighted (including the invisible end of the 1st line)
- right click and choose copy
- return to Gramps and paste the text into the Given name and Surname field.
- click OK button to save this alternate name.
- run the space removal tool.
- notice that this alternate name is not found.
- now make the alternate name the default name
- rerun the tool
- notice that as the default name to issue is detected.

SNoiraud:
> run the space removal tool.
> notice that this alternate name is not found.

It works for me. We have no alternate name here. The Given and Surname names are displayed.

> now make the alternate name the default name
It works.

I attached the .gramps file for which it works. I manually edit to add the new line.
when I import it, this new line is removed only for the Given name.

So, my question now is:
Why given names are processed differently than surname or alternate names?

Perhaps solving this kind of problem would be sufficient and make this tool obsolete.

remove .txt in the file name before import it.
[PR783.gramps.txt](https://github.com/gramps-project/gramps/files/3075942/PR783.gramps.txt)

lordemannd:
In your it looks like the "newline" (\n) is in the middle of the name text. Did you try the "newline" character as the 1st or last character in the name?

SNoiraud:
> Did you try the "newline" character as the 1st or last character in the name?
Yes. in the provided file, given, surname and alternate begin with a new line.

SNoiraud:
I have a patch to suppress all invisible characters in the following filed:
call_name, nick_name, family_nick_name, first_name, suffix, title, surname and prefix

I think this is a better solution

[try_1.patch.txt](https://github.com/gramps-project/gramps/files/3076128/try_1.patch.txt)

with this patch, this tool is obsolete for the Person.

SNoiraud:
The best is perhaps to modify the MonitoredEntry class.

SNoiraud:
This is the new patch for the MonitoredEntry.
It is now impossible to enter 
line feed, carriage return, tab, ... in a field

[try_2.patch.txt](https://github.com/gramps-project/gramps/files/3076278/try_2.patch.txt)

Nick-Hall:
This seems to be progressing well.

SNoiraud:
I think we do not need this tool if we modify the MonitoredEntry with the try_2 patch.
It works on all views: people, events, places, media, sources and repositories

I close this PR. I'll create a new one. Can it be a bug for gramps 5.0 ?

lordemannd:
it can be a bug if the we think the Gramps specification says: leading and trailing white space are not allowed in field text and will be removed.

also saying that white space is any non-printing character to include space, nl, cr, tab, vtab, bell, etc.
(i personally like this and would further say that all white space strings within the text should be replaced with a single space)

Does this apply directly for text blocks like in a Note? or should blocks of text have different rules?

BUT i think this tool is still needed:
1 - in case there are bugs in the bug fix
2 - for users that are upgrading from a previous version (like 3.4 to 5.0), they will need to be able to clean up things that older Gramps versions allowed. The upgrade I refer to is where an existing BSD DB is converted in place instead of the backup and import method.

Nick-Hall:
> Can it be a bug for gramps 5.0 ?

Yes, I think that allowing control characters to be pasted into an entry field could be considered a bug.  Please open another PR.

We probably should allow a non-breaking space because it is used to indicate that multi-part names in Arabic and is used by the Given Names Cloud gramplet.  However, you only seem to be removing control codes in the ASCII range 0-31.

Nick-Hall:
> Does this apply directly for text blocks like in a Note?

Notes are handled differently and may contain whitespace.

lordemannd:
for Arabic, is the non-breaking space ever the 1st or last character in the field?

SNoiraud:
> Please open another PR.

See #811

Nick-Hall:
@SNoiraud What was the decision about this?  Do we still need the tool?

SNoiraud:
> What was the decision about this? Do we still need the tool?

I think we need this tool after an import.
Perhaps the problem is without the name: this plugin also replace commas and not only leading and trailing spaces.
What could be the name? (standardization of input fields ?)

Nick-Hall:
> What could be the name? (standardization of input fields ?)

I suppose it is a form of data cleansing, so something like "Clean input data"?

Nick-Hall:

SNoiraud:
We can merge it.

sam-m888:
Note to self
[ ]  the tool link to the user manual needs to be updated from
https://gramps-project.org/wiki/index.php?title=Gramps_5.1_Wiki_Manual_-_Tools#Remove_leading_and_trailing_spaces
to
https://gramps-project.org/wiki/index.php?title=Gramps_5.1_Wiki_Manual_-_Tools#
Clean_input_data

You access it from the menu "Tools > Utilities > Clean input data..."

See:

https://sourceforge.net/p/gramps/mailman/message/36905316/

PR #784, opened 2019-02-05
Deal with SQLite db corrupted by None name mapping

prculley:
Fixes [#11011](https://gramps-project.org/bugs/view.php?id=11011)

Submitter points out that just fixing the original bug dealt with in #771 is not sufficient.  The db may have been corrupted, and worse yet, XML backups of the corrupted db may have been made.

This PR fixes dbapi to ignore corrupted entries in the current db, and fixes XML import to ignore a bad group_as in the XML file.

### DIFF
--- a/gramps/plugins/db/dbapi/dbapi.py
+++ b/gramps/plugins/db/dbapi/dbapi.py
@@ -330,9 +330,11 @@ class DBAPI(DbGeneric):
         """
         Return the defined names that have been assigned to a default grouping.
         """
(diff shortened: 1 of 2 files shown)


johnbickmore:
@Nick-Hall @prculley This is probably worth an earlier point release of Gramps especially if you want to encourage the testing and use of the SQLite backend 👍

sam-m888:
@johnbickmore Hi, this questions is best asked on the Gramps developers mailing list. In general a new release of Gramps comes out every 3 months or so.

sam-m888:
Thank you

PR #785, opened 2019-02-06
Fix example gramps files so media is found

prculley:
Fixes [#11015](https://gramps-project.org/bugs/view.php?id=11015)

On Windows the location of the example files is "c:\program files\grampsaio64-5.0.1\share\doc\gramps\example\gramps".

On my Ubuntu install from the .deb file these files are in "/usr/share/doc/gramps/example/gramps"

Since the default GRAMPS_RESOURCES value for each system is the 'share', to find the media files we need to append "/doc/gramps/example/gramps" to have the example files work properly.

### DIFF
--- a/example/gramps/data.gramps
+++ b/example/gramps/data.gramps
@@ -15,7 +15,7 @@
       <resphone>(555)123-4567</resphone>
       <resemail>anyone@someplace.com</resemail>
     </researcher>
-    <mediapath>{GRAMPS_RESOURCES}/example/gramps</mediapath>
+    <mediapath>{GRAMPS_RESOURCES}/doc/gramps/example/gramps</mediapath>
   </header>
   <tags>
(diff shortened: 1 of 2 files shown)


sam-m888:
And on Apple Mac the [Example.gramps](https://gramps-project.org/wiki/index.php?title=Example.gramps) is database is not included in the MacOS AIO(All in one package) . But does come with Homebrew and Macports.

sam-m888:
Thank you

emyoulation:
This is not working in 5.2 beta

prculley:
Looks ok to me. https://github.com/gramps-project/gramps/commits/maintenance/gramps52/example/gramps/example.gramps
The photos show up fine in my latest AIO.  I'm not sure you are looking at the right branch when you do the history button.  Make sure you are on maintenance/gramps52.

PR #787, opened 2019-02-12
Fix PythonDeprecationWarning for collections.abc

prculley:
This also requires a minimum Python version of 3.3. #789


Nick-Hall:
This needs to be moved to the _master_ branch.

We should also update the python version in a separate PR.

prculley:
I am concerned that one or more Gramps 5.0.x versions will still be very active when Python 3.8 becomes available on the leading edge distros in/after October (and the pre 3.3 deprecations are removed).  Which is why I set this to the gramps50 branch.  I've been periodically merging the gramps50 branch into master, so it ends up there as well.

I pulled the Python min version to #789

PR #788, opened 2019-02-12
Fix Python time.clock deprecation

prculley:
This also requires pushing our minimum Python version requirement to 3.3, but that is done in #789.


Nick-Hall:
This needs to be moved to the _master_ branch.

PR #789, opened 2019-02-13
Bump required Python version to 3.3

prculley:
To be consistent with #787 #788

### DIFF
--- a/gramps/grampsapp.py
+++ b/gramps/grampsapp.py
@@ -153,7 +153,7 @@ from .gen.mime import mime_type_is_defined
 #
 #-------------------------------------------------------------------------
 
-MIN_PYTHON_VERSION = (3, 2, 0, '', 0)
+MIN_PYTHON_VERSION = (3, 3, 0, '', 0)
 if not sys.version_info >= MIN_PYTHON_VERSION:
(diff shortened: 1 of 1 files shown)


QuLogic:
3.3 is EOL; is there any reason not to bump to 3.4 at least?

Nick-Hall:
Yes.  We should move to 3.5, but not in a maintenance branch.

sam-m888:
- [ ] Update mention in Readme

Nick-Hall:
> 3.3 is EOL; is there any reason not to bump to 3.4 at least?

@QuLogic The problem I have at the moment is that Travis CI fails when I upgrade to xenial.  Do you have any Travis experience?

Nick-Hall:
There are two problems:

1. Ubuntu xenial doesn't have the _python3-pyicu_ package, but I'm not sure that we need it in the unit tests.

2. We are getting the following error:

> Failed to connect to Mir: Failed to connect to server socket: No such file or directory
> Unable to init server: Could not connect: Connection refused

I attempted to fix it by setting the DISPLAY environment variable, but had no luck.

Nick-Hall:
The build is now passing, but pyicu is not installed and we are still getting the errors.

PR #790, opened 2019-02-23
fix finddupes tool when run from Match Threshold screen again after merge. 

prculley:
Avoids HandleError because potentials dict is not cleared and tries to display merged out person object.

Fixes [#11044](https://gramps-project.org/bugs/view.php?id=11044)


sam-m888:
Thank you

PR #791, opened 2019-02-24
Fix Graphs that use graphdoc pdf via Ghostscript with multi-page

prculley:
for poor font rendering of some characters

Fixes [#11047](https://gramps-project.org/bugs/view.php?id=11047)


Nick-Hall:
Is this the patch supplied by archont in [this post](https://sourceforge.net/p/gramps/mailman/message/36594482/)?  If so, you should probably credit him by setting him as the author.  You may also like to reference the suggestion by Ken Sharp in the commit message.

prculley:
Done.

sam-m888:
Thank you

PR #792, opened 2019-02-25
Fix relationshippathbetween filter rule when parent is missing

prculley:
Fixes [#11049](https://gramps-project.org/bugs/view.php?id=11049)

Another HandleError issue, this one hidden by the raw except.  Filter rule just did not work if a family in the path had a missing parent.


sam-m888:
Thank you

PR #793, opened 2019-02-26
Unhandled exception in Geography editor

SNoiraud:
Fixes [#11048](https://gramps-project.org/bugs/view.php?id=11048)

better pylint score
remove unused code


sam-m888:
Thank you

PR #794, opened 2019-02-26
Add option to change the date error color

SNoiraud:
Fixes [#11046](https://gramps-project.org/bugs/view.php?id=11046)

PR 775 must be applied first.

### DIFF
--- a/gramps/gui/widgets/validatedmaskedentry.py
+++ b/gramps/gui/widgets/validatedmaskedentry.py
@@ -26,6 +26,7 @@ __all__ = ["MaskedEntry", "ValidatableMaskedEntry"]
 # Standard python modules
 #
 #-------------------------------------------------------------------------
+from gramps.gen.config import config
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
 import string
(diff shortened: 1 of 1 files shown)


Nick-Hall:
We should use the standard Gtk functionality as I mention in the bug report.

SNoiraud:
I tried, but it doesn't work. The Gtk documentation is insufficient: 
"GtkEntry has a main node with the name entry. Depending on the properties of the entry, the style classes .read-only and .flat may appear. The style classes .warning and .error may also be used with entries."
I am trying to find some examples, but nothing, so I am looking in the gtk source code...

Nick-Hall:
@SNoiraud Thanks for looking at this.  I have created PR #796. There is also some same CSS in a comment explaining how users can customise the appearance to suit their own preferences.

prculley:
Nicks suggestion in the bug report worked fine for me on Windows Gtk 3.24 and 3.18.

I just replaced the 'fade' with 'error' in the set_valid and set_invalid routines, and it used the css entry.error definition (which on Adwaita is #cc0000, which is darn red, also made the entry box outline red as well).  Looked OK in both light and dark variants of the Adwaita theme.

If we want to be fancy, I'll bet we could do some css trick to fade in the 'error' color instead, got some reading to do to make that happen though...

SNoiraud:
OK. I close this PR.
The same changes didn't work for me. I suppose I forgotten something.
I remove the patch on PR775

PR #797, opened 2019-02-28
Fix Person Editor Events to properly update during external changes

prculley:
like Event delete or update.

Fixes [#11051](https://gramps-project.org/bugs/view.php?id=11051)

Looks like someone simply forgot to register the handles with the callman...

I checked other similar cases and did not see any other issues.


sam-m888:
Thank you

PR #799, opened 2019-03-01
Fix Gedcom import so it doesn't create completely empty Birth events

prculley:
Fixes [#10145](https://gramps-project.org/bugs/view.php?id=10145)

Note: a completely empty birth event is removed by Check and Repair; this is wrong in that sometimes we know a person is born, but not date/place.  In Gedcom this is indicated by "1 BIRT Y" lines.

So this PR adds a description to the birth event.  This is not an issue for other event types, the C&R tool sees the non-default type and leaves it alone, but birth is the default type.

### DIFF
--- a/data/tests/imp_sample.gramps
+++ b/data/tests/imp_sample.gramps
@@ -3,7 +3,7 @@
 "http://gramps-project.org/xml/1.7.1/grampsxml.dtd">
 <database xmlns="http://gramps-project.org/xml/1.7.1/">
   <header>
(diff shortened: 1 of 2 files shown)


sam-m888:
Hi Paul,

FYI Travis failed (failures=1) for this PR see:

https://travis-ci.org/gramps-project/gramps/builds/500431710#L2017

Kind regards
Sam

sam-m888:
Thank you

PR #800, opened 2019-03-07
Store objects as JSON rather than pickled blobs

Nick-Hall:
This is an item on the [v5.2 roadmap](https://gramps-project.org/wiki/index.php/5.2_Roadmap).  See also feature request [#9392](https://gramps-project.org/bugs/view.php?id=9392).

So far only dbapi databases have been converted for testing.  There is no upgrade yet.  A family tree must be imported into an empty database.

#### To Do:

* Database upgrade
* BSDDB backend
* Gedcom import
* Progen import
* Rebuild gender statistics
* Check and repair tool
* Media manager
* Change names tool
* Session log gramplet

#### Known Issues:

* The view tree models are passed the raw data in dictionary format instead of a string.  This involves an extra step when a object is required.
* The old serialization sometimes stores empty dates and sometimes stores None.  At the moment the JSON always contains empty dates.
* The JSON only stores the string for a GrampsType.  Perhaps we should also store the value.

### DIFF
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -57,6 +57,7 @@ from ..lib.researcher import Researcher
 from ..lib import (Tag, Media, Person, Family, Source, Citation, Event,
                    Place, Repository, Note, NameOriginType)
 from ..lib.genderstats import GenderStats
+from ..lib.serialize import from_json
 from ..config import config
 from ..const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
@@ -1258,7 +1259,7 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
             raise HandleError('Handle is empty')
         data = self._get_raw_data(obj_key, handle)
         if data:
-            return obj_class.create(data)
+            return from_json(data)
         else:
             raise HandleError('Handle %s not found' % handle)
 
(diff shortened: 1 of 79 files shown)


prculley:
When this is complete do you see a need to keep the gen.lib.object.serialize() and unserialize() methods?

Nick-Hall:
> When this is complete do you see a need to keep the gen.lib.object.serialize() and unserialize() methods?

The `unserialize()` method will be needed for database upgrades, but both methods can be deprecated and eventually removed.

PQYPLZXHGF:
@Nick-Hall This was promising, was hoping this was in 5.1 :( what remains to be done?

Nick-Hall:
The next step is to convert the BSDDB backend.  Then we have to determine if there are any performance implications.

prculley:
Spent a bit of time looking at this and tried a few performance timings 😢.
I had to make a few fixes to get it going, but when I did here are the results.
Run at full speed on my Windows installation, time is in seconds.

| test | master | json |
| ------------ | ------------- | -------------- |
| import example.gramps: | 10.0 | 12.8 |
| export XML: | 2.0 | 8.2 |
| export XML w privacy filter on | 6.2 | 27.2 |
| Check&Repair | 11.0 | 55.5 |

That is quite a bit of performance loss.  I will try some more detailed profiling later to see if anything jumps out at me.

I also looked at the memory usage peak during the Check&Repair run to see if there was a noticeable difference, Process Explorer showed a couple of extra megabytes (out of 170M or so) so maybe 1% more.  Difficult to tell for sure because the sampling was every couple of seconds and the data did not stay around long.

Nick-Hall:
@prculley Thanks for carrying out some performance tests.  The results are interesting.  I didn't expect such a performance drop considering that we no longer pickle the objects.

Unfortunately, there is a another issue.  The BSDDB backend uses DbShelf objects which will not work well with JSON.

prculley:
I just pushed a few tweaks I made to the branch to make it work and to get some better performance.
Sorry it appears to have messed up the commit list, I forgot that I had rebased onto master first. I will get master a bit more up to date, might fix this.

I ran some profile tests on person_read_from_handle to see if I could spot any potential improvements.  There were several.  The biggest was to avoid the __init__ step on all our objects, since all the data created by that step was overwritten anyway.  Obviously it was not quite that simple, but... The second improvement was to notice that setattr has a fair amount of overhead to deal with all of the potential cases.  And it was in the inner loop of the serialize __object_hook function.  I reasoned that we could perhaps just merge in the json obj_dict with the Gramps object __dict__ with a dict .update.  Again, not that simple, but I made it work.

The net result was to improve the json loads version of the Check&Repair from 55.6 sec to 13.1 sec as compared to standard master performance of 11.0 sec.  We might be able to live with this, although there may be some more tweaks available I have not found yet.

Regarding the BSDDB; do we need to continue support for BSDDB?  The upgrade process could just move us to SQLITE/JSON and be done with it.  Food for thought.

Nick-Hall:
Closed in favour of pull request #1786.

PR #803, opened 2019-03-10
Update Croatian translation

milotype:
update with current maint50 pot file


Nick-Hall:
Thanks.  That looks much better!

sam-m888:
Thank you

PR #807, opened 2019-03-30
Fix failure after Gedcom import if missing objects were found

prculley:
Fixes [#11082](https://gramps-project.org/bugs/view.php?id=11082)


sam-m888:
Thank you

PR #808, opened 2019-04-07
 Properly detect missing birth and death years.

jelmer:
Without this change, birth and death dates get set to "0" in the fan chart, if missing.

(creating new pull request against maintenance/gramps50)


jelmer:
Unfortunately I can no longer reproduce the issue; I'm very sure I was originally able to easily reproduce it, but perhaps the underlying bug in gramps has since been fixed.

sam-m888:
Thank you for checking.

PR #811, opened 2019-04-14
Avoid invalid characters in the entry fields

SNoiraud:
In this patch:
 1 - If you copy/paste strings from another application,
     you can add CR, LF, TAB and other special characters. I remove them.
 2 - suppress all leading and trailing spaces for these entry field.

This PR modify the MonitoredEntry class used by all views and the SurnameTab used in the person editor and in the EditName class.

### DIFF
--- a/gramps/gui/editors/displaytabs/surnametab.py
+++ b/gramps/gui/editors/displaytabs/surnametab.py
@@ -50,6 +50,9 @@ from ...ddtargets import DdTargets
 from gramps.gen.lib import Surname, NameOriginType
 from ...utils import match_primary_mask, no_match_primary_mask
 
+# table for skipping illegal control chars
+INVISIBLE = dict.fromkeys(list(range(32)))
+
 #-------------------------------------------------------------------------
 #
 # SurnameTab
@@ -281,7 +284,8 @@ class SurnameTab(EmbeddedList):
         colnr must be the column in the model.
         """
         node = self.model.get_iter(path)
-        self.model.set_value(node, colnr, new_text)
+        text = new_text.translate(INVISIBLE).strip()
+        self.model.set_value(node, colnr, text)
         self.update()
 
     def on_orig_edited(self, cellr, path, new_text, colnr):
--- a/gramps/gui/widgets/monitoredwidgets.py
+++ b/gramps/gui/widgets/monitoredwidgets.py
@@ -64,6 +64,9 @@ from gramps.gen.errors import ValidationError
 _RETURN = Gdk.keyval_from_name("Return")
 _KP_ENTER = Gdk.keyval_from_name("KP_Enter")
 
+# table for skipping illegal control chars
+INVISIBLE = dict.fromkeys(list(range(32)))
+
 #-------------------------------------------------------------------------
 #
 # MonitoredCheckbox class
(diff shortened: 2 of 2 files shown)


lordemannd:
testing, looks good so far, now more tests

lordemannd:
testing looks good, I am a little uncertain about the no-break space.
but right now, it might be working as good as possible.
the no-break space isn't currently being removed, but this might be just right,
the no-break space can be leading, trailing or embedded in the text and they are left intact.

(of course this doesn't do anything about the reported issue of the no-break space being indistinguishable from a standard space, except that now they will not be removed)

SNoiraud:
> (of course this doesn't do anything about the reported issue of the no-break space being indistinguishable from a standard space, except that now they will not be removed)

The strip function remove leading and trailing no-break spaces.

lordemannd:
working with the no-break space is a pain.

change the attached backup file from .txt to .gramps

create a new DB and import this backup file. 
notice that some of the fields are handled differently in the name editor, give name vs prefix.

[test-PR-881.txt](https://github.com/gramps-project/gramps/files/3104428/test-PR-881.txt)

SNoiraud:
It works as expected: only leading and trailing no-break spaces are removed.

lordemannd:
I am using Win 10, see the attached file, I can highlight the no-break space. maybe windows is different?

Also notice that after taking the snapshot, the leading/trailing no-break spaces are removed.
So result is, to get them all removed, I have to go through all of the fields, clicking on each one, it looks like the edit happens when the field losses focus. Then I have do to the same with each of the alternate names, stepping through each field.
ScreenHunter_257 Apr  22 14 58

SNoiraud:
Yes, for old data, when you edit them, you need to select the field to modify then select <TAB>.
In this case, the new tool is necessary to find these old data to modify ( #783 )
For new data, it is done automatically when you quit a field.
In reality, when you quit the field, the strip function is applied.
When you enter characters in a field, for each character, we filter special characters.

lordemannd:
This cleans up when the user enters the data, can it also be used when data is imported? 
(from GEDCOM, gramps XML backup, etc.)
I mean is the MonitoredEntry class used during import?

Or does the new tool ( #783 ) need to be used after the import to clean up these bad characters?

(would it be nice if we could report to the log file when characters are removed from the user data?)

SNoiraud:
> I mean is the MonitoredEntry class used during import?

NO. It's only in the GUI editors

SNoiraud:
> Or does the new tool ( #783 ) need to be used after the import to clean up these bad characters?

YES

> (would it be nice if we could report to the log file when characters are removed from the user data?)

The tool gives a list of bad records (person or place). You need to edit these records. This is a manual edit.
So, I am not sure a log will be very useful.

Nick-Hall:
> I don't really understand the implication of the no-break space in Arabic, is the leading/trailing no-break space needed? or is it enough to preserve the ones embedded in the middle of the text string?
> Nick-Hall do you know the answer?

@lordemannd You missed the @ off my username, so I have only just seen this.

In Arabic a single name can consist of two words.  Using just one doen't make any sense.  So a no-break space is placed between two words to indicate that they are a single name.  This is useful in the Given Names Cloud gramplet.

So it is safe to strip off leading and trailing no-break spaces.

sam-m888:

SNoiraud:
Yes

sam-m888:
Thank you

PR #812, opened 2019-04-21
Avoid comma in a lat/lon entry field.

SNoiraud:
Fixes [#11114](https://gramps-project.org/bugs/view.php?id=11114)

### DIFF
--- a/gramps/gui/editors/editplace.py
+++ b/gramps/gui/editors/editplace.py
@@ -195,6 +195,11 @@ class EditPlace(EditPrimary):
         return ""
 
     def _validate_coordinate(self, widget, text, typedeg):
+        if text.find(',') != -1:
+            # coordinates must be in english format. 43,2545 is not allowed.
+            return ValidationError(
+                _('Invalid latitude\n(syntax: '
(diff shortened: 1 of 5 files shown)


gramps/gui/editors/editplace.py
```diff
@@ -195,6 +195,11 @@ def get_latlongitude(self):
         return ""
 
     def _validate_coordinate(self, widget, text, typedeg):
+        if text.find(',') != -1:
+            # coordinates must be in english format. 43,2545 is not allowed.
+            return ValidationError(
+                _('Invalid latitude\n(syntax: '
```
QuLogic: But it should say `longitude` for longitudes? Maybe add this condition to the below checks instead.

SNoiraud: conv_lat_lon call __convert_float_val which convert the comma. I think this is the solution.
needs more tests because conv_lat_lon is used in many places. If we have already commas in lat/lon, we could have problems.

SNoiraud:
If we have incorrect lat/lon format, We have the following in the places view:
bug11114 1

sam-m888:

SNoiraud:
Yes

sam-m888:
Thank you

PR #813, opened 2019-05-10
Add option to show the Gramps ID in parenthesis in the fan chart

SNoiraud:
Fixes [#11045](https://gramps-project.org/bugs/view.php?id=11045)


Nick-Hall:
@SNoiraud Please rebase branches instead of merging them.  A linear history makes it easier to bisect for bugs.

You probably also want to combine your two pylint commits.

PR #814, opened 2019-05-13
Fix crash for multiple deletes in one transaction

prculley:
Fixes [#11117](https://gramps-project.org/bugs/view.php?id=11117)

Found while trying to undo a GetGOV load that brought in several enclosed places.
The basic problem is that several parts of our code were trying to handle deletes one handle at a time after the db had deleted the actual objects and created the multiple delete signal.  For the MRU update code this resulted in attempting to rebuild MRU entries on other already deleted handles.  For the listview code, the one at a time deletion indirectly caused active changed to rows/handles that were also already gone.  Both resulted in HandleError crashes.

### DIFF
--- a/gramps/gui/displaystate.py
+++ b/gramps/gui/displaystate.py
@@ -116,30 +116,6 @@ class History(Callback):
             if initial_person:
                 self.push(initial_person.get_handle())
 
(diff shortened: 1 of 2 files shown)


sam-m888:
Thank you

PR #816, opened 2019-05-16
Fix uimanager, menu translations

prculley:
Fixes [#11151](https://gramps-project.org/bugs/view.php?id=11151)

The new uimanager was not translating menu and other items created with Gtk.Builder fragments.
This corrects that.


sam-m888:
Thank you

PR #817, opened 2019-05-17
Allow an option to not use hexagons for unknown gender

edbennett:
This pull request adds an option to the relationship graph report to not use hexagons to highlight unknown gender individuals.

Resolves [#11112](https://gramps-project.org/bugs/view.php?id=11112)


Nick-Hall:
As this is such a simple enhancement and only affects one plugin, I think we can make an exception and include it in the next release.

edbennett:
Great news, thanks!

PR #818, opened 2019-05-18
Correct English translation strings for unmarried partners

edbennett:
Previously, unmarried male and female partners were referred to as "husband" and "wife" respectively in English, although other languages had correct translations. This PR adjusts this by adjusting the relevant translation strings so that the correct English is after the pipe, with additional disambiguating information before the pipe to allow the translations to keep working.

This fixes [#11109](https://gramps-project.org/bugs/view.php?id=11109)


sam-m888:
Thank you

PR #819, opened 2019-05-22
Fix '_havechildren.py' person filter

Mattkmmr:
Check all families of a person for children and not only the first family:
Right now, if a person has two families and no children in the first family, this person is filtered as having no children, even if there are children in the second family.

### DIFF
--- a/gramps/gen/filters/rules/person/_havechildren.py
+++ b/gramps/gen/filters/rules/person/_havechildren.py
@@ -2,6 +2,7 @@
 # Gramps - a GTK+/GNOME based genealogy program
 #
 # Copyright (C) 2002-2006  Donald N. Allingham
+# Copyright (C) 2019  Matthias Kemmer
 #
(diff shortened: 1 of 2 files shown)


Mattkmmr:
I saw that the test for the _HaveChildren_ rule fails:
```
FAIL: Test HaveChildren rule.
  File "/home/travis/build/gramps-project/gramps/gramps/gen/filters/rules/test/person_rules_test.py", line 578, in test_havechildren
    self.assertEqual(len(self.filter_with_rule(rule)), 901)
nose.proxy.AssertionError: 905 != 901
```
I guess the following four persons are the reason why the test fails, all four have two families and no children in the first family:
* I00245
* I01355
* I00094
* I02110

Nick-Hall:
@Mattkmmr Thanks for the fix.  Please also update the unit test.

sam-m888:
Thank you

PR #820, opened 2019-05-23
Fix '_hasattributebase.py'

Mattkmmr:
Check all values of an attribute type. If a person has several attributes with the same type/key the current behaviour is that only the first value is checked.

Example Person X Atrributes:
* (Key=occupation: Value=farmer)
* (Key=occupation: Value=soldier)
* (Key=occupation: Value=smith)

=> Currently: Filter 'persons with attribut' (Key=occupation, Value=farmer) returns Person X, but no return of Perrson X for (Key=occupation, Value=soldier) or (Key=occupation: Value=smith).

=> After fix: Person X is returned for all three filter searches.


sam-m888:
Thank you

PR #821, opened 2019-06-01
Add 'HasAttribute' filter rule to citations, repositories and sources

Mattkmmr:
Fixes [#0009845](https://gramps-project.org/bugs/view.php?id=9845): Add a filter on attribute for Sources

### DIFF
--- a/gramps/gen/filters/rules/citation/_hasattribute.py
+++ b/gramps/gen/filters/rules/citation/_hasattribute.py
@@ -1,0 +1,49 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2008  Gary Burton
+# Copyright (C) 2019  Matthias Kemmer
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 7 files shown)


gramps/gen/filters/rules/citation/_hasattribute.py
```diff
@@ -0,0 +1,49 @@
...
+from .._hasattributebase import HasAttributeBase
+
+
+# -------------------------------------------------------------------------
+#
+# HasAttribute
+#
+# -------------------------------------------------------------------------
+class HasAttribute(HasAttributeBase):
+    """Rule that checks for a media object with a particular attribute"""
```
sam-m888: > """Rule that checks for a media object with a particular attribute"""

Copy/paste error, should the doc strings in all four be identical?

Mattkmmr: No that is a copy/paste error.

prculley:
As of this moment, Places don't have attributes, so an attempt to use that filter should result in a crash.
If/when GEPS045 is accepted, places will have attributes...

Mattkmmr:
@prculley will you add 'HasAttribute' for places with GEPS045 or should I put it on my ToDo list?

prculley:
> 
> 
> @prculley will you add 'HasAttribute' for places with GEPS045 or should I put it on my ToDo list?

I already have it in my code base (but not pushed lately).

PR #823, opened 2019-06-07
Fix metadataviewer.py

Mattkmmr:
Fix so that metadata is actually detected and displayed (signal was not getting detected).

https://gramps-project.org/bugs/view.php?id=6606#c56776


prculley:
We should probably apply this to gramps50 branch since if fixes a bug.  Need to wait a week or so according to our rules to accept, but it works.

sam-m888:
Thank you

sam-m888:
Cherry-picked to gramps50 as https://github.com/gramps-project/gramps/commit/1b7b20756f200c4a1430bb7c95de90c8cda9280f

Thanks you

PR #828, opened 2019-06-24
Update croatian hr.po file

milotype:
- update with newest/current .pot file found in maint 51 branch
- translation is complete
- corrections in wording and some missing accelerators

### DIFF
--- a/po/hr.po
+++ b/po/hr.po
@@ -8,8 +8,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Gramps 5.0.1\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2018-12-26 11:20-0600\n"
-"PO-Revision-Date: 2019-06-09 23:23+0200\n"
+"POT-Creation-Date: 2019-06-13 17:22-0500\n"
+"PO-Revision-Date: 2019-06-24 19:45+0200\n"
 "Last-Translator: Milo Ivir <mail@milotype.de>\n"
 "Language-Team: \n"
 "Language: hr\n"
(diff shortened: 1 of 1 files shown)


milotype:
I have the feeling, this isn't going to pass and unfortunately I don't know why ... I thought one should always use the highest maintainance branch ...
Do I have to make a new PR to maint 50, or can some move it to the right place?

Nick-Hall:
This is the correct branch now, although we plan do do one final _gramps50_ release.

milotype:
The checks still haven't passed for my commit. Can someone inspect and resolve it. I'm no programmer and don't know how to proceed. Thank you.

prculley:
It looks like the most important check (travis) is passing.  The Codacy seems not to be enabled for gramps51 branch as of yet; even if it was, it is basically a pylint run and for most of our submissions will never completely pass.  I like to look it over to see if there are obvious fixes that can be done.  But a lot of the complaints it has seem to be things that are not easily dealt with (too many whatever, import order etc.).  In some cases the complaints are due to the way Gramps has traditionally done things, and should either be changed everywhere, or left alone (which I am doing).

milotype:
Hmmm … bit too technical for me … should I:
a) do nothing, and things will get resolved at some point
b) make a merge request for the same file into maint50

prculley:
If we do another release on gramps50 branch (Nick says we will) then we will also merge up to gramps51 from there to make sure that bug fixes get brought forward.  Gramps51 and master branches contain new strings (and probably remove some strings that were in gramps50) so it would generally NOT be a good idea to simply merge down from gramps51 or to simply apply the gramps51 file to gramps50.

If you have a tool that can show important changes made to create your latest translations (not just line numbers), then you might use that as a guide to make similar changes to the gramps50 version, being careful not to remove any strings.

Nick-Hall:
> If we do another release on gramps50 branch (Nick says we will)

@prculley You can go ahead with the final v5.0 release when you have time.

prculley:
@Nick-Hall I think we should get https://gramps-project.org/bugs/view.php?id=11195 (which I sent an email about) fixed up first.  But I need advise on how to go to fix it.

Nick-Hall:
@milotype Sorry that we forgot to merge this earlier.  It will be available as part of v5.1.1 shortly.

PR #829, opened 2019-06-26
OSM forward all http resquest to https.

SNoiraud:
After asking the OSM support, I was told OSM forward all http request to
https when we use .org url.
So it should work in all cases.

This has an energy cost, so it would be good for the planet to change http to https.

Fixes [#11057](https://gramps-project.org/bugs/view.php?id=11057)


sam-m888:
Thank you

PR #830, opened 2019-06-26
Geography: pins very big when related to 2 places

SNoiraud:
it's not really a bug, but I think people will have a better rendering when we have two or three places or events for a single pin.

Fixes [#10644](https://gramps-project.org/bugs/view.php?id=10644)


sam-m888:
Thank you

PR #831, opened 2019-06-27
Fix duplicated accelerators

prculley:
This changes the accelerator keys for the Print graph command used in some of the chart views from the original `<PRIMARY>P` to `<PRIMARY><SHIFT>P`.  This avoids a conflict with the `<PRIMARY>P` used for the Previous View.

I noticed that the `<alt>P` mnemonic in the Edit menu was also duplicated between Print and Preferences and have removed that mnemonic for printing.  This doesn't create any new translations as we already have a "Print..." in our gramps.pot.

This also removes the extraneous accelerator `<PRIMARY>q` (for Quit) on OSX only, as that OS already had the same key combination assigned to 'Quit' by the OS.

Fixes [#11186](https://gramps-project.org/bugs/view.php?id=11186)

### DIFF
--- a/gramps/gui/viewmanager.py
+++ b/gramps/gui/viewmanager.py
@@ -353,7 +353,7 @@ class ViewManager(CLIManager):
         Initialize the actions lists for the UIManager
         """
         self._app_actionlist = [
-            ('quit', self.quit, "<PRIMARY>q"),
+            ('quit', self.quit, None if is_quartz() else "<PRIMARY>q"),
             ('preferences', self.preferences_activate),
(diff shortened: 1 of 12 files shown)


Nick-Hall:
We shouldn't change the accelerator keys for Previous and Next.  They are widely used and have been in the product for years.  Users will be familiar with them.

Choose a new key for the newer print functionality.

prculley:
Updated per your comments (and first comment edited).

Nick-Hall:
@prculley Thanks.  It looks good now.

sam-m888:
Thank you

PR #834, opened 2019-07-02
familygroup.py:  Rename husband/wife

Mattkmmr:
- Rename husband/wife to father/mother
- Fixes [#11199](https://gramps-project.org/bugs/view.php?id=11199)

### DIFF
--- a/gramps/plugins/textreport/familygroup.py
+++ b/gramps/plugins/textreport/familygroup.py
@@ -624,7 +624,7 @@ class FamilyGroup(Report):
         self.doc.write_text('', mark)
         self.doc.end_paragraph()
 
-        self.dump_parent(self._("Husband"), family.get_father_handle())
+        self.dump_parent(self._("Father"), family.get_father_handle())
         self.doc.start_paragraph("FGR-blank")
         self.doc.end_paragraph()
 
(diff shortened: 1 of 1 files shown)


Nick-Hall:
I notice that the original reporter of the bug disagrees with this solution.

Also, @romjerome proposed something similar in PR #347.

Related bug reports:  [#3146](https://gramps-project.org/bugs/view.php?id=3146) and [#9899](https://gramps-project.org/bugs/view.php?id=9899).

This probably need to be discussed on the mailing list.

Mattkmmr:
@Nick-Hall This pull is just intended to be a small fix to remove some confusion by renaming. Showing the right relationship role is the better solution, but I'm currently too busy to work on it.

Nick-Hall:
Why don't we just remove these labels?  They don't seem to add much to the report.

This would avoid complications with relationship types, same-sex relationships and couples without children.

romjerome:
> Why don't we just remove these labels? They don't seem to add much to the report.
>
>This would avoid complications with relationship types, 
> same-sex relationships and couples without children.

:+1:

Nick-Hall:
Replaced by PR #935.

PR #835, opened 2019-07-02
Fix zoom via mouse wheel in media event editor selectionwidget

prculley:
also fixes zooming with scroll bars always present

Fixes [#11198](https://gramps-project.org/bugs/view.php?id=11198)

### DIFF
--- a/gramps/gui/widgets/selectionwidget.py
+++ b/gramps/gui/widgets/selectionwidget.py
@@ -297,7 +297,7 @@ class SelectionWidget(Gtk.ScrolledWindow):
             self.original_image_size = (self.pixbuf.get_width(),
                                         self.pixbuf.get_height())
 
-            viewport_size = self.viewport.get_allocation()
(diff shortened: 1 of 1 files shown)


lordemannd:
zoom part works for me (debian,GTK+ 3.24.5)
but I notice strangness, an artifact at bottom left edge when a region of the image is selected.
- add some image
- select a region near right edge of image (like a face)
- area outside the selected region is greyed out.
- except for bottom left edge, continuing to the left edge of the selected region (face)

prculley:
@lordemannd There does seem to be a few float/int rounding errors that result in 1 pixel issues.  I added a pixel to the shading region, and it seems to fix it for my system.  There were several other such adjustments in the original code, I guess the author either missed or did not see this one.

lordemannd:
checked it on windows and debian, looks good.

sam-m888:
Thank you

PR #836, opened 2019-07-03
Fix crash when addon/plugin contains an id with space

prculley:
Fixes [#11202](https://gramps-project.org/bugs/view.php?id=11202)

The updated uimanager and managedwindow uses the plugin id to create actions for plugins that make a window to support the menus. If the plugin id contains a space, Gramps crashes.


sam-m888:
Thank you

PR #840, opened 2019-07-07
Narrative web: new features

SNoiraud:
- Add Stamen map (#05984)
    new maps addes: watercolor, toner and terrain

- Add popup  to manage markers (#11150)
    If you click on a marker in the family map page, you get a popup.
    In this popup associated with a place, if you have several events,
    for each event you see the person and the event type.
    If you click on the person, you go to the related page for this person.
    If you click on the event type, you go to the related page for this event.
    
Fixes #05984, #11150

PR #841, opened 2019-07-07
Narrative web new features

SNoiraud:
This PR follow #840

- Add Stamen map (#05984)
   new maps addes: watercolor, toner and terrain

- Add popup to manage markers (#11150)
    If you click on a marker in the family map page, you get a popup.
    In this popup associated with a place, if you have several events,
    for each event you see the person and the event type.
    If you click on the person, you go to the related page for this person.
    If you click on the event type, you go to the related page for this event.

- Don't link to event page if we don't include them.

Fixes #05984, #11150


SNoiraud:
I don't really understand what happen.
Why do I have all these old patches ?

I tried git rebase, git merge, ... without success. What is the problem ?

SNoiraud:
See next one which solves all this #842.

PR #846, opened 2019-07-11
Update Portuguese (Europe) translation

pmraps:


sam-m888:
Is this translation file actually meant for the gramps51 branch instead of Gramps master branch?

Nick-Hall:
Rebased onto the _gramps51_ branch and merged into the latest template.

PR #851, opened 2019-07-17
Add a new person filter rule 'HasAddressText'

Mattkmmr:
Allows users to filter persons who have an address containing a given text.
Fixes [#9168](https://gramps-project.org/bugs/view.php?id=9168)

### DIFF
--- a/gramps/gen/filters/rules/person/_hasaddresstext.py
+++ b/gramps/gen/filters/rules/person/_hasaddresstext.py
@@ -1,0 +1,54 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2008  Brian G. Matherly
+# Copyright (C) 2008  Jerome Rapinat
+# Copyright (C) 2008  Benny Malengier
+#
+# This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


gramps/gen/filters/rules/person/_hasaddresstext.py
```diff
@@ -0,0 +1,55 @@
...
+
+# -------------------------------------------------------------------------
+#
+# HasAddressText
+#
+# -------------------------------------------------------------------------
+class HasAddressText(Rule):
+    """Rule that checks for a string in personal addresses"""
+
+    labels = [_('Adress:')]
```
Nick-Hall: Please fix spelling.

prculley:
Are you aware that the rule "People with records containing substring" already searches in the person's addresses?  Or did you add this because it is faster than the more general rule?

Mattkmmr:
My thinking was that the "People with records containing substring" rule gives back many results which mached other fields than adresses, especially when you don't know the exact adress. Then you have to edit/open each person to check if the adress was matched or something else, which could be avoided if only the adress fields were checked in a new filter.

Nick-Hall:
Yes, it seems that a separate rule may be useful.

PR #856, opened 2019-07-21
Narrative web: Sort problem with places.

SNoiraud:
Fixes [#10868](https://gramps-project.org/bugs/view.php?id=10868).

Places list is not sorted depending on the selected language.

If you start gramps in english or another language then try to generated
a narrative web report in another language, the navigation alphabet is
incorrect. This is true for the place list and the person list.
The problem was related to pyICU.


lordemannd:
looks like the 5.1 version works as well on windows.

PR #860, opened 2019-07-24
Maintenance/gramps51

astenix:
Ver 5.0.1

It will be useful to add a button [Today] to the day picker interface?

### DIFF
--- a/gramps/gui/managedwindow.py
+++ b/gramps/gui/managedwindow.py
@@ -280,7 +280,7 @@ class GrampsWindowManager:
         return func
 
     def generate_id(self, item):
-        return 'wm/' + str(item.window_id)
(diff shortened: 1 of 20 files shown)


prculley:
Not sure what you are doing here.  We periodically merge gramps50 and gramps51 up into master, so it doesn't take a PR to ask for that.  The question you asked and the 5.0.1 comment don't seem to make sense in the context of the PR.

astenix:
> Not sure what you are doing here. We periodically merge gramps50 and gramps51 up into master, so it doesn't take a PR to ask for that. The question you asked and the 5.0.1 comment don't seem to make sense in the context of the PR.

Sorry, my fault.

PR #861, opened 2019-07-31
Fix GEDCOM import for better support of TMG

prculley:
Fixes [#11224](https://gramps-project.org/bugs/view.php?id=11224)

A users TMG GEDCOM file caused Gramps to crash because it put a DATE structure in an unexpected place which ended up in a Gramps.gen.date object in an event.description (string) field.  SQLITE tries to put the description field in a db column, it did not like the structure in the string field.

It is not clear why BSDDB did not object, but it looks like somewhere in the BSDDB code the object was converted to string.

TMG was using the REFN tag in a decidedly non-standard way which was the root cause of the issue.

In any event, I modified the offending GEDCOM import code to make sure that the situation would be properly recognized as an invalid GEDCOM sequence.

TMG GEDCOM also used the _SDATE GEDCOM tag extensively, so I added it as a synonym for a normal DATE.

### DIFF
--- a/data/tests/imp_notetest_dfs.gramps
+++ b/data/tests/imp_notetest_dfs.gramps
@@ -3,7 +3,7 @@
 "http://gramps-project.org/xml/1.7.1/grampsxml.dtd">
 <database xmlns="http://gramps-project.org/xml/1.7.1/">
   <header>
-    <created date="2016-10-23" version="5.0.0-alpha1"/>
+    <created date="2019-07-30" version="5.0.2"/>
     <researcher>
       <resname>John A. Tester</resname>
     </researcher>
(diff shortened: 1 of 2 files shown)


lordemannd:
confirmed that GEDCOM/GEDCOM2 imports with BSDDB and SQlite DB still work without ErrRptWiz showing up. (but don't have a TMG GEDCOM to test with)

johnbickmore:
> TMG GEDCOM also used the _SDATE GEDCOM tag extensively, so I added it as a synonym for a normal DATE.

@prculley I don't believe that to be a good idea as normally it only used for sorting the event for display in TMG 👎 

> Sort dates
> 
> Sort dates are an advanced TMG feature. While they always exist within TMG, they can only be explictly entered/modified when in TMG’s Advanced (versus Beginner) edit mode. This enhanced export option will export these sort date values as a custom GEDCOM tag type following any tag’s standard DATE tag or any custom _DATE tag added by this enhanced export option.
> 
> 2 DATE date
> 
> 2 _SDATE sort date
> 
> or
> 
> 2 _DATE date
> 
> 2 _SDATE sort date 
> 
> https://www.mjh-nm.net/GED-EXPT.HTML

johnbickmore:
> but don't have a TMG GEDCOM to test with

@lordemannd I'll see if I can work out how to add one here for you.

PR #863, opened 2019-07-31
Fix Edit Link 'New' button to work

prculley:
Fixes [#11225](https://gramps-project.org/bugs/view.php?id=11225)

Fixed a typo


lordemannd:
worked for me on Windows with 5.0.1 (plus current changes on gramps50)

PR #865, opened 2019-08-03
Fix some Gramplets not updating during tree changes after db change.

prculley:
Fixes [#11234](https://gramps-project.org/bugs/view.php?id=11234)

### DIFF
--- a/gramps/gen/plug/_gramplet.py
+++ b/gramps/gen/plug/_gramplet.py
@@ -33,8 +33,10 @@ LOG = logging.getLogger(".Gramplets")
 # Gramps modules
 #
 #------------------------------------------------------------------------
+from ...gui.dbguielement import DbGUIElement
 from ..const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
+
 
 class Gramplet:
     """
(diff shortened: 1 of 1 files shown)


gramps/gen/plug/_gramplet.py
```diff
@@ -397,6 +397,12 @@ def _db_changed(self, db):
             if hasattr(self.gui, 'navtypes') and 'Person' in self.gui.navtypes:
                 self.connect_signal('Person', self._active_changed)
             self.db_changed()
+        # Some Gramplets use DbGUIElement; and DbGUIElement needs to know if
+        # db is changed.  However, at initialization, DbGUIElement is not yet
+        # initialized when _db_changed is called, thus the test for callman
+        if hasattr(self, "callman"):  # get DbGUIElement informed if in use
```
Nick-Hall: This looks error prone.   Perhaps we should call the `_change_db` method in the gramplet, or test that the gramplet is an instance of `DbGUIElement`.  This may involve changing the order of the multiple inheritance.

prculley: I actually spent more time on this than I thought it would warrant.  I started out with a test for DbGUIElement, only to discover that 'callman' had not been initialized when change db was first called.  Then I considered swapping the order of initialization for Gramplet and DbGUIElement, only to discover that would not work because DbGUIElement depended on something from Gramplet.

A 'cleaner' approach would mean making several lines of change (adding these same lines in a change db method) in each of the users of DbGUIElement and Gramplet; several files.

But I thought I would try to sell this first...
If just adding a test for DbGUIElement would satisfy, I'll do that.  Just let me know.

Nick-Hall:
@prculley Thanks.  That looks safer.

PR #866, opened 2019-08-04
Add support for GEDCOM import _FREL/_MREL tags in INDI/FAMC

prculley:
Improve support for GEDCOM export of _FREL/_MREL in INDI/FAMC

Fixes [#10750](https://gramps-project.org/bugs/view.php?id=10750)

A user noted that Gramps could not import its own exported GED file properly when child relationships were different for mother and father.  We have export support for that case, but did not have import support.

This PR allows us to properly support AGES! and a few other programs GEDCOM files.

The export support was enhanced to export all relationship types, the import support was simply added.  Tests were modified to improve test coverage to include these cases, as a side effect, some of the notes in test files have different line numbers.

Found a couple of non-functional lines in the code and removed them while I was analyzing the support.

### DIFF
--- a/data/tests/exp_sample.gramps
+++ b/data/tests/exp_sample.gramps
@@ -1278,7 +1278,7 @@
       <rel type="Unknown"/>
       <father hlink="_0000000b0000000b"/>
       <mother hlink="_0000002600000026"/>
-      <childref hlink="_0000006f0000006f"/>
+      <childref hlink="_0000006f0000006f" frel="Adopted" mrel="Foster"/>
     </family>
     <family handle="_0000001000000010" change="1198222526" id="F0014">
       <rel type="Unknown"/>
--- a/data/tests/exp_sample_ged.ged
+++ b/data/tests/exp_sample_ged.ged
@@ -2,8 +2,8 @@
 1 SOUR Gramps
 2 VERS 5.0.2
 2 NAME Gramps
-1 DATE 31 JUL 2019
-2 TIME 10:58:33
+1 DATE 4 AUG 2019
+2 TIME 15:26:44
 1 SUBM @SUBM@
 1 FILE C:\Users\prc\AppData\Roaming\gramps\temp\exp_sample_ged.ged
 1 COPR Copyright (c) 2019 Alex Roitman,,,.
@@ -629,8 +629,12 @@
 2 TYPE Birth of Lloyd Smith
 2 DATE 13 MAR 1935
 2 PLAC San Francisco, San Francisco Co., CA
+1 ADOP Y
+2 FAMC @F0009@
+3 ADOP HUSB
 1 FAMC @F0009@
-2 PEDI birth
+2 _FREL Adopted
+2 _MREL Foster
 1 FAMS @F0008@
 1 CHAN
 2 DATE 4 SEP 2016
@@ -1040,7 +1044,7 @@
 2 SURN Tester
 1 SEX M
 1 FAMC @F0016@
-2 PEDI Unknown
+2 PEDI Sponsored
 1 CHAN
 2 DATE 29 OCT 2016
 3 TIME 16:27:59
--- a/data/tests/imp_sample.ged
+++ b/data/tests/imp_sample.ged
@@ -83,6 +83,8 @@
 2 DATE 11 AUG 1966
 2 PLAC San Francisco, San Francisco Co., CA
 1 FAMC @F8@
+2 _FREL Adopted
+2 _MREL Foster
 1 CHAN
 2 DATE 21 DEC 2007
 3 TIME 01:35:26
--- a/data/tests/imp_sample.gramps
+++ b/data/tests/imp_sample.gramps
@@ -3,7 +3,7 @@
 "http://gramps-project.org/xml/1.7.1/grampsxml.dtd">
 <database xmlns="http://gramps-project.org/xml/1.7.1/">
   <header>
-    <created date="2019-07-31" version="5.0.2"/>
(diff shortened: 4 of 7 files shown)


lordemannd:
no errors reported on import
but there is a small mapping difference:
Original    -     After import
Birth             -> Birth
Adopted      -> Adopted
Foster          -> Foster
None           -> none 
Sponsored  -> sponsored
Stepchild    -> Adopted
Unknown    -> unknown
user def      -> user def

So Stepchild got converted to Adopted.
And Capital letter dropped on None, Sponsored, Unknown

prculley:
> So Stepchild got converted to Adopted.
And Capital letter dropped on None, Sponsored, Unknown

Fixed

lordemannd:
Looks good.

Nick-Hall:
This is an enhancement rather than a bug and was raised a a feature request, so it should really be applied to the _master_ branch.

Our original policy was that the GEDCOM import/export should only support features in the official specification.  This seems to have changed recently with an attempt to support all widely used extensions.  However, this has the problem that they are not properly documented and can vary between software.

In the future, I think that we should aim to support the FHISO extanded legacy format (ELF).   However, this hasn't been discussed on the _gramps-devel_ list yet as far as I am aware.

lordemannd:
Can I (weakly) make the case it is a bug in the sense that the _FREL and _MREL tags were being supported on export of a GEDCOM file, but import of that same file didn't correctly handle them?

romjerome:
[GedcomExtensions](https://github.com/gramps-project/addons-source/tree/master/GedcomExtensions) addon (and [alternate versions](https://github.com/grocanar/GedcomforGeneanet)) could already be able to handle extensions on export whatever branches?

Nick-Hall:
Yes.  The policy was that core Gramps should import/export according to the GEDCOM 5.5.1 specification and that peculiarities of individual software should be handled by extensions.

prculley:
I  have been responsible for most of the GEDCOM changes.  I only recently payed any attention to the GEDCOM extensions addon.  And I certainly never saw a 'policy' on how to deal with extensions.  I just noted that we had a fair amount of non-standard tag support already in our importer when I joined the developers.

I certainly agree that managing new **export** extensions with the GEDCOM extensions addon is possible, although I suspect that there might be a maintenance issue if we use it a lot.  At the moment the addon only adds two small export functions, both enabled via GUI.

I am NOT easily convinced that we should put import extensions there.  Often these additions require rework of multiple methods and the constant tables; this would make it necessary to duplicate a lot of functionality in the addon, with associated maintenance problems.

New users importing a GEDCOM for the first time would probably not be aware to try an addon to see if their GEDCOM imported better when renamed *.ged2.

In any event, with almost full coverage of the GEDCOM import/export in our tests, adding new import functionality is low risk.

For these reasons I advocate extending our importer as specific needs arise via bug report or user request.

As for export extensions I am a bit concerned about splitting functionality into a addon and the core.  But could make it work either way.

I do wonder if we need to make a policy about making extensions to export under control of the GUI.  On one hand most users would not know much about GEDCOM and probably would not understand why they might want to use of non-standard extensions.  On the other hand, not all programs would understand all of the extensions.  So would be likely to get those programs equivalent of our import "not understood" messages.

The particular tags in this PR are only used in split parent adoptions and similar, so don't get exported very often.  That would NOT be true for the upcoming GEDCOM-L place extensions, which would add a set of new tags/lines for each place.

romjerome:
As far as I can remember issues related to gedcom indempotency (does this set/couple of words exist?) was possible on gramps 1.x. As event, citation and place as top objects cannot really match a model based on individual, since gramps 2.x, some devs worked to improve and to fix gedcom support (import/export), some [limitations have been listed](https://gramps-project.org/wiki/index.php/Gramps_and_GEDCOM) and [users should be warned](https://gramps-project.org/wiki/index.php/Gramps_5.1_Wiki_Manual_-_Getting_started#GEDCOM). 

Anyway, even with the [report environment after a gedcom import](https://gramps-project.org/wiki/index.php/Gramps_5.1_Wiki_Manual_-_Manage_Family_Trees#Reading_the_report), we always get new extensions or handling methods (sometimes the same program generating a gedcom can produce a different file according its version ...). So, a gedcom generated by gramps 5.0.0 will use multiple ways (sometimes fixed ways) on import if we are using gramps 5.0.0 or gramps 5.0.2? We have the same type (_in worst_...) with some commercial products (e.g., see some TMG exceptions on code) and even open model (like geneanet, using geneweb engine) are using a flexible custom 'gedcom model' (that's why 'GedcomforGeneanet' also exists!), with :

* export for witness on event
* handling non-marriage relations
* including path to media
* excluding name of repository
* better support for source (visual/display on geneanet)
* remove absolute media path
* remove "Gramps ID merged" attribute
* add quay source on source's description
* create zip archive for media
* improve alternate names 
* remove anychar support (Gedcom 5.5)
etc...

Agreed with @prculley about maintenance issues with the(se) addon(s).
Also, as gramps aims to properly import and export as much as possible data written on the gedcom, it is also a big challenge to be able to understand logical behind all gedcom files.

Nick-Hall:
@prculley The original policy changed a long time ago, possibly when @kulath was maintaining the GEDCOM import/export.

The current situation is that the importer should attempt to import any GEDCOM that is exported by common software products.  As such, I don't see why we shouldn't consider unsupported custom tags as bugs.

The export is a slightly different situation, that we probably need to discuss on the mailing list.  However, the changes in the export functionality for this PR are so minor that I think that it can go into a maintenance branch.

PR #869, opened 2019-08-15
Fix CSV import for multiple place enclosed by on multiple imports

prculley:
Fixes [#11259](https://gramps-project.org/bugs/view.php?id=11259)


Nick-Hall:
@prculley Please move this to the _gramps51_ branch.  There are no further releases planned for _gramps50_.

PR #870, opened 2019-08-18
Fix exception when merging with active sidebar filter

prculley:
Fixes #11262
Fixes improperly done previous fix...

Should probably be done to gramps50 branch, except we said no more releases there...


prculley:
Merging early to get this in the Gramps 5.1.0 release

PR #871, opened 2019-08-20
Fix GEDCOM importer for SOUR/REFN combinations

prculley:
Fixes #11251

A previous fix tried to treat REFN tags as Gramps Attributes, I forgot that Sources used SrcAttribute as the underlying data object, so had attached a regular Attribute to the Source object.

### DIFF
--- a/gramps/plugins/lib/libgedcom.py
+++ b/gramps/plugins/lib/libgedcom.py
@@ -2438,7 +2438,7 @@ class GedcomParser(UpdateCallback):
             TOKEN_RNOTE  : self.__source_note,
             TOKEN_TEXT   : self.__source_text,
             TOKEN_ABBR   : self.__source_abbr,
-            TOKEN_REFN   : self.__source_refn,
+            TOKEN_REFN   : self.__source_attr,
             TOKEN_RIN    : self.__source_attr,
             TOKEN_REPO   : self.__source_repo,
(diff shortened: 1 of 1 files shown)


lordemannd:
I tested REFN on windows and debian, both bsd and sqlite, looks GOOD.
Check and repair passed in all cases.
Unused items did find a small number of notes left unconnected.

lordemannd:
I don't have the TMG file to test if that problem (11224) is still solved.

I can now see what the reporter in https://gramps-project.org/bugs/view.php?id=11224#c57000
was talking about. There are dates of the form "7 June 1842" or "November 1874", after import they show up in Bold black, then if you edit the event and click OK, the date gets converted to a gramps normal form of "1842-06-07" or "1874-11-00".  And the user just wanted a way to do the conversion automatically.

prculley:
@lordemannd The Gramps GEDCOM import support for dates complies with the GEDCOM specs.  The dates you mentioned do NOT.  So Gramps import drops them into the date field as text.  Since the Gramps editor uses the Gramps date code to look at the entered text, it con sometimes interpret text dates better.  So it gets converted to the internal Date object, and will display according to the Date preferences.

In the ideal case, the TMG sw would export the dates according to the GEDCOM standard.

I suppose someone could write an addon that scanned all date fields and tried the conversion again if it was a text date...

PR #872, opened 2019-08-22
Fix Unstable Release message code for 5.1.0 and easier future use

prculley:
Fixes [#11274](https://gramps-project.org/bugs/view.php?id=11274)

Foobar in creating a release meant that we have an unstable release message on a real release.  Decided to make this a bit easier for future maintainers.  Should only be necessary to edit version.py in main gramps code now.  Still need to update other parts of the source.


gramps/gen/config.py
```diff
@@ -133,7 +133,7 @@ def emit(key):
 register('behavior.addmedia-relative-path', False)
 register('behavior.autoload', False)
 register('behavior.avg-generation-gap', 20)
-register('behavior.betawarn', False)
+register('behavior.betawarn', tuple())
```
Nick-Hall: This should be a simple flag to indicate that the warning has been read.  Does it need to be a tuple?  Is there some problem caused by upgrading?

gramps/gui/grampsgui.py
```diff
@@ -490,7 +491,7 @@ def _display_welcome_message(parent=None):
     Display a welcome message to the user.
     (This docstring seems very legacy/historical, not accurate.)
     """
-    _display_generic_message("master", 'behavior.betawarn', parent=parent)
+    _display_generic_message(VERSION, 'behavior.betawarn', parent=parent)
```
Nick-Hall: It's probably better to put an `if DEV_VERSION:` here, rather than modifying the `_display_generic_message` function.

prculley: I expect to update the release procedure anyway, because it never mentioned these files, thus the issue.

I considered the original simple flag, but if it ever got set, the user would never see a warning again for any other development version.  With the tuple, he will get a new warning every time he runs a different development version.

I actually considered simplifying one of the _display_generic_message or _display_welcome_message methods out of existence, since only one bit of code ever calls the latter, and only it calls the former.  And it is not very 'generic'.  Original change to create the generic method courtesy of Paul Franklin.

lordemannd:
I incorporated these changes, then removed the gramps51 directory and restarted Gramps.
The Unstable Release window wasn't shown.
I think problem is solved.

wroldwiedbwe:
Came to point this out, glad to see a fix 👍 works for me.

gramps/gui/grampsgui.py
Nick-Hall: I would prefer an `if` statement around the lines that we previously commented out.

gramps/version.py
```diff
@@ -19,6 +19,7 @@
 #
 
 VERSION_TUPLE = (5, 1, 1)
+DEV_VERSION = False  # True for development, False for a maintenance release
 VERSION_QUALIFIER = ""
```
Nick-Hall: We should really issue the warning if VERSION_QUALIFIER is not empty.

johnbickmore:
Instead of creating an enhancement, why not just use the old method for the current release?

Nick-Hall:
Yes.  We will probably use the existing method for the current release.

prculley:
Closed in favor of https://github.com/gramps-project/gramps/pull/890

PR #874, opened 2019-08-23
Could not convert string to float

SNoiraud:
by using withinarea filter rule
+ difference between sidebar filter and filter rule
+ some pylint improvements

Fixes [#11269](https://gramps-project.org/bugs/view.php?id=11269)

### DIFF
--- a/gramps/gen/filters/rules/place/_withinarea.py
+++ b/gramps/gen/filters/rules/place/_withinarea.py
@@ -40,6 +40,7 @@ import re
 from ....const import GRAMPS_LOCALE as glocale
 from .. import Rule
 from ....utils.place import conv_lat_lon
+from gramps.gen.errors import FilterError
 
 _ = glocale.translation.sgettext
 
@@ -78,6 +79,12 @@ class WithinArea(Rule):
             self.latitude, self.longitude = conv_lat_lon(latitude,
                                                          longitude,
                                                          "D.D8")
+            if self.latitude is None or self.longitude is None:
+                raise FilterError(_("Cannot use the filter 'within area'"),
+                    _("The place you selected contains bad coordinates. "
+                      "Please, run the tool 'clean input data'"))
+                return
+
             val = self.list[1]
             if isinstance(val, str):
                val = re.sub("\D", "", val) # suppress all alpha characters
@@ -107,6 +114,8 @@ class WithinArea(Rule):
             lon = place.get_longitude()
         if lat and lon:
             latit, longit = conv_lat_lon(lat, lon, "D.D8")
+            if latit is None or longit is None:
+                return False
             if (hypot(float(self.latitude)-float(latit),
                       float(self.longitude)-float(longit))
                     <= self.radius):
--- a/gramps/gen/filters/rules/place/_withinarea.py
+++ b/gramps/gen/filters/rules/place/_withinarea.py
@@ -1,3 +1,6 @@
+"""
+ WithinArea : used to verify if a place is contained in a specific area
+"""
 #
 # Gramps - a GTK+/GNOME based genealogy program
 #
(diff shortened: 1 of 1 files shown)


gramps/gen/filters/rules/place/_withinarea.py
```diff
@@ -60,13 +70,19 @@ def prepare(self, db, user):
...
                 return
-            self.longitude = ref_place.get_longitude()
-            value = self.list[1]
-            unit = self.list[2]
+            longitude = ref_place.get_longitude()
+            self.latitude, self.longitude = conv_lat_lon(latitude,
+                                                         longitude,
+                                                         "D.D8")
+            val = self.list[1]
+            if type(val) == str:
```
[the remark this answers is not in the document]
SNoiraud: Yes I agree. I'll do that.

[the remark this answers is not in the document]
SNoiraud: Done

romjerome:
Two remaining issues:
* user can still set some letters (not number) as self.list[1] (value)

> File "gen/filters/rules/place/_withinarea.py", line 72, in prepare
>      value = int(self.list[1])
> invalid literal for int() with base 10: '--5'

maybe:
```diff
...
-           value = int(self.list[1])
+           try:
+               value = int(self.list[1])
+           except ValueError:
+               value = 0
```

* earth_perimeter should ignore self.latitude = None; because it can raise:
TypeError: float() argument must be a string or a number, not 'NoneType'
_e.g., use an active place without coordinates_

maybe:
```diff
            unit = int(self.list[2])
+           if self.latitude == None:
+               return
            # earth perimeter in kilometers for latitude
            # 2 * pi * (6371 * cos(latitude/180*pi))
            # so 1 degree correspond to the result above / 360
            earth_perimeter = 2*pi*(6371*cos(float(self.latitude)/180*pi))
```

Nick-Hall:
@SNoiraud Is this ready to merge now?  Do you want to squash any of the commits?

SNoiraud:
> Is this ready to merge now?

Yes

gramps/gen/filters/rules/place/_withinarea.py
```diff
@@ -78,6 +78,12 @@ def prepare(self, db, user):
             self.latitude, self.longitude = conv_lat_lon(latitude,
                                                          longitude,
                                                          "D.D8")
+            if self.latitude == None or self.longitude == None:
```
Nick-Hall: Use `is` instead of `==` in comparisons to singletons such as `None`.  See the [PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) programming recommendations section.

gramps/gen/filters/rules/place/_withinarea.py
```diff
@@ -107,6 +113,8 @@ def apply(self, dummy_db, place):
             lon = place.get_longitude()
         if lat and lon:
             latit, longit = conv_lat_lon(lat, lon, "D.D8")
+            if latit == None or longit == None:
```
Nick-Hall: Same as previous comment.

gramps/gen/filters/rules/place/_withinarea.py
```diff
@@ -78,6 +78,12 @@ def prepare(self, db, user):
             self.latitude, self.longitude = conv_lat_lon(latitude,
                                                          longitude,
                                                          "D.D8")
+            if self.latitude == None or self.longitude == None:
+                print(_("ERROR: Cannot use the filter 'within area': "
```
Nick-Hall: Raise a `FilterError` here rather than using a print statement.

gramps/gen/filters/rules/place/_withinarea.py
```diff
@@ -60,13 +71,25 @@ def prepare(self, db, user):
...
+                                                         "D.D8")
+            if self.latitude is None or self.longitude is None:
+                raise FilterError(_("Cannot use the filter 'within area'"),
+                    _("The place you selected contains bad coordinates. "
+                      "Please, run the tool 'clean input data'"))
+                return
+
+            val = self.list[1]
+            if isinstance(val, str):
+               val = re.sub("\D", "", val) # suppress all alpha characters
```
prculley: I'm getting a DeprecationWarning: invalid escape sequence \D here.  Either double the '\' or make it a r"\D".

SNoiraud: OK. I'll do that.
Where do you get this DeprecationWarning ?

Nick-Hall: If you plan to fix this now, I can delay the 5.1.1 release for a while.

SNoiraud: Done with PR #904

prculley: I think this is the Python command line argument "-W all" to display warnings.  My IDE (Eclipse/PyDev) has this set by default.

PR #876, opened 2019-08-24
Release 510

RossGammon:
Updated Debian directory for the 5.1.0 release. For merging into the gramps51 branch.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+gramps (5.0.2-1) unstable; urgency=medium
+
+  * New release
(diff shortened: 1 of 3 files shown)


lordemannd:
has the same unstable code warning as the Windows version of 5.1.0

Nick-Hall:
> has the same unstable code warning as the Windows version of 5.1.0

Yes.  We would normally comment out [line 621](https://github.com/gramps-project/gramps/blob/maintenance/gramps51/gramps/gui/grampsgui.py#L621) in the `grampsgui.py` file for a stable release.

PR #884, opened 2019-08-28
Genealogical symbols : Some corrections

SNoiraud:
+ Reduce the size of the sexuality symbol in relview
+ Set good symbols for marriage
+ Set symbols for baptism, burial and cremation
+ suppress the ":" after the symbol
+ set symbols for the active person

Fixes [#11295](https://gramps-project.org/bugs/view.php?id=11295)


SNoiraud:
New translations are needed.

PR #886, opened 2019-08-30
The Clean Input Data tool failed to find en dash

SNoiraud:
Fixes [#11308](https://gramps-project.org/bugs/view.php?id=11308)

### DIFF
--- a/gramps/plugins/tool/removespaces.py
+++ b/gramps/plugins/tool/removespaces.py
@@ -53,6 +53,22 @@ _ = glocale.translation.sgettext
 WIKI_HELP_PAGE = '%s_-_Tools' % URL_MANUAL_PAGE
 WIKI_HELP_SEC = _('manual|Remove_leading_and_trailing_spaces')
 
+def validate_lat_lon(field):
+    """
+    Return True if some characters are found in the field
+    # hyphen (u+2010)
+    # non-breaking hyphen (u+2011)
+    # figure dash (u+2012)
+    # en dash (u+2013)
+    # em dash (u+2014)
+    # horizontal bar (u+2015)
+    """
+    for char in (',', '\u2010', '\u2011', '\u2012',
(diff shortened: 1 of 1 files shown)


lordemannd:
do we need to also look for the em dash?

gramps/plugins/tool/removespaces.py
```diff
@@ -216,11 +216,35 @@ def places(self):
...
+            if plat.find('—') != -1: # em dash (u+2014)
+                found = True
+            if plat.find('―') != -1: # horizontal bar (u+2015)
+                found = True
             plon = place.get_longitude()
             if plon != plon.strip():
                 found = True
             if plon.find(',') != -1:
                 found = True
+            if plon.find('‐') != -1: # hyphen (u+2010)
```
Nick-Hall: It would be neater to write a function rather than to duplicate code.

lordemannd:
i just noticed a change in the editor for the lat and long, any edit to the field always leaves the cursor at the end of the field. this issue appeared between 5.0.1 and 5.0.2. I have created a bug for it.
I might be related to the cleanup changes?

gramps/plugins/tool/removespaces.py
```diff
@@ -53,6 +53,27 @@
 WIKI_HELP_PAGE = '%s_-_Tools' % URL_MANUAL_PAGE
 WIKI_HELP_SEC = _('manual|Remove_leading_and_trailing_spaces')
 
+def validate_lat_lon(field):
```
Nick-Hall: This function could further be simplified to:
```
def validate_lat_lon(field):
    for char in (',', '‐', '‑', '‒', '–', '—', '―'):
        if field.find(char) != -1:
            return True
    return False
```

SNoiraud: I'll do that

PR #888, opened 2019-08-31
Fix missing menus/buttons when operating in non-English languages

prculley:
Fixes [#11292](https://gramps-project.org/bugs/view.php?id=11292)

Actually two bugs; if the users Tag included XML invalid characters, the Tag button would disappear. 

Also there was a hard to find Gtk bug where it appears that too many non-ASCII strings in the XML for Gtk Builder would cause parts of the XML to be lost.  Fixed by making the XML ASCII (all non-ASCII characters in labels, and tooltips had to be converted to their HTML forms).  Fortunately the ElementTree code could do just that.

### DIFF
--- a/gramps/gui/uimanager.py
+++ b/gramps/gui/uimanager.py
@@ -213,8 +213,17 @@ class UIManager():
         # need to copy the tree so we can preserve original for later edits.
         editable = copy.deepcopy(self.et_xml)
         iterator(editable)  # clean up tree to builder specifications
-        xml_str = ET.tostring(editable, encoding="unicode")
-        #print(xml_str)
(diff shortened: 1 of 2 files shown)


lordemannd:
ran through all my test cases on debian & windows, tested languages, sv_SE, fr_FR, de_DE, nl_NL, zh_TW,he_IL and en_US.

prculley:
Turns out that this was a Gtk bug, and it has been fixed in very recent Gtk versions, (more recent than Ubuntu 18.04 LTS is using).  Guess I will have to try a more recent Gtk version on Windows one of these days.

PR #890, opened 2019-08-31
Fix Unstable Release message code for 5.1.0 and easier future use

prculley:
Fixes [#11274](https://gramps-project.org/bugs/view.php?id=11274)

Foobar in creating a release meant that we have an unstable release message on a real release. Decided to make this a bit easier for future maintainers. Should only be necessary to edit version.py in main gramps code now. Still need to update other parts of the source.

Based on comments, (https://github.com/gramps-project/gramps/pull/872) this now will display the message whenever DEV_VERSION is True and something changes in the VERSION_TUPLE or VERSION_QUALIFIER.  A change of the git revision status WILL NOT cause the message to display, as that happens frequently during development on the developers computer.

I decided to re-integrate the so called _display_generic_message, as it wasn't very generic and was not used elsewhere.

To make Codacy happy, I also did two very minor pylint cleanups.  I wish Codacy did not evaluate several lines before/after a change, always seems to want an ever expanding set of fixes.


Nick-Hall:
Following the comment made by @johnbickmore in PR #872, we should use the old method for v5.1 and introduce the new variable in the next major release.

Nick-Hall:
@prculley I've had some further thoughts about this.  Rather than having a warning dialog that can be disabled, it would be useful to have an indication that a beta version is being used in the status bar.  This would act as a reminder to the user that they are running a development version is being annoying.  Clicking on the indicator could bring up the warning dialog with more detailed information.

What do you think?

Nick-Hall:
Implemented as PR #1012 instead.

PR #894, opened 2019-09-07
Fix CLI crash when generating reports

prculley:
Fixes [#11318](https://gramps-project.org/bugs/view.php?id=11318), [#11213](https://gramps-project.org/bugs/view.php?id=11213)

The custom filters were not getting initialized in CLI mode after previous https://github.com/gramps-project/gramps/commit/8366ceb8964b948b53be2d07171b146fff039acf

I need to remember that not everything is about the GUI...


Nick-Hall:
> I need to remember that not everything is about the GUI...

Yes.  It's very easy to forget to test the fix from the command line.

PR #880, opened 2019-08-27
Hebrew translation ~62% done

avma:

PR #897, opened 2019-09-10
Hebrew he.po for v5.1.0

avma:
Almost there...


Nick-Hall:
@avma After merging with the latest template on the _gramps51_ branch, I see the coverage as 62.14% in PR #880, but only 40.79% in this PR.  This looks like last translation by Igal Shapira.

avma:
Strange it shows ~80% on poedit 1329 strings to go...
I probably need to amend the header section on the po file, make Igal Shapira the First Translator...

Nick-Hall:
I see the problem.  This PR contains an updated mo file, we need the corresponding po file.

avma:
ok, that explains it ;) just created a new pr (this time verified that it is a po file), should show about 82% completion.

PR #899, opened 2019-09-11
Bump required Python version to 3.3 on Readme

johnbickmore:
As mentioned on the release announcement
https://gramps-project.org/introduction-WP/2019/08/gramps-5-1-0-released/

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Requirements
 ============
 The following packages **MUST** be installed in order for Gramps to work:
 
-* **Python** 3.2 or greater - The programming language used by Gramps. https://www.python.org/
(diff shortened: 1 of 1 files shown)


johnbickmore:
How did/does the project decide on what version of Python to use?

Can I suggest that Gramps 5.2 have [python 3.5 or even better python 3.6](https://devguide.python.org/#status-of-python-branches) as the min version, as [python 3.3 end of life'd 2017-09-29](https://devguide.python.org/devcycle/#end-of-life-branches) and 3.4 EOL'd 2019-03-18.

Nick-Hall:
@johnbickmore We consider increasing the minimum requirements if we want to use a new feature.  However, we don't want to exclude people using older versions, so we generally look at what current LTS releases are using.

PR #900, opened 2019-09-11
Using regex in the sidebar gives different result

SNoiraud:
from previous gramps release.

Fixes [#11321](https://gramps-project.org/bugs/view.php?id=11321)

Side effect of PR #674.


Nick-Hall:
It seems like you found a good compromise here.

PR #902, opened 2019-09-13
Add new output format to the tree document generator

hahnjo:
* Add options (like `id`) to `sandclock`
 * Add output format "Graph File"

### DIFF
--- a/gramps/gen/plug/docgen/treedoc.py
+++ b/gramps/gen/plug/docgen/treedoc.py
@@ -588,6 +588,47 @@ class TreeDocBase(BaseDoc, TreeDoc):
 
 #------------------------------------------------------------------------------
 #
+# TreeGraphDoc
+#
+#------------------------------------------------------------------------------
+class TreeGraphDoc(TreeDocBase):
+    """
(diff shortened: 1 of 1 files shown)


Nick-Hall:
I remember you mentioning the graph output format in _addons-source_ PR [#178](https://github.com/gramps-project/addons-source/pull/178) when you added the family sandclock report.  Both of these are good ideas.  Perhaps the missing sandclock options should even be regarded as a bug.

hahnjo:
> I remember you mentioning the graph output format in _addons-source_ PR [#178](https://github.com/gramps-project/addons-source/pull/178) when you added the family sandclock report. Both of these are good ideas. Perhaps the missing sandclock options should even be regarded as a bug.

Yes, exactly (it just took longer due to $reasons).

Getting the first commit into a maintenance release would be awesome. Let me know if I should target `maintenance/gramps51` directly where most of the changes are merged at the moment.

Nick-Hall:
@hahnjo First part merged as a bug fix in commit daa845492deeb8cd9585bb94989657d623b0387c.

hahnjo:
ping

PR #903, opened 2019-09-14
Deprecations

prculley:
Various fixes for Gtk and gi deprecations.

- Gtk.Widget.override_font and modify_font:  mostly fixed by using markup instead, but I decided that GenderLabel was no longer needed (it only had effect on Windows systems), so I removed it entirely.  The MonitoredEntry class was doing the same thing that Gtk.Entry.set_placeholder text (as of 3.2), so i modified it to do that instead.
- Misc.set_padding: deprecated in 3.14, uses Widget margins instead.  set_margin_start/end are new as of 3.12 (Gramps51 requires Gtk 3.12) set_margin_top/bottom as of 3.0.
- Widget.reparent: deprecated in 3.14, uses Container.remove/add and Widget.get_parent instead.
- GObject.GError:  convert to GLib.GError Available since Gtk 3.0.
- GObject.PARAM_READWRITE: deprecated ?? switch to GObject.ParamFlags.READWRITE if GLib version is newer than 2.42.  Note that the docs https://lazka.github.io/pgi-docs/#GObject-2.0/flags.html#GObject.ParamFlags don't mention the 2.42, but that the program warning mentioned it.
- ScrolledWindow.add_with_viewport: deprecated 3.8, presumably the ScrolledWindow.add with automatic viewport add function works as of then.
- Menu.set_title:  This was used in conjunction with 'tear-off' menus, which have been deprecated as of 3.10.  I found no evidence that we ever used this feature.
- Gtk positional arguments: deprecated as of the move from PyGtk to the GObject introspection.  Note: I found a few cases where we were using 'parent=' keyword argument which is also deprecated, these were switched to 'transient_for'.  In most case the 'buttons' keyword argument is converted to the extra call widget.add_buttons.  But for MessageDialog the buttons keyword argument can be used with Gtk.ButtonsType values which is still legal.


Nick-Hall:
Thanks for doing this.

The GObject.PARAM_READWRITE deprecation will continue to need a conditional fix until Gtk 3.15 which requires a Glib version > 2.42.

PR #906, opened 2019-09-15
Escape for name, dates and places in graph reports

ailintom:
Currently if the birth date or the death date contain XML special characters (example: birth date = "1867 <?>"), the item is rendered incorrectly in the report, when images are included. Escaping special characters with html.escape solves the problem.

### DIFF
--- a/gramps/plugins/graph/gvfamilylines.py
+++ b/gramps/plugins/graph/gvfamilylines.py
@@ -33,6 +33,7 @@ Family Lines, a Graphviz-based plugin for Gramps.
 #
 #------------------------------------------------------------------------
 from functools import partial
+import html
 
 #------------------------------------------------------------------------
 #
@@ -818,10 +819,10 @@ class FamilyLinesReport(Report):
             if bth_event and self._incdates:
                 date = bth_event.get_date_object()
                 if self._just_years and date.get_year_valid():
-                    birth_str = self._get_date( # localized year
+                    birth_str = self.get_date( # localized year
                         Date(date.get_year()))
                 else:
-                    birth_str = self._get_date(date)
+                    birth_str = self.get_date(date)
 
             # get birth place (one of:  hamlet, village, town, city, parish,
             # county, province, region, state or country)
@@ -834,10 +835,10 @@ class FamilyLinesReport(Report):
             if dth_event and self._incdates:
                 date = dth_event.get_date_object()
                 if self._just_years and date.get_year_valid():
-                    death_str = self._get_date( # localized year
+                    death_str = self.get_date( # localized year
                         Date(date.get_year()))
                 else:
-                    death_str = self._get_date(date)
+                    death_str = self.get_date(date)
 
             # get death place (one of:  hamlet, village, town, city, parish,
             # county, province, region, state or country)
@@ -876,8 +877,7 @@ class FamilyLinesReport(Report):
                 label += '<TD>'
 
             # at the very least, the label must have the person's name
-            name = name.replace('"', '&#34;')
-            label += name.replace('<', '&#60;').replace('>', '&#62;')
+            label += html.escape(name)
             if self.includeid == 1: # same line
                 label += " (%s)" % p_id
             elif self.includeid == 2: # own line
@@ -964,10 +964,10 @@ class FamilyLinesReport(Report):
                         if self._incdates:
                             date = event.get_date_object()
                             if self._just_years and date.get_year_valid():
-                                wedding_date = self._get_date( # localized year
+                                wedding_date = self.get_date( # localized year
                                     Date(date.get_year()))
                             else:
-                                wedding_date = self._get_date(date)
+                                wedding_date = self.get_date(date)
                         # get the wedding location
                         if self._incplaces:
                             wedding_place = self.get_event_place(event)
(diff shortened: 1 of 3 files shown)


Nick-Hall:
Thanks for the fix.  The build is failing due to trailing whitespace.  As it is a bug fix this can be moved to the _gramps51_ branch.

Should we still escape the label if we are not using HTML output?  Perhaps try something like the following just before adding the node:

```
if use_html_output:
    label = html.escape(label)
```

ailintom:
Thank you for your feedback. 
I also thought about not escaping the label when not using HTML output. 
But in the original code the `name ` is escaped regardless of `use_html_output`:
```
            name = name.replace('"', '&#34;')
            label += name.replace('<', '&#60;').replace('>', '&#62;')
```
and 
```
                place_text = place_text.replace('<', '&#60;')
                place_text = place_text.replace('>', '&#62;')
```
(a snippet from the original code; these original self-made escapes failed on `&` in the name; the date was left unescaped, which caused my bug)
 Thus, I think we should escape other elements unconditionally as well. Works well for me with images turned off (which should in theory set `use_html_output` to false) and on.

Escaping the whole label instead of escaping single elements it is composed of as per `label = html.escape(label)` will not work, because label is concatenated of the elements taken from the database (which should be escaped) and additional HTML tags (`line_delimiter = '<BR/>'` and `label = ('<TABLE BORDER="0" ... `), which should not be escaped.

ailintom:
I am not sure what is the correct way to move this to `maintenance/gramps51`  could the maintainers do this?

ailintom:
If this solution is fine for everyone, I would also suggest introducing similar changes in `gramps\plugins\graph\gvhourglass.py` and `gramps\plugins\graph\gvrelgraph.py`

There is also a plethora of disparate escaping techniques used in 
`gramps\plugins\tool\eventcmp.py` line 213
`gramps\plugins\lib\libodfbackend.py ` line 68
`gramps\plugins\lib\libcairodoc.py ` line 186
`gramps\plugins\export\exportxml.py` line 87
`gramps\gui\tipofday.py` line 103
`gramps\gen\utils\docgen\odstab.py` line 388
`gramps\gen\filters\_filterlist.py` line 120
Should one probably standartize escaping across the project ? `html.escape` introduced in Python 3.2 seems like a fine solution working out of the box. Isn't it?

Nick-Hall:
Rebased.  Also added one blank line and a docstring.

@ailintom This looks good now.  It appears that we do need to escape the strings even when not using HTML output.  You are quite correct that we need to avoid escaping the HTML tags.

Nick-Hall:
> Should one probably standartize escaping across the project ? html.escape introduced in Python 3.2 seems like a fine solution working out of the box. Isn't it?

I did replace the deprecated [cgi escape](https://docs.python.org/3/library/cgi.html#cgi.escape) function with the html version in commit e355a93dc171b6eb4ddc62c92ab636f1086f103a.  I'm not sure if we need to replace any more.  Are any of the other techniques deprecated?

ailintom:
In many cases we have just 
```
        name = name.replace('"', '&#34;')
        name = name.replace('<', '&#60;').replace('>', '&#62;')
```
or
```
def escxml(d):
    return escape(d,
                  {'"' : '&quot;',
                   '<' : '&lt;',
                   '>' : '&gt;',
                   }) if d else ""
```

or
```
def fix(line):
    l = line.strip().replace('&','&amp;').replace('>','&gt;')
    return l.replace(l,'<','&lt;').replace(l,'"','&quot;')
```

instead of escaping with a built-in function. This often incomplete escaping causes problems when the "&", which should also be escaped, is omitted. Arguably, using html.escape instead of escaping with replace will also make the code more readable.

prculley:
I recently found yet another escape function that may be better yet;
from gi.repository.GLib import markup_escape_text https://lazka.github.io/pgi-docs/#GLib-2.0/functions.html#GLib.markup_escape_text

Nick-Hall:
> I recently found yet another escape function that may be better yet;
from gi.repository.GLib import markup_escape_text

We shouldn't import from GLib in the _gen_ module.  It is acceptable in the _gui_ code, however I don't see an advantage over the version in the _html_ module.

You will also see the version in [saxutils](https://docs.python.org/3/library/xml.sax.utils.html#module-xml.sax.saxutils) used.  The main difference is how quotes are escaped for attribute values.

In general, we should use library code over our own versions.  Library code is well tested and robust.

sam-m888:
@ailintom Thanks for doing this.  Note that your last commit introduced trailing spaces and caused Travis-ci to [fail](https://travis-ci.org/gramps-project/gramps/builds/585949375#L2826)  with:

> gramps/plugins/graph/gvrelgraph.py:775:        return html.escape(self._get_date(date))    

You might find the following useful

> To check for whitespace errors with git. Just execute
> git diff --check
> 
> https://kparal.wordpress.com/2011/07/04/git-tip-of-the-day-check-for-whitespace-errors-in-diff/

sam-m888:
Thanks for fixing the trailing spaces 👍 

Travis is now showing that one of the test is failing the build

> FAIL: test_one (gramps.plugins.test.db_undo_and_signals_test.TestMyTestClass_SQLite)
> 
> https://travis-ci.org/gramps-project/gramps/builds/586926203#L2025

@Nick-Hall Can you have a look please?

prculley:
Issue with Travis is unrelated to this PR.  The test that is failing got caught because it uses a db modification time that differed by 1 second.  I suspect that the operations usually occur in the same second (pass) but in this case just happened to occur on both sides of the seconds boundary (fail).

We should probably fix the test with some sort of mock on the db time function, so that time doesn't enter into the testing.

PR #907, opened 2019-09-16
Fix 'Go' menu direct object selection, goes to wrong place

prculley:
Fixes [#11341](https://gramps-project.org/bugs/view.php?id=11341)

Fixes up the scrambled 'Go' menu, got messed up in the UIManager upgrade.


sam-m888:
Thank you

PR #908, opened 2019-09-16
Narrative web and Webcal new features

SNoiraud:
+ allow urls in the user's css files
+ allow alternate stylesheets in pages for narrative web and webcal
+ webcal: incorrect results when divorce event.
+ some pylint improvements


johnbickmore:
How often is gramps master updated with changes from the previous branches eg gramps51 ?

prculley:
When someone is willing to do it and a fair number of commits have occurred since the last time it was done.  Looks like the last time was a couple of months back Sept 16. 6 or 7 non-translation commits ago.

PR #910, opened 2019-09-20
Reorder ID tool fixes

prculley:
Two commits:

Fix Reorder ID tool so subsequent db additions used next possible ID:  After using the Reorder ID tool users noted that under some conditions new objects were using IDs beyond the last used ID.  Particularly noticeable on SqLite, but also possible on BSDDB.  This resets the next ID index in the db so that it starts searching at the beginning (and finds the first available ID for next use).  Fixes [#11346](https://gramps-project.org/bugs/view.php?id=11346)

Fix Reorder ID tool to avoid scrambling the order IDs are assigned:  Users noticed that for BSDDB in particular, that fully renumbering the IDs would sometimes scramble the order the IDs were assigned.  This was a result of the db return order of handles not being in any particular order.  The change sorts processing by handle order (which is also creation order) to provide consistency.  Fixes [#10641](https://gramps-project.org/bugs/view.php?id=10641)

### DIFF
--- a/gramps/plugins/tool/reorderids.py
+++ b/gramps/plugins/tool/reorderids.py
@@ -247,7 +247,7 @@ class ReorderIds(tool.BatchTool, ManagedWindow, UpdateCallback):
 
         self.prim_methods, self.obj_methods = {}, {}
         for prim_obj, prim_objs in self.xobjects:
-            iter_handles = "iter_%s_handles" % prim_obj
+            get_handles = "get_%s_handles" % prim_obj
             get_number_obj = "get_number_of_%s" % prim_objs
             prefix_fmt = "%s_prefix" % prim_obj
             get_from_id = "get_%s_from_gramps_id" % prim_obj
@@ -258,7 +258,7 @@ class ReorderIds(tool.BatchTool, ManagedWindow, UpdateCallback):
             self.prim_methods[prim_obj] = (getattr(self.db, prefix_fmt),
                                            getattr(self.db, get_number_obj)(),
                                            getattr(self.db, next_from_id)())
-            self.obj_methods[prim_obj] = (getattr(self.db, iter_handles),
+            self.obj_methods[prim_obj] = (getattr(self.db, get_handles),
                                           getattr(self.db, commit),
                                           getattr(self.db, get_from_id),
                                           getattr(self.db, get_from_handle),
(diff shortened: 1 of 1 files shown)


gramps/plugins/tool/reorderids.py
```diff
@@ -548,6 +551,9 @@ def _execute(self):
                         self.progress.set_pass(
                             _('Reorder %s IDs ...') % _(prim_objs.title()),
                             self.obj_values[prim_obj].quant_id)
+                    # reset the db next_id index to zero so we restart new IDs
+                    # at lowest possible position
+                    setattr(self.db, DB_INDXES[prim_obj] + 'map_index', 0)
```
Nick-Hall: I'm not sure that this is desirable in all cases.  The original idea of this tool was confusingly to change the format of the IDs and not actually the order.  Some users prefer not to reuse the ID of an object that has been deleted.

gramps/plugins/tool/reorderids.py
```diff
@@ -585,8 +591,12 @@ def _reorder(self, prim_obj):
         change = self.obj_values[prim_obj].get_change()
         index_max = int("9" * self.obj_values[prim_obj].width_fmt)
         do_same = False
+        # Process in handle order, which is in order handles were created.
```
Nick-Hall: This is a side-effect of the way we currently generate handles.  It should not be assumed that this will be true on all database backends.  We may even choose to use a UUID as a handle in future versions.

gramps/plugins/tool/reorderids.py
prculley: @Nick-Hall I'm not sure how to deal with this; I fully understand the bug submitter who wants his IDs to be nice and contiguous after a full resort, and remain contiguous with new items.  BSDDB kind of acts that way now, in that after a restart, the new ids will fill in any gaps.  Where now, with SQLite the only way to avoid gaps is to export/import XML with new tree.

Do you think we need an option for this?

gramps/plugins/tool/reorderids.py
prculley: I still think this is a reasonable idea to allow a full reorder/renumber to maintain pretty much the same order as items were originally assigned.  Even if we switch to some other handle method, I believe that many UUIDs include a time element in their creation which would make sorting possible.

At the very least, sorting here makes it such that multiple invocations of the Reorder tool don't get vastly different results.

Nick-Hall: Sorting the handle to make the output idempotent is acceptable.

If we choose to use a UUID for the handle we would use `uuid4()` from the python [uuid](https://docs.python.org/3.5/library/uuid.html) module.  At the same time we could extract a creation timestamp from the old handle format if necessary.

gramps/plugins/tool/reorderids.py
Nick-Hall: No.  We should avoid extra options unless they are necessary

The tool doesn't guarantee an order, so we are free to change the SQLite order to conform the the BSDDB order.

PR #911, opened 2019-09-20
Fix Preferences so <ctrl>PageUp/PageDn doesn't stick on Dates tab

prculley:
Fixes [#11351](https://gramps-project.org/bugs/view.php?id=11351)

Keyboard shortcut conflict in Gtk between the SpinButton and the NoteBook; both used <ctrl>PgUp/PgDn to do things.

Simple fix was to rearrange Dates panel so a SpinButton was not the active item when <ctrl>PgUp/PgDn selected that panel.

### DIFF
--- a/gramps/gui/configure.py
+++ b/gramps/gui/configure.py
@@ -1489,6 +1489,26 @@ class GrampsPreferences(ConfigureDialog):
             align=Gtk.Align.CENTER, bold=True)
 
         row = 1
+        self.add_pos_int_entry(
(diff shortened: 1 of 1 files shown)


sam-m888:
New position for "Markup for invalid date format:"

2019-09-21 11_17_23-Window


https://gramps-project.org/wiki/index.php/Gramps_5.1_Wiki_Manual_-_Settings#Dates

sam-m888:
- [x] User manual image & text updated to match interface.

PR #913, opened 2019-09-21
Upgrade Vcalendar to 2.0 to support utf8; fix export tests to support this

prculley:
Bug [#11316](https://gramps-project.org/bugs/view.php?id=11316) notes that you cannot export a vcs file (turns out it was because the file contained non-ASCII characters).

To correct this I had to upgrade the VCalendar export to V2.0 since v1.0 did not support utf8.  This involved some other changes to make sure the export met specs.  I used the https://icalendar.org/validator.html tool to test the results.

Then I noticed that the export tests had not flagged an error on the extensive changes.  Turns out a small bug was preventing the tests for most file types.

I fixed that, then added mock code to make the VCalendar export of times this year consistent, and added additional tests for both db types, and noticed that we got different results for the db types in some cases.  SqLite doesn't guarantee any particular order when returning a handles list or iter_handles, and neither does BSDDB, so it is not surprising that the order did not match.

So I patched each of the export types to sort on handles before working.

The three commits are separate, but the tests will fail unless all three are commited.


sam-m888:
Looks like travis fails even with all three commits on the following two test.

> FAIL: Run a Geneweb export test
> ...
> FAIL: Run a Geneweb export test
> 
> https://travis-ci.org/gramps-project/gramps/builds/587879002#L2024

prculley:
The Travis failure was because another test had set "behavior.date-after-range" to 9999.  A bit difficult to debug, in that it would only occur when running all of the tests, not just the exports_test.

Nick-Hall:
> SqLite doesn't guarantee any particular order when returning a handles list or iter_handles, and neither does BSDDB, so it is not surprising that the order did not match.

Yes, the handle order is not guaranteed to be consistent.

We had a similar issue with the Gramps XML export back in 2011.  See bug [#4365](https://gramps-project.org/bugs/view.php?id=4365) and commit 4aa7b0710e466539c80bb8e6b19766488767a04b.

PR #915, opened 2019-09-25
Fix tag colors on PedigreeView

romjerome:
It fixes [bug 11363](https://gramps-project.org/bugs/view.php?id=11363)

### DIFF
--- a/gramps/plugins/view/pedigreeview.py
+++ b/gramps/plugins/view/pedigreeview.py
@@ -196,10 +196,10 @@ class PersonBoxWidgetCairo(_PersonWidgetBase):
         if tags and person:
             for tag_handle in person.get_tag_list():
                 # For the complete tag, don't modify the default color
-                # which is black (#000000000000)
+                # which is black
                 tag = dbstate.db.get_tag_from_handle(tag_handle)
-                if tag.get_color() != "#000000000000": # only if the color
-                    self.bgcolor = tag.get_color()     # is not black
+                if tag.get_color() not in ("#000000", "#000000000000"):
+                    self.bgcolor = tag.get_color()
         self.bgcolor = hex_to_rgb_float(self.bgcolor)
         self.bordercolor = hex_to_rgb_float(self.bordercolor)
 
--- a/gramps/gui/utils.py
+++ b/gramps/gui/utils.py
@@ -551,16 +551,17 @@ def color_graph_box(alive=False, gender=Person.MALE):
 
 def hex_to_rgb_float(value):
     """
-    Convert a hexademical value #FF00FF to rgb. Returns tuple of float between
-    0 and 1
+    Convert a 6 or 12 digit hexademical value to rgb. Returns tuple of floats
+    between 0 and 1.
     """
     value = value.lstrip('#')
     lenv = len(value)
-    return tuple(int(value[i:i+lenv//3], 16)/256.0 for i in range(0, lenv, lenv//3))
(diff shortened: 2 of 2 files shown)


gramps/plugins/view/pedigreeview.py
```diff
@@ -197,9 +197,17 @@ def __init__(self, view, format_helper, dbstate, person, alive, maxlines,
             for tag_handle in person.get_tag_list():
                 # For the complete tag, don't modify the default color
                 # which is black (#000000000000)
+                old_black = "#000000000000" #self.__color on new tag
+                new_black = old_black[0:7]
                 tag = dbstate.db.get_tag_from_handle(tag_handle)
-                if tag.get_color() != "#000000000000": # only if the color
-                    self.bgcolor = tag.get_color()     # is not black
+                if tag.get_color() != old_black: # only if the color
```
Nick-Hall: Can't this code be simplified to a single "if" statement?

`if tag.get_color() not in ("#000000", "#000000000000"):`

romjerome: Oh, yes, less narrative but clearer.
Thanks.

Note, maybe hardcoded color values should be limited?
To set [long hexa digit and black color](https://github.com/gramps-project/gramps/blob/master/gramps/gen/lib/tag.py#L62) as default color for new tags 
could be a problem on many places with dark gtk3 themes.

romjerome:
Going further, on [Feature Request 11364](https://gramps-project.org/bugs/view.php?id=11364), I tried to use a pseudo-grey color as fanchart
environment (widget) is able to properly display "dark" backgrounds and text!

gramps/plugins/view/pedigreeview.py
Nick-Hall: Perhaps the default should be `None`, meaning that the tag has not been assigned a colour?

Nick-Hall:
@romjerome I have modified the patch.  Is it sufficient for this bug fix?

romjerome:
@Nick-Hall 

Sorry, my bad, two issues will be fixed on the first patch!

There is one issue reported on bug report [#11363](https://gramps-project.org/bugs/view.php?id=11363), which could be fixed by something like:
```
+                    rgba = Gdk.RGBA()
+                    rgba.parse(tag.get_color())
+                    hexval = "#%02x%02x%02x" % (int(rgba.red * 255),
+                                        int(rgba.green * 255),
+                                        int(rgba.blue * 255))
+                    self.bgcolor = hexval
```
Currently if you set "Shows tags" on PedigreeView, then you do not get the tag colors.
Parsing via Gdk.RGBA() as safety alternative to many tests and it also limits conversions for handling long hexa digit values.

... and a **second** issue, around the black color, ignoring the new short digit hexa value.
As both formats (old and new) are accepted I suppose that no one checked the hardcoded long digits? (maybe check `$ git grep '# ffffffffffff'` too?)

Your modification seems to overwrite the primary fix.

romjerome:
The support for tags (colors) on PedigreeView has been added [via commit f1e3c9](https://github.com/gramps-project/gramps/commit/f1e3c9500ec9fbb7cb4c3b4317641731a24a3d62).

Nick-Hall:
There is a bug in the [hex_to_rgb_float](https://github.com/gramps-project/gramps/blob/maintenance/gramps51/gramps/gui/utils.py#L552) utility function.

sam-m888:
Thank you

PR #916, opened 2019-09-29
NarrativeWeb: dates not localised in place pages

SNoiraud:
Home link translated to lowercase

Fixes [#11357](https://gramps-project.org/bugs/view.php?id=11357), [#11358](https://gramps-project.org/bugs/view.php?id=11358)


sam-m888:
Thank you

PR #918, opened 2019-09-30
Update POTFILES.in/skip for missing translations

romjerome:
Fix translations being missed in some files by updating  POTFILES.in/skip

### DIFF
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -17,6 +17,7 @@ gramps/gen/db/base.py
 gramps/gen/db/exceptions.py
 gramps/gen/db/generic.py
 gramps/gen/display/name.py
+gramps/gen/display/place.py
 gramps/gen/filters/_filterparser.py
 gramps/gen/filters/_genericfilter.py
 gramps/gen/filters/rules/_changedsincebase.py
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -739,6 +739,7 @@ gramps/plugins/tool/rebuildgenderstat.py
 gramps/plugins/tool/rebuildrefmap.py
 gramps/plugins/tool/relcalc.glade
 gramps/plugins/tool/relcalc.py
+gramps/plugins/tool/removespaces.glade
 gramps/plugins/tool/removespaces.py
 gramps/plugins/tool/removeunused.glade
 gramps/plugins/tool/removeunused.py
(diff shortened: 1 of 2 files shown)


romjerome:
Note, if "Full" string is not translated, the hardcoded string will be stored on 'place_format.xml' file without "en" lang code. 

Maybe 'fmt.language' might be used for a 'place_format.xml' file update or as key filter?
My current inherited (5.0-> 5.1.0 -> 5.1.1) 'place_format.xml' file looks like:
```
<place_formats>
  <format language="" levels=":" name="Full" reverse="False" street="0"/>
  <format language="fr" levels="0:p+1" name="Nouveau" reverse="False" street="1"/>
</place_formats>
```
I suppose that now (_with a complete translation for my locale_), it will be something like:
```
<place_formats>
  <format language="fr" levels=":" name="Complet" reverse="False" street="0"/>
  <format language="fr" levels="0:p+1" name="Nouveau" reverse="False" street="1"/>
</place_formats>
```

romjerome:
Ah, to create a new 'place_format.xml' just ignores the lang code!
My new file now looks like:
```
<place_formats>
  <format language="" levels=":" name="Complet" reverse="False" street="0"/>
</place_formats>
```
Maybe at least set a format language for avoiding handling issues in the future?

sam-m888:
@romjerome Is this good to go, your comments while related don't impact this PR right? You can raise an issue if you need regarding pace_formal.xml

romjerome:
@sam-m888 it is indirectly related!
I was not able to understand the hardcoded string on 'place_format.xml' file,
maybe because translated strings on 'gramps/gen/display/place.py'
were skiped?

While entry on localized files is not translated (e.g., during devel stage)
we can generate an harcoded "Full" string on 'place_format.xml' file,
even now after adding the reference on POTFILE.in file (e.g., locales without
a translation for the "Full" string). Sure it should be a new PR.

Note, the entry already exists on translation template:
```
#: ../gramps/gen/plug/docgen/treedoc.py:63
msgid "Full"
msgstr ""
```

So, it could be a minor additionnal issue for most locales and users with an updated translation, otherwise "Full" will be hardcoded for the branch (5.1 -> 5.1.0, 5.1.1, 5.1.2 ,etc ...)

romjerome:
By renaming (or a temp remove) 'POTFILES.skip', then running:
`$ intltool-update -m`, we can see a second file ignored/hidden!
**gramps/plugins/tool/removespaces.glade**

sam-m888:
-  [ ]  PO files need to be regenerated.

sam-m888:
Thank you

PR #920, opened 2019-10-02
Fix various Entry fields so Undo/Redo works

prculley:
Fixes [#11380](https://gramps-project.org/bugs/view.php?id=11380) [#11339](https://gramps-project.org/bugs/view.php?id=11339)

In prior versions of Gramps you could use the undo/redo keys to edit items in the text entries.  This no longer works.
This is related to the changes to suppress odd characters and leading/trailing spaces. The Gtk.Entry.set_text() call is effectively clearing the undo/redo list at every keystroke or when you leave the field.

In this PR I move the odd characters cleanup into UndoableEntry.do_insert_text instead of MonitoredEntry.

I scanned the users of MonitoredEntry and they all appeared to use Glade files, I scanned the glade files for anyone using just plain Entry (none found), they all appeared to use UndoableEntry or ValidatableMaskedEntry, so this should cover everyone.

### DIFF
--- a/gramps/gui/widgets/monitoredwidgets.py
+++ b/gramps/gui/widgets/monitoredwidgets.py
@@ -64,8 +64,6 @@ from gramps.gen.errors import ValidationError
 _RETURN = Gdk.keyval_from_name("Return")
 _KP_ENTER = Gdk.keyval_from_name("KP_Enter")
 
-# table for skipping illegal control chars
-INVISIBLE = dict.fromkeys(list(range(32)))
 
 #-------------------------------------------------------------------------
 #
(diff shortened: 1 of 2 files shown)


sam-m888:
Was this bug fix meant for gramps51 branch?

gramps/gui/widgets/monitoredwidgets.py
```diff
@@ -141,14 +139,10 @@ def connect(self, signal, callback, *data):
         self.obj.connect(signal, callback, *data)
 
     def _on_quit(self, obj, event):
-        text = obj.get_text().translate(INVISIBLE).strip()
-        self.set_val(text)
-        obj.set_text(text)
+        self.set_val(obj.get_text().strip())
```
SNoiraud: The only thing I see for the moment, is we don't see anymore the change after the strip.
I propose to have the obj.set_text(text) available again.

SNoiraud: Same thing in the on_change method.

SNoiraud: This problems occurs when copy/past

sam-m888:
Thank you

PR #922, opened 2019-10-03
Fix issues with RTL languages and LAT/LONG

prculley:
- [x] ~~**This should be accepted only after https://github.com/gramps-project/gramps/pull/920 as it includes the (necessary) commit from that PR.**~~

While discussing and testing for https://gramps-project.org/bugs/view.php?id=11335, several issues came up.  These are all about RTL languages, specifically noted and tested with Hebrew.

- The LAT/LONG entries in the place editors have the '-' sign on the wrong end after editing when using the floating point degrees format.
- The Lat.Long entries are scrambled if the entry contains the translated character for N, S, E, W, which is allowed by our GPS parser.  This is due to the Unicode bidi rules which mangle the reading order if a 'strong' alpha character is part of the string.

Since Lat/Long entries are a number, although they contain the N, S, E, W characters, our Hebrew speaking bug submitter says they should be fully treated as a numeric with the whole string in LTR order.

A third issue will be the subject of a different PR (storing the Lat/Long values when non-English N, S, E, W characters are present).

It took me multiple days to get around the various Gtk/GObject issues I ran into and to find a mechanism that seems to work well.

The basic fix is to tell Gtk that we want to override the normal bidi algorithm for these fields.  Unfortunately Gtk doesn't have CSS support for this like HTML does, so that doesn't work.  So instead I had to surround the text string with the Unicode LRO (LEFT-TO-RIGHT OVERRIDE) and PDF (POP DIRECTIONAL FORMATTING) characters.

For the Place Views this is relatively easy, although a CSV export of the view will contain the LRO/PDF characters; thus I only put them in place for RTL languages.

For the editor entries, which use UndoableEntry widget, it was more difficult.  I only enable the surround characters for the specific fields and only in RTL languages.  Then I had to fix up the entry class to protect the extra characters.

I tested everything I could think of, and the only oddity that results is when cut/copy of the whole field or a portion containing the end characters is pasted into another app.  Then one or both of the LRO/PDF characters come along for the ride.

### DIFF
--- a/gramps/gui/editors/editplace.py
+++ b/gramps/gui/editors/editplace.py
@@ -158,24 +158,30 @@ class EditPlace(EditPrimary):
             self.obj.set_code, self.obj.get_code,
             self.db.readonly)
 
+        entry = self.top.get_object("lon_entry")
+        entry.set_ltr_mode()
         self.longitude = MonitoredEntry(
-            self.top.get_object("lon_entry"),
+            entry,
             self.obj.set_longitude, self.obj.get_longitude,
             self.db.readonly)
         self.longitude.connect("validate", self._validate_coordinate, "lon")
         #force validation now with initial entry
-        self.top.get_object("lon_entry").validate(force=True)
+        entry.validate(force=True)
 
+        entry = self.top.get_object("lat_entry")
+        entry.set_ltr_mode()
         self.latitude = MonitoredEntry(
-            self.top.get_object("lat_entry"),
+            entry,
             self.obj.set_latitude, self.obj.get_latitude,
             self.db.readonly)
         self.latitude.connect("validate", self._validate_coordinate, "lat")
         #force validation now with initial entry
-        self.top.get_object("lat_entry").validate(force=True)
+        entry.validate(force=True)
 
+        entry = self.top.get_object("latlon_entry")
+        entry.set_ltr_mode()
         self.latlon = MonitoredEntry(
-            self.top.get_object("latlon_entry"),
+            entry,
             self.set_latlongitude, self.get_latlongitude,
             self.db.readonly)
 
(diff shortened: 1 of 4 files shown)


Nick-Hall:
@prculley Did the standard method of using `entry.set_direction(Gtk.TextDirection.LTR)` fail to work?

prculley:
The `entry.set_direction(Gtk.TextDirection.LTR)` was sufficient to fix the minus sign on the wrong end.  But not the embedded RTL characters.  It really only sets the default direction of the container, the one that the bidi algorithm uses to start with.  It does not perform an overrride.

One of the many experiments I tried before getting to what I have...

Nick-Hall:
The text direction seems to be set automatically at the pango level.  Unfortunately, the pango layout in the entry widget is read-only.

sam-m888:
> The text direction seems to be set automatically at the pango level. Unfortunately, the pango layout in the entry widget is read-only.

@Nick-Hall  Should I take your comment to mean that something else is required for this PR, or can you or I merge this? I've tested this PR (in Hebrew! & English) as best I can and believe it to do what it says.

Nick-Hall:
@sam-m888 No.  This is ready to merge.  I was just pointing out why the obvious solution doesn't work.

Nick-Hall:
Has @avma tested this?

sam-m888:
Thank you.

avma:
Humm, I missed this one, (went to spam...).
I looks OK now, one small issue though; it transposes the Hebrew abbreviation letters for west and east. 
West should be 'מע' and not 'עמ'.
East should be 'מז' and not 'זמ', as you can see on the at attached screenshot.
Screenshot from 2019-12-08 10-30-16

PR #923, opened 2019-10-04
Fix Gramps -v error when Gtk is not present

prculley:
Via email a user noted a bug when "Gramps -v" was used with Gtk not installed (and gi was).
When the "gi.require_version('Gtk', '3.0')" was added during Gramps 5.0.0 timeframe, a corresponding exception was not.


sam-m888:
Thank you

@johnbickmore  Thanks for reporting this.

PR #924, opened 2019-10-04
Fix up Event Editors Place display for bidi text with Gramps ID

prculley:
Fixes [#10124](https://gramps-project.org/bugs/view.php?id=10124)

Some versions of Gtk/Pango have trouble with text containing both LTR and RTL characters.  The bug notes this shows up in our Event Editor on the 'Place" field, where we have the place displayer title and the Gramps ID concatenated into the same string. 

In older versions of Gtk (3.18.9 tested) the bracket around the Gramps ID would get mangled to the beginning of the string:
image
fixed (somewhat):
image

In newer versions of Gtk (3.24.3 tested) it was better to start with:
image
fixed:
image

The older version of Gtk/Pango doesn't seem to properly interpret all of the potential Unicode bidi control characters, so the fix shown is the best I can do.  The fixed version of the newer Gtk version is what is desired.

I note that the Place displayer should be fixed up to use more appropriate separators than just commas for RTL text, but that is another issue.

### DIFF
--- a/gramps/gui/editors/objectentries.py
+++ b/gramps/gui/editors/objectentries.py
@@ -311,7 +311,10 @@ class PlaceEntry(ObjEntry):
 
     def get_label(self, place):
         place_title = place_displayer.display(self.db, place)
-        return "%s [%s]" % (place_title, place.gramps_id)
+        # When part of the place title contains RTL text, the gramps_id gets
+        # messed up; so use Unicode FSI/PDI and LRM chars to isolate it.
+        # see bug 10124 and PR924
+        return "%s \u2068[%s]\u200e\u2069" % (place_title, place.gramps_id)
 
     def call_editor(self, obj=None):
         if obj is None:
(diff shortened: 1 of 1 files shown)


gramps/gui/editors/objectentries.py
```diff
@@ -311,7 +311,7 @@ def get_from_handle(self, handle):
 
     def get_label(self, place):
         place_title = place_displayer.display(self.db, place)
-        return "%s [%s]" % (place_title, place.gramps_id)
+        return "%s \u2068[%s]\u200e\u2069" % (place_title, place.gramps_id)
```
Nick-Hall: A short comment would be helpful here.

prculley: Comment added

sam-m888:
Thank you

johnbickmore:
How often is gramps master updated with changes from the previous branches eg gramps51 ?

PR #925, opened 2019-10-04
Fix Descendat Tree report for HandleError when no parents on family

prculley:
Fixes [#11365](https://gramps-project.org/bugs/view.php?id=11365)
Yet another HandleError.


sam-m888:
Thank you

PR #927, opened 2019-10-05
Fix CLI parser to accept negative integers as valid

prculley:
Fixes [#11381](https://gramps-project.org/bugs/view.php?id=11381)


sam-m888:
Thank you

PR #928, opened 2019-10-05
Fix improper escaping in odt output for TOC/Bookmark etc.

prculley:
Fixes [#11378](https://gramps-project.org/bugs/view.php?id=11378)

For the index, toc, bookmark etc. normal XML escape is correct, the original code also added sub-tags for line break and tab which was NOT correct.


sam-m888:
Thank you

PR #929, opened 2019-10-08
Narrative web: Ancestor's tree display looks weird

SNoiraud:
Solves the following:
- Person boxes overlap
- Some person boxes partially visible or hidden
- allow very long names

Fixes [#11382](https://gramps-project.org/bugs/view.php?id=11382)

https://github.com/gramps-project/gramps/pull/548


sam-m888:
Thank you

PR #930, opened 2019-10-15
Fix Preferences/Genealogical Symbols when only one font is present that passes

prculley:
Fixes [#11395](https://gramps-project.org/bugs/view.php?id=11395)

If user has only a single font available that passes check for the Genealogical Symbols being all present, the original code crashed.  [Apparently never tested...](https://github.com/gramps-project/gramps/pull/598)


sam-m888:
Thank you

PR #932, opened 2019-10-16
Fix Window family tree title for non-ASCII chars on Windows

prculley:
Fixes [#11390](https://gramps-project.org/bugs/view.php?id=11390)

As it says, if Family Tree name contained non-ASCII characters, the titlebar on Windows would display it wrong.  Turned out to be reading of a utf8 file without the 'encoding' set.  On Windows this results in using the default encoding which is one of the code pages, NOT utf8.

I scanned other uses of file 'open', and some other potential trouble spots are;
gen.plug.docgen.stylesheet (writing the XML stylesheet)
gen.plug.docgen.odstab (also writing the XML stylesheet and meta)
gen.utils.lds (XML temple info, currently contains no non-ASCII chars)
gui.views.listview write_tabbed_data (writes CSV of view)
plugins.docgen.htmldoc (XML CSS of the textdoc)

I did NOT attempt to test/debug each case.

### DIFF
--- a/gramps/cli/grampscli.py
+++ b/gramps/cli/grampscli.py
@@ -280,7 +280,7 @@ class CLIManager:
             # Attempt to figure out the database title
             path = os.path.join(filename, "name.txt")
             try:
-                with open(path) as ifile:
+                with open(path, encoding='utf8') as ifile:
                     title = ifile.readline().strip()
             except:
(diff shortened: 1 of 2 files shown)


sam-m888:
Thank you

PR #936, opened 2019-10-20
Limit Age Stats gramplet settings to appropriate values

Nick-Hall:
* Max ages should be divisible by 5 to avoid out of range
  errors.
* The chart width should be greater than 45 to look right and
  avoid division by zero errors.


romjerome:
#941

sam-m888:
Thank you

PR #939, opened 2019-10-21
Update Hebrew translation

avma:
Although I've completed all strings (using the recent .POT) some strings are still in English; Preference -> Theme tab is all in English...


prculley:
The Preferences/Theme tab is an addon.  It appears that there is no Hebrew translation for that addon as of yet.

Thanks for all your work .

avma:
ok thanks, looks like I'll do that one next... :)

sam-m888:
Thank you

PR #940, opened 2019-10-21
Fix the Preferences 'Age display precision' value getting lost

prculley:
Fixes [#11384](https://gramps-project.org/bugs/view.php?id=11384)
It turns out the changes was actually changing the wrong config setting...

I had to look this one up.  Using a lambda like this is called a 'closure' by some; what is happening is that the value 'constant' is being evaluated at the time the lambda is called, not when it is assigned.  So in this particular bit of code the preference setting was actually changing 'preferences.family-relation-type' (the value of the variable 'constant' set a bit after the lambda definition.

I reverted this particular bit of code, as I think this kind of Python knowledge is pretty obscure (I could have just used a different unique name for the 'constant' variable).


sam-m888:
Thank you

PR #942, opened 2019-10-23
Improve CSV import to allow places that are not in enclosed order

prculley:
Fixes [#11407](https://gramps-project.org/bugs/view.php?id=11407)

The CSV import originally required that places that enclose other places be located first in the file, to facilitate single pass operations.

This PR removes that particular restriction.  This was in response to a bug report that the user was having a difficult time interpreting the 'exception' that was raised when places were out of order in the original code.

Added code to make temporary place if the enclosing place did not exist, and then checked at end of processing to make sure it really happened.  If not, fixed up db to be valid, and reported 'Unknown' places created, with their Gramps IDs.  Similar to what XML and GEDCOM import do.

My IDE also noticed a fatal error at line 273 which I couldn't resist fixing as well.

### DIFF
--- a/gramps/plugins/importer/importcsv.py
+++ b/gramps/plugins/importer/importcsv.py
@@ -60,6 +60,7 @@ from gramps.gen.datehandler import parser as _dp
 from gramps.gen.utils.string import gender as gender_map
 from gramps.gen.utils.id import create_id
 from gramps.gen.utils.location import located_in
+from gramps.gen.utils.unknown import create_explanation_note
 from gramps.gen.lib.eventroletype import EventRoleType
 from gramps.gen.config import config
 from gramps.gen.display.place import displayer as place_displayer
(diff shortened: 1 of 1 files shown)


gramps/plugins/importer/importcsv.py
```diff
@@ -373,6 +377,24 @@ def parse(self, filehandle):
             LOG.debug(msg)
             LOG.debug("New Families: %d" % self.fam_count)
             LOG.debug("New Individuals: %d" % self.indi_count)
```
jstockman67: Can you also add a new debug line for Places?

`LOG.debug("New Places: %d" % self.place_count)`

prculley: Done

jstockman67: @prculley 🌟🌟🌟🌟 Thank you.

PR #943, opened 2019-10-25
Fix Spanish translation for dates

prculley:
Fixes [#10836](https://gramps-project.org/bugs/view.php?id=10836)

Easy enough, even for a non-Spanish speaker like me.  Other examples in the same file starting at line 40645.


sam-m888:
Thank you.

PR #809, opened 2019-04-12
GEPS 045 - a second attempt at enhanced places

prculley:
An upload for GEPS 045: Place Model Enhancements in case anyone wants to see if I'm going in the right direction.

At this point I think that all the code should work, so any remaining issues are bugs.  While it is possible to upgrade a tree with the current db, (and a backup is automatically made of the db first) please use care on your special trees.  Import/export of XML works, as well as the other importers and exporters.  All reports should work.  The https://github.com/gramps-project/addons-source/pull/214 contains the addons also updated to work with this PR, since they have not been built/uploaded, they will have to be installed manually.

A few notes;

* Place names and place types are actually unified lists, not a primary and alternate list like previously. You can still edit the top item in these lists in the top of the dialogs, changes appear in the displaytabs.  Users who don't mess with multiple names and types should have no trouble.

* I made the new editors as much like the old ones as possible, so they should be familiar to previous users.

* The place types list is still the same as always, just the previous Gramps types.  Users of GetGOV will find additional custom types appearing in their data and on the place type combos.

* GOV types are added as numbered custom types.  Custom types are either 'numbered' or fully custom; the numbered types are assigned in a range starting from -1 and counting down, fully custom types while internally numbered (in a different negative range) don't preserve the number for XML/GEDCOM transfer, while numbered types do.

* To deal with the many place types, I am adding the concept of place groups. These groups are used to make the place types combo menu a bit easier, and to allow selection and display of hierarchies and groups for titling. For example, Gramps originally had the concept of Country, County, City type etc. Since there are now multiple place types which could be described as countries, we use the Countries group as a group designation for Countries.  The groups are as follows:
  * Countries - for all nation states and confederations. e.g. Kingdom, Republic.
  * Regions - for all administrative and geographic regions. e.g. State, County.
  * Populated Place - for all types of settlement. e.g. City, Town, Village, Hamlet.
  * Unpopulated Place - e.g. Cemetery, Battlefield.
  * Building - e.g. House, Church, Hospital.
  The old fields COUNTRY and CITY can then map to Country and Populated Place.
  The old fields STATE and COUNTY can be determined by the actual position of a Region within the place hierarchy. STATE will be a top level Region and COUNTY a second level Region. LOCALITY is simply a Region below the Populated Place level.

* The Groups are stored with the place types as an integer bit field. This makes identifying a place type as belonging to a group a simple bitwise 'and' operation.

[...]

PR #945, opened 2019-10-27
Fix dbapi db to warn/prevent opening newer schema versions

prculley:
While testing the [enhanced places GEPS](https://github.com/gramps-project/gramps/pull/809), I inadvertently opened the upgraded db with Gramps51 branch code, and it crashed.  Seems we forgot to perform the check for a newer schema version for dbapi dbs.

I think we should get this into gramps51 branch sometime before we release gramps52 (assuming that gramps52 bumps the schema version), to avoid similar crashes.


prculley:
@sam-m888 Why the Work in progress ??

PR #946, opened 2019-10-30
Narrativeweb: Referenced regions problems

SNoiraud:
When image width > 800, the referenced regions are incorrectly placed

Fixes [#011414](https://gramps-project.org/bugs/view.php?id=11414)


sam-m888:
Thank you

PR #947, opened 2019-10-30
Fix GEDCOM export to avoid translated GrampsType strings

prculley:
Fixes [#11410](https://gramps-project.org/bugs/view.php?id=11410)

The bug submitter noticed that the GEDCOM export contained translated strings in the MEDI lines of his export.  GEDCOM should contain English values for most such cases where 'types' are exported.

I scanned the export code for other potential cases, and found:
- Certain unexpected event types on individuals (Marriage)
- Certain unexpected event types on families (Immigration, Residence, etc.)
- Pedigree types that were not standardized on PEDI,_FREL,_MREL (Sponsored, Unknown)
- Family Attributes other than Number if children (Caste, Description, National etc.)
- Personal Attributes other than Caste, Description, ID, National, number of children, SSN (Occupation etc.)
- The MEDI types of this bug report...

### DIFF
--- a/gramps/plugins/export/exportgedcom.py
+++ b/gramps/plugins/export/exportgedcom.py
@@ -596,7 +596,7 @@ class GedcomWriter(UpdateCallback):
             if val.strip():
                 self._writeln(2, 'TYPE', val)
             else:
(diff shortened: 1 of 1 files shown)


sam-m888:
Thank you

PR #948, opened 2019-11-02
Add index of person in ancestor tree report.

tengel:
This PR draws an index number next to the box of each person of the ancestor chart report. The index number is calculated using the Sosa/Kekule method. It is only printed when enabled in the settings dialog. The default start number of the proband is 1, but the settings dialog allows to set an arbitrary start index. In this way it is possible to print sub-trees with an index related to the complete tree.


Nick-Hall:
Rebased.

Nick-Hall:
I posted to the *gramps-devel* mailing list on [15 Feb 2022](https://sourceforge.net/p/gramps/mailman/message/37610897/) asking for reviews and comments about this PR.  Since there have been no replies, I assume that nobody objects to this.

PR #949, opened 2019-11-04
Add Web Relations report plugin [To be resubmitted as an addon]

tuxofil:
This plugin was at first developed as extension to Narrative Web Report. Now it can extend NWR or can be used as standalone report.

NWR has pedigree charts, but their max depth is often not enough + only ancestors are shown. Although people often want great charts and want not to surf between NWR pages. On the other hand, it's impossible to add *all* ancestors and descendants into NWR because it will destroy page layout.

This report generates single HTML page with interactive SVG graph, showing *all* ancestors and *all* descendants of the selected person. Every person has context menu to switch to nearest relatives and optional link to personal page at Narrative Web Report.

Example can be shown here: http://tuxofil.com/genealogy/example/


prculley:
I suggest you resubmit this as an addon, rather than a standard part of Gramps.  That has the advantage that the acceptance criteria is much easier to meet (pretty much that it is somewhat useful and doesn't crash Gramps easily).  Also when it is accepted it becomes available immediately; otherwise, as new code, it would not be available until next summer at the earliest when Gramps 5.2.x is released.

A quick look suggests to me that the only changes would be to keep your templates with the addon instead of the Gramps data directory.

tuxofil:
Will do. Thanks!

Nick-Hall:
Was this ever submitted as a third-party addon?

PR #950, opened 2019-11-06
Fix duplicated "døde døde" Norwegion Translation for libnarrate

prculley:
Fixes [#11429](https://gramps-project.org/bugs/view.php?id=11429)


sam-m888:
Thank you

PR #951, opened 2019-11-06
Provide an option to "hide" the LDS tab

mathieumd:
As requested by Romjerome in comment [c58148](https://gramps-project.org/bugs/view.php?id=3872#c58148)

Fixes [#3872](https://gramps-project.org/bugs/view.php?id=3872)


gramps/gui/configure.py
```diff
@@ -1350,6 +1350,10 @@ def add_formats_panel(self, configdialog):
             extra_callback=self.cb_grampletbar_close,
             tooltip=_("Show close button to simplify removing gramplets "
                       "from bars."))
+        row += 1
+        self.add_checkbox(grid,
+            _('Use LDS tabs on person and family editors'),
```
Nick-Hall: "Display LDS tabs in person and family editors" reads slightly better.

gramps/gui/editors/editfamily.py
```diff
@@ -793,7 +793,11 @@ def _create_tabbed_pages(self):
                                             self.uistate,
                                             self.track,
                                             self.obj.get_lds_ord_list())
-        self._add_tab(notebook, self.lds_embed)
+
+        if config.get('interface.use-lds') == False and self.lds_embed.is_empty() == True:
```
Nick-Hall: You can simplify this to:
`if not config.get('interface.use-lds') and self.lds_embed.is_empty():`

gramps/gui/editors/editperson.py
```diff
@@ -530,7 +530,11 @@ def _create_tabbed_pages(self):
                                      self.uistate,
                                      self.track,
                                      self.obj.get_lds_ord_list())
-        self._add_tab(notebook, self.lds_list)
```
Nick-Hall: Perhaps we should add the tab anyway and just hide it?

PR #954, opened 2019-11-21
Add person filter "related to <filter>"

Mattkmmr:
Feature requested in [#11449](https://gramps-project.org/bugs/view.php?id=11449)
This filter matches all people who are related to anybody matched by a person filter.
[Addon (#260)](https://github.com/gramps-project/addons-source/pull/260)


prculley:
You might consider making an addon out of this; or adding it to the https://github.com/gramps-project/addons-source/tree/maintenance/gramps51/FilterRules addon.  It would become available much sooner, and if the developers think it is worthwhile would eventually get added to Gramps main rules.

sam-m888:
@Mattkmmr Should this be closed as has been added as an addon?

Mattkmmr:
@sam-m888 Yes it can be closed.

PR #956, opened 2019-12-01
fix tagref support

romjerome:
Add missing code for event, repository, source, citation and place


romjerome:
See [bug report 11458](https://gramps-project.org/bugs/view.php?id=11458).

PR #958, opened 2019-12-02
tmpdir on treedoc generator under linux

romjerome:
Currently the pdf output file format for these reports is looking at a temp dir which seems not the same as my default LaTex converter (LaTex to PDF), under linux.

Need to test it (manual edition via github editor).
Do not push it, just a first commit without local testing!


romjerome:
Sorry. Close it as it seems rather related to some missing dependencies on my configuration.

Start the [linux section on the dedicated wiki documentation](https://gramps-project.org/wiki/index.php/Addon:GenealogyTree#Linux).

PR #960, opened 2019-12-06
[Tree doc Tex] fix issues around localization

romjerome:
* specific non-ASCII glyphs related to our locale are not properly displayed on PDF
* date cannot be parsed (one date format and in english)
* genealogytree macro needs a lang set in english

### DIFF
--- a/gramps/gen/plug/docgen/treedoc.py
+++ b/gramps/gen/plug/docgen/treedoc.py
@@ -347,7 +347,7 @@ class TreeDocBase(BaseDoc, TreeDoc):
         trans = glocale.language[0][:2]
         lang_en = _locale._get_language_string(trans).lower()
         if lang_en in LANG_SUPPORT:
-            self.write(0, '\\gtrset{language=%s}\n' % lang_en.lower())
+            self.write(0, '\\gtrset{language=%s}\n' % lang_en)
         self.write(0, '\\begin{document}\n')
 
         if self.nodecolor == 'preferences':
--- a/gramps/gen/plug/docgen/treedoc.py
+++ b/gramps/gen/plug/docgen/treedoc.py
@@ -47,6 +47,8 @@ from ...constfunc import win
 from ...config import config
 from ...const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
+from ...utils.grampslocale import GrampsLocale
+_locale = GrampsLocale(lang='en_US')
 
 #-------------------------------------------------------------------------
 #
@@ -339,6 +341,8 @@ class TreeDocBase(BaseDoc, TreeDoc):
         self.write(0, '\\usepackage[%s,%s]{geometry}\n' % (paper, margin))
         self.write(0, '\\usepackage[all]{genealogytree}\n')
         self.write(0, '\\usepackage{color}\n')
+        lang_en = _locale._get_language_string(glocale.language[0])
+        self.write(0, '\\gtrset{language=%s}\n' % lang_en.lower())
         self.write(0, '\\begin{document}\n')
 
         if self.nodecolor == 'preferences':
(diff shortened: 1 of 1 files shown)


romjerome:
After forcing to run gramps in english I got an error ...
`\gtrset{language=english (usa)}`
sorry, need to avoid such string as lang set.

romjerome:
before_after
Before change|After

romjerome:
A proper solution might be to use the lang selector available on reports dialog as an option.

romjerome:
Sorry, need to have something like:
`self.write(0, '\\usepackage{lmodern}\n')`
I thought on a separated pull request, but I cannot test local support with an incomplete
characters support for my locale. So it will be added.

romjerome:
latin

romjerome:
Still some specific issues for few characters (not common under my locale!):
"ÿ' on "L'Haÿ-les-Roses", 
Haÿ-les-Roses
'ß' on "Hänßeler"
Hänßeler

gramps/gen/plug/docgen/treedoc.py
```diff
@@ -339,6 +344,12 @@ def write_start(self):
         self.write(0, '\\usepackage[%s,%s]{geometry}\n' % (paper, margin))
         self.write(0, '\\usepackage[all]{genealogytree}\n')
         self.write(0, '\\usepackage{color}\n')
+        trans = glocale.language[0][:2]
+        lang_en = _locale._get_language_string(trans).lower()
```
Nick-Hall: The variable name `lang_en` implies English.  Can we call this `lang`?

gramps/gen/plug/docgen/treedoc.py
```diff
@@ -339,6 +344,12 @@ def write_start(self):
...
+        if lang_en in LANG_SUPPORT:
+            self.write(0, '\\gtrset{language=%s}\n' % lang_en)
+        if lang_en in LATIN:
+            self.write(0, '\\usepackage{lmodern}\n')
```
Nick-Hall: We should check that the package exists as we already do with _libertine_.

romjerome:
I have some issues with libertine package version 5.3.0 (2019-05-20).
Package available under my TeX environment but I need to hide/remove the section.
I do not remember well but it seems that some of these issues were fixed by adding:
`\usepackage{anyfontsize}`. So it has not included on this pull request. (need tests)

Got few issues by only including old 'lmodern' package... 
`\gtrset{language=...}` if for date format and localization.

There is an other issue because I have "german" characters and my locale is french...
So, non-ASCII characters handling is still incomplete but better than before.

gramps/gen/plug/docgen/treedoc.py
romjerome: Ah yes, more clear.

romjerome: There is maybe a problem with code like:
IfFileExists{libertine.sty}! e.g., got an issue with 'shellesc', where " shellesc.sty"
does not exist anymore on recent TeX packages.
There is now (with recent TeX versions) something like compressed "doc and source"
into a "shellesc.dtx" file.
https://www.ctan.org/tex-archive/macros/latex/required/tools

'lmodern' should be a standard (ok, an old one)
https://www.ctan.org/pkg/lm

I did not try 'cm-super'.

gramps/gen/plug/docgen/treedoc.py
romjerome: There is maybe a problem with code like:
IfFileExists{libertine.sty}! e.g., got an issue with 'shellesc', where " shellesc.sty"
does not exist anymore on recent TeX packages.
There is now (with recent TeX versions) something like compressed "doc and source"
into a "shellesc.dtx" file.
https://www.ctan.org/tex-archive/macros/latex/required/tools

'lmodern' should be a standard (ok, an old one)
https://www.ctan.org/pkg/lm

I did not try 'cm-super'.

Nick-Hall:
> There is an other issue because I have "german" characters and my locale is french...

Yes.  Users may wish to display names in a variety of alphabets irrespective of their locale setting.  I found this problem when I attempted to produce a report using the example database which includes names using the cyrillic alphabet.

Libertine was a font recommended by someone of the mailing list.  Their may well be better alternatives available.

PR #961, opened 2019-12-07
[Tree doc Tex] fix "-" char on place name

romjerome:
"-" can lead to confusion, 
generating text out of the box
with PDF file format

### DIFF
--- a/gramps/gen/plug/docgen/treedoc.py
+++ b/gramps/gen/plug/docgen/treedoc.py
@@ -519,7 +519,7 @@ class TreeDocBase(BaseDoc, TreeDoc):
             date_str = date_str + '/' + stop_date
 
         place = escape(_pd.display_event(db, event))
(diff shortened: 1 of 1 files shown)


romjerome:
before
after

romjerome:
sample with patch
sample with patch

romjerome:
Some fixes around date format and non-ASCII char support on screenshots are related to #262

Nick-Hall:
Need to squash the commits prior to merging.

romjerome:
Maybe there is just hyphenation issues on TeX?
https://texfaq.org/FAQ-hyphenaccents
http://www.tug.org/docs/liang/
www.ctan.org/tex-archive/systems/knuth/dist/lib/hyphen.tex
https://www.tug.org/TUGboat/tb05-2/tb10desarm.pdf

PR #962, opened 2019-12-07
Fix import file chooser filter to accept upper case extensions

prculley:
Fixes [#11463](https://gramps-project.org/bugs/view.php?id=11463)

User noted that import file dialog with filter enabled doesn't find all uppercase extensions.  It is possible the original code contained an error, as it was looking for capitalized extension, rather than upper case, and I cannot imagine anyone using capitalized extensions in their file names.

However I decided to simply add the Uppercase extension to the filter criteria.

### DIFF
--- a/gramps/gui/dbloader.py
+++ b/gramps/gui/dbloader.py
@@ -317,6 +317,13 @@ def add_all_files_filter(chooser):
     mime_filter.add_pattern('*')
     chooser.add_filter(mime_filter)
 
+
+def icase(ext):
+    """
+    Return a glob reresenting a case insensitive file extension.
+    """
+    return ''.join(['[{}{}]'.format(s.lower(), s.upper()) for s in ext])
+
 #-------------------------------------------------------------------------
 #
 # Format selectors: explictly set the format of the file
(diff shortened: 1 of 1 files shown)


DomoViridi:
**Fix is not complete.**
- Before the fix, the filter had match patterns "`*.def`" and "`Def`".
- After this fix, the filter has match patterns "`*.def`" and "`DEF`".

Note that "Def" and "DEF" are missing the part for matching the name of the file, and the period.

prculley:
Fixed; sorry about that, got a little too hurried.

prculley:
Once I realized that Brian Matherly had added the filter for Capitalized extension ('Def'), rather than capitalized wildcard, I deliberately decided to leave this in place.  It could have been an error, but I suspected that he know of some use case I don't.  It doesn't hurt anything.

DomoViridi:
> Once I realized that Brian Matherly had added the filter for Capitalized extension ('Def'), rather than capitalized wildcard, I deliberately decided to leave this in place. It could have been an error, but I suspected that he knows of some use case I don't. It doesn't hurt anything.

Ah, I see the misunderstanding. It is not about allowing the capitalized **extension** 'Def' or not. It is the fact that this patch now allows
- All files with the lower case extension 'def' (good)
- All files with the upper case extension 'DEF' (good)
- The single file litteraly called 'Def', *no* extension. (nonsense)

To spell it out: if you also want to allow for the **extension** 'Def' then the line
- `file_filter.add_pattern(plugin.get_extension().capitalize())`

needs to be changed to
- `file_filter.add_pattern("*.%s" % plugin.get_extension().capitalize())`

--------
Speaking more generaly, my personal preference would be to make the matching case-insensitive, rather than setting up a list of possibilities. But I am not sure this is easy to do in this programming environment.

For completeness: in this and previous comments, the extension 'def' and its variants are used as an example, particularly for selecting Pro-Gen files. Gramps allows a choice of many different extensions. The code involved in this bug covers all these extensions.

Nick-Hall:
> Speaking more generaly, my personal preference would be to make the matching case-insensitive, rather than setting up a list of possibilities. But I am not sure this is easy to do in this programming environment.

It's quite easy.  I have added a commit.

PR #964, opened 2019-12-09
Update INCOMPLETE_TRANSLATIONS

sam-m888:
See [Gramps_translations#INCOMPLETE_TRANSLATIONS](https://www.gramps-project.org/wiki/index.php/Template:Gramps_translations#INCOMPLETE_TRANSLATIONS)

* Remove: he (Hebrew)

* Add: br, ga, sr_Latn


Nick-Hall:
Apparently, the Serbian Latin translation can be derived from the Serbian Cyrillic translation.  I remember one of our Serbian translators asking if we could run a script to do this as part of the build process.

PR #965, opened 2019-12-09
Fix Lat/Long translated NSEW display and entry in rtl language;Store Lat/Long in place objects with NSEW always in English

prculley:
https://github.com/gramps-project/gramps/pull/922#issuecomment-562924935 noted that the already commited PR922 is displaying the translated NSEW characters reversed for rtl languages.  This is a consequence of that PRs change to make the entire Lat/Long strings display and enter in ltr mode.

There doesn't seem to be a completely ideal solution to this, as Gtk doesn't handle the mixed Lat/Long strings well at all.  I have decided to manually reverse the rtl portion of the string so the ltr override can display it correctly.  While this looks OK for output, the Lat/Long Entry in the editor doesn't work the way other bidi editors work.  When initially opening the editor with a correct Lat/Long in an rtl language, it looks OK, but if the user moves the cursor through it he will note that the cursor doesn't behave as perhaps expected.

If the user tries to make the text 'look right' while entering translated NSEW rtl characters, he has to enter them in reverse order.  If the user just types blindly, the order will end up correct.  So the lat/long functions are modified to accept the translated NSEW characters for rtl languages in either normal or reverse order.

The second commit modifies the editors to always store the lat/long with the NSEW characters in English.  This is another part of the overall fix for https://gramps-project.org/bugs/view.php?id=11335, it is a start on making sure that the when opening Gramps in another language or XML export/import with Gramps in another language doesn't get bad values for lat/long.

I chose to store the lat/long this way to preserve the users lat/long entries as much as possible; if he enters in Degree, Minutes, Seconds mode, it stays that way , if he enters in decimal degrees it stays that way etc.

Note: the final part of the overall fix is to 'upgrade' the db and modify any current language specific lat/long values to be always in English.  I propose to do this as part of the place enhancement GEPS045.

### DIFF
--- a/gramps/gen/utils/place.py
+++ b/gramps/gen/utils/place.py
@@ -57,19 +57,25 @@ South = South.replace("1", " ").strip()
 East = East.replace("1", " ").strip()
 West = West.replace("1", " ").strip()
 
-# build dictionary with translation en to local language
-translate_en_loc = {}
-translate_en_loc['N'] = North
-translate_en_loc['S'] = South
-translate_en_loc['E'] = East
-translate_en_loc['W'] = West
-
 # keep translation only if it does not conflict with english
 if 'N' == South or 'S' == North or 'E' == West or 'W' == East:
-    translate_en_loc['N'] = 'N'
-    translate_en_loc['S'] = 'S'
-    translate_en_loc['E'] = 'E'
-    translate_en_loc['W'] = 'W'
+    North = 'N'
+    South = 'S'
+    East = 'E'
+    West = 'W'
+# for rtl locales the lat/long strings are always displayed ltr, so we need to
+# reverse the translated NSEW strings to make them correctly display rtl
+# we keep both normal/reversed for comparison on input
+if glocale.rtl_locale:
+    North_ = North
+    North = North[::-1]
+    South_ = South
+    South = South[::-1]
+    East_ = East
+    East = East[::-1]
+    West_ = West
+    West = West[::-1]
+
 # end localisation part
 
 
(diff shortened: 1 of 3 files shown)


romjerome:
Maybe it make sense to use something like UTM coordinates for storage?
http://heuristnetwork.org/using-utm-coordinates-heurist/

Nick-Hall:
I don't think that we should change how latitude and longitude are stored in a maintenance release.  We also shouldn't modify fields that are directly editable by the user.

However, we can consider a new storage format for the next major release.  I favour storing decimals with negative values for South and West.  Some applications already do this, and it is similar to the format proposed in the unofficial Gedcom v5.5.5.

The place editors should then be modified to have a single field for geographic position allowing all current data input formats.  The display format should be chosen by the user.

romjerome:
> I don't think that we should change how latitude and longitude are stored in a maintenance release.

Just thought on an addition for the future. e.g., there is already a RT90 support:
```
    The code below converts to/from the Swedish RT90 koordinate
    system. The converion functions use "Gauss Conformal Projection
    (Transverse Marcator)" Krüger Formulas.
    The constanst are for the Swedish RT90-system.
    With other constants the conversion should be useful for
    other geographical areas.
```
(there is maybe a typo here ? Should be Transverse M**e**rcator, no ?)

and a "gedcom" format:

```
        # The 5.5.1 spec is inconsistent.  Length is supposedly 5 to 8 chars,
        # but the sample values are longer, using up to 6 fraction digits.
        # As a compromise, we will produce up to 6 fraction digits, but only
        # if necessary correct possible roundoff error
```

Nick-Hall:
@prculley This was rejected for the maintenance release.  Please revert and rebase on the _master_branch.

prculley:
Sorry.  I misinterpreted your comment about not suitable for maintenance; I thought you were responding to the Jeromes comment about making the coordinates UTM for storage.

Nick-Hall:
OK.  No problem.  I didn't give it the thumbs up, but I didn't give it the thumbs down either.

Just looking at the last couple of bugs.

PR #966, opened 2019-12-10
options for including occupation data on LaTeX trees

romjerome:


sam-m888:
@romjerome Travis shows that  line: gramps/gen/plug/docgen/treedoc.py:161  has Trailing whitespace in source file.

Nick-Hall:
Looks good.  I haven't tested with multiple occupation events and/or attributes yet though.

romjerome:
About the options (list):
```
_OCCUPATION = [{'name': _("Do not include"), 'value': "no"},
               {'name': _("Only description"), 'value': "basic"},
               {'name': _("Use date"), 'value': "date"},
               {'name': _("Use place"), 'value': "place"},
{'name': _("Use date and place"), 'value': "date place"}]
```
they are extensions, also related to the detail list:
```
_DETAIL = [{'name': _("Full"), 'value': "full"},
           {'name': _("Medium"), 'value': "medium"},
{'name': _("Short"), 'value': "short"}]
```
So either test detail options too or let report check
detail. i.e. "short" will ignore 'profession/occupation' and 'comment'...

About the option list for occupation text maybe to use a more global options?
`from gramps.gen.plug.report import stdoptions`
`...`

I do not remember very well how many occupations were returned,
just made this local improvement for genearting some trees.
Now I have problems for properly get the stable TeX environment,
so testing will be more difficult...

PR #967, opened 2019-12-12
Fixes bug ID 0011470: some date variants are not consider valid

vejrous:
Dates with estimated, calculated, spand and range are not consider valid.


romjerome:
@Nick-Hall 

How can we move it to gramps51 branch as a bug fix?

PR #970, opened 2019-12-17
Simplify get_name_format in the name file of the display module

Baizley:
This PR aims to simplify the get_name_format.

First, the PR brings the method under test coverage.
Second, the duplicated `cmp_to_key` function is 
replaced by the method in functools. 
Third, the sorting method is simplified by abstracting 
it away behind a method that is named after 
the sort order and by collapsing
cases as described [in the commit](https://github.com/gramps-project/gramps/commit/0d292291ff51e0859c29a8a09d9e4b0892046f9e).
Fourth, the steps in the method is reordered so the
name formats are filtered first and then sorted.

### DIFF
--- a/gramps/gen/display/name.py
+++ b/gramps/gen/display/name.py
@@ -495,29 +495,32 @@ class NameDisplay:
         """
         Get a list of tuples (num, name,fmt_str,act)
         """
-        the_list = []
 
-        keys = self.sort_by_ascending_positives_followed_by_negatives_in_reverse_order(self.name_formats)
+        formats = [
+            (index, name, format_string, active)
+            for index, (name, format_string, active, *_) in self.name_formats.items()
+            if (also_default or index) and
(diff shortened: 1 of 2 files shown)


Nick-Hall:
It's nice to see a unit test with this PR.

Baizley:
> It's nice to see a unit test with this PR.

Thanks, it also made it a lot easier to simplify the sorting method.

sam-m888:
FYI travis-ci shows [some  failed test](https://travis-ci.org/gramps-project/gramps/builds/626362071#L2029)

gramps/gen/display/name.py
```diff
@@ -59,6 +59,9 @@
 #-------------------------------------------------------------------------
 import re
 import logging
+from functools import cmp_to_key
+from math import copysign
```
QuLogic: Neither of these imports seem used?

PR #981, opened 2019-12-26
Change Relationship Graph so parents are on same rank

prculley:
A recent change (https://github.com/gramps-project/gramps/pull/733) caused parents to end up on different 'levels' of the graph.

Several users have objected to this.

This PR causes the parents to end up back at the same level.  It doesn't revert the earlier change, the invisible line between parents is still there, which in most cases keeps the parents a bit closer together.


prculley:
I closed this because it did not do a good enough job.  Even though the first family for a parent get put together on the same rank, other families for the same parent do not.

PR #983, opened 2019-12-29
Narrated Web: bad event links on media pages

SNoiraud:
Two problems:

 - Don't use media regions in some case:
  1 - If we don't show families
   2 - If we don't show events

 - When a media is processed in the media display section for an event
page, it's remove from the media_list, so when we try to create the
media regions in the media page, the media list is empty and we
can't display the regions.

Fixes [#11500](https://gramps-project.org/bugs/view.php?id=11500)

### DIFF
--- a/gramps/plugins/webreport/basepage.py
+++ b/gramps/plugins/webreport/basepage.py
@@ -1840,7 +1840,7 @@ class BasePage: # pylint: disable=C1001
                         _linkurl = self.report.build_url_fname_html(_obj.handle,
                                                                     "ppl",
                                                                     linkurl)
-            elif classname == "Family":
+            elif classname == "Family" and self.inc_families:
                 _obj = self.r_db.get_family_from_handle(newhandle)
                 partner1_handle = _obj.get_father_handle()
                 partner2_handle = _obj.get_mother_handle()
@@ -1866,7 +1866,7 @@ class BasePage: # pylint: disable=C1001
                                                                 "ppl", True)
                 if not _name:
                     _name = self._("Unknown")
-            elif classname == "Event":
+            elif classname == "Event" and self.inc_events:
                 _obj = self.r_db.get_event_from_handle(newhandle)
                 _name = _obj.get_description()
                 if not _name:
--- a/gramps/plugins/webreport/basepage.py
+++ b/gramps/plugins/webreport/basepage.py
@@ -1952,57 +1952,16 @@ class BasePage: # pylint: disable=C1001
                     # make a thumbnail of this region
                     newpath = self.copy_thumbnail(photo_handle, photo, region)
                     newpath = self.report.build_url_fname(newpath, uplink=True)
-
                     snapshot += self.media_link(photo_handle, newpath, descr,
                                                 uplink=self.uplink,
                                                 usedescr=False)
                 else:
-
                     dummy_rpath, newpath = self.report.prepare_copy_media(photo)
                     newpath = self.report.build_url_fname(newpath, uplink=True)
-
-                    # FIXME: There doesn't seem to be any point in highlighting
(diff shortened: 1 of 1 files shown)


lordemannd:
I used PR 983, it is working again, Thank You!

Now you can see 2 related issues:
issue 1 - the titles of the Reference Regions are the word "Unknown" or the contents of the Description field for the event. I think the title should be the event title seen in the Reference section of the Media page. With the example DB, follow these steps:
- find Event E0008, edit it and delete the contents of the Description field.
- regenerate the report. (make sure to include the Event pages)
- view the Media page for traditional-dutch-houses-11294081342QxR
- in upper right of the image notice the region with the "Unknown" title.
- at the bottom of the Media page, in the References section, the correct title
   "Residence, Family of LaSalle, Leonard and Cotenoir, Julie"
- I think the reference region should have the same title instead of "Unknown" or the Description field.
- this applies to all of the reference regions attached to Events.

issue 2 - much less important. If the reference region is the is the entire image,
Then it works differently (maybe correctly).
- in Gramps navigate to Family F0002
- to the Family add an event titled "Test-Whole-Image"
- to that event Galarey add the existing image traditional-dutch-houses-11294081342QxR
- don't select any region, just click OK to add the image to the event.
- regenerate the report
- view the Media page for traditional-dutch-houses-11294081342QxR
- notice no title bar for the whole image (maybe OK)
- in the reference section click on "Test-Whole-Image, Family of LaSalle, Leonard and Cotenoir, Julie"
- notice at the top only a small thumbnail image and the inventory of reference regions included within the entire image.

SNoiraud:
> * at the bottom of the Media page, in the References section, the correct title
>   "Residence, Family of LaSalle, Leonard and Cotenoir, Julie"
> 
> * I think the reference region should have the same title instead of either "Unknown" or the Description field.

Difficult to do. Which reference do you want to see?
When we set the field to unknown, We are on the event record. We need to get the backref list of this event. Do I take the first one ? the last one ?

lordemannd:
Is it possible to pick the title where the text up to the 1st comma matches the Event name?

SNoiraud:
In the present case, we have 3 references. gramps is unable to select one of them;
So I need to select either the first one or the last one. These references are not sorted so the reference type is undefined. In your exemple, look at the attached snapshot it seems you want to use the family event


If you want to select an event type, which one? and if we have no references with this event type, what do we do ?

lordemannd:
I took your Commit on Jan 3, 2020, but didn't see any difference.

I tried to use Media Notes to work around the issue, but I can't find a way to get Narrated Web Report to show the notes on the Media pages or with the Event images. I tried the shared and non-shared notes, neither gets displayed.

SNoiraud:
The media notes works correctly for me:

lordemannd:
I will try to reproduce what I am seeing. Can this be in the difference between 5.1.1 and 5.2?

SNoiraud:
There is no differences for this. I changed nothing.
Be careful, the browser cache make me become crazy sometimes.

lordemannd:
When I add an un-shared note to a small region of an image, the note gets a type of "Media Reference Note", these notes don't get displayed. If I add the note to the shared part they get the type of "Media Note" and they get displayed with the original full image.

What I was hoping to do was have an image (like a group of houses or a church book page) that has several small regions, and that the "Media Reference Note" could have text associated only with the small region. For example, a church book page that has several baptisms, and the small referenced region could be displayed large enough to read, with a "Media Reference Note" that can have a transcription and translation of only that small region. And there maybe many such regions on one church book image.

lordemannd:
@SNoiraud, please commit the current changes. I now think my additional comments should be in a feature request and should not hold up the corrections to basepage.py

PR #984, opened 2020-01-01
Update LDS Temple list

prculley:
As it says.  The list has obviously not been updated lately (and had a bug in the XML).


Nick-Hall:
The Travis CI build is failing.

PR #985, opened 2020-01-03
Fix custom paper size in LaTeX reports

rosswhitfield:
The height was always set to the same as width


johnbickmore:
Thanks but it looks like this has already been fixed in the gramps51 branch by https://github.com/gramps-project/gramps/pull/963 and usually the developers merge the changes to master!

PR #987, opened 2020-01-07
Fix GEDCOM export of compound dates estimated/calculated modifers

prculley:
Current Gramps allows the estimated/calculated prefix (EST/CAL) to be added to GEDCOM export for all the other modifier options (before after range span text about).  This is illegal GEDCOM.

```
DATE_VALUE:=
[
<DATE> |
<DATE_PERIOD> |
<DATE_RANGE>|
<DATE_APPROXIMATED> |
INT + space + <DATE> + space + (<DATE_PHRASE>) |
(<DATE_PHRASE>)
]
DATE_APPROXIMATED:=
[
ABT + space + <DATE> |
CAL + space + <DATE> |
EST + space + <DATE>
]
```

See bug [#11513](https://gramps-project.org/bugs/view.php?id=11513).

### DIFF
--- a/data/tests/exp_sample_ged.ged
+++ b/data/tests/exp_sample_ged.ged
@@ -1,12 +1,12 @@
 0 HEAD
 1 SOUR Gramps
-2 VERS 5.0.2
+2 VERS 5.1.2
 2 NAME Gramps
-1 DATE 4 AUG 2019
-2 TIME 15:26:44
+1 DATE 7 JAN 2020
+2 TIME 13:39:43
 1 SUBM @SUBM@
 1 FILE C:\Users\prc\AppData\Roaming\gramps\temp\exp_sample_ged.ged
-1 COPR Copyright (c) 2019 Alex Roitman,,,.
+1 COPR Copyright (c) 2020 Alex Roitman,,,.
 1 GEDC
 2 VERS 5.5.1
 2 FORM LINEAGE-LINKED
@@ -911,7 +911,7 @@
 2 DATE I think 1970 to 1971
 2 PLAC San Francisco, San Francisco Co., CA
 1 SLGC
-2 DATE EST ABT 1999
+2 DATE ABT 1999
 2 FAMC @F0016@
 2 TEMP DENVE
 2 PLAC Denver, Denver Co., CO, Denver Co., Colorado, USA
@@ -1424,7 +1424,7 @@
 0 @N0018@ NOTE Another Citation Note
 0 @N0019@ NOTE A bad photo for sure
 0 @O0000@ OBJE
-1 FILE c:\users\prc\workspace\grampsm\main\data\tests\O0.jpg
+1 FILE d:\users\prc\documents\gramps\data\tests\O0.jpg
 2 FORM jpg
 2 TITL Michael O'Toole 2015-11
 1 NOTE @N0019@
--- a/gramps/plugins/export/exportgedcom.py
+++ b/gramps/plugins/export/exportgedcom.py
@@ -1250,20 +1250,14 @@ class GedcomWriter(UpdateCallback):
         if start != Date.EMPTY:
             cal = date.get_calendar()
             mod = date.get_modifier()
-            quality = date.get_quality()
-            if quality in libgedcom.DATE_QUALITY:
-                qual_text = libgedcom.DATE_QUALITY[quality] + " "
-            else:
-                qual_text = ""
+            quality = None if mod else date.get_quality()
             if mod == Date.MOD_SPAN:
-                val = "%sFROM %s TO %s" % (
-                    qual_text,
+                val = "FROM %s TO %s" % (
                     libgedcom.make_gedcom_date(start, cal, mod, None),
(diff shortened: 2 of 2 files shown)


lordemannd:
I notice that with the xml backup created by 5.1.1, it no longer works to import the 5.1.1 backup into gramps 4.2.8. I get error reporting wizard:
  File "C:\Program Files (x86)\GrampsAIO32-4.2.8\gramps\plugins\importer\importxml.py", line 2547, in start_datestr
    date_value = self.placeref.get_date_object()
AttributeError: 'NoneType' object has no attribute 'get_date_object'

maybe this should be ignored as a problem with 4.2.8, but it could point at some issue with the dates export in 5.1.1.

prculley:
> 
> 
> I notice that with the xml backup created by 5.1.1, it no longer works to import the 5.1.1 backup into gramps 4.2.8. I get error reporting wizard:
> File "C:\Program Files (x86)\GrampsAIO32-4.2.8\gramps\plugins\importer\importxml.py", line 2547, in start_datestr
> date_value = self.placeref.get_date_object()
> AttributeError: 'NoneType' object has no attribute 'get_date_object'
> 
> maybe this should be ignored as a problem with 4.2.8, but it could point at some issue with the dates export in 5.1.1.

This is not related to this PR, so please submit questions like this elsewhere.  I was able to import a current XML to 4.2.8 with example db, so will need a bit more to figure out what is going wrong.

Nick-Hall:
@prculley I seem to remember that this was discussed a few years ago and we decided to allow date quality to apply to date ranges and spans.  This is supported in GEDCOM with a small non-standard extension. I see that someone has linked bug [#8355](https://gramps-project.org/bugs/view.php?id=8355) which fixed an earlier issue.

lordemannd:
done, bug #0011515,
the discussion of the 4.2.8 import thing should be deleted here?

prculley:
GEDCOM allows the quality to be exported on a single date.  DATE ABT 1999, or DATE CAL 1999 or  DATE EST 1999 (the 1999 can have month/day as well).  This we can still do.

What GEDCOM doesn't allow is EST/CAL combined with BEF/AFT/FROM ... TO ... /... TO ... or even DATE EST ABT 1999 which Gramps could generate prior to this fix.  I firmly believe we should meet the GEDCOM specs on export.  Even if we accept some non-standard GEDCOM on import.

If we want non-standard extensions of export data, we should have a way to disable these, or do something to make the user allow them.  I linked the earlier bug report with its patch by kulath as I had looked up the patch that added this.  There was no reason given why this was added to the export...

Nick-Hall:
I wasn't directly involved in the discussion, but I seem to remember that people wanted to indicate that a range was calculated.  This happens quite often when a source contains an age on a given date.

I don't think that we are the only software that supports this feature.  GEDCOM is quite limiting.

Perhaps we could use custom tags of _CAL and _EST instead?

prculley:
Unfortunately GEDCOM custom tags cannot be used for this.  They are only legal as the primary tag, just after the number, not as a modifier to the date.  We could export dates two ways, legally with the usual DATE tag, and with an Gramps specific tag like _DATE.
2 DATE AFT 1999
2 _DATE EST AFT 1999
I don't see the point.  Who (what program) would be able to read this?

I can understand the desire to have Gramps have the estimate/calculated available for all kinds of dates, but that is in our sandbox.  I don't really expect other programs to support such an esoteric feature.

prculley:
If you want to dig back into the history of EST/CAL see https://gramps-project.org/bugs/view.php?id=001210 https://gramps-project.org/bugs/view.php?id=001211 https://gramps-project.org/bugs/view.php?id=001848.  These are all about getting EST/CAL properly supported on standard dates.  I suspect that no one though about whether they should also be applied to all date forms at the time.  The GEDCOM 5.5.1 spec is quite clear on this issue, it may be earlier specs were not.

Nick-Hall:
I was reading the "Extending the Form" section on page 121 of the new GEDCOM 5.5.5 specification.  The GEDCOM specification has always been open to interpretation, and every product seems to have its peculiarities, not just Gramps.  As you noticed, the feature was introduced about 10 years ago, but it has been discussed more recently.

I have created a new [forum topic](https://gramps.discourse.group/t/date-quality-in-gedcom-exports/130) to discuss this further.  I am not a GEDCOM expert and don't use the export myself, so I am happy to listen to the views of the community.

Nick-Hall:
A reply to my forum post suggests exporting these values as a date phrase to prevent data loss.  For example:
`2 DATE (EST AFT 1999)`
I also read in the Gedcom 5.5.5 specification that PAF exports date phrases without the parentheses.

Nick-Hall:
Since there was no response to my forum post that suggested anyone is actually using this extended functionality, I think we should conform to the Gedcom 5.5.1 specification using the unofficial 5.5.5 specification for clarification where necessary.

Translator5:
Btw. whats the current status of gedcom 6.0 support?

Nick-Hall:
Gedcom 6.0 was only a draft specification.  The latest, but unofficial, version is [5.5.5](https://www.gedcom.org/gedcom.html).

PR #988, opened 2020-01-08
Try to fix exceptions on ManagedWindow close

prculley:
Issues #10252, #10642, #10821, #11163, #11440, #11476, #11482, #11508

This one has been annoying me for quite a while.  What all these bugs have in common is that they happen during a ManagedWindow.close.  And that they are not easily repeatable.  In fact I have never been able to repeat any of these despite a lot of trying.  The bug submitters also indicate that they happen intermittently.

There are two basic stack dumps in the list of bugs, when I analyzed each, the only way I can think of that the error can occur is if somehow the ManagedWindow.close has gotten called more than once, so that the code gets exceptions trying to do something that has already been done.

So assuming that this is indeed what is happening, I have added a test to the close code to see if it has already been done, if so skip it.  I have tested this and it seems to have no effect on normal close, stacked window close etc.  But if my analysis is correct, it should stop the exceptions.

### DIFF
--- a/gramps/gui/managedwindow.py
+++ b/gramps/gui/managedwindow.py
@@ -32,6 +32,14 @@ the create/deletion of dialog windows.
 import os
 from io import StringIO
 import html
+import logging
+
+#-------------------------------------------------------------------------
+#
+# Set up logging
+#
+#-------------------------------------------------------------------------
+_LOG = logging.getLogger(".")
 #-------------------------------------------------------------------------
 #
 # GNOME/GTK
@@ -575,6 +583,9 @@ class ManagedWindow:
 
         Takes care of closing children and removing itself from menu.
         """
+        if hasattr(self, 'opened') and not self.opened:
+            _LOG.warning("Tried to close a ManagedWindow more than once.")
+            return  # in case close somehow gets called again
         self.opened = False
         self._save_position(save_config=False)  # the next line will save it
         self._save_size()


lordemannd:
in this case, would it make sense to instrument it somehow, to see some record if it ever happens?
like a one line message written to the log file?

lordemannd:
I just got this while trying to import a gramps backup file:
32715: ERROR: grampsapp.py: line 157: Unhandled exception
  File "C:\Program Files\GrampsAIO64-5.1.1-working\gramps\gui\viewmanager.py", line 945, in import_data
    self.db_loader.import_file()
  File "C:\Program Files\GrampsAIO64-5.1.1-working\gramps\gui\dbloader.py", line 132, in import_file
    callback=self.set_info)
  File "C:\Program Files\GrampsAIO64-5.1.1-working\gramps\gui\dbloader.py", line 483, in __init__
    self.close()
  File "C:\Program Files\GrampsAIO64-5.1.1-working\gramps\gui\managedwindow.py", line 578, in close
    if not self.opened:  # in case close somehow gets called again
AttributeError: 'GrampsImportFileDialog' object has no attribute 'opened'

prculley:
@lordemannd Looks like you got the PR just as I was updating it to fix the bug.  Seems that some ManagedWindows don't use the 'show()' and thus don't get 'self.opened' created.

Regarding the 'instrumentation', just how far should I go?  If I just print out the stack in a LOG.warning, I just get the stack I got in the original bug reports, which are not particularly helpful.  Of more help might be to store a stack dump on the first (normal) close, and then print out or LOG.warning the first and current stack dump in the second (abnormal) close.  But even then it would require users to write up a report if they saw the warning.

I'm open to suggestions.

lordemannd:
Regarding the 'instrumentation': i was thinking simpler, just a line that says the correction was done.
Then if a similar issue is ever reported, and you can see the line that shows the correction action was taken, then you know it didn't solve the issue. (I guess not very well thought out)

Nick-Hall:
Just an indication that the close method was called twice is probably sufficient.

prculley:
Log.warning added...

PR #989, opened 2020-01-08
EditPrimary Windows: Always Call _do_close Before callback

stevenyoungs:
There are several types of window derived from EditPrimary, all of which take an optional callback to be called after saving the object. They differ in that

- EditRepository does not initalise self.callback and does not attempt to.
- EditFamily, EditPerson, EditPlace and EditSource close their GUI elements before calling the callback
- EditCitation, EditEvent, EditMedia and EditNote call the callback before closing their GUI elements

This PR ensures that all types of EditPrimary consistently close their GUI elements before calling the callback. This allows the callback to have certainty, for example if it wishes to updates its own GUI elements. This capability is used in PR267 of the addons-source repo.

A separate PR, #986, updates EditEvent to initialise self.callback.

### DIFF
--- a/gramps/gui/editors/editcitation.py
+++ b/gramps/gui/editors/editcitation.py
@@ -341,9 +341,9 @@ class EditCitation(EditPrimary):
                         self.obj.set_gramps_id(self.db.find_next_citation_gramps_id())
                     self.db.commit_citation(self.obj, trans)
 
+        self._do_close()
         if self.callback:
             self.callback(self.obj.get_handle())
-        self._do_close()
 
     def data_has_changed(self):
         """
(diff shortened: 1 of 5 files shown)


stevenyoungs:
Some types of EditPrimary have
`self.callback = None`
after calling the callback. Is this necessary?

prculley:
> self.callback = None

These were added about 11 years ago in an effort to reduce memory leaks due to circular references.  Early versions of Python had a hard time garbage collecting these.  Modern versions can clean them up without such 'fixes', but the memory gets released faster (and probably with less overhead) with the circular reference broken.

I generally don't try to change stuff like this unless it is obviously broken, as I sometimes miss unexpected side effects.

PR #990, opened 2020-01-10
bug 0011078: 'Default person' instead of 'Home person'

giansalvo:
- modified button hints
- modified filter rules names and descriptions
- modified .pot and .po files

PR #991, opened 2020-01-11
Fix bug 0011078: second part

giansalvo:
modified pot and po files
see first part in PR https://github.com/gramps-project/gramps/pull/990


Nick-Hall:
Please add this commit to PR #990 instead of opening a new PR.

giansalvo:
added to PR https://github.com/gramps-project/gramps/pull/990

PR #994, opened 2020-01-12
Add option that allows to use the name of the individual as title.

tengel:
The Complete Individual Report uses as first title 'Complete Individual Report' and as second title the name of the person. When printing the report for many persons, every page starts with these two titles.
This patch adds an option to use the name of the person as first title and omit the title 'Complete Individual Report'.


Nick-Hall:
This looks like a reasonable enhancement.

PR #995, opened 2020-01-12
Fix StyledText so serialize will match for style list order changes

prculley:
Fixes [#11529](https://gramps-project.org/bugs/view.php?id=11529)

A minor bug; if a note has a lot of styled text (several different types) it can fail the EditNote (EditPrimary) check for data_has_changed, even though the data has not in fact been changed.

This is due to the fact that the SyledTextBuffer code puts our StyledText into the Gtk model and then pulls it back out.  The StyledText tag list gets recreated in the process, from data in the buffer, put into a dict and eventually back into the tag list.  Since the dict ordering (and possibly the Gtk returned tag ordering) was not guaranteed the returned tag list is often different in order than when it was first saved in the db.

When EditPrimary test for data changes in data_has_changed, it uses the object.serialize() to standardize the original and updated objects for comparison.  The sometimes reordered tag list doesn't match in this test.

I considered several ways to fix this, but because the db and XML files contain tag lists in various orders, I decided the best way was simply to sort the tag list as it is serialized.  The order of the tag list makes no difference to the actual functionality.

### DIFF
--- a/gramps/gen/lib/styledtext.py
+++ b/gramps/gen/lib/styledtext.py
@@ -299,6 +299,7 @@ class StyledText:
         """
         if self._tags:
             the_tags = [tag.serialize() for tag in self._tags]
+            the_tags.sort()
         else:
             the_tags = []
 
--- a/gramps/gen/lib/test/styledtext_test.py
+++ b/gramps/gen/lib/test/styledtext_test.py
@@ -64,11 +64,11 @@ class Test1(unittest.TestCase):
         C = self.C.join([self.A, self.S, deepcopy(self.B)])
         C = C.replace('X', StyledText('_', [self.T3]))
         _C = ('123_456\ncleartext\nabc_def',
-              [((1, ''), 'v1', [(0, 2), (2, 3)]),
-               ((0, ''), 'v3', [(3, 4)]),
+              [((0, ''), 'v3', [(3, 4)]),
+               ((0, ''), 'v3', [(21, 22)]),
+               ((1, ''), 'v1', [(0, 2), (2, 3)]),
                ((1, ''), 'v1', [(4, 6)]),
                ((2, ''), 'v2', [(19, 21), (18, 21)]),
-               ((0, ''), 'v3', [(21, 22)]),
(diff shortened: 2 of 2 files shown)


wroldwiedbwe:
one of the test failed 

[FAIL: test_replace (gramps.gen.lib.test.styledtext_test.Test1)](https://travis-ci.org/gramps-project/gramps/builds/636053158#L2026)

Nick-Hall:
> The order of the tag list makes no difference to the actual functionality.

Did you check this?  Suppose you have the text "ABC" and you set "AB" to red and then "BC" to blue.  Does it change the range of the red characters to just "A"?

prculley:
You are correct, overlapping ranges of colors are sensitive to ordering.  Which I did not think of when I created this.

I suspect that we have not noticed color changes when this occurs over different versions of Python.  I understand recent versions of Python dict (3.6+) now behave as ordered dict.  And I understand the Gtk itself does a sort on the tags by the priority that they are created according to the docs.  So I will have to go back and work a bit harder to figure out why what is going into the Gtk TextBuffer is not matching what comes out.

The travis test failure was due to my sorting, and easily fixed.

prculley:
After further review, overlapping style ranges of the same type (example color) are not possible.  The StyledTextBuffer code removes any StyledTextTag that matches the newly set StyledTextTagType prior to setting the new StyledTextTag.  This means that for any given range in the text, there can be only one LINK, FONTFACE, FONTCOLOR, HIGHLIGHT, or FONTSIZE.

As a result, I now stand by my original patch, which sorts the various StyledTextTags in the list to make sure that when serializing, the list is always the same, no matter what Python, Gtk and our code does with it.

P.S. As an aside, I found that coming out of the text buffer the tags end up put into the dict in the order found in the text as the string is scanned, regardless of the order the tags were added.  For recent versions of Python 3.6+ they should have stayed that way when converted to a list, so I suspect that the example.gramps N0013, N0014 tag order was created with an older version of Python.

Nick-Hall:
Thanks for checking.

PR #996, opened 2020-01-13
Symbols enhancement.

SNoiraud:
Add the possibility to set all default values to a string.
Gui configuration improvement.
Possibility to drag and drop a symbol from the symbol list.
In this case, the glyph must be present in the current font to be displayed.


SNoiraud:
Here is a snapshot of what we can see:

symbols 1

prculley:
I tried this out and I like parts of it.  But I decided that the Preferences tab could use some work, it was too large to fit on smaller screens.
I also added more tooltips, including on the symbol characters.
I've created a commit with the changes https://github.com/prculley/gramps/commit/526f89fb757e82575cfddb67a9f4b75dbad82d5d 
The commit only affects the Preferences.

I remain concerned that using the symbols sets all of Gramps to the specified font; this is somewhat counter to the style guidelines.  I note that in most cases this is not really necessary, the Gtk/fontconfig font substitution seems to find the necessary character glyphs, in fact I have not seen a case where this did NOT work.

But if you think we really need to specify a font, wouldn't it be better to use markup with the symbols, rather than make the whole GUI change?  I realize that this would require changes in a lot of places, replacing 'set_text' methods with 'set_markup' or similar.  But I think the end result might be better.  We could include the necessary markup in the symbol strings...

Nick-Hall:
Yes.  We shouldn't really change the font for the entire interface.

Perhaps we could create a subclass of `Gtk.Label` to display symbols, or the text fallback depending on the preference settings?  We would probably also need the markup for use in charts which write directly to a `Gtk.DrawingArea`.

SNoiraud:
> in fact I have not seen a case where this did NOT work.

Did you try this on old OSes without unicode in standard ?
Many users continue tu use XP machines or something else.

SNoiraud:
Symbols small 1
Symbols small 2
Symbols default 1

prculley:
> 
> 
> > in fact I have not seen a case where this did NOT work.
> 
> Did you try this on old OSes without unicode in standard ?
> Many users continue to use XP machines or something else.

The AIO (for Windows) has not supported winxp for several years now, probably sometime in the Gramps 4.0.x timeframe.  I would be more concerned about OSX or possibly some older versions of Linux, if they did not support fontconfig.

jralls:
Gnome and MacOS X/10 have fully supported Unicode since their inception, no worries there. 

On MacOS fontconfig is a [small problem](https://gramps-project.org/bugs/view.php?id=11644) as I didn't know about the dependency on [python fontconfig](https://pypi.org/project/Python-fontconfig/) and don't include it in the Mac bundle. It's unfortunate that it's necessary because fontconfig has only a minimal annotation for gobject-introspection that doesn't include the API you need.

On the other hand Apple provides a typeface,  Apple Symbols, that I think provides glyphs for all of the genealogy code points. It has only symbols so it's not suitable for setting application-wide, but if you can fix Gramps so that it uses the symbol font only for the symbols then we can hard-code the Apple Symbols font for MacOS and skip the search and its dependency on fontconfig.

prculley:
I spent some time trying to figure out how we might avoid changing the font for the whole interface; and I understand that the changes would be at least extensive, bordering on impossible.  Too many places that don't support markup.

I decided to test out your latest commit (using the default font).  I observed a problem with the font section combo, due to a mixup in the row to which it was attached.  I fixed that and then decided to expand on the idea of the default font.  
I added an item to the font selector to allow the user to revert to the default.
I changed the code to allow the symbols to be used even if fontconfig was not installed.  And allow death symbol selection even if no font was found/selected.  Also simplified the death symbol code a bit.
I fixed up the fontconfig 0.6.0 code to work right (need dict.items() the way you used it).
I moved the 'no fonts' warning to be next to the 'Find font' button, and updated the text a bit.
I fixed up the self.sym_text.get_iter_at_position code to use the returned iter correctly, apparently different versions return these in a different list order; but using the dict value should always work.

I also changed the text of some messages a bit; I think some of these didn't sound right to my American English ear.
My changes are https://github.com/prculley/gramps/commit/184d587a5521f29c5899066e6a7af9948f447ec4
Thanks for starting the work on this; I was not looking forward to trying to get this improved.  But I hope that between your ideas and mine, the users will be happier.

gramps/gui/configure.py
```diff
@@ -2069,17 +2058,19 @@ def can_we_use_genealogical_fonts(self, obj):
...
                         fontname = fam
                 if not fontname:
                     fontname = font.family[0][1]
             else:
+                # fontconfig version 0.6.0 uses dict
                 if local in font.family:
                     fontname = font.family[local]
                 else:
-                    for lang, name in font.family:     # version 0.6.0 use dict
+                    for lang, name in font.family:
```
prculley, line 2073: I think this needs to be `for lang, name in font.family.items()` otherwise you only get the key, not key and data.

gramps/gui/configure.py
```diff
@@ -2168,8 +2150,6 @@ def activate_change_font(self, obj=None):
                 self.find.set_sensitive(True)
                 if self.choosefont:
                     self.choosefont.set_sensitive(True)
-                if self.combo:
-                    self.combo.set_sensitive(True)
```
prculley: Also need to remove the other `if combo:` above at about 2145.

I guess you did not use a "git cherry-pick" to grab my commit ;-).  Thanks for putting up with my comments.

SNoiraud: done

SNoiraud: done

gramps/gui/configure.py
SNoiraud: done

SNoiraud:
When I started this, a few fonts contained all the symbols. It appears now this is no more the case.
I propose to remove the python-fontconfig pre-requisite.
We use the standard font defined for you gui.
In this case, perhaps we could apply this PR to 5.1.3 instead of 5.2
This will solve the new bug request (#11736) for new debian and ubuntu and the problem for mac.

jralls:
You don't really need a python wrapper for most fontconfig commands, they're stand-alone programs. (Try `ls /usr/bin/fc-*`.) Just shell out (`subprocess.Popen`) and run them.

SNoiraud:
> You don't really need a python wrapper for most fontconfig commands, they're stand-alone programs.

Yes, I know that and began to write a patch on that. I saw I had 7900 fonts on my system and only 76 didn't have all the symbols. This is why I propose to suppress that and use the standard font defined for the user desktop. Could this generate problems on macs ?

jralls:
Depends on what you mean by problems. On a stock Mac most of the miscellaneous symbols block (U+2600-U+26FF) are provided only by that Apple Symbols face I mentioned earlier.

I guess the simplest work-around is to find some Free TTF that supports the whole basic plane, include it in the bundle, and make it the default for MacOS.

prculley:
@jralls On the MAC installation does Gtk use fontconfig to do the font substitution?  On my systems this seems to work pretty well, as it pulls the necessary glyphs from the closest font that contains the symbol.  So I would expect that you would end up with Glyphs from 'Apple Symbols' even when a different font was used as the primary one.

jralls:
@prculley It doesn't do font substitution at all on windows or MacOS unless one forces pango to use the ft2 backend; normally pango uses the native backends (pangowin32 and pangocoretext) neither of which do substitution.

prculley:
@jralls I guess I should be more specific; the way Gramps AIO is built on Windows (with msys2/mingw) it looks a lot like Linux.  It DOES use the fontconfig backend as the default backend.  I don't think anyone has ever successfully built Gramps with the MSoft toolset, so the backend used there is not interesting.

But that answers my question about MacOS; no font substitution.  So we need a good answer for there.

jralls:
@prculley, Not necessarily. Pango does have a `use_fontconfig` argument to force using the fontconfig backend instead of the CoreText one. That's a change from the autotools build where it would build pango-ft2 if it found it *as well as* the native backend. I can try that, but I need a good test to see if it does any good. Do you know of one, ideally in C, or can you sketch out how one would work?

prculley:
What I know about 'C' and building is just enough to get me in trouble... I've got the Windows code to force fontconfig by inserting:
`    if(win() and ('PANGOCAIRO_BACKEND' not in os.environ)):
        os.environ['PANGOCAIRO_BACKEND'] = "fontconfig"
`
In grampsapp.py run function.  Maybe the easiest way to do this is to include MacOS in the 'if'???  But only if the necessary code is included in the build...

jralls:
That might work,` libpango-ft2` is in the bundle already and `libpangocairo` links to it. If not I can try building pango with forcefontconfig set.

But like I said, I need a way to test it. Is there a string somewhere in in `data.gramps` with one or more of the symbols?

prculley:
This PR in its current form has the string of all the symbols in the Genealogical Symbols tab of Preferences.  Just make sure that the font selected is 'default'.  Or paste
♀  ♂  ⚪  ⚢  ⚣  ⚤  ⚥  ⚦  ⚲  ⊛  *  ~  ⚬  ⚭  ⚮  ⚯  ⚰  ⚱  ⚔  ‡  ☠ ☥☦ ☧ ☨ ☩ ☪ ♰ ♱  ✚  ✝  ✞  ✠  ✡  Dead 
into a Note.  BTW, in my browser, some of these look odd, colored etc.  But they look more natural in Gramps.

jralls:
Well, I tried setting `os.environ['PANGOCAIRO_BACKEND'] = "fontconfig"` in the environment and it crashed harfbuzz, so I guess I need to rearrange the build a bit.

jralls:
FWIW I pasted your string into an unadulterated Gramps Notes window and most of them worked. The ones that didn't were
⚪,  ⚢ ,⚣, ⚤, ⚦, ⚭, ⚮, ⚯, ✝, ✞, ✠, and ✡. The rest are supported by a more full-featured face, "menlo", and that might be the one that Gramps is using. I'll have to dig a bit to figure that out.

SNoiraud:
I close this PR and make a new one for 5.1.3

PR #998, opened 2020-01-14
Fix Date Display so that it uses LC_TIME if defined

prculley:
Fixes [#11230](https://gramps-project.org/bugs/view.php?id=11230)

The user had a system of environment variables set up so that LC_TIME did not agree with LANG.  A bit unusual, but possible.

Gramps date parser was parsing dates so that it was according to LC_TIME language settings, however the date display was still using the overall LANG setting.  As a result, all dates displayed in LANG, and, when Editing an Event, the displayed date was shown as invalid (red) if it contained any words.

This causes Gramps to use the LC_TIME setting for date display as well as parsing.

### DIFF
--- a/gramps/gen/datehandler/__init__.py
+++ b/gramps/gen/datehandler/__init__.py
@@ -35,6 +35,7 @@ import logging
 # Gramps modules
 #
 #-------------------------------------------------------------------------
+from ..utils.grampslocale import GrampsLocale
 from ..const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.sgettext
 # import prerequisites for localized handlers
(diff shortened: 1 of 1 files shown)


SNoiraud:
It works partially for me : 
export LC_TIME=fr_FR.utf8
LANG=en_GB.utf8 python3 Gramps.py
Here what I get ;
bug11230 1

The event view is correct. But If I click on the event, The EditEvent show the invalid date.

SNoiraud:
To work correctly, you must do the same in the init_strings method in the _dateparser.py

Here is the patch:
diff --git a/gramps/gen/datehandler/_dateparser.py b/gramps/gen/datehandler/_dateparser.py
index d123279d5..2519bf1b7 100644
--- a/gramps/gen/datehandler/_dateparser.py
+++ b/gramps/gen/datehandler/_dateparser.py
@@ -398,6 +398,7 @@ class DateParser:
         .. seealso:: :class:`.DateParserRU` as an example.
         """
         _ = self._locale.translation.gettext
+        self._locale = GrampsLocale(lang=self._locale.calendar)
         self.__init_prefix_tables()
 
         self._rfc_mon_str = '(' + '|'.join(list(self._rfc_mons_to_int.keys())) + ')'

SNoiraud:
And now we have:
bug11230 2

prculley:
@SNoiraud Sigh...  I guess I should not have tested with Swedish, it appears that it has some differences from the French and others.  I found a slightly better place for the patch that covers both cases.

I think it works now, tested with five different languages.

SNoiraud:
OK. it works for me.

PR #999, opened 2020-01-14
Update tag.py

Mattkmmr:
- translation for menu strings added
- escape illegal characters in tag_name


Nick-Hall:
This seems like a bug rather than an enhancement to me.  Please rebase onto the _gramps51_ branch.

Mattkmmr:
@Nick-Hall The bug was introduced with #832 and [merged ](https://github.com/gramps-project/gramps/commit/ab52d0988e4ca29e8d3df8cf41d2561585769806) into _master_ not _gramps51_.

Nick-Hall:
Sorry, I forgot that this feature was only in _master_.

PR #1001, opened 2020-01-16
Add Person Gallery Vertical

giansalvo:
I added the new class PersonGalleryVertical which inherit from PersonGallery and the new gramplet Person Gallery Vertical to show media object vertically. If this idea is accepted I will do the same for the other galleries (Places, Families, etc)

See bug [#11527](https://gramps-project.org/bugs/view.php?id=11527).

### DIFF
--- a/gramps/plugins/gramplet/gallery.py
+++ b/gramps/plugins/gramplet/gallery.py
@@ -1,7 +1,8 @@
 # Gramps - a GTK+/GNOME based genealogy program
 #
 # Copyright (C) 2011 Nick Hall
-# Copyright (C) 2011       Tim G L Lyons
+# Copyright (C) 2011 Tim G L Lyons
+# Copyright (C) 2020 Giansalvo Gusinu
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


gramps/plugins/gramplet/gallery.py
```diff
@@ -49,7 +51,10 @@ def build_gui(self):
         Build the GUI interface.
         """
         self.image_list = []
-        self.top = Gtk.Box(spacing=3)
+        if self.direction == "Horizontal":
+            self.top = Gtk.Box(spacing=3)
+        else:
+            self.top = Gtk.VBox(spacing=3)
```
Nick-Hall: This has been deprecated.  Use `Gtk.Box(spacing=3, orientation=Gtk.Orientation.VERTICAL)` instead.

Nick-Hall:
Thanks for your contribution, however, I'm not sure that creating a new plugin is the best way to solve this issue.  I'll investigate and come up with a suggestion.

giansalvo:
> 
> 
> Thanks for your contribution, however, I'm not sure that creating a new plugin is the best way to solve this issue. I'll investigate and come up with a suggestion.

@Nick-Hall  Thanks for your feedback. Another way would be to make the gramplet differentiate if it is loaded in the sidebar. This would require to maybe change the opts structure of the gramplets. I don't know if this is a good way either. Let me know your directions and I'll be happy to implement the changes.

prculley:
Another idea would be to consider using a Gtk.IconView.  I think this would work in either orientation, filling the available space the way file managers display icons.

Nick-Hall:
Good idea.  This is certainly worth trying for the gallery gramplets.  Other gramplets may benefit from a horizontal and vertical format, but we can think about that later.

Nick-Hall:
Replaced by PR #1370.

PR #1002, opened 2020-01-17
GEDCOM export, add _UID when not present

prculley:
Many programs read and understand the _UID tag on INDI and FAM GEDCOM files.  This tag allows the program to positively identify a person or family when merging.  Gramps has been able to import these tags and add them as attributes.  A recent PR https://github.com/gramps-project/gramps/pull/1000 takes advantage of this in our own Find Possible duplicate people tool.

This extends our own export to attach these tags, when the user selects the option on GEDCOM export.
The _UID records are added to the GEDCOM file, and the matching _UID attribute is added to any Persons or Families that do not already have the attribute in the db so that future GEDCOM imports of derived trees can use the Find Possible duplicate people tool to assist merging.

The addition can be Undone after the GEDCOM export if desired.

### DIFF
--- a/data/tests/exp_sample_ged_uid.ged
+++ b/data/tests/exp_sample_ged_uid.ged
@@ -1,0 +1,1502 @@
+0 HEAD
+1 SOUR Gramps
+2 VERS 5.1.2
+2 NAME Gramps
+1 DATE 31 DEC 1969
+2 TIME 18:00:00
+1 SUBM @SUBM@
+1 FILE C:\Users\prc\AppData\Roaming\gramps\temp\exp_sample_ged_uid.ged
+1 COPR Copyright (c) 1969 Alex Roitman,,,.
+1 GEDC
+2 VERS 5.5.1
+2 FORM LINEAGE-LINKED
+1 CHAR UTF-8
+1 LANG English
+0 @SUBM@ SUBM
+1 NAME Alex Roitman,,,
+1 ADDR Not Provided
+2 ADR1 Not Provided
+1 PHON 666-555-4444
+1 EMAIL an_email@@gmail.com
+0 @I0000@ INDI
+1 NAME Anna /Hansdotter/
+2 GIVN Anna
+2 SURN Hansdotter
+2 NICK Annanana
+1 NAME Anna Nana /Hansdotter/
+2 TYPE aka
+2 GIVN Anna Nana
+2 SURN Hansdotter
+1 SEX F
+1 BIRT
+2 TYPE Birth of Anna Hansdotter
+2 DATE 2 OCT 1864
+2 PLAC Löderup, Malmöhus Län, Sweden
+1 DEAT
+2 TYPE Death of Anna Hansdotter
+2 DATE 29 SEP 1945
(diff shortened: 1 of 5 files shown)


Nick-Hall:
I don't think that we should write to the database during an export.  The UID should be added either when a new object is created or when an object with an existing UID is imported.

Consider using the python `uuid` module to create the UID and add the checksum on export.

We could also consider adding a new `uuid_list` field to Person and Family objects.

prculley:
I considered several of your objections when designing this.

As to when to add the UID, what about people who already have an extensive database?  If we only add at object creation, when will these get a UID?

I considered using the UUID module, even read its code.  One concern was that, depending on which version used, you either get a high overhead (it runs a system command to get a MAC address, which opens up questions about anonymity), does high overhead hash commands, or just gets a random number (which cannot be seeded, making it difficult to create a repeatable export test).  And it returns a format which must be condensed (removing "{}-" chars).  Which is why I chose to just use a random number.

Our current import pays no attention to checksum, it just adds it as part of the stored string.  So if we only add checksum on export, we need to be ready to deal with two types of UID, with/without.

My general strategy was to NOT modify the UID string found, even if invalid (I never actually look at the checksum, since some programs don't add one).  Guarantees that what comes in, goes out.  This code is in the other PR.

I had read Tamuras comments, and the somewhat conflicting comments determined by the GEDCOM L group, and decided that for matching purposes and maximal ability to use what comes in this is the best way to go.

Nick-Hall:
> As to when to add the UID, what about people who already have an extensive database? If we only add at object creation, when will these get a UID?

During the database upgrade.

> I considered using the UUID module, even read its code. One concern was that, depending on which version used, you either get a high overhead (it runs a system command to get a MAC address, which opens up questions about anonymity), does high overhead hash commands, or just gets a random number (which cannot be seeded, making it difficult to create a repeatable export test). And it returns a format which must be condensed (removing "{}-" chars). Which is why I chose to just use a random number.

Yes.  We would just use it like a random number generator.  The integer value is available in `uuid.int`.  Unit tests would mock the uuid interface.

> Our current import pays no attention to checksum, it just adds it as part of the stored string. So if we only add checksum on export, we need to be ready to deal with two types of UID, with/without.

It probably should validate the checksum, if present, and warn/reject on failure.  No hurry to implement this though.  All valid formats should be accepted.

> My general strategy was to NOT modify the UID string found, even if invalid (I never actually look at the checksum, since some programs don't add one). Guarantees that what comes in, goes out. This code is in the other PR.

My preference would be for Gramps to store the UID only and export to Granps XML as a formatted UUID.  Tamura Jones recommends exporting to Gedcom with the checksum for maximum compliance, but in the future we could offer different formats.  The FHISO ELF format may recommend another format for example.

We should ignore the checksum when matching.  Two people collaborating and using different software may export the same UID in different formats.

prculley:
OK.  Would you mind if I base this off of the GEPS045 (Place enhancements) PR?  It already has upgrade code, and if I'm going to put UID in the base person/family objects I will need to adjust the schema stuff as well.  Might be less work in the final merge.

Nick-Hall:
We should keep this PR separate.

PR #1007, opened 2020-01-25
Optional img_dpi param for add_media in libcairodoc

Mattkmmr:
This optional parameter allows adding images with a different DPI resolution than the document DPI in cairodocgen. This is needed to fix issue [11492](https://gramps-project.org/bugs/view.php?id=11492) where it was found that media included in PDF documents had a low quality and is blurred.

As far as I understand it, [plugins/docgen/cairodoc.py at line 70](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/docgen/cairodoc.py#L70) defines the cairo document DPI which is also used in [plugins/lib/libcairodoc.py class GtkDocPicture](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/lib/libcairodoc.py#L1036). Default value is 72 DPI so larger images get scaled down and blurred.

To bypass this problem I added a new optional parameter img_dpi to [add_media](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/lib/libcairodoc.py#L1584) which is then used in GtkDocPicture to scale the loaded pixelbuf image differently to the rest of the document. So far this solution seems to work fine, but it needs more testing.


prculley:
Instead of yet another parameter to confuse our users, can't we pick abetter default value, like 300dpi?  I think that would print well on the reports, and look ok even when zooming into a pdf.  The current code has been that way forever, and I think the bug submitter and you are the only ones who ever really complained.

Mattkmmr:
The optional parameter only concerns developers. Users don't have to be confused with it at all. It depends on the design of an addon e.g. I added it as a NumberOption to MediaReport, so the new paramater can easily be tested, but it is also possible to just use a hard-coded value.

I think so far the only use for images in reports were very small thumbnail-like pictures, so the problem wasn't obvious. If a bigger file size is no problem for PDF-reports then we could just pick 300dpi instead. I have no objections against that.

Mattkmmr:
@prculley @Nick-Hall Should I rewrite this pull request to a fixed DPI value or keep the optional parameter?

lordemannd:
My narrow view is to request keeping this, and if we want to limit it, just use a fixed value (300dpi) in [MediaReport]: New image DPI option #285. I have been using this since Jan 26, and haven't seen any side effect. (but have not done extensive testing of all reports) If we set it to 300 here, wouldn't that make all reports with images larger? (instead of just the MediaReport)

lordemannd:
can this get in to the upcoming release?

lordemannd:
It looks like we tried to merge this into Gramps51 but I can't find it, what am I doing wrong?

jralls:
@lordemannd You're reading that last commit backwards: It's @Mattkmmr trying to sync his branch with gramps51, not merging the PR into gramps51.

@Mattkmmr Please rebase to get rid of the merge and force-push. Merging the target into a feature branch creates undesirable history loops.

Though perhaps @prculley should weigh in first, no point in doing more work if the PR is rejected.

prculley:
First of all, I see this as an new feature, which means it must (according to our rules) be in a major release, not a maintenance release.  So maybe 5.2.  As I said earlier in the comments, I would prefer a more reasonable DPI by default, rather than a new parameter.  But I'm not the only one who's opinion matters...

It does seem to be set for 5.2 milestone, so should be rebased to the master branch.

Nick-Hall:
Yes.  I agree that this is a new feature and should be rebased on the _master_branch.

Would a better default, as suggested by @prculley, solve the problem?  If not, and an extra parameter would prove useful, then I have no objections.

Mattkmmr:
@Nick-Hall @prculley changing the default DPI works too and is an easier solution, but it will affect the file size of all reports, while the optional parameter only changes the file size of reports where it was implemented. As far as I know people haven't recognized that images in other reports are also blurry, so a change there is propbably not necessary, but could always added later if needed. Some file size tests:

Reports | 72 DPI |  300 DPI
---|---|---
Media (1 image) | 227 kB | 3,500 kB
Individual complete (1 person, 45 images) | 1,3 MB | 18,3 MB
Individual complete (~1000 persons, ~500 images) | 10,5 MB | 127 MB

Mattkmmr:
> Please rebase to get rid of the merge and force-push. Merging the target into a feature branch creates undesirable history loops.

@jralls Sorry, I'm not so familiar with git and all commands. How do I rebase back to clean the history loop or should I just open a new PR based on the master branch?

jralls:
If you're going to switch the base branch to master then I think you do need a new PR. Just create a new branch from master and cherry-pick 374339d over, push to your GitHub repo, and make a new PR.

To get rid of the gramps51 merge commit the easiest way is
```
  git reset --hard HEAD^
  git rebase maintenance/gramps51 optional-db-parm
```
In theory doing just the rebase will get rid of the merge commit but sometimes conflicts get in the way so resetting first is more sure-fire.

PR #1008, opened 2020-01-27
Upgade from BSDDB to SQLITE

prculley:
This is the initial attempt at an upgrade, assuming we will not be using BSDDB anymore.  It seems to work, but I know we have more to do.  It should be able to convert any bsddb tree that current 5.1.x can open.

One thing I did, is to modify the dbloader code so that we don't get several sequential popups asking if it is ok to upgrade.  I did not like the user aspects, and was a bit concerned that with only a partial bsddb open process, we would leave the db in an unstable state if the user did not accept all the upgrade prompts.  So with my changes, the only time a user can get two prompts is if his bsddb library version is older than the version last used to create the tree.  Then he gets the 'downgrade' warning.

ToDo:
* make sure CLI read of BSDDB db either works, or fails nicely
* Strip out most bsddb code from API etc.  Everything not needed for upgrade.
* Maybe reduce the number of zip backups made; could be 4 I think, if upgrading a very old bsddb
* Fix up config database.backend setting to exclude bsddb (in preferences and in case was set in ini file)
* Make sure we don't default to bsddb anywhere (cli db create?)
* Look into callbacks during upgrade process, no sign of progress on GUI, and numbers seem to be all over the place.
* Fix up other PRs (UID, GEPS045 Place enhancements) to use this if accepted.

### DIFF
--- a/gramps/gen/db/exceptions.py
+++ b/gramps/gen/db/exceptions.py
@@ -360,6 +360,35 @@ class DbConnectionError(Exception):
                     'message': self.msg,
                     'settings_file': self.settings_file}
 
+
+class DbSupportedError(Exception):
+    """
+    Error used to report that a database is no longer supported.
+    """
+    def __init__(self, msg):
+        Exception.__init__(self)
+        self.msg = msg
+
+    def __str__(self):
+        return _('The Family Tree you are trying to load is in the %(dbtype)s '
+                 'database, which is no longer supported.\nTherefore you '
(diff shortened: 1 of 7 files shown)


Nick-Hall:
You should have discussed this on the list first.  I have a better way of doing this that will work even after we remove the BSDDB backend. I'll submit a PR shortly.

sam-m888:
https://github.com/gramps-project/gramps/pull/1009

https://github.com/gramps-project/gramps/pull/1010

PR #1009, opened 2020-01-27
Upgrade from BSDDB to SQLite

Nick-Hall:
Essentially just a raw table copy between a BSDDB database and a SQLite database.  It also updates references and secondary values.

This will work without any of the BSDDB backend code.  It should even recover databases with corrupt environments!

Still need to do some version checks.

### DIFF
--- a/gramps/gui/dbman.py
+++ b/gramps/gui/dbman.py
@@ -320,10 +320,8 @@ class DbManager(CLIDbManager, ManagedWindow):
             self.repair_btn.set_sensitive(False)
         else:
             self.close_btn.set_sensitive(False)
-            dbid = config.get('database.backend')
-            backend_type = self.get_backend_name_from_dbid(dbid)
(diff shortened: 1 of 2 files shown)


prculley:
How does this work if the bsddb needs to load from a much older version, something with an older schema version, or even a python pickel?

Nick-Hall:
@prculley It doesn't.  It will load v4.2.0 and above, and then it will use the SQLite update code for future updates.

prculley:
I think it would be good if we handled any of the older bsddb versions.  Perhaps we could use your method and add/modify the original bsddb upgrade code while running under dbapi?  Maybe create the 'commit_raw' method in dbapi to support that upgrade code (I think we need it anyways for other post 19 upgrades.

Feel free to assign work to me on this...

Nick-Hall:
Thanks for the offer to take this over.  That would be very helpful.

I can see that a private `__commit_raw` method would be useful for upgrades.

PR #1010, opened 2020-01-31
Upgrade from bsddb to SQLITE v2

prculley:
Nick;  This was a combination of your work and mine.  I've coded this to allow the user to open a bsddb normally (with upgrade message).  I did not attempt to make 'Convert' work, it fails (nicely, with message) because the tree needs updating.

The bsddb tree can be rather old, I tested with a bsddb db tree (painfully) created in Gramps 3.0.4, as well as with a tree from 4.2.8.  The test process was to take example.gramps from the Gramps release, import it into the older Gramps.  Then start up modern Gramps with PR code and load tree.  The conversion makes a single zip file backup of the original bsddb db directory.  A single upgrade message, regardless of how old.  I then exported XML to give me something to compare.

I also imported/exported to XML the two example.gramps into sqlite, and exported again.  The modern 4.2.8 upgrade gave identical XML exports.  The older one was not identical, because there were a lot of new Citation objects that had been created by the upgrade from older SourceRef?  But everything else looked ok.  I also used DB Browser for SqLite to examine the metadata in detail, where I found (and fixed) several issues.  

I used Python 'coverage' to see how good my tests were, I got 86% on upgrade.py, and 88% on bsddb.py.  To do better will take either or both of more stuff in the various example.gramps files, or more test dbs from other versions of Gramps.

I based the convert code on your work, but it needed a bunch of extra stuff to manage the older dbs.  Particularly around the metadata.  To deal with schema upgrades, I edited the original bsddb upgrade code to use dbapi calls (add a new commit_raw to dbapi). 

A couple of other commits;
* switch some tests to use sqlite instead of bsddb
* Several changes to make sqlite the default type in all cases, even if config setting was originally set to bsddb
* I was not happy with progress 'freezes' during rebuild of secondary tables and reference maps so modified these to be smoother.


gramps/plugins/test/exports_test.py
```diff
@@ -206,16 +206,6 @@ def setUp(self):
...
-        """ Run a csv export test """
-        set_format(0)   # Use ISO date for test
-        config.set('database.backend', 'bsddb')
-        src_file = 'exp_sample_csv.gramps'
-        tst_file = 'exp_sample_csv.csv'
-        msg = do_it(src_file, tst_file)
-        if msg:
-            self.fail(tst_file + ': ' + msg)
-
     def test_csv_sq(self):
```
Nick-Hall: We can drop the "_sq" suffixes now.

gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -628,6 +629,32 @@ def _commit_base(self, obj, obj_key, trans, change_time):
...
+        part of the transaction.
+        data is serialized primary object (must have handle in position 0)
+
+        This is intended to be use only for db upgrades.
+        """
+        table = KEY_TO_NAME_MAP[obj_key]
+        handle = data[0]
+
+        if self._has_handle(obj_key, handle):
+            # update the object:
```
Nick-Hall: These two comments are not really necessary.

gramps/plugins/test/exports_test.py
prculley: done

gramps/plugins/db/dbapi/dbapi.py
prculley: done

sam-m888:
Brilliant 🥇

emyoulation:
Will this break if the BSDDB support is removed from the core, as has been discussed elsewhere? Would it change to a 'prerequisites' add-on?  

Also if BSDDB support is removed, would the troubleshooting workflows change? I note that a CLI process that specifies using a test.grdb file.

emyoulation:
@prculley 

>The bsddb tree can be rather old, I tested with a bsddb db tree (painfully) created in Gramps 3.0.4, as well as with a tree from 4.2.8. The test process was to take example.gramps from the Gramps release, import it into the older Gramps. Then start up modern Gramps with PR code and load tree. The conversion makes a single zip file backup of the original bsddb db directory. A single upgrade message, regardless of how old. I then exported XML to give me something to compare.

Is there a collection of testing ZIPped tree folders from grampsdb in various versions? If not, should we slowly collate such a beast? And with what naming conventions for the ZIP and Tree?

Like you described above, I had to created database files to test Gramps 6.0rc2. 

Each was an example.gramps files imported into the default db format for 4.28 (BSDDB schema 18) and 5.24 (SQL schema 20)

prculley:
I added https://github.com/gramps-project/gramps/pull/1993 which contains several small old version dbs for test purposes.  This is the only collection I am aware of.  Might make sense to save some in the wiki for test purposes.

PR #1012, opened 2020-02-04
Move beta warning into the status bar

Nick-Hall:
Alternative to PR #890.

Moves the beta warning into the status bar.  The idea is that developer or beta tester has a constant reminder that they are running a development version, without it being irritating.  Clicking on the version displays the standard beta warning.

Would it be useful to display this for all versions?  Perhaps clicking on it could display other version information.


prculley:
Tried it out and it seems OK; simpler than the other way.  However, when DEV_VERSION is an empty string (or None) as I would expect for normal release, the button shows up with no label.  I think we either need to hide it in displaystate, or not create or show it in statusbar if DEV_VERSION evals as True.

I was also wondering why the version in the button was not showing the git commit; perhaps you might want to "from ..gen.const import VERSION, DEV_VERSION" in displaystate to get that version of VERSION.

I think the button should only be shown on development versions; user can always look at help/about to find out version otherwise.

Should the button be made to stand out in some way, maybe a color or text in red or something?

Nick-Hall:
@prculley Good suggestions.  I have updated the code.

PR #1018, opened 2020-03-05
Remove no_magic support; which was not working for dbapi

prculley:
**History:**
'no_magic' is an optional parameter for the db 'transaction_begin' for example:
        with DbTxn(_("Gramps XML import"), self.db, batch=True,
                   no_magic=no_magic) as self.trans:

It appears to be a legacy of the bsddb, doesn't show up in any documents on the wiki, and is not listed in the gen.db.base.py.

It was an indicator to the bsddb to allow backlinks to be updated on a commit (when set to True) despite operating in a 'batch' transaction, where backlinks are not normally updated until the end of the transaction.

I found a bug where the GEDCOM importer was not properly fixing up the Sqlite db when the GEDCOM file was not consistent.  Turns out that when the importer tried to 'make_unknown' a family, it was doing a find_backlink_handles still within the batch transaction.  It had previously set no_magic to True, which for bsddb allows the backlinks to be updated as part of the batch transaction.

DBAPI seems to have no code to support the no_magic parameter, and always disables the backlink updates during the batch transaction.

So when make_unknown tried to do the backlink search to find the handles to attach to the family, it came up empty, generating an inconsistent Gramps db.

**This PR:**
I decided to remove support for no_magic.  I also decided to allow the backlinks feature (db references table) to work in batch mode.  To make this a bit less of a burden, I optimized the backlink creations during batch a bit by removing operations that were not needed.

The result is that batch mode transactions now process backlink creation during each commit, rather than at the end of the transaction in a mass update.  For operations like imports which do a lot of commits this slows them down slightly, for operations with few commits it is a bit faster;

Operation | original time | time with PR
---- | ---- | ----
Import file example.gramps | 10.1 sec | 10.7 sec
Import file my.ged (1000 persons) | 28.0 sec | 29.3 sec
Check&Repair my db | 47.6 sec | 45.6 sec

**P.S.**  I also noticed an odd pause in the Check & repair progress meter during processing; traced it to a missing progress update.  Separate commit for that.

### DIFF
--- a/gramps/gen/utils/db.py
+++ b/gramps/gen/utils/db.py
@@ -261,14 +261,8 @@ def get_participant_from_event(db, event_handle, all_=False):
     """
     participant = ""
     ellipses = False
-    try:
-        result_list = list(db.find_backlink_handles(event_handle,
-                                 include_classes=['Person', 'Family']))
-    except:
-        # during a magic batch transaction find_backlink_handles tries to
-        # access the reference_map_referenced_map which is closed
-        # under those circumstances.
-        return ''
+    result_list = list(db.find_backlink_handles(
+        event_handle, include_classes=['Person', 'Family']))
 
     #obtain handles without duplicates
     people = set([x[1] for x in result_list if x[0] == 'Person'])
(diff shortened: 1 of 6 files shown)


Nick-Hall:
Rebased.

PR #1022, opened 2020-03-14
Fix Windows GUI mode startup for crash with some languages

prculley:
Fixes: #11612. #11490, #11518, #9179, #9201, #9266
The symptom was a crash on startup when running under some (but not all) languages.  And it was occurring in a print statement, a UnicodeEncodeError with the system code page, often cp1252.

This one was difficult to debug because it wouldn't occur under debug, so I could not see exactly what was happening in my debugger.  It only seemed to appear when running in the AIO and started with the no-console shortcut.  Initially I thought that this was some kind of Windows/Python issue because in that mode, stdout normally goes nowhere (NUL device).

But it turned out to be the gramps51.log file; created only when running under Windows without a console.  It was opened with the wrong encoding.

### DIFF
--- a/gramps/grampsapp.py
+++ b/gramps/grampsapp.py
@@ -126,7 +126,7 @@ if win():
             pass  # ok
         elif not os.path.isdir(HOME_DIR):
             os.makedirs(HOME_DIR)
-        sys.stdout = sys.stderr = open(logfile, "w")
+        sys.stdout = sys.stderr = open(logfile, "w", encoding='utf-8')
 stderrh = logging.StreamHandler(sys.stderr)
 stderrh.setFormatter(form)
 stderrh.setLevel(logging.DEBUG)


lordemannd:
excellent work, thank you, tested with zh_TW, sl_SI, he_IL, hu, etc.

lordemannd:
also solves #11611
testing note: it is a little hard to compare after and before this change, when you back it out to compare you find that windows continues to use utf-8 encoding, it apparently it uses what ever the last encoding was. 
So when you back out this change, you still get utf-8 encoding. If you wait long enough the encoding will revert to the apparent default (for english windows) encoding of cp1252.
(don't know if it is just time or some other activity resets the encoding)
To remove any uncertainty an easy way is to use the new code, but change to explicitly set the encoding to cp1252.

prculley:
@lordemannd I cannot replicate your issue about 'backing out'.  There SHOULD be no 'history' around the log file encoding.  It is opened each time in 'w' mode (write, replace) so SHOULD be written in the encoding in the Gramps file (default system encoding [cp1252 on my machine] if not specified).  I assume when you 'back out' the code changes, you are careful to save your work and restart Gramps each time.  I note that on my machine, making these changes requires administrative mode on my text editor, as the files are in 'program files' which is protected.  Otherwise the changes don't save.  A mystery...

lordemannd:
regarding the mystery: Win 10 Home edition, Python 3.6.4.
I was deleting the log file before starting Gramps, is there any part of Gramps/python that is cached and would be reused instead of being recompiled/interpreted?

prculley:
Well, Python compiles into bytecode on the fly and usually stores the code in __pycache__/*.pyc files to speed up loading the next time.  But it SHOULD still be able to work correctly from the memory copy even if it cannot write the pyc file (as would normally be the case in Windows Program Files directory).  The AIO stores pre-compiled pyc files in the AIO, so if these were always getting used, you would never have been able to get it to change to utf-8 encoding.  This stuff seems pretty robust, I've never had any issues like this that were not my own fault (forgetting that I had created links to some other Gramps code on my machine for instance).  My most common mistake is simply forgetting to save the file changes...

lordemannd:
my actual practice is having 2 installs of 5.1.2, one is clean/as installed, the other a working directory, has the changes from PRs. Using the clean version, I am not able to reproduce the original issue and then find that the encoding of the log file is utf-8. If I modify the working directory code to set the encode back to cp1252, continue with Gramps and reproduce the issue, then close Gramps and start the clean version, it can once again reproduce the issue. (now that I've said that, I will go make sure it is exactly true. we have probably spent enough energy on this mystery)

lordemannd:
Oops, now everything is working as it should. I may continue to figure out what I was doing wrong, but lets call it weak testing procedures, and say mystery disappeared/solved. My apologies for the extra time.

lordemannd:
Oops, now everything is working as it should. lets call it weak testing procedures, and say mystery disappeared/solved. My apologies for the extra time. 
But I will continue to figure out what else was perturbing things. I have clue...

lordemannd:
mystery understood: a testing interaction between issue 11611 and the log file encoding change.
i was testing with/with-out GeoAncestor: trying to remove the dancing map. pr292.
after pr292, GeoAncestor no longer writes the person name into the log file. hence utf-8 encoding doesn't mater. Instead GeoAncestor, with the troublesome person names, now writes a warning message similar to:
C:\Program Files\GrampsAIO64-5.1.2-working\lib\python36.zip\gi\overrides\GObject.py:553: Warning: ../../glib-2.54.3/gobject/gsignal.c:2641: instance '000000000dba9870' has no handler with id '42677'

lordemannd:
also fixes #11611, #11702

@prculley, I propose deleting the entries related to my testing craziness regarding the caching of python, I think they will only distract from the real issue and possibly delay merging. what do you think?

prculley:
I 'hid' them, as they are indeed off topic from the bug at hand.  Are you allowed to delete/hide your own comments?  Or just us privileged few?

PR #1023, opened 2020-03-16
Fix Genealogy Tree reports for crash in CLI

prculley:
Fixes #11621

Typo in file caused crash in CLI mode for Genealogy Tree

The second commit avoids some warning messages where the EnnumeratedListOption can sometimes have list elements different than the default initial value (seems to occur for empty string as initial value).  This depends on the usage of the option.  I decided that if an the initial value for the option was considered ok, then when it appears as a value it is ok, even if it is not in the current list.

### DIFF
--- a/gramps/cli/plug/__init__.py
+++ b/gramps/cli/plug/__init__.py
@@ -519,8 +519,8 @@ class CommandLineReport:
                         self.format = tree_format["class"]
             if self.format is None:
                 # Pick the first one as the default.
-                self.format = tree_format.FORMATS[0]["class"]
-                _chosen_format = tree_format.FORMATS[0]["type"]
+                self.format = treedoc.FORMATS[0]["class"]
+                _chosen_format = treedoc.FORMATS[0]["type"]
         else:
             self.format = None
         if _chosen_format and _format_str:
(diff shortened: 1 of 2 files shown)


lordemannd:
on linux, all previous problems are gone.
But on Linux I used a different tree DB which happened to have a user defined Tag named Religious.
Before PR1023, the user defined tag didn't cause a problem, now it reports a warning:
Using options string: name=TodoReport,show=all
2020-03-16 11:07:45.317: WARNING: _enumeratedlist.py: line 146: Value 'Religious' not found for option 'Tag'
2020-03-16 11:07:45.317: WARNING: _enumeratedlist.py: line 147: Valid values: [('', '')]
   Available options:

prculley:
@lordemannd The warnings for the ToDo report and some other similar situations are going to happen, the way the code works now.  The valid values for the ennumerated list (they are selectors in the GUI) depend on the contents of the users db.  But different dbs may have different valid values.  Since the report stores the last value selected in the report_options.xml, it may not be valid for the current db, and thus generates the warning.

One choice would be to convert this from a printed warning, to a LOG.debug warning, which would only show up in debug mode.  Another choice would be to eliminate the warning altogether.  Another would be to add a parameter in the init for the class that says "ignore warnings", for those uses that might create this issue.  I'm open to suggestions.

lordemannd:
since this issue doesn't seem to effect the normal operation of the report,
and is likely to only add confusion for most users I would pick LOG.debug.
(but a weak preference)

PR #1028, opened 2020-03-25
Suppress age = 0 days in events list

SNoiraud:
If the reference event date is equal to the event date, don't show the age
except if the date is estimated, calculated, ...

Fixes #0007745


prculley:
What does the Age column show for death event if the person was born dead (stillborn) or dies the first day?  Seems like it would be nice to make this obvious with a "0 Days"...

PR #1030, opened 2020-03-30
Fix some Python syntax errors that appear in v3.8.x

prculley:
Fixes #11641

The 'is not 1' expression works with cpython since the cpython implementation has already created small integers as an optimization and doesn't have to create them as it does other objects. And they are used whenever needed. This is not true for all python implementations, so Python 3.8.x decided to add this warning.

### DIFF
--- a/gramps/gen/plug/_manager.py
+++ b/gramps/gen/plug/_manager.py
@@ -83,7 +83,7 @@ class BasePluginManager:
 
     def __init__(self):
         """ This function should only be run once by get_instance() """
-        if BasePluginManager.__instance is not 1:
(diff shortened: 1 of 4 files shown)


Mattkmmr:
@prculley Here is another [line of code](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/lib/maps/messagelayer.py#L116) that needs to be fixed.
Fixes: [#11653](https://gramps-project.org/bugs/view.php?id=11653)

lordemannd:
I don't think it fixes the crash in #11653, But it does address the warning messages like others fixed by this PR, In #11653, the warning message appears independent of the crash.

PR #1031, opened 2020-03-31
Fix RemoveUnused tool for crash caused by PyGObject problem

prculley:
Fixes #11634  AttributeError: 'ListStore' object has no attribute 'sort_new_with_model'

PyGObject 3.26.1 (and possibly earlier) has mistakenly removed the original function in trying to fix something else.  It appears that it will be fixed again in a newer release, but for now we need to try both ways of creating a child model.

### DIFF
--- a/gramps/plugins/tool/removeunused.py
+++ b/gramps/plugins/tool/removeunused.py
@@ -172,7 +172,11 @@ class RemoveUnused(tool.Tool, ManagedWindow, UpdateCallback):
                                         GObject.TYPE_STRING,
                                         GObject.TYPE_STRING,
                                         GObject.TYPE_STRING)
-        self.sort_model = self.real_model.sort_new_with_model()
+        # a short term Gtk introspection means we need to try both ways:
+        if hasattr(self.real_model, "sort_new_with_model"):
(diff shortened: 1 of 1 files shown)


lordemannd:
after adding this, it seems to work with the New Libs version of 5.1.2 (python3.8.2) but when I go back to original 5.1.2 (python3.6.4) I get the following error:

16086: ERROR: tool.py: line 257: Failed to start tool.
  File "C:\Program Files\GrampsAIO64-5.1.2-working\gramps\gui\plug\tool.py", line 253, in gui_tool
    callback = callback)
  File "C:\Program Files\GrampsAIO64-5.1.2-working\gramps\plugins\tool\removeunused.py", line 127, in __init__
    self.init_gui()
  File "C:\Program Files\GrampsAIO64-5.1.2-working\gramps\plugins\tool\removeunused.py", line 175, in init_gui
    self.sort_model = Gtk.TreeModelSort.new_with_model(self.real_model)
AttributeError: type object 'TreeModelSort' has no attribute 'new_with_model'

lordemannd:
how should prerequisite checker adapt to this?

PR #1033, opened 2020-04-03
Improve rst manpage syntax

bentley:
`rst2man` expects a particular format to properly generate a manpage from a `.rst` file. Currently `data/man/en.rst` is essentially the output of `man gramps` (from a pre-`rst` manpage) with minor formatting applied. This results in visibly weird output when the manpage is viewed from the command line or a web viewer (such as [Debian’s](https://manpages.debian.org/buster/gramps/gramps.1.en.html)):

* Spacing
    * The original `man` output converted to `rst` was width‐justified, so there are extra spaces in between some words.
    * Properly written manpages begin sentences at the start of a line, so that troff will double space between sentences.
    * There are unnecessary spaces between flags and punctuation (e.g., `-f , --format= FORMAT` should be `-f, --format=FORMAT`)
* Hyphenation
    * The original `man` output was hyphenated, leading to problems such as the first line of text, “gramps - Genealogical Research and Analysis Management Programming Sys‐ tem.”
* Structure
    * `rst2man` has its own idea of section structure that it uses to populate `NAME`, etc. Currently there are multiple nested NAME sections, the manpage title is “English” rather than “Gramps,” and the document title and version are embedded in the text instead of in the header and footer.

This pull request fixes these problems. I also improved the grammar in a separate commit.

Of course, the same changes will have to be made to Czech, French, etc. But before I start that, does this look good?

### DIFF
--- a/data/man/en.rst
+++ b/data/man/en.rst
@@ -284,7 +284,7 @@ AUTHORS
 #######
 
 Donald Allingham <don@gramps-project.org>
-http://gramps-project.org/
+https://www.gramps-project.org/
 
 This man page was originally written by:
 Brandon L. Griffith <brandon@debian.org>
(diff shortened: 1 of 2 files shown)


Nick-Hall:
Thanks.  This looks good.

PR #1034, opened 2020-04-03
Narweb: incorrect place index if alternate names

SNoiraud:
In the narrative web, the first phase is to select all places.
For each object, if we have a place, we memorize that depending on date.
When we create the place pages, we start from these information in the place table. At this moment, we have no date. So the sort by primary names gives an incorrect index list and the place names are incorrect.

If the primary name is different, from the found name, that means it's an alternate name.
In this case, we set the place name to the primary name in place of the alternate name.

Fixes #11645


SNoiraud:
Too complex to be in 5.1. This is a feature request so for 5.2

PR #1037, opened 2020-04-06
Narweb: incorrect place index if alternate names

SNoiraud:
The places page index doesn't show the alternate names used.

In the narrative web, the first phase is to select all places.
For each object, if we have a place, we memorize that depending on date.
When we create the place pages, we start from these information in the place
table. At this moment, we have no date.
So the sort by primary names gives an incorrect index list and the alternate
place names are missing.

Fixes #11645


SNoiraud:
The new index for the gramps database test (see the bug report):

PR #1038, opened 2020-04-08
Narrated Website Google Maps Output JS Warning

SNoiraud:
SensorNotRequired:
The sensor parameter is no longer required for the Maps JavaScript API.
It won't prevent the Maps JavaScript API from working correctly, but we
recommend that you remove the sensor parameter from the script element.

Fixes #011654


rogerrosentreter:
Thank you...

PR #1040, opened 2020-04-12
Relationship Graph; Add option for 'offset parents', replacing/expanding 'use subgraph'

prculley:
The relationship graph was changed in https://github.com/gramps-project/gramps/pull/733 to make it ‘more compact’ and ‘beautiful’ according to the author.  I don't care for it and several others have filed bug reports and complained about it.
https://gramps.discourse.group/t/spouses-misplaced-on-relationship-graph/299/4
https://gramps-project.org/bugs/view.php?id=11550
https://gramps-project.org/bugs/view.php?id=11494
https://sourceforge.net/p/gramps/mailman/gramps-users/thread/CABN8Op8RGCcDV90o9kZsiMASLU0-HiPZHxb4rP3WJP%3DC1UvDkA%40mail.gmail.com/#msg36977242

This PR makes that change an option.  The original 'Use subgraphs' option has been removed from the GraphViz Options tab.  There is now a 3-way choice ('normal', 'Parents together', 'Parents offset') in the 'Graph Style' tab.  It also fixes the user bug mentioned in the sourceforge thread.

Because the original 'Use subgraphs' option was removed from the GraphViz Options tab, the Family Lines graph was also slightly affected.  I put the 'Use subgraphs' option back into the 'Report Options (2)' tab for that report.

### DIFF
--- a/gramps/gen/plug/docgen/graphdoc.py
+++ b/gramps/gen/plug/docgen/graphdoc.py
@@ -258,13 +258,6 @@ class GVOptions:
                            "between columns."))
         menu.add_option(category, "ranksep", ranksep)
 
-        use_subgraphs = BooleanOption(_('Use subgraphs'), True)
-        use_subgraphs.set_help(_("Subgraphs can help Graphviz position "
-                                 "spouses together, but with non-trivial "
-                                 "graphs will result in longer lines and "
-                                 "larger graphs."))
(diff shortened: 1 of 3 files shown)


Nick-Hall:
It looks like this has been discussed in some detail.  I'll merge it now.

PR #1042, opened 2020-04-13
Move master branch to gramps52 5.2.0 status

prculley:
Master branch is already incompatible from a db version standpoint; as further changes are made, (UIDs, place enhancements etc.) the addons will need to start pulling from its master branch which should be built for gramps52.


prculley:
Looks like Nick_H did this one on his own.  https://github.com/gramps-project/gramps/commit/bb1c784d04d89af55244de42fce6dec583bb8874

Nick-Hall:
@prculley Sorry, I didn't notice this.  I suppose I should really review all the open pull requests.

PR #1045, opened 2020-04-23
Fix typo in configure

SNoiraud:
Fixes #11665


romjerome:
Maybe an additional [complement](https://gramps-project.org/wiki/index.php/Coding_for_translation#Changing_translated_text_message_in_the_source_code)?
`sed -i 's/specifed/specified/g' *.po`
This should also fix entry (msgid) on all translation file.

SNoiraud:
Yes, you are right. done

lordemannd:
maybe not important, but in most/all files this line "Specifed tag will be added on import.\n"
is changed to "specified tag will be added on import.\n"

The spelling was corrected but the upper case "S" was lost.

prculley:
Rebased and merged manually...

PR #1048, opened 2020-04-28
NarrativeWeb should show patronymic in individuals

SNoiraud:
In the individuals and in surnames pages, we should show the complete name
like defined in the display tab from the narrative web configuration.

Fixes #04404


baznikin:
I waited for this for ages :)

PR #1050, opened 2020-05-03
enable "file:///" URI paths for addons location

CWSchulze:
I failed to open addons from a local folder. I didn't see a reason why a locally built addons folder should not be opened. I failed to do this with the current master. Did I miss a way to do this?

To change this I tried to use file:/// paths, which pass the first checks. I had to check if only a file was found (which doesn't have a return code).
Then opening fails when trying to open the file with open(), because it starts with file:///. So I removed that.
Additionally It always failed to open the addon tgz. I had to switch to "read binary" -> "rb". Did this ever work without 'rb' using python 3?

So for me now this works to load addons:
grafik

### DIFF
--- a/gramps/gen/plug/utils.py
+++ b/gramps/gen/plug/utils.py
@@ -215,12 +215,12 @@ def available_updates():
                 LOG.warning("Failed to open addon metadata for {lang} {url}: {err}".
                         format(lang=lang, url=URL, err=err))
                 fp = None
-        if fp and fp.getcode() == 200: # ok
+        if fp and (fp.getcode() == 200 or fp.file):
             break
 
     pmgr = BasePluginManager.get_instance()
     addon_update_list = []
-    if fp and fp.getcode() == 200:
+    if fp and (fp.getcode() == 200 or fp.file):
         lines = list(fp.readlines())
         count = 0
         for line in lines:
@@ -282,7 +282,8 @@ def load_addon_file(path, callback=None):
     import tarfile
     if (path.startswith("http://") or
         path.startswith("https://") or
-        path.startswith("ftp://")):
+        path.startswith("ftp://") or
+        path.startswith("file://")):
         try:
             fp = urlopen_maybe_no_check_cert(path)
         except:
@@ -291,7 +292,7 @@ def load_addon_file(path, callback=None):
             return False
     else:
         try:
-            fp = open(path)
+            fp = open(path, 'rb')
         except:
             if callback:
                 callback(_("Unable to open '%s'") % path)


SNoiraud:
It doesn't work on linux. We must keep the last "/".
On linux: path = path[7:]

gramps/gen/plug/utils.py
```diff
@@ -292,7 +292,12 @@ def load_addon_file(path, callback=None):
             return False
     else:
         try:
-            fp = open(path)
+            if path.startswith('file:///'):
```
Nick-Hall, line 295: You should only need a double slash here.

CWSchulze:
@Nick-Hall:
According to Wikipedia I need three slashes for an absolute reference to the file system.
https://en.wikipedia.org/wiki/File_URI_scheme

Linux examples:
```
file://localhost/etc/fstab
file:///etc/fstab
```

Windows examples:
```
file://localhost/c$/WINDOWS/clock.avi
file:///c:/WINDOWS/clock.avi
```


And if I drag and drop a folder onto my browser I get a three slash address:
grafik

SNoiraud:
file://localhost/etc/fstab is not a local file for me. localhost must be fully qualified.
It's a pseudo file and you need to go through the local network.
file://localhost/... is called a non-local file

Nick-Hall:
> According to Wikipedia I need three slashes for an absolute reference to the file system.

Correct.  The format is `file://[host]/path`.  The triple slash is a special case where the host is not specified.

CWSchulze:
maybe it should test for os.name == 'nt' instead, to enable darwin...

prculley:
I'm not sure you need to mess with the file:/// at all; on my system at least, the file opens correctly without any modifications to the URL, when you use the correct URL starting with file://.  Python seems to do it right.  On my Windows machine it accepts "file:///c:/addons.gramps51" or "file://localhost/c:/addons.gramps51" or "file://prc_dev/c:/addons.gramps51".  You can leave off the 'c:/' if the program is running off of drive c.

We do need to fix up the `fp.getcode() == 200` assumption, as there is no HTTP return code when opening local files.

Furthermore, you also need to modify the gen.plug.utils load_addon_file routine to allow the file:// prefix.

CWSchulze:
This one also works for me.

Which are the cases the else-branch is used for? ```open(path)``` didn't work in my case to open the tgz file, even if the path was correct. I had to add the ```, 'rb'```.

Nick-Hall:
> Which are the cases the else-branch is used for?

It was used in the old plugin manager install page.  I can't see it used anywhere now.

emyoulation:
is the specification for the `\listings\addons-en.json` file exist somewhere.  Specifically, can the 'z' parameter be fully qualified and include a URI?

"z":'PostgreSQL.addon.tgz'

https://raw.githubusercontent.com/gramps-project/addons/master/gramps52/listings/addons-en.json

PR #1053, opened 2020-05-07
Install icons into the hicolor theme

cjmayo:
I started working on this because the Gramps icon on my Linux GNOME desktop looked a bit fuzzy and gramps.png was the only file that was installed in the /usr/share/icons/ directory itself.

Later I realised that .gramps and .gpkg files didn't have icons in GNOME Files (nautilus).

These commits fix all of that for me. I've recorded how I created the 128x128 and 256x256 png's in that commit message.


cjmayo:
Updated to be compatible with v5.2.0-dev2.

Nick-Hall:
Thanks.  I'd forgotten about this one.

PR #1054, opened 2020-05-09
Add godparent as event role

CWSchulze:
I added a new event role "Godparent" to enhance visibility, simplify access and improve navigation through christening events. Since many times this is a relevant information in church books to trace back relations, this is a valuable addition to gramps.

I also think it is more consistent with other events. If you can setup a marriage event and add witnesses to this event, then the role of godparent to a christening is similar, even though there is the specific name "godparent" for it.

**The flaw of the current approach to link godparents via "association" has a few drawbacks:**
* There is no clear listing of all godparents at one visible place
    There is a listing of the association, but this includes all types of associations and doesn't tell which type of association this is:
    grafik
* Navigation in both directions is not possible
   Only on one of the sides I can see the type of association, but there is no button to navigate there:
    grafik
* Nothing of this is visible in the christening event
    So the event doesn't have the relevant information, which belongs to this event.
    grafik

**With this change in the code you just have to share the christening event with event role "godparent":**
    grafik

**It will be visible in the following places:**
* On the Godparent side prominently with date, place, and "main participant"
    grafik
* In the references of the event (enables navigation), there will be no other persons in this list
    grafik
* In the list of events, there will be no other persons in this list
    grafik


Nick-Hall:
Perhaps "Godparent" would be better.  If you add "Godfather" then people will also want "Godmother".

This seems like a reasonable request though.  Godparents are likely to be present at a christening.

emyoulation:
We need a tool/process to convert an deprecated Association to a newly supported Role.

Both Witness and Godparent have been added to standard roles. Both were previously stored as an Association.

It is likely that this will be an often situation as Roles evolve.

It would probably be best accomplished as a SuperTool script. (Since it is used infrequently.  Usually once and not used again until adapted to a new Association/Role.)

PR #1055, opened 2020-05-10
New tools for media manager

CWSchulze:
I wrote two tools for the media manager:
1. **Fix path separator**
  This tool can be used to align the path separator of media files
2. **Add date information to media**
  This tool can parse a date from a filename or path, or use the exif data to set the date information.

**More info about path separator:**
I had an issue with wrong separators or mixed separators on windows. Usually windows uses backslash, but msys2/mingw uses forward slash (https://github.com/msys2/MSYS2-packages/issues/1591). I.e. in some dialogs in gramps (such as adding an image in the person editor gallery) the paths were added with backslash, in others with forward slash (e.g. basically everything handled with os.path...).
This tool can set the separator to a user defined value.
grafik
grafik

**More info about date detection**
I needed dates assigned to photos for https://github.com/gramps-project/addons-source/pull/335.
grafik
In the configuration view you can try out different regular expression to detect the date based on the filename / path name. It is also possible to read the exif information (Exif.Photo.DateTimeOriginal). Optionally already set data can be overwritten. The default regular expression searches for a four-digit year at the beginning of the filename.
grafik
The list on the bottom is continuously updated and shows the resulting dates for different files.
grafik

**General remark on threading**
I also implemented a thread to handle the continuous evaluation of user input. However, it turned out that the database cannot be accessed easily from other threads. The thread now only communicates via a queue of native python types. The thread now lives after opening the tool window, is deleted when opening a new one.
After implementing this, I realized that there are not too many thread implementations in gramps. Is there a specific reason for that?

### DIFF
--- a/gramps/plugins/tool/mediamanager.py
+++ b/gramps/plugins/tool/mediamanager.py
@@ -132,6 +132,9 @@ class MediaMan(ManagedWindow, tool.Tool):
         """
         Close the assistant.
         """
+
+        for op in self.batch_ops:
+            op.on_close()
         position = self.window.get_position() # crock
         self.assistant.hide()
         self.window.move(position[0], position[1])
@@ -495,6 +498,12 @@ class BatchOp(UpdateCallback):
     def _prepare(self):
         print("This method needs to be written.")
         print("Preparing BatchOp tool... done.")
+
+    def on_close(self):
+        """
+        This method is called when the dialog is closed
+        """
+        pass
 
 
 #------------------------------------------------------------------------
@@ -969,8 +978,11 @@ class AddDateInformation(BatchOp):
 
     def __del__(self):
         # also kill the thread if this batchop is deleted
-        self.worker_thread.stop()
-        self.worker_thread.join()
+        if self.worker_thread:
+            self.worker_thread.stop()
+            self.worker_thread.join()
+            del self.worker_thread
+            self.worker_thread = None
 
     def check_queue(self):
         """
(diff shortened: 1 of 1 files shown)


prculley:
Threading is not generally used in Gramps because our code is not thread safe.  Not even a little bit.  We don't have any way to lock our modifications to the db that would allow other threads to continue seeing a coherent view of the data.  For instance a change to an object that might refer to other objects doesn't require that the other objects even be present; our standard code would eventually get around to creating the needed objects in the db, but another thread would not know when it was safe to examine the db.  Our GUI uses signals from the db to try to remain consistent with the db, but is has taken literally years of effort to keep the GUI somewhat up to date.  We are still finding issues there.  Adding in changes from other threads would make this even more 'interesting'.

CWSchulze:
I thought the txn.batch was intended for defining groups of changes, that have to be done sequentially and block access to the database. It seems that the database itself doesn't allow access from different threads. So there could be a database thread, which is feed with commands via queues. Actually it could even be a separate process communicating via gRPC. At least this could improve the GUI response time.

I guess you had already some discussions about GTK. I guess replacing the GUI with Qt or some React.js would be a better idea, than making this gtk gui threadsafe. But there seems to be no way in the middle, since neither gtk.plug is available, nor a webview in gtk.

The code I implemented here is not really interacting with the db. It starts with reading all the information from the db, and after that the worker thread handles the user input. The main thread polls changes every 100ms.
Different from my description above, I had to change the termination of the thread. Now the thread is terminated when the window is closed. Maybe a context for the tools would make this more reliable.

Nick-Hall:
I'm not sure why we would need the **Fix path separator** tool.  Why not just use the existing **Replace substrings in the path** tool instead?

The **Add date information to media** seems like a reasonable enhancement.  I'm sure that some people would find it useful.  The use of threads now looks safe, however the `GExiv2` package is an optional dependency.  You need to check that it is installed.

I recommend that this enhancement starts life as a third-party addon.

PR #1056, opened 2020-05-11
Teach DateParser about calendar quarter dates

stevenyoungs:
The calendar year can be divided into four quarters, often abbreviated
as Q1, Q2, Q3, and Q4. Q1 is 1 January to 31 March, Q2 1 April to
30 June, Q3 is 1 July to 30 September and Q4 is 1 October to 31 December

This commit teaches DateParser to parse quarter dates and represet them
as a date range. An example quarter date is "Q2 2020" which is converted
to "between 1 April 2020 and 30 June 2020". This is a one way conversion

Rationale
For a user it is much easier to enter a date using quarter notation than the verbose range equivalent. Quarter dates are used for English civil registation indexes.

### DIFF
--- a/gramps/gen/datehandler/_dateparser.py
+++ b/gramps/gen/datehandler/_dateparser.py
@@ -4,6 +4,7 @@
 #
 # Copyright (C) 2004-2006  Donald N. Allingham
 # Copyright (C) 2017       Paul Franklin
+# Copyright (c) 2020       Steve Youngs
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -445,6 +446,9 @@ class DateParser:
         self._range = re.compile(
             r"(bet|bet.|between)\s+(?P<start>.+)\s+and\s+(?P<stop>.+)",
             re.IGNORECASE)
+        self._quarter = re.compile(
+            r"[qQ](?P<quarter>[1-4])\s+(?P<year>.+)",
+            re.IGNORECASE)
         self._modifier = re.compile(r'%s\s+(.*)' % self._mod_str,
                                     re.IGNORECASE)
         self._modifier_after = re.compile(r'(.*)\s+%s' % self._mod_after_str,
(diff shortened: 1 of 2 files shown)


stevenyoungs:
This change currnetly assumes a new year of Jan 1. 

Does it make sense to support the alternative new year dates supported by DateParser?
If so, for a newyear of Mar 25, is Q1 simply Mar 25 to June 25?

tomhughes:
I would very much appreciate this but I'd like to see syntax for quarters as both ranges and spans... When dealing with english civil registration indexes I use a range (between X and Y) for the event date and a span (from X to Y) for the date on the source citation as a page in the index covers a set of events in that range.

stevenyoungs:
> I would very much appreciate this but I'd like to see syntax for quarters as both ranges and spans... When dealing with english civil registration indexes I use a range (between X and Y) for the event date and a span (from X to Y) for the date on the source citation as a page in the index covers a set of events in that range.

Technically it is easy to generate either a range or span date but currently there is no user interface available to identify which should be generated in each circumstance. 

Do you have any suggetions how a user could indicate whether the quarter date should generate a range or a span? You have mentioned a range for an event date, a span for a citation date. I would appreciate comments from other users about this. If the rule is widely acceptable I can look into implementing it as a second change

Similar decisions would need to be made to all points where a date is created from a string, regardless of if that is a dialog box, text from an import file, ....

Perhaps a compromise of "Q1 2020" generates a range date whilst "from Q1 2020" generates a span date would work?

tomhughes:
Yes I realise the tricky part is finding syntax to distinguish them ;-)

Qualifiers like "in Q1 2020" or "during Q1 2020" would work for a range but as you say defaulting to a range is probably better and qualifiers for a span are a bit harder to come up with. About the best I have is "covers Q1 2020" but I don't really like that.

stevenyoungs:
Agreed. I think the best way forward is to keep this PR as is, adding basic support for "quarter dates". That adds value to both of us plus, I hope, other gramps users using British civil registration indexes.
This can then be enhanced in future PRs which can succeed on their own merits.

sam-m888:
Related to feature request

* [6626: Date Field (particularly for events) to accommodate the quarter as an option.](https://gramps-project.org/bugs/view.php?id=6626)

* [11981: Support calendar quarter dates for entry and display](https://gramps-project.org/bugs/view.php?id=11981)

Nick-Hall:
Added some unit tests.

PR #1057, opened 2020-05-12
Update portuguese (Brazil) translation

Juliano-SC-BR:
I translated the whole file and corrected some other things.
It is my first pull request, I hope everything is ok.

### DIFF
--- a/po/pt_BR.po
+++ b/po/pt_BR.po
@@ -7,20 +7,22 @@
 # lcc <lcc@6zap.com>, 2009-2009.
 # André Marcelo Alvarenga <andrealvarenga@gmx.net>, 2011, 2012, 2013, 2014.
 # Paulo Henrique Paiva de Moraes <paulohpmoraes@gmail.com>, 2016, 2017, 2018...
+# Juliano Camargo <julianosc@pm.me>, 2020.
+#
 msgid ""
 msgstr ""
 "Project-Id-Version: GRAMPS 5.0\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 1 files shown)


prculley:
It looks like you may have changed some strings that had already been translated.  Is this because (in your opinion) the original translations were not correct?

Did you create the .mo file and test it with your installation?

FYI, I note that you fixed up some bad string format problems in the original translations (thanks).

Juliano-SC-BR:
> It looks like you may have changed some strings that had already been translated. Is this because (in your opinion) the original translations were not correct?
> 
> Did you create the .mo file and test it with your installation?
> 
> FYI, I note that you fixed up some bad string format problems in the original translations (thanks).

Hi o/  I am not an advanced user, just someone tired of seeing texts in Portuguese and English all messed up in your genealogy application. I'm not really proficient in English as well but I have little knowledge about coding, with that in mind, I thought someone had incorrectly translated a variable that could no longer be called by its original name, so I changed it back to English. If you think there is something strange, it is possibly because it is wrong.

About the .mo file, no. I didn't find any option that would generate a .mo file here in Gtranslator.

About testing it in my installation, no. I imagine that I lack knowledge on how to update the installation because I ran "python update_po.py -m pt_BR.po" in the po folder that I had cloned but nothing changed on Gramps that was already installed.

Juliano-SC-BR:
I see now that the master branch already has a pt_BR.po file translated better than mine (as I said it is my first time trying to translate an application, firts time on github etc). I learned how to create a .mo file and will use the master branch version until Gramps is updated on its own. Thanks.

PR #1062, opened 2020-05-15
Add Abbreviation columns to Source and Citation selectors

DaveSch-gramps:
Bug [#11710](https://gramps-project.org/bugs/view.php?id=11710)


prculley:
It looks like you are mixing up the Welcome Gramplet changes with the Source/Citation selectors.  You should really be making these changes in separate branches (based on master) and then use separate PRs.

DaveSch-gramps:
I have been waiting for a reply from Nick H (sent vie Discourse) telling me
what I am doing wrong (or not doing).

Back to the youtube videos.

DaveSch-gramps:
The welcomegramplet has its own pull request now #1063 

Do I need to delete this pull and do a separate for the abbreviations? Doing it correctly this time.

PR #1063, opened 2020-05-15
Update welcome gramplet

DaveSch-gramps:
lets try it again


emyoulation:
The updated content will suffer instant linkrot with the targeted release.  The links need to be Generic Names without a VERSION embedded in the URL or have REDIRECT pages that are linkrot resistant. 

See bug report [https://gramps-project.org/bugs/view.php?id=12009](https://gramps-project.org/bugs/view.php?id=12009)

Nick-Hall:
Needs proof reading.

PR #1066, opened 2020-05-17
Improve EditLink's usability: Preselect the active objects for each type

SvenWe:
Background story: In my private family tree I transcribe quite a lot of documents. And in these transcriptions notes I usually want to link people that are mentioned there. Often this is easy, because I usually want to link the active person, which is often preselected, or a person from the same family, which is close to the active person in the list. Unfortunately, quite often it's cumbersome to select the correct person, because the dialog starts not with a person, but with a place. Then I have to change the type to "Person" and find the right person starting from scratch.

This patch makes Gramps gives "Person" the highest priority, and not "Media" as it used to be. And if the user changes the type, Gramps selects the active of that type. This should make it easier to link people and other types.

(Note: This behaviour could be improved even further if Gramps would use the default type according to the currently active category/view.)

### DIFF
--- a/gramps/gui/editors/editlink.py
+++ b/gramps/gui/editors/editlink.py
@@ -208,17 +208,28 @@ class EditLink(ManagedWindow):
                                    (object_class, prop, value))
 
     def _on_type_changed(self, widget):
-        self.selected.set_text("")
         if self.uri_list.get_active() == WEB:
             self.url_link.set_sensitive(True)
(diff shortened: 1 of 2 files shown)


Nick-Hall:
This seems like a reasonable idea.

sam-m888:
linked to 

0005024: By default, to display [Last Recent Used] and [bookmarked] objects on selectors
https://gramps-project.org/bugs/view.php?id=5024

PR #1068, opened 2020-05-22
replace file() -> open()

CWSchulze:
This seems to be Python 2. But even there it is not recommended:
> When opening a file, it’s preferable to use open() instead of invoking this constructor directly. file is more suited to type testing (for example, writing isinstance(f, file)).

https://docs.python.org/2/library/functions.html?highlight=file#file

There is another one in [./gramps/gui/dbloader.py](https://github.com/gramps-project/gramps/blob/47a83a82d74b713558d4717ccbb07a82c484796c/gramps/gui/dbloader.py#L466).

### DIFF
--- a/gramps/gen/plug/utils.py
+++ b/gramps/gen/plug/utils.py
@@ -116,7 +116,7 @@ class Zipfile:
                 os.mkdir(fullname)
         for name in self.get_files(names):
             fullname = os.path.join(path, name)
-            outfile = file(fullname, 'wb')
+            outfile = open(fullname, 'wb')
             outfile.write(self.zip_obj.read(name))
             outfile.close()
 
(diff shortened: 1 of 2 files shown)


gramps/gen/plug/utils.py
```diff
@@ -116,7 +116,7 @@ def extractall(self, path, members=None):
                 os.mkdir(fullname)
         for name in self.get_files(names):
             fullname = os.path.join(path, name)
-            outfile = file(fullname, 'wb')
+            outfile = open(fullname, 'wb')
             outfile.write(self.zip_obj.read(name))
             outfile.close()
```
QuLogic, line 121: Even better to use a context here:
```suggestion
            with open(fullname, 'wb') as outfile:
                outfile.write(self.zip_obj.read(name))
```

prculley:
rebased and merged...

PR #1069, opened 2020-05-26
Symbols enhancement.

SNoiraud:
Add the possibility to set all default values to a string.
Gui configuration improvement.
Possibility to drag and drop a symbol from the symbol list.
In this case, the glyph must be present in the current font to be displayed.
Remove the python-fontconfig prerequisite and use the default font.

This replace PR #996 

Fixes #11769, #11736, #11299


prculley:
I'd like to get this into gramps51 branch, but it unfortunately has a new string to translate.  I think that makes this a candidate for master instead...

Nick-Hall:
Is this a bug fix or an enhancement?

There is some code tidy up that we wouldn't normally accept in a maintenance branch.  This make the change look bigger than it really is.

prculley:
I looked at this again; I think that if we are no longer choosing a font, that the "What you will see" box with the fonts is no longer necessary.  And likely to cause confusion.

I would have liked to get this into 5.1.x, but with the 4 new strings to translate, I don't think we should.

SNoiraud:
> I would have liked to get this into 5.1.x, but with the 4 new strings to translate, I don't think we should.

Is this really a problem? There were no french translation until 5.1.2.

Nick-Hall:
In general, bug fixes should be as small as possible and avoid extra translation strings or changes to the user interface.  However, sometimes this is not possible, especially when the bug relates to the UI.

If the bug is causing significant problems and the fix has been well tested then we can allow it.

prculley:
If we want this in 5.1.3, we should decide soon; I already tagged a commit with the v5.1.3 tag, did NEWS etc.  But have not yet completed releasing the AIO or announced it.  I think we could clear the tag and set it to a new commit, or failing that go immediately to 5.1.4.

There are several bugs around this, including a new one where we cannot find a version number for the darn python-fontconfig prerequisite, which is difficult enough to get anyway.  This PR eliminates that entirely.

jralls:
>  I think we could clear the tag and set it to a new commit

Don't do that. Once you've pushed something to the public repo you have to assume that others have pulled it and if you delete and re-use the tag you'll create trouble for them.

> But have not yet ... announced it.

You may not have announced it on the mailing lists, but everyone who's subscribed to the repo got an email with the release announcement on github, and everyone who visits the releases page sees it.

PQYPLZXHGF:
@prculley Yep too late cats already out of the bag, it was announced over on [reddit](https://www.reddit.com/r/gramps/comments/i8mrav/gramps_v513_released/) linking to your groups announcement on the Gramps blog and here on github.  Personally I was hoping #1040 would have made it in 😢  A big thanks for the release everyone, very much appreciated 👍

jralls:
Now it's [really out there](https://blog.eogn.com/2020/08/18/gramps-5-1-3-released/). Guess I'd better get the macOS bundle up soon!

SNoiraud:
@prculley Is there an easy way to change the destination of this PR ?
(replace gramps51 by master)

Nick-Hall:
> Is there an easy way to change the destination of this PR ? (replace gramps51 by master)

@SNoiraud Change the base by clicking the "Edit" button to the right of the PR title.  You will also have to rebase your branch.

Nick-Hall:
@SNoiraud I'm having problems resolving the conflicts in the `gramps/gui/configure.py` file.  If you can fix the conflicts, I'll give it a quick test and then merge.

PR #1070, opened 2020-06-02
Add date to events in backlink gramplet

SNoiraud:
With the backlink gramplets, it's difficult to select an event because we don't know the date.

This PR add a date column which will be displayed only if we have one event or more.

This will works with the following views: Notes, Places, Citations


SNoiraud:
backlinks 1

SNoiraud:
Already in master with PR #1185. I close this PR

PR #1072, opened 2020-06-06
Update Simplified Chinese translation for gramps51 branch

tiansworld:
Updated translation for gramps51.
Please review.

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -1,5 +1,5 @@
 # Chinese (simplified) translation for Gramps
-# Copyright (C) 2009, 2010, 2011, 2012, 2014, 2015
+# Copyright (C) 2009, 2010, 2011, 2012, 2014, 2015, 2020
 # This file is distributed under the same license as the Gramps package.
 #
 # 陈欣(Celia), 2009.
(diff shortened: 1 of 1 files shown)


tiansworld:
Hello everyone,
I updated the translation in gramps51 branch, The updated file is generated by using the command:
`msgmerge zh_CN_master.po zh_CN_51.po -o zh_CN.po`, which I merged my master branch translation into the gramps51 translation.  
But I don't know why the one of the two checks failed.  

I checked the original zh_CN translation in gramps51, I found that the old zh_CN.po doesn't use the updated gramps.pot in gramps51.
The latest in gramps51 is `POT-Creation-Date: 2019-10-17 11:43+1100\`
But the old zh_CN.po uses `POT-Creation-Date: 2019-09-10 17:16+0100`
Also many source strings positions in po file are different from their counterparts in source file.
For example:
This string from the old zh_CN.po:
 `#: ../gramps/plugins/webreport/webcal.py:1671
msgid "Filter"
msgstr "筛查器"` 

But this string is in line 1672 of webcal.py in gramps51 branch.
`1672         self.__filter = FilterOption(_("Filter"), 0)`

By the way, this problem occurs in zh_TW.po, zh_HK.po and ja.po too.

So I reverted the previous commit d800202 today.

I used translation from master branch,  and back ported it to gramps51, and fixed the problem above by using the command:
`msgmerge -C zh_CN_51.po zh_CN_master.po gramps.pot -o zh_CN.po`.

Source strings positions are fixed and the translation is updated. But the continuous-integration/travis-ci/pr check still failed with the same error:
`if git --no-pager grep --color -n --full-name '[ 	]$' -- \*.py; then echo "ERROR - Trailing whitespace found in source file(s)"; exit 1; fi`

Judging from the error message, I don't think it's from po file. 
Can you teach me how to find the trailing whitespace?

prculley:
I fixed the Travis issue, it was not a whitespace problem.

prculley:
Manually rebased and merged...

PR #1073, opened 2020-06-15
[PedigreeView] fix font color in boxes (bug 11799)

vantu5z:
Detect contrast color for text depending on background color (see [wiki](https://en.wikipedia.org/wiki/YIQ)).
Fixes [11799](https://gramps-project.org/bugs/view.php?id=11799)

### DIFF
--- a/gramps/plugins/view/pedigreeview.py
+++ b/gramps/plugins/view/pedigreeview.py
@@ -65,7 +65,8 @@ from gramps.gen.config import config
 from gramps.gui.views.bookmarks import PersonBookmarks
 from gramps.gen.const import CUSTOM_FILTERS, URL_MANUAL_PAGE, URL_WIKISTRING
 from gramps.gui.dialog import RunDatabaseRepair, ErrorDialog
-from gramps.gui.utils import color_graph_box, hex_to_rgb_float, is_right_click
+from gramps.gui.utils import (color_graph_box, hex_to_rgb_float,
+                              is_right_click, get_contrast_color)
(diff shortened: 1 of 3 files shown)


vantu5z:
Снимок экрана от 2020-06-15 14-19-26
Снимок экрана от 2020-06-15 14-20-05

vantu5z:
Add experimental detecting of contrast color.
Maybe we should use it instead of theme text color?
Need testing! If all is fine it can resolve another issues about contrast text color.

vantu5z:
It works like this if we taged some persons.
Снимок экрана от 2020-06-17 13-09-09

gramps/plugins/view/pedigreeview.py
```diff
@@ -253,6 +253,15 @@ def _boxpath(context, alloc):
                                   0, alloc.height-8)
             context.close_path()
 
+        def get_contrast_color(color):
+            """
+            Choose contrast white or black text color for provided background.
+            """
+            yiq = ((color[0]*299)+(color[1]*587)+(color[2]*114))/1000
```
Nick-Hall, line 260: You don't really need to divide by 1000 here.

Nick-Hall:
I've used this algorithm before, and it works quite well.  Perhaps we should add it to our utilities?

The original intention with tag colour was that it would be used as a foreground colour.  I can see that this may cause problems when used as a background colour or for users who change themes.

vantu5z:
As the PR creates a new function in ( gramps/gui/utils ) the PR branch rebased to Gramps master.

PR #1075, opened 2020-06-17
Events View: "Main Participants" column does not

SNoiraud:
show the full list of participants when expanded.

Fixes #11805


Nick-Hall:
If we want to change this column to display all participants rather than just the main participants then we should really change the column heading as well.

PR #1077, opened 2020-06-27
Remove xdg-utils dependency

a-andre:
This is for opening files in external programs, not location of cache/user data.

It simplifies the code and gets rid a a dependency.  We would have to test it on Linux, Mac and Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,6 @@ The following three packages with GObject Introspection bindings (the gi package
 * **pangocairo** - Allows you to use Pango with Cairo http://www.pango.org/
 
 * **librsvg2** - (SVG icon view) a library to render SVG files using cairo. http://live.gnome.org/LibRsvg
-* **xdg-utils** - Desktop integration utilities from freedesktop.org
 * **bsddb3** - Python bindings for Oracle Berkeley DB https://pypi.python.org/pypi/bsddb3/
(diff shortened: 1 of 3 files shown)


prculley:
I am curious why you believe this is necessary or beneficial?  Any bug related?

Nick-Hall:
It simplifies the code and gets rid a a dependency.  We would have to test it on Linux, Mac and Windows.

wroldwiedbwe:
Does this mean you are removing what little Gramps supports of the xdg spec?

* [8025: Please use the XDG Basedir Spec](https://gramps-project.org/bugs/view.php?id=8025)

* [I'm tired of .folders littering my home directory -- want to do something about it.](https://www.reddit.com/r/linux/comments/971m0z/im_tired_of_folders_littering_my_home_directory/)

QuLogic:
This is for opening files in external programs, not location of cache/user data.

Nick-Hall:
@sam-m888 Why was the WIP label added?

@a-andre Is this ready to merge?

a-andre:
From my side it is ready.

PQYPLZXHGF:
Should a note for xdg-utils be added to the "No longer needed" section for to keep things consistent?

> Since Gramps 5.2: **xdg-utils**

PR #1082, opened 2020-07-05
Add Catalan holidays

pacoriviere:
Holiday file in Catalonia
according to https://ajuntament.barcelona.cat/calendarifestius/ca/ and https://beteve.cat/economia/calendari-laboral-2019-barcelona-catalunya/#festius-catalunya-2019

### DIFF
--- a/data/holidays.xml
+++ b/data/holidays.xml
@@ -81,6 +81,23 @@
     <date name="Christmas" value="*/12/25" type="national" />
     <date name="Boxing Day" value="*/12/26" type="national" />
   </country>
+  <country name="Catalonia">
+    <date name="Any Nou (estatal)" value="*/1/1" type="religious" />
+     <date name="Reis (estatal)" value="*/1/6" type="religious" />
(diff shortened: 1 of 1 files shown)


wroldwiedbwe:
Thanks for adding this 👍 but your new file needs to be merged into the standard [holiday.xml.in](https://github.com/gramps-project/gramps/blob/master/data/holidays.xml.in) file as currently it is standalone and in the wrong directory so gramps will not see your holidays.

eg yours in in 
`gramps/holidays.xml.in`
instead of the standard

`gramps/data/holidays.xml.in `

also the default country name 
`<country _name="Catalunya">`
needs to be in English and later on you can translate it.

Nick-Hall:
@jcreusand Does this look OK to you?

jcreusand:
Sorry for the delay. It's fine with me, but two of the holidays are local, just for Barcelona. If this applies to the whole of Catalonia, we should perhaps remove the two lines _Mare de Déu de la Mercè(a Barcelona)_ and _Dilluns de Pasqua Granada (a Barcelona)_

PR #1085, opened 2020-07-09
Fix XML export when 'Group-as" name contains XML invalid chars

prculley:
Fixes #11834


lordemannd:
confirmed it works, both BSD and SQLite DBs.

PR #1088, opened 2020-07-15
Fix and add objects to Polish translation of Gramps

izikeros:
Here is PR with enhanced Polish translation of Gramps.


Nick-Hall:
Thanks for the updated translation.

PR #1093, opened 2020-07-27
Check & Repair: search and fix bad "links" in StyledTextTags

prculley:
Issue #11855

User discovered an issue with his data when running the Deep Connections Gramplet.  This Gramplet scans into Notes and even uses any StyledTextTag.LINK found to make possible connections.  Unfortunately some Notes contained LINKs pointing to deleted (merged out) persons.  These caused a HandleError exception in the Gramplet.  There was no easy way to fix this.  User ended up manually editing his notes, once he was able to figure out what was happening from exception log.

This PR extends Check & Repair to scan for and remove any such bad links.


lordemannd:
tested with Gramps 5.1.2 on Windows with SQLite.
confirmed by noting the Deep Connections mushroom/litebulb after the merge, then doing a backup,
using the cleanup tool, then doing a Database differences report. Examined the merged Note and see that the link name text is still present, but the linkage to the deleted/merged person has been removed.

PR #1095, opened 2020-07-31
[New Si translation] Add language

PushKK:
Add SI language.
Test from new branch (for rule 'one commit - one PR').


prculley:
As I mentioned before, you have commits unrelated to the new language present.  You need to get those out of here.

There are other thing to do to install a new language; do a search on some likely strings for another language like 'pt_BR' and see what turns up.  You will have to make equivalent changes for that.

And lastly; I don't think we can accept another language unless it contains translations for most of Gramps core; a mostly empty po file isn't really much use.

Nick-Hall:
When I updated the Japanese translation, I added to a list of translations that are over 70% complete.

I suppose a translation could be useful even if it just translated a few reports.  If this is just the start of a translation then we could also consider including it.  All new translators are very welcome.

PushKK:
Yes, minimum 50% must be translated in new language for initial work.
This commit I try do after reading issue https://gramps-project.org/bugs/view.php?id=10889
May be this file can help to translate Gramps.

Nick-Hall:
This translation is only 0.35% complete.

@PushKK Do you intend to translate any more?

PushKK:
Not used language.

PR #1094, opened 2020-07-28
INSTALL file: update python version, small fix, etc.

PushKK:
1. Update version: python3.4 > python3.8,
2. Trim text rows,
3. 'gramps' > 'Gramps' (as program's name only),
4. 'e.g.' > 'e.g.,',
5. Sorted directories, delete doubled '.../mime-info'
6. Small fixes.
7. New language 'Sinhalese'

PR #1096, opened 2020-08-03
Bug10853m

prculley:
Several users with very large trees note that auto-backups can make Gramps unresponsive, particularly noticeable when resuming a system from sleep.  Because a sleeping system has usually gone beyond the backup time limit, an auto-backup is triggered as soon as the system wakes.  This is also when the system is also busy doing other maintenance tasks.

This PR checks for the a longer time than expected since last backup when the  auto-backup timer fires.  If a significantly longer time has elapsed (indicating probable sleep), we add a further delay before starting the auto-backup.

A second commit makes a change to the code that detects if anything has been done, requiring an auto-backup.  Originally the auto-backup would go if any commit had occurred since Gramps opened the db.  With this change, we monitor the number of commits since Gramps opened the db; if no new commits have occurred since the last auto-backup, we skip it.  This will stop backups when the db is just open for reference without changes.

There have been requests for longer intervals; this PR includes a 12 and 24 hour interval.  Which was taken from another https://github.com/gramps-project/gramps/pull/1094 and included to make sure this would work with that other PR if it was accepted.

### DIFF
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -537,7 +537,7 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
         self.transaction = None
         self.abort_possible = False
         self._bm_changes = 0
-        self.has_changed = False
+        self.has_changed = 0  # Also gives commits since startup
         self.surname_list = []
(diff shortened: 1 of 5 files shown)


Nick-Hall:
All good points.

PR #1097, opened 2020-08-06
Increase info in summary text report

Jmuccigr:
Added other types to report. This more closely matches the info in, for example, a post-import report.


emyoulation:
Does this address
https://gramps-project.org/bugs/view.php?id=11881
"Feature Request 11881: [Database Summary Report] Include counts of all item types"

Nick-Hall:
Yes.  I didn't realise that there was an associated bug report.

PR #1098, opened 2020-08-10
Get Note LINK references updated for object deletes or merges

prculley:
A recent bug [11855](https://gramps-project.org/bugs/view.php?id=11855) noted that a report failed when a Note contained a LINK to something that no longer was in the database.

This problem can occur for either of two reasons, the Note contained a LINK to an object that was 1) deleted, or 2) merged out of existence.

I submitted https://github.com/gramps-project/gramps/pull/1093 to scan for these in Check and Repair.

But this has bothered me for a while and I decided to see what it took to get this fixed.

After studying the code, I decided to do several things.
1) rework the view delete code to be more consistent across the views.  This was so I did not have to make modifications to all views/DeleteQuery.  By taking more advantage of backlinks and the new db.method, I was able to make much of the code work in ListView, with only a few minor overrides.
2) Support note LINKs in the backlink code, References Gramplet etc.
3) Support note LINKs when another object was deleted; the Note LINK is simply removed, with the LINK text left in place.
4) Support note LINKs when another object was merged out of existence.  The Note LINK handle is updated to the new handle.

In the process I found a couple of bugs in the MultiSelectDialog; they are covered in a separate commit.

I added test code for test the core NOTE functions and updated the merge tests to check that all the modified merge code was exercised.


Nick-Hall:
> Support note LINKs in the backlink code, References Gramplet etc.

This has been discussed previously and rejected by my predecessor.  The original design specified that links in notes should be soft links.  I'll see if I can find a reference to the original decision.

prculley:
Back before we had HandleError, bad links in Notes only didn't work.  Now they cause crashes.  I really believe that we should fix things that crash the program or just not work.

Seems like you are the boss now; you should be able to recognize that conditions have changed.  Since this PR fixes the issue, why not just accept it?

Nick-Hall:
@prculley I've been looking back at the previous discussions.  There were a couple of points raised.

Firstly, there was a desire to avoid circular references.  This seems sensible, but there are already a few existing examples:

Person -> Person  (as in association)
Media -> Citation -> Media (possibly a bad design inherited from GEDCOM)
Family -> Person -> Family (also a part of the GEDCOM design)

We should avoid this in the future, but we could still regard links in notes as soft references.  So, for example, adding citations to the note structure would be wrong, but referencing a citation within a note would be acceptable.  I can't see why adding backlinks for soft references would cause a problem.

Secondly, in the original design broken links were allowed.  I can see that this could be useful in the future for certain types of notes.

In biographical notes, perhaps intended for inclusion in reports, links are intended to be followed.  I've actually considered extending my experimental _Combined View_ to take advantage of this functionality.

In transcripts, it may be useful to tag a person as a person even if they are not currently in the database, or keep them tagged as a person even if they are deleted.  We could create a new type of tag for this use case though.  Broken links shouldn't actually cause a problem if we don't try to follow them.

Finally, when it comes to merging objects, I can't think of any situation where you wouldn't want to update links.

In conclusion, it seems that this PR can be merged.  I'll probably just check with the _gramps-devel_ list first though in case I am missing something that another developer can remember.

Nick-Hall:
In the definition of the NOTE_STRUCTURE in Gedcom 7, the specification reads:
"A NOTE_STRUCTURE (p.53) can contain a SOURCE_CITATION (p.56), which in turn can contain a NOTE_STRUCTURE , allowing potentially unbounded nesting of structures."

As I mentioned above, this is something we would have previously tried to avoid.  This is another reason that I think that this PR is acceptable.

I'll look at it again tomorrow.

Nick-Hall:
Rebased.

@prculley This PR has a few failing unit tests.

Nick-Hall:
@prculley Thank you.

PR #1106, opened 2020-08-24
Remove the resource-path file

Nick-Hall:
Attempt to derive the resource path from the package path assuming that
one of the three main installation schemes has been used.

The package path will be one of the following:

  _prefix_/lib/pythonX.Y/site-packages
   _prefix_\Lib\site-packages

  _home_/lib/python

  _userbase_/lib/pythonX.Y/site-packages
  _userbase_\PythonXY\site-packages

  Where _prefix_, _home_ and _userbase_ are the resource paths used in the
  Prefix, Home and User installation schemes.

The use of the command line option "--install-data" in the setup script
is no longer supported.

This change is intended to allow a Gramps core package to be created
as a wheel and installed via pip.

### DIFF
--- a/gramps/gen/utils/resourcepath.py
+++ b/gramps/gen/utils/resourcepath.py
@@ -2,6 +2,7 @@
 # Gramps - a GTK+/GNOME based genealogy program
 #
 # Copyright (C) 2013       John Ralls <jralls@ceridwen.us>
+# Copyright (C) 2020       Nick Hall <nick-h@gramps-project.org>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -35,6 +36,25 @@ class ResourcePath:
 
     It should be called only by const.py; other code should retrieve the
     paths from there.
+
+    Attempt to derive the resource path from the package path assuming that
+    one of the three main installation schemes has been used.
+
+    The package path will be one of the following:
+
+    <prefix>/lib/pythonX.Y/site-packages
+    <prefix>\Lib\site-packages
+
+    <home>/lib/python
+
+    <userbase>/lib/pythonX.Y/site-packages
+    <userbase>\PythonXY\site-packages
+
+    Where <prefix>, <home> and <userbase> are the resource paths used in the
+    Prefix, Home and User installation schemes.
+
+    The use of the command line option "--install-data" in the setup script
+    is no longer supported.
     """
     instance = None
     def __new__(cls):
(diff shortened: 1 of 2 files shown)


gramps/gen/utils/resourcepath.py
```diff
@@ -47,9 +67,9 @@ def __new__(cls):
     def __init__(self):
         if self.initialized:
             return
-        resource_file = os.path.join(os.path.abspath(os.path.dirname(
-            __file__)), 'resource-path')
-        installed = os.path.exists(resource_file)
+        package_path = os.path.join(os.path.abspath(os.path.dirname(
+            __file__)), '..', "..", "..")
```
jralls, line 71: You need to normalize out the trailing `..`:
```Python
        package_path = os.path.normpath(os.path.join(os.path.abspath(os.path.dirname(
            __file__)), '..', "..", ".."))
```
Otherwise `tail` below is always `..`.

Nick-Hall: Thanks.  According to the [documentation](https://docs.python.org/3/library/os.path.html) the `abspath` function returns a "normalized absolutized version of the pathname", so I just swapped the order of the `join` and `abspath` calls.  It should be fixed now.  I tested with a User scheme installation.

DavidMStraub:
I tested this with the following change in `setup.py`:

```python
from distutils.core import setup, Command
```

to 

```python
from setuptools import setup
from distutils.core import Command
```

This way, I was able to create a wheel with 
```
python3 setup.py bdist_wheel
```

Installing this wheel with `pip` in a virtual environment, the correct resource path was found and starting Gramps with `gramps` worked!

Also, I tested installing the cloned repository into the virtual env with
```
python3 -m pip install -e .
```

Also in this case it worked!

RossGammon:
Fantastic!
And don't forget to update INSTALL/README as the instructions will now be different.

Nick-Hall:
@RossGammon Thanks for reminding me.   Whilst reading the INSTALL file I realised that Debian can install to _dist-packages_ instead of _site-packages_.  I'll need to update the code to cope with this.

DavidMStraub:
Is anything blocking this (except adding `dist-packages`)? Any help needed?

Nick-Hall:
@DavidMStraub I was just waiting for anyone else who wanted to comment.

wroldwiedbwe:
> from distutils.core import Command

FYI In Python 3.10 _distutils_ will be deprecated and in Python 3.12 it will be removed (see [PEP 632](https://www.python.org/dev/peps/pep-0632/)).

Thanks for the new Gramps release 👍

Nick-Hall:
@wroldwiedbwe Thanks.  It looks like we should migrate to `setuptools` in the next major release.

PR #1109, opened 2020-08-30
Support msgctxt in po files instead of a vertical bar in the msgid

Nick-Hall:
For developers, an extra optional context parameter has been
added to the gettext functions. For example, _('point size|pt')
should now be written as _('pt', 'point size').

For translators this would appear in the message catalog as:

msgctxt "point size"
msgid "pt"
msgstr "pt"

Implements [#3416](https://gramps-project.org/bugs/view.php?id=3416).


Nick-Hall:
The changes to the code and _po_ files were mainly scripted.

I copied the `pgettext` function from python 3.8 so that it runs on earlier python versions.

The implementation of deferred translations is not particularly elegant.  I construct a string consisting of the context and msgid separated by `\x04`.  In this way the `sgettext` function gets passed a string as before and can detect an embedded context.

Now that we don't use the vertical bar convention for context, `sgettext` is probably called more than is necessary, but I wanted to make minimal changes to the code.

The lexeme translations still contain a vertical bar in the message.  I'm not sure why a double bar was necessary.

prculley:
Did we ever use the vertical bar context message in any of our XML files? Glade, embedded XML etc.

Nick-Hall:
Yes.  I converted a few glade files.

In the "Edit Text" dialog in the glade editor there is a field called "Context for translation".

jralls:
>  I'm not sure why a double bar was necessary.

I don't remember either, though I do remember working with @vassilii when he was implementing the lexeme stuff.

PR #1114, opened 2020-09-03
Completing pip installability

DavidMStraub:
Based on the changes by @Nick-Hall in #1106, this PR will make Gramps (the Python library) installable via `pip` and will push the source distribution and a platform-independent wheel to PyPI automatically for tagged commits.

For clarification, this won't be a supported installation method for the desktop application, but will be very useful for things like a [web API](https://github.com/gramps-project/web-api).

In detail:

- It replaces `disutils.setup` by `setuptools.setup` in `setup.py` to provide the `bdist_wheel` command
- It adds `extras_require` to be able to install extra Python requirements
- It adds an automatic deploy stage to the Travis CI configuration. In order to work, this needs defining a PyPI deploy token as Travis environment variable (I'll be happy to assist there. I tried it successfuly on TestPyPI.)

After deployment, the Gramps Python library can be installed, e.g. in a venv, like this:

```
python3 -m pip install gramps
```

Some extras require non-Python dependencies. For instance, for locale support, on Ubuntu one would need
```
sudo apt install build-essential libicu-dev
python3 -m pip install gramps[i18n]
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -122,3 +122,25 @@ after_success:
 # picks up coverage from /usr/bin, rather than from
 # /home/travis/virtualenv/python3.3.6/bin/
  - PATH=/usr/bin:$PATH bash <(curl -s https://codecov.io/bash)
+
+stages:
+  - test
+  # Only execute deployment stage on tagged commits, and from our repository
+  # (e.g. not PRs).
+  - name: deploy
+    if: tag IS PRESENT AND repo = gramps-project/gramps
+
+env:
+  global:
+    - TWINE_USERNAME=__token__
+
+jobs:
+  include:
+    # Deploy source distribution
+    - stage: deploy
+      name: Deploy source distribution and wheel
+      install: skip
+      script: python3 setup.py sdist bdist_wheel
+      after_success: |
+        python3 -m pip install twine
+        python3 -m twine upload --repository testpypi --skip-existing dist/*
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,8 @@ if sys.version_info < (3, 2):
     raise SystemExit("Gramps requires Python 3.2 or later.")
 
 from distutils import log
-from distutils.core import setup, Command
+from setuptools import setup
+from distutils.core import Command
 from distutils.util import convert_path, newer
 from distutils.command.build import build as _build
 import os
@@ -400,6 +401,13 @@ setup(name = 'gramps',
       cmdclass = {'build': build, 'test': test},
       packages = packages,
       package_data = {'gramps': package_data},
+      extras_require={
+          "bsddb": ["bsddb3"],
+          "image": ["Pillow"],
+          "i18n": ["PyICU"],
+          "GUI":  ["PyGObject", "pycairo"],
+          "testing": ["jsonschema", "mock", "lxml"],
+      },
       data_files = data_files,
       scripts = ['scripts/gramps'],
       classifiers = [


Nick-Hall:
> It replaces disutils.setup by setuptools.setup in setup.py to provide the bdist_wheel command

Should we also replace the other `disutils` imports with their `setuptools` equivalents?

Nick-Hall:
@DavidMStraub What needs to be done next?  Do we plan to deploy 5.2.0 as a pre-alpha release?

DavidMStraub:
> Should we also replace the other disutils imports with their setuptools equivalents?

That would probably make sense, but I would suggest to do it one by one to make it easier to track any issues that might arise. In particular, I would suggest that the packagers try if any problems arise with just switching `setup` (I certainly don't expect any, but...)

> @DavidMStraub What needs to be done next? Do we plan to deploy 5.2.0 as a pre-alpha release?

Having a deployed pre-release would be great (as a workaround for experimentation, it's also possible to directly pip-install from git as done [here](https://github.com/gramps-project/web-api/blob/f7f61f239f42aac283d92a2f3c9a3b8d7235fa3e/.travis.yml#L8)). 

Pushing a pre-alpha to PyPI would require appending ".dev1" to the version number ([see packaging guide](https://packaging.python.org/guides/distributing-packages-using-setuptools/#pre-release-versioning)). Then the release would be automatically marked as "pre-release" and it wouldn't be possible to install it via `pip install gramps`, only via `pip install --pre gramps` or `pip install gramps==5.2.0.dev1`. I think this would be the desired behaviour.

Then the remaining steps would be:

1. Set `VERSION_QUALIFIER = ".pre1"` in `version.py`
2. Possibly set `VERSION_TUPLE = (5, 2, 0)`? I mean, it should be a `5.2.0.dev1`, not `5.1.4.dev1`, shouldn't it?
3. Add the PyPI deploy token to Travis
4. Tag the pre-release. It will be automatically deployed.

I could do 1. and 2. as part of the PR, if desired.

Instructions for 3.:

- Register an account at PyPI
- [Add an API token](https://pypi.org/manage/account/token/) (named Gramps PyPI)
- Go to the repo settings in Travis and look for the section "Environment variables"
- Add the variable `TWINE_PASSWORD` and add the API token as value. Unselect "display value in build log" (important!)

Nick-Hall:
I have created a main PyPI account and also an identical one on the test server.  I generated the API token on the test server and set the `TWINE_PASSWORD` environment variable on all branches.

Our convention is to use a hyphen to separate the version qualifier so the version would be `5.2.0-dev1`, but that seems to be an acceptable format.

The only issue with setting the version to `5.2.0` is that we would need to create a new branch for the addons repository.  This will need to be done at some point anyway.

DavidMStraub:
Perfect! I changed the version to `5.2.0-dev1` in the PR. (If you prefer to make the version increment separetely let me know, then I'll revert it.)

DavidMStraub:
Thanks for merging! Really happy to have my first commit to Gramps after 13 years or so of being a user.

Just a heads-up: at the moment, there is still an empty `VERSION_QUALIFIER` in the master branch, so a wheel that will be pushed will be marked as `5.2.0`. For the moment it's just on TestPyPI, but still, a bit dangerous.

Another thing I noticed: my above statement 

>  as a workaround for experimentation, it's also possible to directly pip-install from git 

was partly incorrect; while it can be installed like that, I realized that installing from `git` does not build the translations. So to have a properly working `pip`-installed version, the PyPI wheels (pre-release) are actually necessary.

Nick-Hall:
I was about to test this with a small commit setting the `VERSION_QUALIFIER` to '-dev1' and tagging it 'v5.2.0-dev1'.

You have already uploaded a version to the TestPyPI server.  Will this cause a problem?  Do you need to transfer ownership?

Let me know when you want to run the first upload test?

DavidMStraub:
Good question; I will simply send you an API token for the "gramps" account I created on TestPyPI. I also deleted the test releases I created there. Actually, even when deleting them, it won't be possible to re-upload a version with the same version number. I had one "5.2.0" release, but this should not conflict with the "5.2.0-dev1" test.

Nick-Hall:
@DavidMStraub The Travis CI unit tests are not being run on the _master_ branch anymore.  What do we need to change to fix this?

DavidMStraub:
Oh, that's really strange. I will investigate. Can you please check if you can see the "Requests" tab on Travis [mentioned here](https://docs.travis-ci.com/user/common-build-problems/#i-pushed-a-commit-and-cant-find-its-corresponding-build)? I can't see it, probably because I don't have permisson.

DavidMStraub:
Nevermind, found the tab. It says "Build config did not create any jobs". Need to find out why now. I will experiment in my fork.

DavidMStraub:
OK that PR should fix it. Sorry for breaking things :slightly_frowning_face:

DavidMStraub:
@Nick-Hall I completely forgot that this workflow was never migrated to Github Actions.

Should I do this now, so we can test it with v5.2.0-beta2?

Nick-Hall:
@DavidMStraub I tried to upload the release manually using twine, but it complained about the name "gramps".  Is it possible that it is conflicting with something that you have uploaded?  If not what name do you suggest that we use?

I have applied for an organization account for the Gramps project.  Hopefully I should get a response soon.

DavidMStraub:
Hi,

there is only a gramps on `test.pypi.org`, created by the old Travis workflow. I don't have anything on the production PyPI and I also don't find anything there, so it's weird.

Once the upload with twine works and you have stored the tokens as Github secrets, I can open a PR with a workflow to automate it.

DavidMStraub:
https://test.pypi.org/project/gramps/

Nick-Hall:
I'm running:

`twine upload dist/*`

and I get the error:
```
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/                        
         The name 'gramps' isn't allowed. See https://pypi.org/help/#project-name for more      
         information.                                                                           
```

It also complains about the line "Natural Language :: German (Austrian)" in the setup, but that is easy to remove.

DavidMStraub:
Really strange. Perhaps you need to request a name transfer to find out why the name is blocked. I double-checked that I don't have any related projects except `gramps-webapi`.

Nick-Hall:
I'll wait a couple of days to see if our Gramps organization gets approved, then I'll ask about the best way to solve this.

bartfeenstra:
Hi! I'm the maintainer of [Betty](https://github.com/bartfeenstra/betty) and would like to replace the current, rather old code that parses Gramps XML by a completely new integration where I call Gramps APIs directly, so Betty can pull existing family trees straight out of Gramps without the need for an export, or import any file in a format that Gramps supports.

@Nick-Hall Did you hear back about organization/project name approval?

DavidMStraub:
Thanks for digging up this old thread, indeed it would be good to make sure PyPI integration works by the time 5.2 is released. By the way, if you find some useful code in https://github.com/gramps-project/gramps-webapi to realize your  new Gramps-to-Betty interface, maybe there will be some synergy.

Nick-Hall:
> Did you hear back about organization/project name approval?

No.  I can see one pending organization request in my account settings as follows:
```
gramps "COMMUNITY" "REQUEST SUBMITTED"
Submitted Aug 1, 2023
```
I'm not sure what is the best way to chase this.  Should I open a support ticket or is there a mailing list or forum that would get a quicker response?

sam-m888:
@Nick-Hall Investigated and the issue seems to be that in 2020 @DavidMStraub unintentionally was allowed to create a **gramps** account on test.pypi.org ( https://test.pypi.org/user/gramps/ )  while the main official account I created back in 2018 for the project is https://pypi.org/user/gramps/   this most likely happened during the period pypi was being migrated off the legacy infrastructure and they had not enforced unique usernames! 

I've contacted EWDurbin the Director of Infrastructure at The Python Software Foundation  and emailed security@pypi.org  to see if they can fix this issue.

DavidMStraub:
Hi @sam-m888, is the account on test.pypi.org interfering with the main account somehow? I assumed they are two completely isolated instances.

Would it help if I just hand this account over to you (or Nick(?

sam-m888:
@DavidMStraub Thank you,  Probably better if @Nick-Hall has it.   I believe Nick also has another account , so we may need to get rid of few accounts and get that organization account that is mentioned above.

DavidMStraub:
Ok, no problem. @Nick-Hall please let me know which e-mail address to set for the account, then we can switch it over.

DavidMStraub:
I don't think so, since Betty is a static site generator in Python, no need for a server. What I meant was rather reusing/sharing some Python functions we have, like especially in https://github.com/gramps-project/gramps-webapi/blob/master/gramps_webapi/api/resources/util.py.

bartfeenstra:
> Thanks for digging up this old thread, indeed it would be good to make sure PyPI integration works by the time 5.2 is released. By the way, if you find some useful code in https://github.com/gramps-project/gramps-webapi to realize your new Gramps-to-Betty interface, maybe there will be some synergy.

Using the API would be a bit much initially because I dont expect all Betty users to have an instance running. If I could pull in the Gramps Python code, I may hopefully benefit from some of its parsing and  conversion powers by just importing the modules.

emyoulation:
Have you downloaded the Isotammi addon SuperTool? Maybe its engine and utility modules would be lightweight alternatives?

bartfeenstra:
You mean the Gramps plugin? I'm not sure what it does, but it seems similarly unavailable for a pip install, which at the moment is how all Betty dependencies are managed.

emyoulation:
The SuperTool is a general purpose scripting tool that can be used to do "ad-hoc" queries against a Gramps family tree/database. The queries are expressed in the Python programming language so the tool is most useful for programmers. 

I'm pretty certain I have at least SOME wrong assumptions. So take the following with more than 1 grain of salt....  The SuperTool.py is an interface module for the scripting and output. And it has uses modules that read gramps objects into proxy objects, manipulates those and writes back to the database.  Instead of having to explore dozens of modules, you only have to read through 3. 

https://github.com/Taapeli/isotammi-addons/blob/master/source/SuperTool/README.md

Nick-Hall:
@DavidMStraub My account is named "gramps-project" on both the live and test servers.  I intended this to be for use by the project rather than for me personally, but it is clear that we shouldn't do this.

I have applied for "gramps" as an organization on the live server only.  If our request for an organization is accepted, I can delete the "gramps-project" account.  Then I can add you and I as owners/administrators.

The registered email for both my accounts is "nick-h@gramps-project.org".

bartfeenstra:
Any luck so far? :)

Nick-Hall:
Yes.  The ownership of the "gramps" project name has now been transferred to my "gramps-project" account.  I haven't heard anything about the progress of our organisation account yet though.

It's probably time for another maintenance release, so we will have the opportunity to test this shortly.

PR #1115, opened 2020-09-05
Internet tab added to Event category.

janskarvall:

### DIFF
--- a/data/tests/imp_MediaTest.gramps
+++ b/data/tests/imp_MediaTest.gramps
@@ -3,7 +3,7 @@
 "http://gramps-project.org/xml/1.7.1/grampsxml.dtd">
 <database xmlns="http://gramps-project.org/xml/1.7.1/">
   <header>
-    <created date="2019-03-05" version="5.0.2"/>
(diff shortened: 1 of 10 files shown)


prculley:
Looks like Travis failed; and it appears that adding the URLs to the Event has caused the import_tests to fail.  This is not surprising, since those tests compare the XML export of imported files to reference XMLs.  So URLs in the Event might appear in the XML export.  Looks like imp_MediaTest.ged had a media object that was of BIRT/OBJE/FORM 'URL' at line 52, which since it was attached to an event, got included as an URL.   This was reflected in the XML export.  To fix, make sure that the XML is really what is expected (I think it is) and replace the imp_MediaTest.gramps file with the one generated when you run the test locally (should have been left in the .gramps/temp directory after the test.  And add this to the PR.

Nick-Hall:
This and PR #1116 seem like a reasonable enhancement.  However, I have some concerns about adding `Url` objects to sources and citations.  I'll add a comment to the appropriate PR.

Nick-Hall:

PR #1118, opened 2020-09-05
Internet tab added to Citation category.

janskarvall:

### DIFF
--- a/gramps/gen/db/upgrade.py
+++ b/gramps/gen/db/upgrade.py
@@ -50,12 +50,35 @@ LOG = logging.getLogger(".upgrade")
 def gramps_upgrade_20(self):
     """
     Placeholder update.
+
+    1. Add web link to citation objects.
     """
     length = 0
     self.set_total(length)
     self._txn_begin()
 
     # uid and place upgrade code goes here
+
+    # ---------------------------------
+    # Modify Citation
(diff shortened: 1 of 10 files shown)


Nick-Hall:
As opposed to people and events where it is useful to have multiple urls, a citation will normally only have one url, which should be included in the formatted citation together with the date accessed.  For example, `(https://www.findagrave.com/memorial/123456789 accessed: 13 Sep 2020)` could form part of a citation for a burial.

In discussions about citations on the [FHISO](https://fhiso.org/) public forum, the opinion was to use key-value pairs already supported in the GEDCOM `SOUR.PAGE` tag. I can see the future ELF format using this convention.

Some work was started along these lines in [GEPS 018: Evidence style sources](https://gramps-project.org/wiki/index.php/GEPS_018:_Evidence_style_sources).  However, that project was rather ambitious.  Perhaps we could try something simpler?

Just as an initial thought, suppose we stored three attributes:

url:  "https://www.findagrave.com/memorial/123456789"
accessed:  "13 Sep 2020"
format:   "%{page}s (%{url}s accessed: %{accessed}s)"

If an attribute called "format" exists, then we could use it to generate the formatted citation.

Nick-Hall:

PR #1119, opened 2020-09-05
Allow setting DB read-only in CLI DB handler

DavidMStraub:
This is a very simple change that adds a keyword argument `mode` to the methods `CLIDbLoader.read_file` and `CLIManager.open_activate`. At the moment, whether the database is open in readonly mode by the CLI is determined only by whether the directory is writable or not. However, there might be use cases where the DB should be opened in readonly mode even if it could be writable in principle. In particular, opening it in readonly mode avoids writing a lock file, so that multiple such calls could be run in parallel.

My specific use case is the use of the CLI DB handler for the [web API](https://github.com/gramps-project/web-api): while in principle the use of the database by a web server is completely analogous to the CLI, so the class can be used directly, there is one difference: for read-only operations, it is preferable to allow simultaneous read access, which requires *not* writing a lock file. With this (very minor) change, this is achieved by simply using `open_activate(..., mode="r")`.

In the future, this could in principle even be used to add a `--readonly` switch to the CLI to allow batch operations without locking the DB.

But for the moment, merging this would tremendously simplify work on the web API without having to monkey-patch classes :wink:

### DIFF
--- a/gramps/cli/grampscli.py
+++ b/gramps/cli/grampscli.py
@@ -47,7 +47,7 @@ LOG = logging.getLogger(".grampscli")
 from gramps.gen.display.name import displayer as name_displayer
 from gramps.gen.config import config
 from gramps.gen.const import PLUGINS_DIR, USER_PLUGINS
-from gramps.gen.db.dbconst import DBBACKEND
+from gramps.gen.db.dbconst import DBBACKEND, DBMODE_R, DBMODE_W
 from gramps.gen.db.utils import make_database
 from gramps.gen.errors import DbError
 from gramps.gen.dbstate import DbState
@@ -128,7 +128,7 @@ class CLIDbLoader:
         """
         pass
 
-    def read_file(self, filename, username, password, mode=None):
+    def read_file(self, filename, username, password, mode=DBMODE_W):
         """
         This method takes care of changing database, and loading the data.
         In 3.0 we only allow reading of real databases of filetype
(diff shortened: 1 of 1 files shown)


gramps/cli/grampscli.py
```diff
@@ -128,7 +128,7 @@ def _end_progress(self):
         """
         pass
 
-    def read_file(self, filename, username, password):
+    def read_file(self, filename, username, password, mode=None):
```
Nick-Hall, line 131: Possibly default to "w" rather than None?

DavidMStraub: Of course, you're right: we default to "w" unless the directory is read-only or it is set manually. Good point.

DavidMStraub: Indeed the code is more concise now. Used `DMODE_W` and `DBMODE_R` while I was at it.

DavidMStraub:
Please note that even though this is displayed as "failed", this is only due to strange behaviour of codecov. I did not add test, but I actually made the code shorter than before.

In my view this can be merged anytime.

Nick-Hall:
Loading a database in read-only mode doesn't actually prevent write access.  It sets a flag that editors use to prevent data entry and it also prevents a schema update and writing metadata back when the database is closed.  However, I can't find any code that prevents a plugin such as a tool modifying the database.

If we made this change it would have to be used with caution.

DavidMStraub:
Yes you're right, "opened in readonly mode even if it could be writable in principle" might have been a bit misleading; indeed `db.mode` at the moment is just an attribute that hints to downstream code that it *shouldn't* write to the DB, but all the methods inherited from `DbWriteBase` are still there.

So what I meant was that this change gives any code using `CLIDbLoader.read_file` the possibility to indicate that it does not *intend* to do any DB writes and consequently skips the creation of a lock file. The web API use case is a good example: for read-only operations, we can instantiate a handler that does not write a lock file since it will not modify the DB.

But indeed my idea 

> In the future, this could in principle even be used to add a --readonly switch to the CLI to allow batch operations without locking the DB.

was a bad idea, since such a flag would make the user rightly assume that their DB will not be touched under any circumstances, but we cannot guarantee that, e.g. in the case of a bug in a plugin etc. This would require adding a check in front of every method of `DbWriteBase` (and that would be quite a substantial change). So please forget this idea :slightly_smiling_face: 

But I think the original use case still stands. If you feel uneasy about using the `readonly` attribute, an alternative would be to introduce a new parameter `write_lock=True` that could be disabled on demand. But this would require changing `DbGeneric.load` as well, since this currently has

https://github.com/gramps-project/gramps/blob/c567b9e399514cfbf3bc6e5dd3996a04b7f2ed97/gramps/gen/db/generic.py#L590-L591

Finally, the third option would be to close this PR and to move a copy of `CLIDbLoader` with this change to `web-api`.

DavidMStraub:
Let's close this, no real benefit right now.

PR #1120, opened 2020-09-06
Add __main__.py

DavidMStraub:
Another very simple PR which adds a single file: `gramps/__main__.py`, which is nothing but a copy of `Gramps.py` or `scripts/gramps` (apart from the `if` statement).

Rationale: it allows to start Gramps (the GUI or the CLI) on any system alternatively with `python3 -m gramps` instead of `gramps`. This has a small benefit if the `gramps` script does not happen to be in the `PATH` for some reason. But the main motivation for this PR is to be able to start Gramps without problems when installed in a virtual environment, e.g. with pip (cf. #1114). Using the `-m` notation avoids any ambiguities which Python interpreter is used to run the GUI/CLI.

### DIFF
--- a/gramps/__main__.py
+++ b/gramps/__main__.py
@@ -1,0 +1,31 @@
+#! /usr/bin/env python3
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2012       Benny Malengier
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301  USA
+#
+
+"""This is a stub to start Gramps.
+
+It is provided to be able to start gramps as `python3 -m gramps`,
+e.g. in a virtual environment.
+"""
+
+import gramps.grampsapp as app
+
+if __name__ == "__main__":
+    app.main()
--- a/gramps/cli/argparser.py
+++ b/gramps/cli/argparser.py
@@ -54,7 +54,7 @@ from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
 
 _HELP = _("""
-Usage: gramps.py [OPTION...]
+Usage: gramps [OPTION...]
   --load-modules=MODULE1,MODULE2,...     Dynamic modules to load
 
 Help options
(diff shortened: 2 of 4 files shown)


DavidMStraub:
Forgot to mention: I also changed the usage hint in argparse from `Usage: gramps.py [OPTION...]` to `Usage: python3 -m gramps [OPTION...]`. There is no `gramps.py` anyway ...

Nick-Hall:
Users should start Gramps from the CLI using the installed script.  So the help should read:  `Usage: gramps [OPTION...]`.  This is consistent with the _man_ page and most of the user manual.

The `gramps.py` file was created by the old build system.

The `Gramps.py` file is provided for the convenience of developers.  We may choose to remove it.

gramps/__main__.py
```diff
@@ -0,0 +1,31 @@
...
+
+"""This is a stub to start Gramps.
+
+It is provided to be able to start gramps as `python3 -m gramps`,
+e.g. in a virtual environment.
+"""
+
+import gramps.grampsapp as app
+
+if __name__ == "__main__":
```
Nick-Hall, line 30: Will this ever be something other than "__main__"?

DavidMStraub: I just checked that if it is imported by another script, it is set to `"gramps.__main__."`. Of course, it's a bit silly to assume that this will happen, so one can also just drop it.

DavidMStraub:
>  So the help should read: Usage: gramps [OPTION...]

Agreed, changed it.

DavidMStraub:
Can this be merged or are there still doubts? It would solve some problems I have with [web-api](https://github.com/gramps-project/web-api) unit tests.

gramps/__main__.py
Nick-Hall: The `if` statement seems redundant.  Is it necessary?

DavidMStraub: It's to some extent convention, but the motivation is that it does not run the code on importing the module (`import gramps.__main__`).

See e.g. here: https://docs.python.org/3/library/__main__.html

Nick-Hall: The `_main__.py` file isn't run on import anyway, just the `__init__.py`  file.

DavidMStraub: I meant when importing `__main__`. I mean, this `if` statement is just there so the code below is only ever executed when this is run as a script. And to be honest I have never seen a `__main__.py` that does *not* have this `if` statement. Prominent examples:

- https://github.com/pypa/pip/blob/master/src/pip/__main__.py#L25
- https://github.com/python/mypy/blob/master/mypy/__main__.py#L22
- https://github.com/pytest-dev/pytest/blob/master/src/pytest/__main__.py#L4
- https://github.com/pypa/twine/blob/master/twine/__main__.py#L52
- https://github.com/python/cpython/blob/master/Lib/ensurepip/__main__.py#L4

It's also a common pattern to have a `main` function in `__main__.py` *before* the block which can be imported.

Nick-Hall: OK.  Thanks.  I'll merge this now.

DavidMStraub: Thanks! :+1:

Nick-Hall:
@DavidMStraub I just noticed that the new file will need an entry in `po/POTFILES.skip`.

DavidMStraub:
@Nick-Hall just in case you missed it, I did add the missing entry in `POTFILES.skip`.

PR #1124, opened 2020-09-14
Added support for open spans.

janskarvall:
I chose to add separate phrases in gramps.pot rather than make complicated patterns in DateParser, and as I thought it be more safe, being unsure if the open span phrases always are sub phrases of the closed span phrase. See line 1949 and line 1954 in gramps.pot.

Changed the date comparison operators '==', '!=', '<', and '>' to use match instead of sortval in case any of the dates is a span, and added a '!=' comparison in match. Checked that the change does not break the display order of events in gramps itself and the webreport.
 
Besides date.py, _datedisplay.py, and _dateparser.py, the change affected all _date__XX.py and XX.po files.
_date_ja.py also failed to match calendar names as the re.compile truncated the match at 200 characters, so I added match_calendar.

I also had to make a small change to uk.po due to a clash for 'before date' and 'from date'.

Changed DateHandlerTest to run tests for a static list of languages and to not stop at first failure.
Doing this required changes in the DateParser to accept date strings as output by DateDisplay. Thus I added the _qual_match_str, _mod_match_str, and _cal_match_str.

Most languages passes the test, but some failes due to lacking _date_XX.py files in combination with that DateParser does not adapt to the target language (as DateDisplay does), and thus expects dates in 'en' language.
Some of the languages that lacks _date_XX.py should possibly fail, but as they passes they are included among the tested languages.

Added a DateComparisonTest to date_test.py. Doing that was what made me change the comparison operators as described above, as I thought that the sortval based comparisons for spans where a bit awkward, especially for a span open to the left, i.e. 'to *date*'. Some tests, not related to spans, have an outcome that is likely wrong. Those are anyway included in the test.

### DIFF
--- a/gramps/gui/editors/editdate.py
+++ b/gramps/gui/editors/editdate.py
@@ -31,32 +31,32 @@ The DateEditor provides a dialog in which the date can be
 unambiguously built using UI controls such as menus and spin buttons.
 """
 
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 #
 # Python modules
 #
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 
 
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 #
 # set up logging
 #
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 import logging
 
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 #
 # GNOME modules
 #
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 from gi.repository import Gtk
 
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 #
 # gramps modules
 #
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 from gramps.gen.config import config
 from gramps.gen.lib.date import Date, DateError, calendar_has_fixed_newyear
@@ -68,42 +68,47 @@ from ..glade import Glade
 
 LOG = logging.getLogger(".EditDate")
 _ = glocale.translation.sgettext
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 #
 # Constants
 #
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 MOD_TEXT = (
-    (Date.MOD_NONE, _('Regular')),
-    (Date.MOD_BEFORE, _('Before')),
-    (Date.MOD_AFTER, _('After')),
-    (Date.MOD_ABOUT, _('About')),
-    (Date.MOD_RANGE, _('Range')),
-    (Date.MOD_SPAN, _('Span')),
-    (Date.MOD_TEXTONLY, _('Text only')))
+    (Date.MOD_NONE, _("Regular")),
+    (Date.MOD_BEFORE, _("Before")),
+    (Date.MOD_AFTER, _("After")),
+    (Date.MOD_ABOUT, _("About")),
+    (Date.MOD_RANGE, _("Range")),
+    (Date.MOD_FROM, _("From")),
+    (Date.MOD_TO, _("To")),
+    (Date.MOD_SPAN, _("Span")),
+    (Date.MOD_TEXTONLY, _("Text only")),
+)
 
 QUAL_TEXT = (
-    (Date.QUAL_NONE, _('Regular')),
-    (Date.QUAL_ESTIMATED, _('Estimated')),
-    (Date.QUAL_CALCULATED, _('Calculated')))
+    (Date.QUAL_NONE, _("Regular")),
+    (Date.QUAL_ESTIMATED, _("Estimated")),
+    (Date.QUAL_CALCULATED, _("Calculated")),
+)
 
 CAL_TO_MONTHS_NAMES = {
-    Date.CAL_GREGORIAN  : displayer.short_months,
-    Date.CAL_JULIAN     : displayer.short_months,
-    Date.CAL_HEBREW     : displayer.hebrew,
-    Date.CAL_FRENCH     : displayer.french,
-    Date.CAL_PERSIAN    : displayer.persian,
-    Date.CAL_ISLAMIC    : displayer.islamic,
-    Date.CAL_SWEDISH    : displayer.swedish}
+    Date.CAL_GREGORIAN: displayer.short_months,
+    Date.CAL_JULIAN: displayer.short_months,
+    Date.CAL_HEBREW: displayer.hebrew,
+    Date.CAL_FRENCH: displayer.french,
+    Date.CAL_PERSIAN: displayer.persian,
+    Date.CAL_ISLAMIC: displayer.islamic,
+    Date.CAL_SWEDISH: displayer.swedish,
+}
 
 WIKI_HELP_PAGE = URL_MANUAL_SECT1
-WIKI_HELP_SEC = _('Editing_dates', 'manual')
+WIKI_HELP_SEC = _("Editing_dates", "manual")
 
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 #
 # EditDate
 #
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 class EditDate(ManagedWindow):
     """
     Dialog allowing to build the date precisely, to correct possible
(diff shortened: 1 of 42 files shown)


Nick-Hall:
I'd like to see two new modifiers added for FROM and TO.  This would make FROM, TO and SPAN more consistent with BEFORE, AFTER and RANGE.  It should also make the implementation slightly simpler.

We also need to check that the solution works with inflected date formats.  Perhaps @PushKK could help us out with some Russian examples.

PushKK (commented):
I insert this code, run Gramps.
But what I need do?
I try run
`$ python3 -m unittest discover -p 'date_test.py'
`
I small change code on 'date_test.py':
`
830    i=0
842        print('Row=', self.i+510, failstr, 'but=', date1.match(date2))
865            self.i += 1
`
And receive error:

```
Row= 552 before 1960-01-01 = 1959-12-31 should be True but= False
F.....s................
FAIL: test_comparison (gramps.gen.lib.test.date_test.DateComparisonTest)
  File "./gramps/gramps/gen/lib/test/date_test.py", line 864, in test_comparison
    self.do_case(*testdata)
  File "./gramps/gramps/gen/lib/test/date_test.py", line 844, in do_case
    self.assertEqual(date1.match(date2), expected, failstr)
AssertionError: False != True : before 1960-01-01 = 1959-12-31 should be True
Ran 24 tests in 1.964s

FAILED (failures=1, skipped=1)
```

If I change True to False in row 552 and run test again:

```
Row= 553 before 1960-01-01 = 1960-01-01 should be False but= True
F.....s................
FAIL: test_comparison (gramps.gen.lib.test.date_test.DateComparisonTest)
  File "./gramps/gramps/gen/lib/test/date_test.py", line 864, in test_comparison
    self.do_case(*testdata)
  File "./gramps/gramps/gen/lib/test/date_test.py", line 844, in do_case
    self.assertEqual(date1.match(date2), expected, failstr)
AssertionError: True != False : before 1960-01-01 = 1960-01-01 should be False
Ran 24 tests in 1.986s

FAILED (failures=1, skipped=1)
```
Row 553 have bad(not) value.

I think - problem in work 'date1.match(date2)' with 'before' and other textual dates.
Test ended on 
```
    def match(self, other_date, comparison="="):
        if (other_date.modifier == Date.MOD_TEXTONLY or self.modifier == Date.MOD_TEXTONLY):
            if comparison == "=":
                return self.text.upper().find(other_date.text.upper()) != -1
```
Date 'before 1960-01-01' have Date.MOD_TEXTONLY and test have '='. Return.

P.S. Linux Mint 20.1, Xfce
```
$ python3 -V
Python 3.8.5

$ locale
LANG=ru_RU.UTF-8
LANGUAGE=ru_RU:ru
LC_CTYPE="ru_RU.UTF-8"
LC_NUMERIC=ru_RU.UTF-8
LC_TIME=ru_RU.UTF-8
LC_COLLATE="ru_RU.UTF-8"
LC_MONETARY=ru_RU.UTF-8
LC_MESSAGES="ru_RU.UTF-8"
LC_PAPER=ru_RU.UTF-8
LC_NAME=ru_RU.UTF-8
LC_ADDRESS=ru_RU.UTF-8
LC_TELEPHONE=ru_RU.UTF-8
LC_MEASUREMENT=ru_RU.UTF-8
LC_IDENTIFICATION=ru_RU.UTF-8
LC_ALL=
```

janskarvall:
I believe that  @PushKK's ran into problems with 'date_test.py' because the 'DateComparisonTest' can only be run in the English locale.
I suppose that I should have inserted the line:
`@unittest.skipUnless(ENGLISH_DATE_HANDLER,
        "This test of Date() matching logic can only run in English locale.")
`
just before the 'DateComparisonTest' class.

janskarvall:
Referring to @Nick-Hall's suggestion to introduce modifiers for FROM and TO:
I have tried two ways to implement his wish:
- representing an open span as a compound date, one of the dates being Date.EMPTY. (This is was I originally did.)
- representing an open span as a non-compound date (just like BEFORE and AFTER)

After having done that, I would put my vote on representing FROM and TO dates as non-compound dates. Doing it the other way gave me problems with the handling of Date.sortval. I can not say that I have come to full understanding of the intention and overall use of the sortval, but I think that when sorting dates in the GUI, it is better to find a Date = TO 1999 in the vicinity to a DATE = 1999 rather than at the top (or bottom) of the list.

One improvement with  @Nick-Hall's suggestion is that entering open spans in the Date Editor (editdate.py) becomes more intuitively. You may only enter a single date, rather than entering a spanned date with one of the dates empty. FROM and TO becomes two new selectable date type values. in the Date Editor. However, that behaviour in Date Editor can also be implemented when representing an open span as a compound date, but it is not as straightforward.

Additionally, I have understood that the addition of open spans have impact on database export and import, and I now think I have come to grips with the necessary changes. I see that the GeneWeb format does not distinguish between "between .. and ..." and "from .. to ..", so in that case the import from GeneWeb has to make a choice, and I chose mapping onto FROM and TO if open ended, as already a two date entry is currently mapped to a closed span.

What I really needs advice upon, is to how to handle gramps.pot and all .po-files, now that WebLate have been introduced.
My "improvement" of datehandler_test.py, which now runs its test for most of the "supported" languages, revealed a lot of issues, some of which needed corrections in the .po files. That is besides that the introduction of open spans needs additions to the gramps.pot file (both for open span strings, and now also for the Date editor date type menu additions).

To achieve a successful Travis run, my (suggested) changes to the .po files needs to be part of this PR.
Should I:

1. Make minimum changes to gramps.pot and the .po files to make them stand out.
2. Regenerate gramps.pot and update the .po files to match. This will clutter up the diff views.
3. Cut down on the languages tested by datehandler_test.py, and somehow get translators to accept my suggestions. Whenever WebLate produces the .po file required by datehandler_test.py, enable testing of that language. A bit mundane...
4. Something else. I have not enough knowledge about what could be done using WebLate, to achieve a successful run of the Travis test of this PR.

I see that gramps master get a lot of WebLate related updates. Perhaps I should stay low until that activity recedes?

Besides this, I tentatively added an AgeTest to date_test.py, just to assure myself that I did not break anything with the introduction of open spans. I stopped as I found out that the age calculation seems a bit flawed. I have not tried to correct this, but the tests  I added so far had the same (bad) outcome after adding open spans. Here is an extract of the test, with the expected values are the ones output by the master branch (commit b85f5c1c8bdaf7b0ddd5cda583698caadd62f8c3):

<pre>
class AgeTest(BaseDateTest):
    """
    Age tests.
    """
    tests = [("1960", "1960", "0 days"), # NOK. No "()". Logically (less than 1 year).
             ("1960-01", "1960-01", "unknown"), # NOK. No "()". Logically (less than 1 month)
             ("1960-01", "1960-02", "unknown"), # NOK. No "()". Logically (between 1 month and 1 month, 27|28) days (2 months?)).
             ("1960-01-01", "1960-01-02", "(1 days)"), # OK
             ("1960-01-01", "1960-01-31", "(30 days)"), # OK
             ("after 1960-01-01", "1960-01-31", "(less than 29 days)"), # OK
             ("1960-01-01", "1960-02-02", "(1 months, 1 days)"), # OK
             ("1960-01-01", "1961-02-02", "unknown"), # NOK, (1 years, 1 months, 1 days)
             ("1960-01-01", "1965-01-01", "(5 years)"), # OK
             ("1960-01-01", "1965-02-02", "(5 years, 1 months)"), # OK if 1 days is ignored..
             ("about 1960-01-01", "before 1965-02-02", "(less than about 5 years, 1 months)"), # OK
             ("about 1960-01-01", "after 1965-02-02", "(more than about 5 years, 1 months)"), # OK
             ("1960", "2000", "(40 years)"), # OK? between 39 years, 1 days and 40 years, 364 days.
             ("1760", "", "greater than 110 years"), # OK. No "()".
             ("1960", "", "greater than 110 years"), # NOK. No "()".
             ("2020", "", "greater than 110 years"), # NOK. No "()".
             ("3020", "", "greater than 110 years"), # NOK. No "()".
             ("", "2000", "(1999 years)"), # NOK. (unknown)?.
           ]

    def convert_to_date(self, d):
        return d if isinstance(d,Date) else _dp.parse(d)
    
    def do_case(self, d1, d2, exp_age):
        span = Span(self.convert_to_date(d1), self.convert_to_date(d2))
        age = span.as_age()
        failstr = "Birth at " + d1 + ", death at " + d2 + " gives age " + age + ". Should be {}".format(exp_age)
        self.assertEqual(age, exp_age, failstr)
        
    def test_age(self):
        for testdata in self.tests:
            self.do_case(*testdata)
 </pre>

I believe that it should not be allowed to enter a span, open or not, as a birth, death, ... date, i.e. any event that is connected to a single date. Around, before, after, between is OK in such cases, but not spans. I understand that such a restriction needs thorough investigation on gramps itself, import/export plugins and whatever ...

Jan, hoping for that my "Internet tab added to XXX category" can be put to rest in favour of web located media. I can live with that Media itself would lack the addition of web links, but I hope for that Media can still be added to a Place (as opposed to what is possible in GEDCOM 7.0 - maybe it should be a suggestied addition to the GEDCOM 7.0 spec.)

Nick-Hall:
> I would put my vote on representing FROM and TO dates as non-compound dates.

Yes.  That's the way to go.

> What I really needs advice upon, is to how to handle gramps.pot and all .po-files, now that WebLate have been introduced.

 Generate a new pot file and add new strings to the po files.  I'll need to merge the PR manually.

> Jan, hoping for that my "Internet tab added to XXX category" can be put to rest in favour of web located media.

Yes.  I'm also in favour of allowing urls in media objects.

janskarvall:
> Generate a new pot file and add new strings to the po files. I'll need 
> to merge the PR manually.
>
So I will run 'python3 update_po.py -p' to generate gramps.pot.

Should I also update the .po files by running 'python3 update_po.py -m all'?

Jan

Nick-Hall:
> So I will run 'python3 update_po.py -p' to generate gramps.pot.

Yes.  You need to generate a new template file when translated strings are added or changed.

> Should I also update the .po files by running 'python3 update_po.py -m all'?

For this PR it may be the easiest way.   Normally Weblate will do this when the template file is updated,

I just need the new strings.  I'll lock Weblate prior to the merge and use `msgcat` to update the po files.

Nick-Hall:
@janskarvall I've updated and simplified the code.

I've also removed the translations.  The "from" and "to" will now need translating.  This simplifies the merge.

A couple of `date_test` unit tests are failing and I had to exclude 'eo', 'he', 'ja', 'tr' and 'vi' from the `datehandler_test`.  Have a look at the output.  We need to remove the print statements before merging.

The import and export code looks like it may need changing.  There is still a reference to a stop date.

gramps/gui/editors/editdate.py
```diff
@@ -74,13 +74,15 @@
...
-    (Date.MOD_ABOUT, _('About')),
-    (Date.MOD_RANGE, _('Range')),
-    (Date.MOD_SPAN, _('Span')),
-    (Date.MOD_TEXTONLY, _('Text only')))
+    (Date.MOD_NONE       , _('Regular')),
+    (Date.MOD_BEFORE     , _('Before')),
+    (Date.MOD_AFTER      , _('After')),
+    (Date.MOD_ABOUT      , _('About')),
+    (Date.MOD_RANGE      , _('Range')),
+    (Date.MOD_FROM       , _('From')),
```
SNoiraud, line 82: Why did you add spaces ? It's nicer but pylint says it's not good:

$  pylint3 gramps/gui/editors/editdate.py
Using config file /home/gramps/pull-gramps-1/.pylintrc
************* Module gramps.gui.editors.editdate
C: 77, 0: Exactly one space required after comma
    (Date.MOD_NONE,    _('Regular')),
                  ^ (bad-whitespace)

Nick-Hall: Fixed.  Thank you.

Nick-Hall:
> I think that date comparison "assertion" ("from 1960", "!=", "from
> 1961", False) looks right, i.e. that the two spans should have no date
> in common to be True, so I guess that there is a bug in the comparison
> implementation.

I changed the operators '==', '!=', '>=' and '<=' back to use the sort value.  I've updated the unit test accordingly.

> I also think that date matching "assertion" ("from 1950", "jan 1, 1950",
> True) looks right too, i.e.that the two spans should have a date in
> common to be True, so I guess that there is a bug in the match
> implementation.

This was a bug that I have now fixed.

> Do you know why 'eo', 'he', 'ja', 'tr' and 'vi' failed?

I think that 'eo', 'he', 'tr' and 'vi' are well translated, but use the default date handler.  As such we don't expect a round trip test to work.

We should be able to get 'ja' to work, but I removed you fixes.

Nick-Hall:
I've fixed the Japanese date parser.  All unit tests are now passing.

This PR now needs testing.

I may split off the enhancements to the date handler tests and merge it separately.

Nick-Hall:
Split off the date handler test into PR #1356.

Nick-Hall:
Rebased.

Nick-Hall:
I think we are now in a position to merge this.  I'll ask users to test this further in the beta release.

PR #1126, opened 2020-09-18
Update Croatian

milotype:
... this is my first try to send a file via GitHub Desktop via my Mac OS computer. Please excuse, if something doesn't happen as expected!


milotype:
I'm sorry, but I found more things I'd like to change in the "hr.po" file.
So this PR is irrelevant !
I'll delete my gramps repository, so I can start off with a fresh one next time.

PR #1131, opened 2020-09-29
Do not round rotation in SVG

tammojan:
When making >8 generation fan charts, the rounding of rotation becomes visible. SVG does not force rotation to be integers, so better not round this.

### DIFF
--- a/gramps/plugins/docgen/svgdrawdoc.py
+++ b/gramps/plugins/docgen/svgdrawdoc.py
@@ -128,7 +128,8 @@ class SvgDrawDoc(BaseDoc, DrawDoc):
         self.buffer.write(
             '<text ' +
             'x="%4.2f" y="%4.2f" ' % (xpos, ypos) +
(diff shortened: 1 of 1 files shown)


gramps/plugins/docgen/svgdrawdoc.py
```diff
@@ -128,7 +128,7 @@ def rotate_text(self, style, text, x, y, angle, mark=None):
         self.buffer.write(
             '<text ' +
             'x="%4.2f" y="%4.2f" ' % (xpos, ypos) +
-            'transform="rotate(%d %4.2f %4.2f)" ' % (angle, centerx, centery) +
+            'transform="rotate(%4.2 %4.2f %4.2f)" ' % (angle, centerx, centery) +
```
QuLogic, line 131: I think you mean:
```suggestion
            'transform="rotate(%4.2f %4.2f %4.2f)" ' % (angle, centerx, centery) +
```

tammojan: Yes I did, thanks for checking! I amended the commit with the extra `f`.

PR #1133, opened 2020-10-03
Fanchart overhang

tammojan:
This makes it possible to create fan charts with an overhang, which seems to be a commonly used way to make optimal use of the available paper. The title is put in the available space below the chart.

I'm not sure 'overhang' is the right word for this; suggestions are welcome.

I tested this for SVG and PDF export, however the code is generic enough that I trust also other export formats will work.

For example:
fanchartexample

### DIFF
--- a/gramps/plugins/drawreport/fanchart.py
+++ b/gramps/plugins/drawreport/fanchart.py
@@ -30,7 +30,7 @@
 # python modules
 #
 #------------------------------------------------------------------------
-from math import pi, cos, sin, log10, acos
+from math import pi, cos, sin, log10, acos, asin, degrees, radians
 
 def log2(val):
     """
@@ -69,6 +69,7 @@ from gramps.gen.display.name import displayer as _nd
 FULL_CIRCLE = 0
 HALF_CIRCLE = 1
 QUAR_CIRCLE = 2
+OVERHANG = 3
 
 BACKGROUND_WHITE = 0
 BACKGROUND_GEN = 1
@@ -96,10 +97,9 @@ def draw_wedge(doc, style, centerx, centery, radius, start_angle,
 
     path = []
 
-    degreestoradians = pi / 180.0
-    radiansdelta = degreestoradians / 2
-    sangle = start_angle * degreestoradians
-    eangle = end_angle * degreestoradians
+    radiansdelta = radians(0.5)
+    sangle = radians(start_angle)
+    eangle = radians(end_angle)
     while eangle < sangle:
         eangle = eangle + 2 * pi
     angle = sangle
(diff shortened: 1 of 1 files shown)


tammojan:
I see the test coverage decreases, but I have trouble finding the place where this should be tested. Could someone point me at the place where I could add tests for the added fanchart option?

prculley:
At this time we don't have tests that cover ANY of the GUI code.  A problem that bites us every now and then...  Designing test code for GUIs is a non-trivial challenge that none of us developers has yet taken on.  So your submission is fine as it is, at least as far as testing is concerned.

emyoulation:
### Did a few comparisons and diagrams.  Here is a labeled chart comparing the new layout to FamilySearch & Gramps original:
FanLayout pnd

### captured some text that has various descriptors for the chart characteristics:

traditional semicircle Ancestor [Fan Chart](https://en.wikipedia.org/wiki/Family_tree#Fan_chart), a quarter circle, or a full circle


The full arc of a [semicircle](https://en.wikipedia.org/wiki/Semicircle) always measures 180° (equivalently, π radians, or a half-turn). It has only one line of symmetry (reflection symmetry). In non-technical usage, the term "semicircle" is sometimes used to refer to a half-disk, which is a two-dimensional geometric shape that also includes the diameter segment from one end of the arc to the other as well as all the interior points. 


<hr />

A [circular sector](
https://en.wikipedia.org/wiki/Circular_sector) or circle sector (symbol: ⌔), is the portion of a disk enclosed by two radii and an arc, where the smaller area is known as the minor sector and the larger being the major sector. In the diagram, θ is the central angle, r {\displaystyle r} r the radius of the circle, and L {\displaystyle L} L is the arc length of the minor sector. 

A sector with the central angle of 180° is called a half-disk and is bounded by a diameter and a semicircle. Sectors with other central angles are sometimes given special names, these include quadrants (90°), sextants (60°) and octants (45°), which come from the sector being one 4th, 6th or 8th part of a full circle, respectively. Confusingly, the arc of a quadrant can also be termed a quadrant. two-thirds

The angle formed by connecting the endpoints of the arc to any point on the circumference that is not in the sector is equal to half the central angle.

tammojan:
Thanks for your investigation. The angle in the screenshot is 25˚ (so the sector is 180˚ + 25˚ + 25˚ = 230˚), but in a later commit I changed the angle to make optimal use of the available paper. So depending on whether you're choosing A4 or letter paper, a different angle will be used. This can of course be changed, or the angle can be made into a user parameter.

emyoulation:
> or the angle can be made into a user parameter

A parameter would be wonderful.  

Another wonderful thing would be to be able to have a [knockout](https://www.cutpasteandprint.com/knockout-mean-printing/) option for the [center person](https://cdn.shopify.com/s/files/1/0080/2937/2476/products/1_kids_fan_chart_white_1024x1024@2x.jpg?v=1600370356). Leaving a blank would let us re-use the same chart on webpages for every child of that family group.

And a separate knockout option for Privacy (Living, Flagged Private or both).
So... say I want a wedding fan chart for my  parents.  I set myself, the parents & grandparents as 'Private' and the knockouts (rather than color filled blanks) leave space to insert their wedding photo where they were flanked by the grandparents.

fanchart

emyoulation:
I just applied the patches and found a minor problem with the chart subtitle.  The new overhang overwrites the subtitle.

image

PR #1136, opened 2020-10-10
Update Dutch translation

Nick-Hall:


gramps/gen/datehandler/_date_nl.py
```diff
@@ -147,17 +149,17 @@ class DateDisplayNL(DateDisplay):
...
-    _mod_str = ("", "voor ", "na ", "rond ", "", "", "")
+    _mod_str = ("", "voor ", "na ", "omstreeks ", "", "", "")
 
     _qual_str = ("", "geschat ", "berekend ")
 
     _bce_str = "%s v. Chr."
 
     formats = (
-        "JJJJ-MM-DD (ISO)", "Numerisch DD/MM/JJ", "Maand Dag, Jaar",
+        "JJJJ-MM-DD (ISO)", "Numerisch DD/MM/JJJJ", "Maand Dag, Jaar",
```
bartfeenstra, line 162: I believe this should be "Numeriek".

gramps/gen/datehandler/_date_nl.py
```diff
@@ -147,17 +149,17 @@ class DateDisplayNL(DateDisplay):
 
     calendar = (
         "", "juliaans", "hebreeuws",
-        "franse republiek", "persisch", "islamitisch",
+        "frans republikeins", "persisch", "islamitisch",
         "zweeds" )
 
-    _mod_str = ("", "voor ", "na ", "rond ", "", "", "")
+    _mod_str = ("", "voor ", "na ", "omstreeks ", "", "", "")
```
bartfeenstra, line 155: Either word would work here :)

gramps/plugins/rel/rel_nl.py
```diff
@@ -69,9 +70,207 @@
...
+                   "overgrootouders 45e graad",
+                   "overgrootouders 46e graad",
+                   "overgrootouders 47e graad",
+                   "overgrootouders 48e graad",
+                   "overgrootouders 49e graad",
+                   "overgrootouders 50e graad", ]
+
+_siblings_level = ["",
+                   "broers en zussen",
+                   "ooms/tantes",
```
bartfeenstra, line 127: No "en" instead of a slash, like with "broers en zussen"?

gramps/gen/datehandler/_date_nl.py
```diff
@@ -83,35 +83,37 @@ class DateParserNL(DateParser):
...
-        'z'               : Date.CAL_SWEDISH,
+        'gregoriaans'         : Date.CAL_GREGORIAN,
+        'greg.'               : Date.CAL_GREGORIAN,
+        'juliaans'            : Date.CAL_JULIAN,
+        'jul.'                : Date.CAL_JULIAN,
+        'hebreeuws'           : Date.CAL_HEBREW,
+        'hebr.'               : Date.CAL_HEBREW,
+        'islamitisch'         : Date.CAL_ISLAMIC,
+        'isl.'                : Date.CAL_ISLAMIC,
+        'frans republiekeins' : Date.CAL_FRENCH,
```
bartfeenstra, line 106: This is the correct label, although it, along with the other calendar names, should be capitalized. However, the fact this is a conversion map requires me to ask: does this change break backwards compatibility?

gramps/gen/datehandler/_date_nl.py
bartfeenstra, line 155: Either word works here.

gramps/plugins/rel/rel_nl.py
```diff
@@ -69,9 +70,207 @@
...
+                   "kleinkinderen 45e graad",
+                   "kleinkinderen 46e graad",
+                   "kleinkinderen 47e graad",
+                   "kleinkinderen 48e graad",
+                   "kleinkinderen 49e graad",
+                   "kleinkinderen 50e graad", ]
+
+_nephews_nieces_level = ["",
+                         "broers en zussen",
+                         "neven/nichten",
```
bartfeenstra, line 221: No "en" instead of a slash, like with "broers en zussen"?

Nick-Hall:
@janpython Should I make the changes suggested by @bartfeenstra and then merge?

gramps/plugins/rel/rel_nl.py
janpython: Yes, good idea. Let's change it into:
"neven en nichten",

gramps/gen/datehandler/_date_nl.py
janpython: Not any more, in new Dutch translation I have changed 'rond' into 'omstreeks'. So datehandler should contain 'omstreeks' too. Otherwise you could get an error when trying to fill in 'omstreeks'.
Maybe it's possible to have both words?

gramps/gen/datehandler/_date_nl.py
janpython: As far as I can oversee this only changes the way it is translated into Dutch. This change should not change the functionality.
I have this datehandler running now for a few months and don't experience difficulties yet.

gramps/plugins/rel/rel_nl.py
janpython: Good idea, let's change it into:
"ooms en tantes",

gramps/gen/datehandler/_date_nl.py
janpython: Not any more, in new Dutch translation I have changed 'rond' into 'omstreeks'. So datehandler should contain 'omstreeks' too. Otherwise you could get an error when trying to fill in 'omstreeks'.
Maybe it's possible to have both words?

gramps/gen/datehandler/_date_nl.py
janpython: Indeed, very well found!
Please, change it into "Numeriek".

janpython:
Hi Nick,
Yes, Bart's suggestions are very good. I have commented all his suggestions above, if you're willing to change them and merge the files I will be very happy!!!

Thank you very much for your good work!!!
Jan Sparreboom

gramps/gen/datehandler/_date_nl.py
bartfeenstra: Ah, keeping those two components in sync makes sense. Does this not break BC, though? E.g. will existing date entries using "rond" stop being parseable?

gramps/plugins/rel/rel_nl.py
Nick-Hall: Do you want the whole list changed in the same way?

"oudooms/tantes" -> "oudooms en tantes"
"overoudooms/tantes" -> "overoudooms en tantes"
"overoudooms/tantes 5e graad" -> "overoudooms en tantes 5e graad"
etc...

gramps/plugins/rel/rel_nl.py
Nick-Hall: Do you want the whole list changed in the same way?

gramps/plugins/rel/rel_nl.py
ennoborg: This is wrong, because it should be oudooms en -tantes, where the dash means that oud as a prefix is left out.

Nick-Hall: So you are suggesting:

"oudooms/tantes" -> "oudooms en -tantes"
"overoudooms/tantes" -> "overoudooms en -tantes"
"overoudooms/tantes 5e graad" -> "overoudooms en -tantes 5e graad"

and further down:

"achterneven/nichten" -> "achterneven en -nichten"
"achterneven/nichten 4e graad" -> "achterneven en -nichten 4e graad"

Is that correct?

bartfeenstra: Good suggestion! That looks correct.

ennoborg: Yes, that's correct.

janpython: Yes, I agree. Thanks all for good help!!

gramps/gen/datehandler/_date_nl.py
janpython: Well, if you mean it doesn't change old reports or something I guess it will break BC as a translation is a 1-1 relationship.
This is a translation of the word 'about' which is translated most correctly as 'omstreeks'. The Dutch word 'rond' should be used in case the word 'around' was used and that's not the case here.
I've created a long list of words which I use now as a kind of convention for translation of all different addons.
So, maybe it breaks something, but at least it will break everywhere in same way. Hopefully the translation into Dutch will improve this way.
But, maybe it is possible to use both words by a little programming.

Nick-Hall:
Thanks to all three of you.

PR #1138, opened 2020-10-20
In familysidebarfilter, search on each part of name

prculley:
Instead of requiring that the entire search string matches a single
one of the Person's names, the function will require that each word
in the search string matches any of the Person's name fields.

Fixes #12023

This same enhancement was done to person sidebar in https://github.com/gramps-project/gramps/commit/dd9ddab84920538cf475b3bd700a7a1288f30043 several years ago.

### DIFF
--- a/gramps/gui/filters/sidebar/_familysidebarfilter.py
+++ b/gramps/gui/filters/sidebar/_familysidebarfilter.py
@@ -163,16 +163,43 @@ class FamilySidebarFilter(SidebarFilter):
                 generic_filter.add_rule(rule)
 
             if father:
-                rule = RegExpFatherName([father], use_regex=regex)
-                generic_filter.add_rule(rule)
+                # if the name is not empty, split the name in multiple part if
+                # we don't use regexp. if the regexp is used, don't split the
+                # field
+                if not regex:
+                    name_parts = father.split(sep=" ")
+                    for name_part in name_parts:
+                        rule = RegExpFatherName([name_part], use_regex=regex)
+                        generic_filter.add_rule(rule)
+                else:
+                    rule = RegExpFatherName([father], use_regex=regex)
+                    generic_filter.add_rule(rule)
 
             if mother:
-                rule = RegExpMotherName([mother], use_regex=regex)
-                generic_filter.add_rule(rule)
+                # if the name is not empty, split the name in multiple part if
+                # we don't use regexp. if the regexp is used, don't split the
+                # field
+                if not regex:
+                    name_parts = mother.split(sep=" ")
+                    for name_part in name_parts:
+                        rule = RegExpMotherName([name_part], use_regex=regex)
+                        generic_filter.add_rule(rule)
+                else:
+                    rule = RegExpMotherName([mother], use_regex=regex)
+                    generic_filter.add_rule(rule)
 
             if child:
-                rule = RegExpChildName([child], use_regex=regex)
-                generic_filter.add_rule(rule)
+                # if the name is not empty, split the name in multiple part if
+                # we don't use regexp. if the regexp is used, don't split the
+                # field
+                if not regex:
+                    name_parts = child.split(sep=" ")
+                    for name_part in name_parts:
+                        rule = RegExpChildName([name_part], use_regex=regex)
+                        generic_filter.add_rule(rule)
+                else:
+                    rule = RegExpChildName([child], use_regex=regex)
+                    generic_filter.add_rule(rule)
 
             if etype:
                 rule = HasEvent([etype, '', '', '', ''], use_regex=regex)


emyoulation:
The performance hit for searching on words (instead of phrase) on across all name fields has made the Filter unbearably slow for large Trees.  (More than 15 seconds for a 38K tree on a 1.5 GHz AMD A6-3420M APU with 8GB)

The current workaround in the People View is to temporarily use the Search instead. It takes less than half the processing time because it only does a phrase search and ONLY on the concatenated name as set in Preferences for "Name format:"

Also, the existing phrase search allows finding all the "Robert George" individuals without finding the "George Robert", "Smith, George Robert", "Roberts, George" or "George, Robert"

SNoiraud:
> The performance hit for searching on words

Did you try to set the use regexp option ? In this case, the search isn't split into words.

emyoulation:
Thanks for that lead.  It is useful that the RegEx switch can be used to force the phrase search.  I have not been using that option. (It always seems to be hard to make the mental shift to RegEx) 

But the search time is still just as impacted with RegEx as without. (Probably because of all the various Name fields. Are they search sequentially?)

Would the Families filter take a triple hit because of the Father Mother & Child would then be doing sequential Name searches for 3 people?

vantu5z:
Maybe we should implement thread for search (like GraphView do). Than we see processed results immediately, progress can be shown at the bottom bar.

emyoulation:
It there a way to make this Enhancement reversible? Or maybe co-exist as an option? The original Phrase search doesn't have the performance penalty.

emyoulation:
Is it possible to profile the "multiple search term" patched version and the original "phase" search?  Since the RegEx switch doesn't affect performance, maybe there is another way to have a phrase mode which doesn't suffer the penalty? 

The current workaround for a fast name filter is to turn off the sidebar to use the Searchbar instead. But a proposed patch to the Searchbar would extend the same multi-term searching AND PERFORMANCE PENALTY ... thereby disabling the workaround.

"Gramps has a convenience hook to allow you to do profiling. Profiling gives an overview how many times methods are called in a code fragment and how long each one took. It helps to find performance bottlenecks." 
https://gramps-project.org/wiki/index.php/Debugging_Gramps#Use_profiling 


https://gramps-project.org/wiki/index.php?title=GEPS_016:_Enhancing_Gramps_Processing_Speed

prculley:
I think a better approach would be to vastly improve the performance of the search.  The current method is to run the underlying rule once for every search term (word) in the search, thus scanning all the people (or families) once for every word, and then combining the results with an 'AND' function.  One of the slowest ways of getting the job done I can imagine.

Instead we should rewrite the underlying rule to do the work during a single scan.  What I have in mind is to allow the user to either use single space separated words, or a combination of space separated words and double quoted strings, which would be matched exactly.  Note that the quoted phrase part of this would still only match a name part (like given) , not a given+surname combination.  For that you would still have to enter `"John Robert" Smith` for example.
Also, if using regex, we would revert to the original search method, without the multi word feature at all.

So notes to self:
return all(sub_search(s_t) for s_t in search_terms)
where search_terms is list of terms in the disassembled search string (see https://stackoverflow.com/questions/79968/split-a-string-by-spaces-preserving-quoted-substrings-in-python)
where sub_search scans list of name parts of a name and matches s_t, returning true for a match
remember that these searches are case insensitive...

Nick-Hall:
Yes, it could certainly benefit from a re-write.  The rule should match a Name object.

emyoulation:
If there is a possibility of having the Search interruptible, then we might gain another dividend.

After the load & filter when switching from another view, the next step is often a sort or a filter reset. If applying the filter became fast, that is no impediment. But should you need to change the parameter and sort, then that can absorb 3 full (and long) repeats of the filtering.

prculley:
@emyoulation It seems I was mistaken about the multi-word performance hit; it doesn't require scanning the whole person list for each word.  So when I rewrote the rule to speed it up, there was only a minor performance improvement.

I'm beginning to wonder if you are seeing some other performance issue.  Or if you are comparing current performance with the multi-word PR with some past memory when your tree was smaller or some other issue changed the results.

Could you perform a test for me?  Start your Gramps from the command line with the additional switch like
 `Gramps -d ".gui.listview"`. then perform some searches and watch the command window after each search, it will report a "build_tree" number of seconds.

A sidebar search reset and find should be fastest, as no searching is done.
A single word search is nearly the same as before the multi word PR was accepted.
Please report these as well as a two or three word search; I am curious what the actual performance degradation you are seeing.

If you have an older version of Gramps that you think doesn't have the performance hit, please try the same tests.

Note that when sidebar search is not present, the topbar search is available, but it doesn't do anywhere near as complete a search as the sidebar.  It only compares the name field that is shown in the view against the search term, where the sidebar version looks at all the names, surnames, and other fields associated with a person.  The topbar search is much faster, in that it can use the cached version of the names from the view as opposed to looking up in the db the entire person.

emyoulation:
Currently trying to trick 3.4.9 into importing an XML from 5.1.3-2 

Otherwise they won't be on equal data footing

emyoulation:
It is going to take me a LOT more effort to get an older version of Gramps up for testing.  But here are the baseline timings for the 5.1.3-2 on a 43k person Tree.

[Gramps513-2 filter timings 20210724.txt](https://github.com/gramps-project/gramps/files/6873114/Gramps513-2.filter.timings.20210724.txt)

I cannot use GrampsPortable because their launcher doesn't allow passing CLI parameters.  And when i launch directly from the gramps.exe inside their installation, then Gramps uses the User Directory in the normal location within %AppData%  rather than the relative path to the PortableApps launcher.

And the Gramps 3.4.9 doesn't have a Gramps.exe where I can use the CLI switches.  [And it doesn't look like the python shell version offers a timings option](https://gramps-project.org/wiki/index.php/Gramps_3.4_Wiki_Manual_-_Command_Line)

prculley:
Wow, when you were talking about older version, I did not think you were talking that far back...  I thought you were objecting to the performance of the multi-word search.  In your results attachment the difference between the sidebar one word and two word search is slight (two word looks a bit faster, probably due to fewer results to display).  So you are really not seeing performance issues related to this PR or the Person view version of the PR at all.

It seems your issue is related to the sidebar search in general.  Which, except for the multi-word PRs hasn't changed since Gramps 4.1.x.  Perhaps you should submit your issue to the bug tracker and disconnect it from this PR.

emyoulation:
Well... I wasn't PLANNING on going back that far.  I starting using Gramps during version. So this experiment with 3.4.9 was an eye-opener. (The interface was antiquated but it was screaming fast opening the database.)  And I really only wanted to go back to the version before the multi-term Name search enhancement to the filter gramplet.

Unfortunately, there was no version 4.x.x of GrampsPortable. And I did not feel confident I could keep my configuration customizations isolated with other installs.  (Previously, I had used another machine to have a separate test environment.  But I'm down from 4 desktops & laptops to 1 faltering laptop in the last 2 years due to aging hardware failures.  And pandemic austerity precludes correcting that difficulty.)

I agree that the general optimization of the Filter gramplet is a very important goal and should be divorced from this thread.  Certainly divorced from any decision whether to commit this Pull Request.

However, my greatest concern is that the queued up patches to the searchbar not be rolled-into 5.2 without benchmarking the performance penalty.

And a benchmarking requirement could only be justified by determining how much the patches to the Filter gramplet have already degraded performance.  **A key item that I do not believe was communicated is: every search using the Filter gramplet was massively slowed.** Gender filters with NO other parameters used to be fast.  Now that simple EQUALS matching is just as slow (well, within 2 seconds) as multiple term searching for Names. The only thing worse is a multi-stage custom filter.

This could be related to the enhancement in the name search. It could be related to the expansion to allow Rule add-ons. It could be something else. (I might have broken through a record count threshhold that demanded allocating resources differently.)

PR #1140, opened 2020-10-30
Update Appdata to pass `appstream-util validate`

a-andre:


a-andre:
This PR might fix https://gramps-project.org/bugs/view.php?id=9680

PR #1143, opened 2020-11-01
Use reverse DNS name scheme

a-andre:


Nick-Hall:
@OzarkShepherd is currently working on Flatpak support in the _flatpak_ repository.

This PR should enable the Flatpak to reuse our appdata and desktop files.

Maybe we can backport this and the updated appdata file to the _gramps51_ branch.

Nick-Hall:
Rebased.

One typo fixed in desktop reference in appdata file.

PR #1144, opened 2020-11-04
Explicitly set birth name in GEDCOM export

Jmuccigr:
Following discussion at https://gramps.discourse.group/t/distinguish-birth-and-married-names-in-ged-file/815/7 .

### DIFF
--- a/data/tests/exp_sample_ged.ged
+++ b/data/tests/exp_sample_ged.ged
@@ -20,6 +20,7 @@
 1 EMAIL an_email@@gmail.com
 0 @I0000@ INDI
 1 NAME Anna /Hansdotter/
+2 TYPE birth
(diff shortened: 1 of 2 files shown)


Jmuccigr:
Not really sure why this failed. Any help?

prculley:
It failed because the exports_test is seeing differences in the exported Gedcom file between the prior version and what your are exporting with this change.  The added lines in the Gedcom are a difference, and need to be accounted for in the template file(s) for the tests.

You should run the plugins/test/exports_test.py on your system, and look at the difference file and current export in your Gramps temp directory.  If a careful examination of the differences are exactly what you expect then you can replace the template .ged file(s) in the data/tests directory with the new output files in the temp directory, making the updated .ged files part of the PR.

PR #1145, opened 2020-11-05
New Austrian German translation

cedecode:
a new Austrian German translation based on the work of the German translation (only a new file de_at.po)


Nick-Hall:
Rebased.

Nick-Hall:
@Leonhaeuser Does this look reasonable to you?

Nick-Hall:
@cedecode Thank you. Please make any further updates to this translation on [Weblate](https://hosted.weblate.org/projects/gramps-project/gramps/de_AT).

cedecode:
@Nick-Hall Thank you too! Now I reached 100% in Weblate for de_at.

cedecode:
@Nick-Hall I tried to build a version with de_at. It seems to display all the strings from de_at.po correctly in the GUI program.
**When I export a report it still uses the wrong long months and short months.**

I assume that there is a _date_de_AT.py missing. Unfortunately, I didn't succeed in changing this.
Two changes from language de (German) should be enough: long month is Jänner (instead of Januar) and short month is Jän (instead of Jan).

Could you give me an advice please?

Nick-Hall:
See fix in PR #1264.

To test this comment out the month definitions.  Your translations will be inherited from the base class.

PR #1146, opened 2020-11-06
Add a note gramplet for NoteView

Mattkmmr:
Requested in Gramps [discourse forum](https://gramps.discourse.group/t/view-notes-in-note-list-without-opening-them/901).


emyoulation:
I tried this  patch with the 5.1.4 release but decided to remove it.

Instead of the Notes gramplet that shows the Current note

image

It was showing the Note gramplet that has an Editor
image

Mattkmmr:
@emyoulation The problem was that there is a third-party addon with the same id which was overwriting it, when installed. I renamed my gramplet's id and it should now work again as expected.

PR #1147, opened 2020-11-07
Skip time-consuming calls when debug logging is disabled

DavidMStraub:
Profiling the performance of some of the code in [web-api](https://github.com/gramps-project/web-api), I noticed that in some calls, more than 90% of the time was spent evaluating `inspect.stack` merely for the purpose of debug logging.

In fact, this is appreciated in part of the codebase by wrapping this call inside `if __debug__`:

https://github.com/gramps-project/gramps/blob/976edfbf1d7f7fc0d1bd26019d8a3121364be549/gramps/gen/db/txn.py#L126-L128

But there were also some cases where this was missing. So my first commit 9df8d56  adds the `if __debug__` in the remaining places.

However, this is not the whole story. Gramps uses `python -O` in the executable script to disable all the `if __debug__` blocks and this is certainly a great idea. But when using Gramps as a library - as is done for web-api - this does not work as we cannot rely on flags being set on the Python interpreter by whatever code imports our library, and `__debug__` is `True` by default.

Fortunately, there is a very simple way to remedy this without changing anything about how things work at the moment. Namely, `logging` has a well-defined way how to skip calls that are only necessary for a logging output that is not shown anyway: `isEnabledFor`, [see the logging docs](https://docs.python.org/3/howto/logging.html#optimization).

So my second commit 6c75218 replaces all occurrences of `if __debug__:` where `inspect.stack` is called by

```python
if __debug__ and _LOG.isEnabledFor(logging.DEBUG):
    ...
    _LOG.debug(...)
``` 

What will this do?

- For a normal Gramps user installation, `__debug__` is false (thanks to `-O`) and the second part of the `if` statement will never be executed - no change!
- For code using `gramps.gen` as a library and *not* using `-O`, it will now depend on the log level set by that code: if it's set to `DEBUG`, the block will be executed. If it's set to a higher level, it will be skipped entirely.

This way, we can speed up some endpoints in web-api by a factor of 10 and even make Gramps itself a little bit faster (by adding the missing `__debug__` statements).

### DIFF
--- a/gramps/gen/db/dummydb.py
+++ b/gramps/gen/db/dummydb.py
@@ -157,7 +157,7 @@ def wrapper(method):
         This 'wrapped' method logs the original function that was called, and
         where it was called from.
         """
-        if __debug__:
+        if __debug__ and LOG.isEnabledFor(logging.DEBUG):
             class_name = args[0].__class__.__name__
             func_name = method.__name__
             frame = inspect.currentframe()
--- a/gramps/gen/db/dummydb.py
+++ b/gramps/gen/db/dummydb.py
@@ -157,15 +157,16 @@ def wrapper(method):
         This 'wrapped' method logs the original function that was called, and
         where it was called from.
         """
-        class_name = args[0].__class__.__name__
-        func_name = method.__name__
-        frame = inspect.currentframe()
-        c_frame = frame.f_back
-        c_code = c_frame.f_code
-        LOG.debug('calling %s.%s()... from file %s, line %s in %s',
-                  class_name, func_name, c_code.co_filename, c_frame.f_lineno,
-                  c_code.co_name)
-        return method(*args, **keywargs)
+        if __debug__:
+            class_name = args[0].__class__.__name__
+            func_name = method.__name__
+            frame = inspect.currentframe()
+            c_frame = frame.f_back
+            c_code = c_frame.f_code
+            LOG.debug('calling %s.%s()... from file %s, line %s in %s',
+                      class_name, func_name, c_code.co_filename,
+                      c_frame.f_lineno, c_code.co_name)
+            return method(*args, **keywargs)
     return wrapped
 
 
--- a/gramps/gen/db/txn.py
+++ b/gramps/gen/db/txn.py
@@ -77,7 +77,7 @@ class DbTxn(defaultdict):
             self.db.transaction_abort(self)
 
         elapsed_time = time.time() - self.start_time
-        if __debug__:
+        if __debug__ and _LOG.isEnabledFor(logging.DEBUG):
             frame = inspect.currentframe()
             c_frame = frame.f_back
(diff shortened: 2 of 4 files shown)


Nick-Hall:
The first commit clearly fixes a bug and can the applied to the _gramps51_ branch.

The second commit would require us to change out current [debugging](https://gramps-project.org/wiki/index.php/Debugging_Gramps#Add_debug_statements) advice.  This change would need further discussion.

DavidMStraub:
> The second commit would require us to change out current [debugging](https://gramps-project.org/wiki/index.php/Debugging_Gramps#Add_debug_statements) advice. This change would need further discussion.

But concerning the logging output and debugging, unless I'm misunderstanding something, there should be absolutely no change.

When running `python Gramps.py` without `-O`, `__debug__` is `True`, so the current behaviour is to evaluate everything in the block including the `LOG.debug` call. The new behaviour is the same, except that it executes the block and the `LOG.debug` call only if the log level *in this context and of this specific logger* is set to `DEBUG`. So it only skips the block in cases when the logging function anyway wouldn't have had any effect. Do you see what I mean?

BTW, found another use of `isEnabledFor` in the code even if totally unrelated :wink:

https://github.com/gramps-project/gramps/blob/a265f1fd3813f0cbde662f5f5e174004c3fe4c74/gramps/plugins/importer/importxml.py#L482-L485

Nick-Hall:
I was just pointing out that we would need to change the policy and related documentation.  I'm not necessarily against the idea.

Are there any other consequences?  For example, I think that running `__debug__` set to `True` will make _UNSTABLE_ plugins available.

DavidMStraub:
> I was just pointing out that we would need to change the policy and related documentation. 

Right, I guess I just didn't get why the policy and the documentation need to change, I believe they can stay the same, or am I missing something?

> Are there any other consequences? For example, I think that running `__debug__` set to `True` will make _UNSTABLE_ plugins available.

Good question, I browsed the code and, apart from logging and some `print` statements, the relevant change in the presence of `__debug__` indeed seems to be this in the `PluginResgister` constructor:

https://github.com/gramps-project/gramps/blob/c567b9e399514cfbf3bc6e5dd3996a04b7f2ed97/gramps/gen/plug/_pluginreg.py#L1142-L1143

While this is not directly related to this PR, the question is what is the best way to handle this when using Gramps as a library, going forward.

A clean way would be to introduce a new globale variable `GRAMPS_DEBUG`, but this would obviously be a significant change, so it will certainly require some discussion.

In the meantime, I think it's best to open an issue about the fact that doing `import gramps` in any code that is not run with the `-O` flag loads unstable plugins.

Nick-Hall:
Rebase and fix conflicts.

Nick-Hall:
It is also usual to run unit tests without the `-O` flag.  This change could possibly speed up some tests.

PR #1148, opened 2020-11-09
Add Stillbirth as a pre-defined event type

DaveSch-gramps:
Add Stillborn as a fallback event for both Birth and Death

### DIFF
--- a/gramps/gen/lib/eventtype.py
+++ b/gramps/gen/lib/eventtype.py
@@ -82,6 +82,7 @@ class EventType(GrampsType):
     .. attribute DIV_FILING:     Divorce Filing
     .. attribute ANNULMENT:      Annulment
     .. attribute MARR_ALT:        Alternate Marriage
+    .. attribute STILLBIRTH:      Stillbirth
     """
     UNKNOWN = -1
     CUSTOM = 0
@@ -129,13 +130,14 @@ class EventType(GrampsType):
     RESIDENCE = 42
     RETIREMENT = 43
     WILL = 44
+    STILLBIRTH = 45
 
     # _T_ is a gramps-defined keyword -- see po/update_po.py and po/genpot.sh
     def _T_(value, context=''): # enable deferred translations
         return "%s\x04%s" % (context, value) if context else value
 
     _MENU = [[_T_('Life Events'),
-              [BIRTH, BAPTISM, DEATH, BURIAL, CREMATION, ADOPT]],
+              [BIRTH, BAPTISM, DEATH, STILLBIRTH, BURIAL, CREMATION, ADOPT]],
              [_T_('Family'),
               [ENGAGEMENT, MARRIAGE, DIVORCE, ANNULMENT, MARR_SETTL, MARR_LIC,
                MARR_CONTR, MARR_BANNS, DIV_FILING, MARR_ALT]],
@@ -205,6 +207,7 @@ class EventType(GrampsType):
         (DIV_FILING, _("Divorce Filing"), "Divorce Filing"),
         (ANNULMENT, _("Annulment"), "Annulment"),
         (MARR_ALT, _("Alternate Marriage"), "Alternate Marriage"),
+        (STILLBIRTH, _("Stillbirth"), "Stillbirth"),
         ]
 
     _ABBREVIATIONS = {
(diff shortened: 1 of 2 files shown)


DaveSch-gramps:
Feature Request #0011554

Nick-Hall:
The new Gedcom 7.0 standard suggests using the TYPE tag to provide further classification of an event.  “Stillborn” as a qualifier to a BIRT event is given as a specific example:

```
1 BIRT
2 TYPE Stillborn
```

Nick-Hall:
In the associated Discourse [topic](https://gramps.discourse.group/t/stillborn-as-a-coded-event-type/181), four people were in favour of this change.  One stated that he would continue to use two separate events, but I don't see that he would be affected by this change.

gramps/gen/lib/eventtype.py
```diff
@@ -253,7 +256,8 @@ def _T_(value, context=''): # enable deferred translations
         ENGAGEMENT: _T_("engd.", "Engagement abbreviation"),
         DIVORCE: _T_("div.", "Divorce abbreviation"),
         DIV_FILING: _T_("div.f.", "Divorce Filing abbreviation"),
-        ANNULMENT: _T_("annul.", "Annulment abbreviation")
+        ANNULMENT: _T_("annul.", "Annulment abbreviation"),
+        STILLBIRTH: _T_("Stillbirth abbreviation|still.")
```
Nick-Hall, line 260: Needs to use the new translation context:

`STILLBIRTH: _T_("still.", "Stillbirth abbreviation")`

gramps/gen/lib/eventtype.py
```diff
@@ -280,6 +284,13 @@ def is_death(self):
         """
         return self.value == self.DEATH
 
+    def is_stillborn(self):
```
Nick-Hall, line 287: Is this method needed?

gramps/gui/views/treemodels/peoplemodel.py
```diff
@@ -265,7 +265,9 @@ def _get_birth_data(self, data, sort_mode):
             event = self.db.get_event_from_handle(er.ref)
             etype = event.get_type()
             date_str = get_date(event)
-            if (etype in [EventType.BAPTISM, EventType.CHRISTEN]
+            if (etype in [EventType.STILLBIRTH,
```
Nick-Hall, line 268: Use `etype.is_birth_fallback()` instead.

gramps/gui/views/treemodels/peoplemodel.py
```diff
@@ -322,7 +324,8 @@ def _get_death_data(self, data, sort_mode):
             event = self.db.get_event_from_handle(er.ref)
             etype = event.get_type()
             date_str = get_date(event)
-            if (etype in [EventType.BURIAL,
+            if (etype in [EventType.STILLBIRTH,
```
Nick-Hall, line 327: Use `etype.is_death_fallback()` instead.

gramps/gen/lib/eventtype.py
DaveSch-gramps: Fixed in [15a0085](https://github.com/gramps-project/gramps/pull/1148/commits/15a0085b6316bb277400a9133eabf32b47be2558)

gramps/gen/lib/eventtype.py
DaveSch-gramps: Removed the def section and it still worked so obviously was not needed.

Fixed in [15a0085](https://github.com/gramps-project/gramps/pull/1148/commits/15a0085b6316bb277400a9133eabf32b47be2558)

gramps/gui/views/treemodels/peoplemodel.py
DaveSch-gramps: The same issue appeared at line 367 so also made that change.

Fixed in [15a0085](https://github.com/gramps-project/gramps/pull/1148/commits/15a0085b6316bb277400a9133eabf32b47be2558)

gramps/gui/views/treemodels/peoplemodel.py
DaveSch-gramps: The same issue appeared at line 407

Fixed in [15a0085](https://github.com/gramps-project/gramps/pull/1148/commits/15a0085b6316bb277400a9133eabf32b47be2558)

Nick-Hall:
I asked for feedback on the [mailing list](https://sourceforge.net/p/gramps/mailman/message/37631851/).  The only reply pointed out that there are some places where fallback events are not used.  We may need to check this, but I don't see any reason not to merge this PR now.

PR #1139, opened 2020-10-26
Add toolbar icons for Preferences and Plugin Manager

DaveSch-gramps:
Add Icons to open Preferences and the Plugin Manager

Feature Request [#12030](https://gramps-project.org/bugs/view.php?id=12030)

PR #1149, opened 2020-11-10
Update configure.py

DaveSch-gramps:
Feature Request [#12049](https://gramps-project.org/bugs/view.php?id=12049)

Clean up of existing Preferences Tabs

### DIFF
--- a/gramps/gui/configure.py
+++ b/gramps/gui/configure.py
@@ -362,11 +362,11 @@ class ConfigureDialog(ManagedWindow):
             text.set_text(label)
         text.set_halign(align)
         text.set_justify(justify)
-        text.set_hexpand(True)
         grid.attach(text, start, index, stop - start, 1)
         return text
 
-    def add_button(self, grid, label, index, constant, extra_callback=None, config=None):
+    def add_button(self, grid, label, index, constant, extra_callback=None,
+                   config=None):
         if not config:
             config = self.__config
         button = Gtk.Button(label=label)
@@ -558,16 +558,18 @@ class GrampsPreferences(ConfigureDialog):
 
     def __init__(self, uistate, dbstate):
         page_funcs = (
-            self.add_behavior_panel,
+            self.add_data_panel,
+            self.add_general_panel,
+            self.add_addons_panel,
             self.add_famtree_panel,
-            self.add_formats_panel,
-            self.add_text_panel,
-            self.add_prefix_panel,
-            self.add_date_panel,
-            self.add_researcher_panel,
-            self.add_advanced_panel,
+            self.add_import_panel,
+            self.add_limits_panel,
             self.add_color_panel,
-            self.add_symbols_panel
+            self.add_symbols_panel,
+            self.add_idformats_panel,
+            self.add_text_panel,
+            self.add_warnings_panel,
+            self.add_researcher_panel,
             )
         ConfigureDialog.__init__(self, uistate, dbstate, page_funcs,
                                  GrampsPreferences, config,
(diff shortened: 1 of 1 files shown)


DaveSch-gramps:
https://github.com/gramps-project/addons-source/pull/472

Brings Themes into agreement  in the _panel Tabs structure.

lordemannd:
I tried it on Windows and got following
80171: ERROR: grampsapp.py: line 157: Unhandled exception
  File "C:\Program Files\GrampsAIO64-5.1.3-V2-rework\gramps\gui\viewmanager.py", line 672, in preferences_activate
    GrampsPreferences(self.uistate, self.dbstate)
  File "C:\Users\dave\AppData\Roaming\gramps\gramps51\plugins\Themes\themes.py", line 111, in __init__
    self.add_behavior_panel,
AttributeError: 'GrampsPreferences' object has no attribute 'add_behavior_panel'

DaveSch-gramps:
You also need to modify your Themes addon. If installed, it actually set
the Tabs.

https://github.com/gramps-project/addons-source/pull/472

DaveSch-gramps:
NOTE: There is a pull to install icons to the toolbar that **IF** merged will have added settings to configure.py

https://github.com/gramps-project/gramps/pull/1139

**IF** that pull is merged, those additional settings will need to be added to this reworking of configure.py

        row += 1
        # Show Plugins Icon:
        self.add_checkbox(
            grid, _("Show Plugins icon on toolbar *"),
            row, 'interface.toolbar-plugin', start=0, stop=3,
            tooltip=_("Show or hide the Plugins icon on the toolbar.\n"
                      "Requires Gramps restart to apply."))

        row += 1
        # Show Preferences icon:
        self.add_checkbox(
            grid, _("Show Preferences icon on toolbar *"),
            row, 'interface.toolbar-preference', start=0, stop=3,
            tooltip=_("Show or hide the Preferences icon on the toolbar.\n"
                      "Requires Gramps restart to apply."))

I post this so more than myself can be aware of this.

Nick-Hall:
When I go into "Edit -> Preferences" I get the following error:
```
 ... gramps/gen/utils/configmanager.py", line 381, in get
    raise AttributeError("No such config setting name: '%s.%s'" %
AttributeError: No such config setting name: 'preferences.calendar-format-input'
```
Were these changes the ones discussed on the Discourse forum thread?

From what I can see in the images in the feature request, the bold headings need to be left justified to conform to our [UI Style](https://gramps-project.org/wiki/index.php/UI_style) guide.  In particular, the Gnome Human Interface Guidelines section.

DaveSch-gramps:
The only new config setting added was one for @SNoiraud #1161 where he adds register('preferences.calendar-format-input', 0) to gen\config.py now accessed by my gui\configue.py.

All other changes were all only a shuffling of the existing options. A few descriptions alterations are proposed.

In the existing Preferences, the major headings are all center justified. You want them Left justified?

Nick-Hall:
> You want them Left justified?

Yes.  Have a look at the Gnome editor preferences as an example:

gedit_preferences

The indents are multiples of 6 pixels.

DaveSch-gramps:
@Nick-Hall 

I have reformatted the tabs with left-justified headers and indented selectors. All tabs were been modified so that they all share the theme with the exception of Colors where I left the tab's header centered. This better reflects having the "Dark" and "Restore"  buttons centered directly below. I also made no changes to the Symbols tab.

In reviewing UI Style guide and GNOME Human Interface Guidelines they strongly recommend right-justified  labels on the Drop-down and Box options. I have made this change in the current version. Modifying BasicLabel also affected the Relationships and Combined views negatively. The only other file that takes advantage of BasicLabel is _sidebarfilter.py

I created a RightLabel class in gui\widgets\labels.py and used it in gui\configure.py to make labels right-justified.

The problem is an inconsistency in labels. Most notably the Report and Tool/Gramplets configure windows remain left-justified. As they do not use the BasicLabel code any reworking of the code to make them right-justified would be a lot of work and beyond my coding talents. 

Notably, the settings in all the filter rules are all right justified.

So the question: Keep the right-justified labels or revert to left-justified?

DaveSch-gramps:
data
general
addons
famtree
import
idformats
limits
researcher
text
warnings

Nick-Hall:
> So the question: Keep the right-justified labels or revert to left-justified?

I've reverted to left-justified labels, and made a few other minor adjustments.  The title in the colour tab is now left justified.

DaveSch-gramps:
Looks great. Thank you for merging.

@prculley 's Theme Gramplet _panel structure must now agree with the changes made here.  

Fix already filed

 https://github.com/gramps-project/addons-source/pull/472

PR #1150, opened 2020-11-14
Install MIME icons in global hicolor icon theme

pinotree:
At the moment, the icons of MIME types are installed in a location specific for GNOME, which is not usable when not using a GNOME icon theme. Also, the naming is a deprecated GNOME one, that might not be even working anymore.

Instead, rename the icons according to the MIME they represent, and install them in the global hicolor icon theme: this way, they are used with any icon theme that does not provide own versions of them (which is the vast majority of the themes).


pinotree:
Ooops, I just sent this, without noticing that there was #1053 already. OTOH, this also moves the icons to a better location in the sources, following the pattern of the other hicolor icons (apps and actions).

No big deal though, let me know :)

Nick-Hall:
Thanks for this PR.  I decided that #1053 should be classified as a bug and have merged it into the _gramps51_ maintenance branch.

pinotree:
@Nick-Hall thanks! I've opened #1153 for the changes I had that were not part of #1053, in case they can be useful.

I'll close this, then.

PR #1151, opened 2020-11-18
Move temp directory out of HOME_DIR

a-andre:
Using the system's temp directory allows keeping temporary files in RAM.

### DIFF
--- a/gramps/cli/argparser.py
+++ b/gramps/cli/argparser.py
@@ -48,7 +48,7 @@ from glob import glob
 #
 #-------------------------------------------------------------------------
 from gramps.gen.const import (LONGOPTS, SHORTOPTS, USER_PLUGINS, VERSION_DIR,
-                              HOME_DIR, TEMP_DIR, THUMB_DIR, ENV_DIR, USER_CSS)
+                              HOME_DIR, THUMB_DIR, ENV_DIR, USER_CSS)
 from gramps.gen.utils.cast import get_type_converter
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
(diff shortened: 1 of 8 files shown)


Nick-Hall:
The unit test [`test3_files_in_import_dir`](https://github.com/gramps-project/gramps/blob/master/gramps/cli/test/cli_test.py#L100) is failing.  Files are not being removed from the temp directory.

a-andre:
The temp directory created in [line 101](https://github.com/gramps-project/gramps/blob/0b04c8ab77a3de82f4a480708f12f97c32539690/gramps/cli/test/cli_test.py#L101) is different from the directory created (and cleaned up) by [line 115](https://github.com/gramps-project/gramps/blob/0b04c8ab77a3de82f4a480708f12f97c32539690/gramps/cli/test/cli_test.py#L115). So cleanup will not change the content of the former.

`tempfile.mkdtemp(prefix="gramps-")` creates directories like `/tmp/gramps-yfm0864k` where the suffix is random.
This allows multiple users to run different instances of gramps in parallel without conflicts because of a shared temp directory.
(The mode of the temp directory is 700, so only the user who created the directory can read it.)

Nick-Hall:
OK.  It looks like we need to modify that unit test.  However, the temp directories are not removed when exiting Gramps.

a-andre:
The temp directory is now removed when exiting Gramps.

Tests now use `tempfile.TemporaryDirectory()` instead of `TEMP_DIR` for auto clean-up.
As `file.py` was the only remaining user of `TEMP_DIR`, I replaced also it. So `TEMP_DIR` is gone.

The only addon affected by this change seems to be [HtmlView](https://github.com/gramps-project/addons-source/blob/1304b65a7d758bfe17339c26260473ac3e9c4061/HtmlView/htmlview.py#L60), where `TEMP_DIR` is imported, but not used.

For now, I removed the `test3_files_in_import_dir` unit test.

Nick-Hall:
@a-andre Thanks.  This looks good.

PR #1153, opened 2020-11-18
Move MIME icons together with other hicolor icons

pinotree:
Move the icons of MIME types in the same location/structure used for
the other hicolor icons (application icon, action icons).
This way, all the hicolor icons are logically in the same place.


pinotree:
ping?

Nick-Hall:
@pinotree Thanks for reminding me.  This looks good, but needs to be in the _master_ branch.

pinotree:
> This looks good, but needs to be in the _master_ branch.

This is a followpup of #1150, which was done only in `maintenance/gramps51`. Also, that branch has not been merged to `master` in months, which means that:
- #1150 is not present in `master`
- I cannot switch this PR alone to `master`, as the changes it refers to do not exist

Nick-Hall:
Rebased onto _master_ branch.

PR #1160, opened 2020-12-28
Gramps crashes on import of large KML file

SNoiraud:
Fixes [#11954](https://gramps-project.org/bugs/view.php?id=11954).

If we have more than 5 places in a kml file, we add all those places in the database without editing them.
You'll need to edit them manually to add the place type, enclosing, alternate  names...

As the kml example was a testing file, I think this bug is an enhancement and needs more testing.

### DIFF
--- a/gramps/plugins/lib/maps/geography.py
+++ b/gramps/plugins/lib/maps/geography.py
@@ -48,6 +48,7 @@ from gi.repository import OsmGpsMap as osmgpsmap
 # Gramps Modules
 #
 # -------------------------------------------------------------------------
+from gramps.gen.db import DbTxn
 from gramps.gen.lib import EventType, Place, PlaceRef, PlaceName
 from gramps.gen.display.name import displayer as _nd
 from gramps.gen.display.place import displayer as _pd
@@ -1049,7 +1050,8 @@ class GeoGraphyView(OsmGps, NavigationView):
             if val:
                 kmlfile = Kml(val)
                 points = kmlfile.add_points()
-                for place in points:
+                kml.destroy()
+                def init_place(place):
                     (name, coords) = place
                     latlong = coords.pop()
                     (lat, lon) = latlong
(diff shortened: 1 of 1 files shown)


SNoiraud:
For the kml file associated with this bug report, here is the message you could see:
large-kml-message

gramps/plugins/lib/maps/geography.py
```diff
@@ -1017,11 +1019,33 @@ def add_place_from_kml(self, menu, event, lat, lon):
...
-        kml.destroy()
+                    return new_place
+
+                if len(points) > 5:
+                    from gramps.gui.dialog import WarningDialog
+                    WarningDialog(
+                        _('You have more than 5 places in this kml file.'),
+                        _("You will need to edit each place manually "
+                          "to add place type, ...\n"
+                          "You have %d places to edit." % len(points)),
```
Nick-Hall, line 1030: You will need to code for plural forms here.  For example, the Serbian translation will be a problem if you don't.

SNoiraud: What is the problem ? I don't understand.

Nick-Hall: The word "places" may change depending on the number.  For example, in Croatian: "Imate 50 mjesta za uređivanje." and "Imate 51 mjesto za uređivanje.".  I think that this is the way that slavic languages work.

SNoiraud:
I asked this to Milo Ivir

milotype:
@SNoiraud @Nick-Hall 
Nick-Hall is wright. BTW, in my opinion plural forms for different languages should be supported in general **when a string has a placeholder for counting**.

Croatian and other languages (not just slavic ones) use 3 forms: one, few, other. Other languages use even more forms.

I'm not a programmer, so I can't help with coding. Here are two pages about plural forms I found:
- https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html
- http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html?id=l10n/pluralforms

I hope that helps.

PR #1161, opened 2020-12-31
Allow user to set default calendar in date editor

SNoiraud:
Fixes [#11809](https://gramps-project.org/bugs/view.php?id=11809)

This PR must be applied after PR #1149


SNoiraud:
This PR must be applied after PR #1149 because the preferences tab has been restructured
You must remove the configure.py patch which will be included in the PR #1149

See https://gramps.discourse.group/t/brainstorming-preferences-interface/893

PR #1163, opened 2021-01-03
Calendarreport

mbalk:


SNoiraud:
You should use the symbols functions:
get_symbol_for_string or get_symbol_fallback for all symbols except the death symbol.
You have an example in gramps/gui/widgets/fanchart.py

From 5.1, we can select our proper death symbol, so you must use:
  death_symbol_idx = config.get('utf8.death-symbol') or death_symbol_idx = self.uistate.death_symbol
  self.dth = self.symbols.get_death_symbol_for_char(death_symbol_idx)

mbalk:
> You should use the symbols functions [...]

Thank you for your review! I pushed some changes.

Nick-Hall:
Rebased.

This looks good to me.  I'll ask for people who use this report to test it.

Nick-Hall:
Testing should include:

* Check that the new "Include death dates" option works.
* Check that the birth, marriage and death symbols appear correctly.
* Check that no existing functionality is adversely affected.

cdhorn:
@mbalk if "Include only living people" is checked and "Include death dates" is checked then I would expect "Include only living people" to still be honored but all the deceased people are showing up on the calendar anyway. Other than that this looks good to me.

PR #1164, opened 2021-01-03
Pdf preview

PushKK:
For https://gramps-project.org/bugs/view.php?id=12126

I have 4 problems with this PDF view:
1. I can't resize big PDF,
2. I can't show preview in top window for select part of PDF image,
3. After add/delete PDF file I can't renew view in bottom main screen (in Galery gramplet),
4. After close GRAMPS I receive error message in terminal.

Pls, check it...
All needed info in code inside.


Nick-Hall:
Perhaps using Poppler and Cairo would be a better alternative?

```python
import cairo
from gi.repository import Poppler

SCALE = 0.1
doc = Poppler.Document.new_from_file('file:///absolute/path/to/file.pdf')
page = doc.get_page(0)
width, height = page.get_size()
surface = cairo.ImageSurface(cairo.FORMAT_ARGB32, int(width*SCALE), int(height*SCALE))
context = cairo.Context(surface)
context.scale(SCALE, SCALE)
page.render(context)
surface.write_to_png('thumb.png')
```

@jralls Would Poppler be acceptable as an extra dependency?  I thought that cairo used it, but apparently not.

Nick-Hall:
> After add/delete PDF file I can't renew view in bottom main screen (in Galery gramplet)

This is a known issue for all thumbnails.

PushKK:
> Perhaps using Poppler and Cairo would be a better alternative?
Yes, of course. You are developer, you must use better software for GRAMPS.
I had problem with "convert" from ImageMagick (memory is out) and use PyMuPDF as fast and working for me software.
After insert PyMuPDF in GRAMP I saw - it's work too.

> After add/delete PDF file I can't renew view in bottom main screen (in Galery gramplet)
> This is a known issue for all thumbnails.
And I can't show first page image in "Preview" in "Select media" dialog:
PDF_select_media

P.S. Maybe - change "get_thumbnail_image" or create new function can resolve in all OS (Win, Linux without Gnome, MacOS and other) problems with thumbnails from image, PDF and ODF files  (and other as DjVU):

def all_image_from_file(filename, mtype, ...)
# If image
if mtype.startswith('image/'):
    # code not needed
# If PDF
elif mtype.endswith('/pdf'):
    code for create first page image from PDF file
# If ODF
elif mtype = ...
    code for create first page image from ODF file
# not have thumbnail 
else:
        filename = os.path.join(IMAGE_DIR, "document.png")

return GdkPixbuf.Pixbuf.new_from_file(filename)

jralls:
@Nick-Hall Looks like it doesn't build cleanly. I tried today's git master and the last two releases and all three errored out with pretty serious C++ coding errors.  If you think it's worth the effort then I'll dig into it. Both Homebrew and MacPorts supposedly have popper 20.12.1 so I might be able to steal patches from them.

Cairo has a PDF canvas that write a PDF but is very limited in config options and post-creation manipulations. I'm not surprised that you'd need something else to generate thumbnails.

Nick-Hall:
@PushKK I have added support for thumbnailer plugins in PR #1363.  This allows you to write a plugin to create pdf thumbnails.  You just need to make sure the `is_supported()` method always returns `False` if the _pymupdf_ package is not installed.

Nick-Hall:
I'm going to close this now.

@PushKK Let me know if you need any help writing a PDF thumbnailer plugin.

PR #1165, opened 2021-01-03
An experimental fix for bug12048

prculley:


jralls:
No joy:
```
2021-01-03 10:32:44.392: WARNING: dummydb.py: line 1420: database is closed
Problem: multiple Toolbar updates before idle
Segmentation fault: 11
```

jralls:
To explore this a bit I did
```diff
diff --git a/gramps/gui/uimanager.py b/gramps/gui/uimanager.py
index 3fe1896f7..0ed4efede 100644
--- a/gramps/gui/uimanager.py
+++ b/gramps/gui/uimanager.py
@@ -134,8 +134,8 @@ class UIManager():
         self.et_xml = ET.fromstring(initial_xml)
         self.builder = None
         self.toolbar = None
-        self.old_toolbar = None  # holds previous toolbar until Gtk is idle
-        self.old_toolbar_items = None
+        self.old_toolbar = []  # holds previous toolbar until Gtk is idle
+        self.old_toolbar_items = []
         self.action_groups = []  # current list of action groups
         self.show_groups = []  # groups to show at the moment
         self.accel_dict = {}  # used to store accel overrides from file
@@ -244,11 +244,9 @@ class UIManager():
         # the following updates the toolbar from the new builder
         toolbar_parent = toolbar.get_parent()
         tb_show = toolbar.get_visible()
-        if not self.old_toolbar:
-            self.old_toolbar = toolbar
-            self.old_toolbar_items = toolbar.get_children()
-        else:
-            print("Problem: multiple Toolbar updates before idle")
+        self.old_toolbar.append(toolbar)
+        self.old_toolbar_items.append(toolbar.get_children())
+        print("Saved %d toolbars" % len(self.old_toolbar))
         toolbar_parent.remove(toolbar)
         toolbar = self.builder.get_object("ToolBar")  # new toolbar
         if config.get('interface.toolbar-text'):
@@ -266,8 +264,9 @@ class UIManager():
         idle.  To avoid an issue with the toolbar being removed before all
         references were removed.
         """
-        self.old_toolbar = None
-        self.old_toolbar_items = None
+        print("Removing %d retained toolbars" % len(self.old_toolbar))
+        self.old_toolbar.clear()
+        self.old_toolbar_items.clear()
         return False

     def add_ui_from_string(self, changexml):
```
The output was
```
2021-01-03 10:45:38.482: WARNING: dummydb.py: line 1420: database is closed
Saved 1 toolbars
Removing 1 retained toolbars
Saved 1 toolbars
Saved 2 toolbars
Removing 2 retained toolbars
Removing 0 retained toolbars
```
Family Trees window displayed
```
Saved 1 toolbars
Removing 1 retained toolbars
Saved 1 toolbars
Removing 1 retained toolbars
Saved 1 toolbars
Removing 1 retained toolbars
```
Tree loaded, at Dashboard page
```
Saved 1 toolbars
Removing 1 retained toolbars
```
On People View. Shrank the window, clicked the reveal arrow, clicked the People item
```
Saved 1 toolbars
Removing 1 retained toolbars
Segmentation fault: 11
```

The retain/release is working:
```
Removing 1 retained toolbars
Process 62781 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 2.1
    frame #0: 0x00000001174bb54c libgtk-3.0.dylib`gtk_tool_button_finalize(object=0x00007fd059525bc0) at gtktoolbutton.c:802:27
   799 	static void
   800 	gtk_tool_button_finalize (GObject *object)
   801 	{
-> 802 	  GtkToolButton *button = GTK_TOOL_BUTTON (object);
   803
   804 	  g_free (button->priv->stock_id);
   805 	  g_free (button->priv->icon_name);
Target 0: (python3) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 2.1
  * frame #0: 0x00000001174bb54c libgtk-3.0.dylib`gtk_tool_button_finalize(object=0x00007fd059525bc0) at gtktoolbutton.c:802:27
    frame #1: 0x000000010b7286a5 libgobject-2.0.0.dylib`g_object_unref(_object=0x00007fd059525bc0) at gobject.c:3499:11 [opt]
    frame #2: 0x000000010b506cb7 _gi.cpython-38-darwin.so`pygobject_dealloc at pygobject-object.c:1244:13 [opt]
    frame #3: 0x000000010b506c88 _gi.cpython-38-darwin.so`pygobject_dealloc(self=0x000000011d078e40) at pygobject-object.c:1135 [opt]
```
But it's going too soon because your `GtkNotebookPage` `switch-page` signal handler runs the event-loop
```
     frame #27: 0x000000010b59e776 libglib-2.0.0.dylib`g_main_context_iteration(context=0x00007fd028c23510, may_block=1) at gmain.c:4108:12 [opt]
    frame #28: 0x000000011732668b libgtk-3.0.dylib`gtk_main_iteration at gtkmain.c:1432:3
    frame #29: 0x000000010b784d95 libffi.7.dylib`ffi_call_unix64 + 85
    frame #30: 0x000000010b784223 libffi.7.dylib`ffi_call_int + 771
    frame #31: 0x000000010b51f1b3 _gi.cpython-38-darwin.so`pygi_invoke_c_callable(function_cache=<unavailable>, state=<unavailable>, py_args=0x00007ffee70eeab8, py_kwargs=<unavailable>) at pygi-invoke.c:690:9 [opt]
    frame #32: 0x000000010b5200d7 _gi.cpython-38-darwin.so`pygi_function_cache_invoke(function_cache=<unavailable>, py_args=<unavailable>, py_kwargs=<unavailable>) at pygi-cache.c:863:12 [opt]
    frame #33: 0x0000000108b34e86 libpython3.8.dylib`_PyObject_MakeTpCall(callable=<unavailable>, args=<unavailable>, nargs=<unavailable>, keywords=0x0000000000000000) at call.c:159:18 [opt]
    frame #34: 0x0000000108c0bd7c libpython3.8.dylib`call_function(tstate=<unavailable>, pp_stack=0x00007ffee70eec10, oparg=<unavailable>, kwnames=<unavailable>) at abstract.h:0:12 [opt]
    frame #35: 0x0000000108c08759 libpython3.8.dylib`_PyEval_EvalFrameDefault(f=<unavailable>, throwflag=<unavailable>) at ceval.c:3469:23 [opt]
    frame #36: 0x0000000108b3583d libpython3.8.dylib`function_code_fastcall(co=<unavailable>, args=<unavailable>, nargs=<unavailable>, globals=<unavailable>) at call.c:283:14 [opt]
    frame #37: 0x0000000108c0bc52 libpython3.8.dylib`call_function [inlined] _PyObject_Vectorcall(callable=<unavailable>, args=0x000000011d7ce5d8, nargsf=9223372036854775810, kwnames=<unavailable>) at abstract.h:127:11 [opt]
    frame #38: 0x0000000108c0bc46 libpython3.8.dylib`call_function(tstate=<unavailable>, pp_stack=0x00007ffee70eee20, oparg=<unavailable>, kwnames=<unavailable>) at ceval.c:4963 [opt]
    frame #39: 0x0000000108c0873c libpython3.8.dylib`_PyEval_EvalFrameDefault(f=<unavailable>, throwflag=<unavailable>) at ceval.c:3486:23 [opt]
    frame #40: 0x0000000108b3583d libpython3.8.dylib`function_code_fastcall(co=<unavailable>, args=<unavailable>, nargs=<unavailable>, globals=<unavailable>) at call.c:283:14 [opt]
    frame #41: 0x0000000108b381e1 libpython3.8.dylib`method_vectorcall [inlined] _PyObject_Vectorcall(callable=<unavailable>, args=0x00007ffee70eef90, nargsf=4, kwnames=0x0000000000000000) at abstract.h:127:11 [opt]
    frame #42: 0x0000000108b381b5 libpython3.8.dylib`method_vectorcall(method=<unavailable>, args=0x000000011edcaaf8, nargsf=<unavailable>, kwnames=0x0000000000000000) at classobject.c:89 [opt]
    frame #43: 0x0000000108b351d4 libpython3.8.dylib`PyVectorcall_Call(callable=0x000000011c530940, tuple=<unavailable>, kwargs=<unavailable>) at call.c:199:24 [opt]
    frame #44: 0x000000010b51e6df _gi.cpython-38-darwin.so`pygi_signal_closure_marshal(closure=0x00007fd018dc9810, return_value=0x0000000000000000, n_param_values=3, param_values=0x00007ffee70ef320, invocation_hint=<unavailable>, marshal_data=<unavailable>) at pygi-signal-closure.c:197:11 [opt]
    frame #45: 0x000000010b72303d libgobject-2.0.0.dylib`g_closure_invoke(closure=0x00007fd018dc9810, return_value=<unavailable>, n_param_values=3, param_values=0x00007ffee70ef320, invocation_hint=0x00007ffee70ef2a8) at gclosure.c:810:7 [opt]
    frame #46: 0x000000010b739540 libgobject-2.0.0.dylib`signal_emit_unlocked_R(node=<unavailable>, detail=0, instance=<unavailable>, emission_return=0x0000000000000000, instance_and_params=0x00007ffee70ef320) at gsignal.c:3742:8 [opt]
    frame #47: 0x000000010b73a342 libgobject-2.0.0.dylib`g_signal_emit_valist(instance=<unavailable>, signal_id=<unavailable>, detail=0, var_args=0x00007ffee70ef560) at gsignal.c:3498:5 [opt]
    frame #48: 0x000000010b73aa82 libgobject-2.0.0.dylib`g_signal_emit(instance=<unavailable>, signal_id=<unavailable>, detail=<unavailable>) at gsignal.c:3554:3 [opt]
    frame #49: 0x0000000117365206 libgtk-3.0.dylib`gtk_notebook_switch_page(notebook=0x00007fd0591406a0, page=0x00007fd0490fd8f0) at gtknotebook.c:6264:3```

```
frame #49: 0x0000000117365206 libgtk-3.0.dylib`gtk_notebook_switch_page(notebook=0x00007fd0591406a0, page=0x00007fd0490fd8f0) at gtknotebook.c:6264:3
   6261
   6262	  page_num = g_list_index (priv->children, page);
   6263
-> 6264	  g_signal_emit (notebook,
   6265	                 notebook_signals[SWITCH_PAGE],
   6266	                 0,
   6267	                 page->child,
```
The old `GtkToggleToolButton` function is way downstream:
```
    frame #87: 0x00000001174b1016 libgtk-3.0.dylib`menu_item_activated(menu_item=0x00007fd05926d180, toggle_tool_button=0x00007fd059401520) at gtktoggletoolbutton.c:299:7
```

**Comment by jralls** [comment]
@prculley Have you punted on this?

**Comment by prculley** [comment]
I admit I have been distracted lately.  Looks like you did some more analysis since I last payed attention.  If some other code is draining Gtk events before the whole screen is updated, I'm back to trying to figure out a way to tell when it is ok to release the reference to the button/menu.  At a bit of a loss here.  I suppose we could switch to a timer, although redrawing views can sometimes take a very long time in large dbs.  The only downside of this would be a bit of memory being held longer than necessary.  I'm open to suggestions.

**Comment by prculley** [comment]
I am trying out the idea of using a timeout, rather than idle_add.  It appears to be working better on my system, in that I no longer get a warning message.  I updated my code with your tweaks, and switched to timeout, as well as finally removing last reference to toolbars one at a time, when the timeout for that toolbar occurs.

Will want to kill off the print statements before final commit, but leaving them for now in case we are still getting seg faults.

**Comment by prculley** [comment]
I don't see any sign of the delay you mention; perhaps it's a difference in Gtk versions.  In 3.18 everything is fine, in 3.24.14, I get an odd bug where I get no arrow until I move the cursor over the icon where the arrow should be.  Then the icon disappears and the arrow appears.  I'm inclined to attribute that one to Gtk.  Both of these test done on Win10.

**Comment by jralls** [comment]
I tested again and you're right, it's the move-in event not the timer; in fact the timer's diagnostic print fires in rather less than 10 seconds. I guess that gtk needs an event to trigger the redraw, which is a little odd. I also notice that clicking on the menu even after `self.delete_old_toolbar` runs doesn't crash even if one reduces the timeout to 100ms--but 10ms does. Something is racing, I haven't figured out what yet.

**Comment by jralls** [comment]
Forget the timer, the correct fix is to run `ViewManager.__change_page` in an idle:
```
diff --git a/gramps/gui/viewmanager.py b/gramps/gui/viewmanager.py
index e2f4d7f67..2fbe9ef4c 100644
--- a/gramps/gui/viewmanager.py
+++ b/gramps/gui/viewmanager.py
@@ -846,6 +846,11 @@ class ViewManager(CLIManager):
         """
         Called when the notebook page is changed.
         """
+        def page_changer(self, page_num):
+            self.__change_page(page_num)
+            self.view_changing = False
+            return False
+
         if self.view_changing:
             return
         self.view_changing = True
@@ -868,8 +873,7 @@ class ViewManager(CLIManager):
         config.save()

         self.navigator.view_changed(cat_num, view_num)
-        self.__change_page(page_num)
-        self.view_changing = False
+        GLib.idle_add(page_changer, self, page_num)

     def __change_page(self, page_num):
         """
```
It's the function that calls `UIManager.update_menu` and then correctly iterates the main loop. Doing that in an event instead of directly in the signal handler chain ensures that the signal handlers can all complete and clean up after themselves before the redraw.

prculley:
It's clear you have a better understanding of the interior of Gtk activities...  I'm wondering if this fix might cause some delays in getting the GUI updated.  Some Gramplets also use idle_add to handle slower or long running activities, like web lookups or tree searches.  I'm concerned that if we use default priority, the user could see a delay on view change, while other idle events from previous view or active object changes get handled.  I have not tested this yet.  Maybe we just need to use a slightly higher priority?  Or am I concerned about nothing?

jralls:
It doesn't seem likely to me that an event that causes a `__change_page` would also launch a search or a network retrieval, but there's no harm in making the priority `G_PRIORITY_DEFAULT`  or [even lower](https://gitlab.gnome.org/GNOME/glib/-/blob/master/glib/gmain.h#L280) if you think that it would matter... but it's a poor design if it does and that should be looked at. Event handlers should be as self-contained as possible.

This isn't particular to Gtk: All GUIs work this way. The [GLib docs](https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html) have an excellent explanation of how the main event loop works (with one significant caveat: On macOS the poll function actually runs *all* of the events queued on the OS's run loop plus one idle event if there is one). Just keep in mind that once an event handler starts it owns the main thread unless it--and that includes any callbacks--iterates the main context. If it does that then the *next* event (and all system events including redraws on macOS) own the main thread until they return and then execution of that first handler resumes. Signal handlers are a form of callback, run when the signal they're connected to is emitted. They're part of the event handler's execution. They shouldn't delete objects that they didn't create and they should be very careful about iterating the main context: By extension, anything like `__change_page` that deletes GUI elements that it didn't create (e.g. a *modal* dialog box can be created and deleted in a signal handler) should run in its own event, not in a signal handler. With that in mind you should consider removing the main context iteration from `__change_page`. It's not necessary, it will happen on its own as soon as `__change_page` finishes, and it creates the possibility of some other event that's oblivious to the fact that it's running in the middle of a `__change_page` doing something to the wrong window.

prculley:
I tried an alternate fix based on your idea; yours did not work for me because some necessary work was not done immediately.
https://github.com/gramps-project/gramps/pull/1274

I think this fixes the original problem, and don't see any side effects yet...

Nick-Hall:
@prculley Can this PR be closed now?

prculley:
Fixed by #1274

PR #1172, opened 2021-01-13
Display a tooltip warning if a window saves data immediately

SNoiraud:
Fixes [#12117](https://gramps-project.org/bugs/view.php?id=12117)

We have the same in the following windows:
- clipboard
- organize tags
- {Familiy, Event, Note, Person, Place, ...} filter editor


Nick-Hall:
This is not a good way to use tooltips.  In fact the Gnome Human Interface Guidelines advise against it:

"Otherwise, try to keep tooltip usage to a minimum, since they can get in the way when inadvertently displayed, and aren’t always accessible (such as when using a touch device). Don’t rely on tooltips to communicate essential information."

The main purpose of the Esc key is to close a container.  It can be linked to a cancel action if available, but otherwise should just close a dialog.

Looking at other Gnome applications, it seems quite common for changes to preferences to take effect immediately.

Nick-Hall:
Rebased.

Nick-Hall:
Changed the wording slightly and added an option to suppress the warning.

PR #1177, opened 2021-02-03
Added buttons for adding a new person and removing the selected person to the editpersonref dialog

codernarchy:
Just added two buttons to the editpersonref dialog:

grafik
grafik

### DIFF
--- a/gramps/gui/editors/editpersonref.py
+++ b/gramps/gui/editors/editpersonref.py
@@ -46,9 +46,11 @@ from gi.repository import Gtk
 #-------------------------------------------------------------------------
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.sgettext
+from gramps.gen.config import config
 from gramps.gen.display.name import displayer as name_displayer
 from .editsecondary import EditSecondary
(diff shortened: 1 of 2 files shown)


prculley:
You have a trailing whitespace error, found by the Travis test.

chrmue94:
Thanks for the hint, I just removed it

Nick-Hall (changes requested):
A better way to implement this would be to add a `PersonEntry` class to the [objectentries](https://github.com/gramps-project/gramps/blob/master/gramps/gui/editors/objectentries.py) module.  This would ensure a consistent interface.

Have a look at the [event editor](https://github.com/gramps-project/gramps/blob/458772d1e6eb2d04317487777a734a5f8a28c6a1/gramps/gui/editors/editevent.py#L139) to see how the `PlaceEntry` class is used.

Nick-Hall:
Added an alternative in PR #1371.  It looks like we may have used this approach before.  I need to investigate why we may have decided on a different approach.  Until then I'll leave this PR open.

Nick-Hall:
Closed in favour of PR #1371.

PR #1178, opened 2021-02-05
[translation] Hebrew update for v 5.2

avma:
haven't done this for some time now... hope I'm hitting the proper branch ; \


Nick-Hall:
@SNoiraud Thanks for spotting this.  It can be easily fixed by running `sed -i 's/".*lexeme|/"/' po/*.po`.

I can't remember if I sent a message to all the translators about the context changes in the _master_ branch.  Perhaps I just posted to the mailing list.

I've been meaning to email everyone to ask their opinions on us using an online translation service.  I can also update them on the changes in v5.2 at the same time.

PR #1180, opened 2021-02-13
citations gramplet: add date, page, confidence

SNoiraud:
Fixes #09224
discourse: https://gramps.discourse.group/t/enhanced-citations-gramplet/983

This add date, page and confidence to the citation gramplet.
In addition, I modified all citation tabs for people, family, place, ...
I added the date, the publisher and confidence
For that, I also modify:
gramps/gui/editors/displaytabs/citationembedlist.py
gramps/gui/editors/displaytabs/citationrefmodel.py

### DIFF
--- a/gramps/gui/editors/displaytabs/citationembedlist.py
+++ b/gramps/gui/editors/displaytabs/citationembedlist.py
@@ -77,13 +77,13 @@ class CitationEmbedList(EmbeddedList, DbGUIElement):
     #index = column in model. Value =
     #  (name, sortcol in model, width, markup/text, weigth_col
     _column_names = [
-        (_('Title'), 0, 150, TEXT_COL, -1, None),
-        (_('Author'), 1, 150, TEXT_COL, -1, None),
-        (_('Date'), 2, 150, TEXT_COL, -1, None),
-        (_('Publisher'), 3, 150, TEXT_COL, -1, None),
-        (_('Confidence Level'), 4, 100, TEXT_COL, -1, None),
-        (_('Page'), 5, 80, TEXT_COL, -1, None),
-        (_('ID'), 6, 75, TEXT_COL, -1, None),
+        (_('Title'), 0, 350, TEXT_COL, -1, None),
+        (_('Author'), 1, 200, TEXT_COL, -1, None),
+        (_('Date'), 2, 180, TEXT_COL, -1, None),
+        (_('Publisher'), 3, 200, TEXT_COL, -1, None),
+        (_('Confidence Level'), 4, 120, TEXT_COL, -1, None),
+        (_('Page'), 5, 100, TEXT_COL, -1, None),
+        (_('ID'), 6, 80, TEXT_COL, -1, None),
         (_('Private'), 7, 30, ICON_COL, -1, 'gramps-lock')
     ]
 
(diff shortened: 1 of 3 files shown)


emyoulation:
I _**like**_ it!  However, here is a variant in the sequence and widths of columns.  I think it works for both book & periodical citations.


image

citationsembedlist.py  
Line 79-87
```
   _column_names = [
        (_('Title'), 0, 300, TEXT_COL, -1, None),
        (_('Author'), 1, 200, TEXT_COL, -1, None),
        (_('Page'), 2, 75, TEXT_COL, -1, None),
        (_('Date'), 3, 100, TEXT_COL, -1, None),
        (_('Publisher'), 4, 200, TEXT_COL, -1, None),
        (_('ID'), 5, 100, TEXT_COL, -1, None),
        (_('Confidence Level'), 6, 100, TEXT_COL, -1, None),
        (_('Private'), 7, 30, ICON_COL, -1, 'gramps-lock')		
    ]


```
citationsembedlist.py  
line 127
`        return ((1, 0), (1, 1), (1, 2), (1, 3), (1, 4), (1, 5), (1, 6), (1, 7))`


citationsrefmodel.py 
line 45-53

```
        for handle in citation_list:
            citation = self.db.get_citation_from_handle(handle)
            src = dbgsfh(citation.get_reference_handle())
            confidence = citation.get_confidence_level()
            self.append(row=[src.title, src.author, citation.page, 
                             get_date(citation), src.get_publication_info(),
                             citation.gramps_id, _(conf_strings[confidence]), 
                             citation.get_privacy(),
                             handle, ])
```

emyoulation:
discovered that the Sort on the new Date column in the Citation tab seems to be sorting on the formatted Text of the Date rather than the date value. (I am using ‘dd mmm yyyy’ date format.)

This seems to be similar to bug [0010893](https://gramps-project.org/bugs/view.php?id=10893): [Citations List View]"Source: Last Changed" only alpha sorts - Gramps - Bugtracker – Free Genealogy Software

SNoiraud:
> citationsembedlist.py
> Line 79-87

The sizes I put for the columns should respect our limit 800x600 window size.
I know this is difficult to understand but that's a constraint.
As we have a scrollbar, I accept that.

> citationsembedlist.py
> line 127

For the variant, I agree to put the lock at the end. The lock was displayed first, so I had not changed the order.
I changed the page to be displayed just after the author. Don't need to change the append directive.

SNoiraud:
> discovered that the Sort on the new Date column in the Citation tab seems to be sorting on the formatted Text of the Date rather than the date value. 

Do you speak about when you click on the date column ?
If yes, this is not a sort. The list is inverted but not sorted.

emyoulation:
Yes!  The new sort works very nicely for the improved facility in the Citations Tab. Thank you!

I will submit a series of Periodical Source Citations (I think adding 3 Public Notices to creditors for an added Probate event on the Home Person's wife) for inclusion in example.gramps and to be used for Citation Tab illustrations in the wiki.

emyoulation:
This PR also addresses:
	[0010608](https://gramps-project.org/bugs/view.php?id=10608): could Source Citations inherit Citations view preferences?

PR #1181, opened 2021-02-19
Geography cleanup

zradoslav:
Cleaning up popup menu builder for Geography plugins and class members as well - they're not used at class level, thus can be substituted with corresponding local variables

### DIFF
--- a/gramps/plugins/lib/maps/geography.py
+++ b/gramps/plugins/lib/maps/geography.py
@@ -437,7 +437,7 @@ class GeoGraphyView(OsmGps, NavigationView):
         # Add specific module menu
         self.add_specific_menu(menu, event, lat, lon)
         # Add a separator line
-        add_item = Gtk.MenuItem()
+        add_item = Gtk.SeparatorMenuItem()
         add_item.show()
         menu.append(add_item)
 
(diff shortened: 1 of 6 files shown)


Nick-Hall:
Rebased.

gramps/plugins/lib/maps/geography.py
```diff
@@ -397,8 +394,7 @@ def build_nav_menu(self, osm, event, lat, lon):
         add_item = Gtk.MenuItem(label=title)
         menu.append(add_item)
 
-        self.changemap = Gtk.Menu()
```
Nick-Hall, line 409: We had an issue in the past, where a menus failed to display if they were stored in local variables.  I'm not sure if this has been fixed on all supported Gtk versions yet.

Nick-Hall:
@SNoiraud What is your opinion of this PR?

SNoiraud:
@Nick-Hall If I remember, I tested this PR and it worked for me. The test was on my ubuntu 18.04.

gramps/plugins/lib/maps/geography.py
SNoiraud: When I tested this PR, I didn't see problems. I suppose this has been solved.

PR #1184, opened 2021-02-27
Gender_stats gender swapped

Darigar:
After opening the DB file with "DB Brower for SQLite@, I discovered that the "gender_stats" table counts males in the female column (and females as males).
I use the Russian last version.


wroldwiedbwe:
Known issue 

11314: Gender statistics swapped after database conversion to SQLite

read comment  https://gramps-project.org/bugs/view.php?id=11314#c57526

> @prculley Yes. All users would have to rebuild the gender stats. Since this only affects people reading the database directly, I agree with your suggestion of waiting until the next major release. 

So please change this to the master branch for Gramps 5.2 👍

gramps/gen/utils/grampslocale.py
```diff
@@ -525,6 +525,8 @@ def __init_first_instance(self):
         # with locale instead of gettext. Win32 doesn't support bindtextdomain.
         if self.localedir:
             if not sys.platform == 'win32':
+                # bug12278, _build_popup_ui() under linux and macOS
+                locale.textdomain(self.localedomain)
```
jralls, line 529: @romjerome Don't do this, it will make a conflict when @prculley next merges-up gramps51. That merge-up will take care of bug 12278 in master.

prculley:
It looks like this got based on something that didn't match master branch.  I'd suggest either cherry-picking the real commits to a new branch based off of master, or rebasing (I suspect cherry-picking will be easier in this case).  And then redoing the PR...

Nick-Hall:
I'm going to close this now.  There are no commits that seem to be related to gender statistics.

@Darigar This issue has been reported as bug [#11314](https://gramps-project.org/bugs/view.php?id=11314).  Please create another PR with your fix.

PR #1190, opened 2021-03-16
Allow entering place latlong without space

benediktwerner:
When pasting latitude/longitude combinations from Google Maps URLs they don't have a space after the comma and use a dot as decimal separator.

To avoid issues we first still check for `, ` and split there if possible but otherwise try to just split at a comma. If the split doesn't produce exactly two values the unpacking will raise an exception and abort as before.


SNoiraud:
You must also do that in  gramps/gui/editors/editplaceref.py

benediktwerner:
Done

PR #1192, opened 2021-03-21
Fix References Gramplet for inadequate updates when other objects change

prculley:
Fixes #12248

The original bug reports an exception from the backlinks gramplet when editing a backlink following a merge.  The root cause was the Gramplet not updating the references list when an object was deleted; thus trying to edit the object generated the HandleError exception.

On further investigation, I saw that the References were mostly not getting updated when new objects were added with a reference, or most other objects were edited with a new reference.

This should fix these issues.


lordemannd:
works for me on quick test of exact issue in 12248

PR #1197, opened 2021-04-05
Limit print statement by using logging module

romjerome:
```python3 Gramps.py -d "gui.uimanager"```


romjerome:
Hello @prculley
I suppose it could be easier for anyone to debug uimanager (if need), 
with less print() statements and comment markers on code?

Just a proposal after tests on a gramps session with the debug flag.
`$ python3 Gramps.py -d "gui.uimanager"`

Nick-Hall:
Rebased onto _master_ branch.

@prculley Do we want to merge this?  It looks reasonable to me.

prculley:
I usually just uncomment a statement as needed; reduces the amount of output to analyze.  I guess I can do the same with this, just commenting the log statements.  If you think this is better, accept it.

PR #1198, opened 2021-04-12
Need to set locale.textdomain under linux

romjerome:
_build_popup_ui() ignores translated strings without locale.textdomain set.

### DIFF
--- a/gramps/gen/utils/grampslocale.py
+++ b/gramps/gen/utils/grampslocale.py
@@ -525,7 +525,7 @@ class GrampsLocale:
         # with locale instead of gettext. Win32 doesn't support bindtextdomain.
         if self.localedir:
             if not sys.platform == 'win32':
-                # bug12278, _build_popup_ui() under linux
+                # bug12278, _build_popup_ui() under linux and macOS
                 locale.textdomain(self.localedomain)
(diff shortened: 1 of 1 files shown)


romjerome:
@jralls 
I do not know if we need to set locale.textdomain
under MacOS too, for avoiding a specific issue
around context menu on Editors (Person, Family).

See also [bug report 12278](https://gramps-project.org/bugs/view.php?id=12278).

jralls:
@romjerome This *does* set the text domain for macOS, the conditional is `if not sys.platform == 'win32'` and macOS isn't win32.

romjerome:
@jralls 
Sorry, as the proposal makes the change for macOS too, I just wonder if this issue was specific to linux or both OS should have the fix. I do not have a macOS available now for testing...
macOS could have a specific handling for GLib's g_dgettext wrapper (or so) and only linux os should have the fix.
I just want to avoid a bug for macos users in relation with this section.

jralls (approved):
@romjerome, there's very little macOS customization in glib. I tested on macOS: It does have the problem and this fixes it.

romjerome:
Thanks!
I will commit the change and close [BR12278](https://github.com/gramps-project/gramps/tree/BR12278)

PR #1202, opened 2021-05-05
macOS: Use File Logging instead of stderr Logging when there's no terminal

jralls:
When an app like Gramps is launched using LaunchServices on macOS there is no stderr to log to so any log messages are lost if Gramps quits unexpectedly in a way that the GUI Error handler doesn't catch. In those cases where it's Python quitting rather than a low-level crash there isn't a crash report either, leaving no clue about what went wrong.

This change sets up a FileHandler default log handler instead of the StreamHandler one when running on macOS when there's no attached tty and associated stderr.


jralls:
There being no objections, rebased and merged.

PR #1208, opened 2021-05-20
Update Simplified Chinese translation

tiansworld:
Another update. Please review.

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -15,7 +15,7 @@ msgstr ""
 "Project-Id-Version: Gramps 4.2.0-dev\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2020-07-15 22:20+0200\n"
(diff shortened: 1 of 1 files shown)


tiansworld:
> @tiansworld Great work! The Simplified Chinese translation is now 84.7% complete.
> 

Thanks, there are still so many strings needs to be improved or translated, I will try my best.

> We have started to use [Weblate](https://hosted.weblate.org/projects/gramps-project/gramps/) for our translations for v5.2 onwards. However, I suggest that you continue updating this branch for now so that your changes will be included in the next maintenance release. I'll merge your changes into the _master_ branch later.

OK.
BTW, will you accept translation update from Git as an alternative way after 5.2? Or weblate only?

Nick-Hall:
> BTW, will you accept translation update from Git as an alternative way after 5.2? Or weblate only?

Probably Weblate only to avoid merge conflicts.

PR #1210, opened 2021-05-20
Tips.xml: Old string changed

PushKK:
1. Old menu changed:
Database Processing > Family Tree Processing
Compare and Merge... > Merge...
2. Add string "The function ".

PR #1213, opened 2021-05-20
Various fixes to user-visible strings

tosky:
The change includes various fixes:
- removal of trailing whitespaces;
- consistent usage of the final dot (instead of a question mark);
- fix a few acronyms (uppercase when needed, correct spelling).


Nick-Hall:
This and #1210 will be an interesting test of a POT file change in Weblate.

tosky:
Does anything prevent from this change to be merged?

PR #1214, opened 2021-05-21
Remember last calendar type for data-entry (copy)

PushKK:
Addition to
"Remember last calendar type for data-entry #1161"
For https://gramps-project.org/bugs/view.php?id=11809


SNoiraud:
Duplicate of #1161 ?

Nick-Hall:
This PR sets a calendar type in the settings.   Do we want to use the last calendar type or default calendar type in the date editor?

SNoiraud:
>  Do we want to use the last calendar type or default calendar type in the date editor?

Yes. The default calendar is always the Gregorian calendar. When you enter French republican dates, Jewish dates, Islamic dates ...
it is necessary to change at each time the "Gregorian" choice to the desired calendar. This is tedious when you are always working with a different type of calendar than the Gregorian calendar.

Nick-Hall:
This PR lets you set the calendar in the preferences.  Do we need this, or can we just use the last calendar as in your previous PR?

Nick-Hall:
I've just been looking at #1161.  It depends on another PR that also adds a configuration setting.

I'm closing this because it does seem to be a duplicate.

PR #1212, opened 2021-05-20
Viewmanager.py: add check disk space

PushKK:
For https://gramps-project.org/bugs/view.php?id=12306
I add check disk space before Quit and BackUp.

Test this code please.

PR #1218, opened 2021-05-23
sqlite3.OperationalError: database or disk is full #2

PushKK:
For https://gramps-project.org/bugs/view.php?id=12306

1. Add check disk space before close db, save config file, autobackup and XML
backup.
2. Add new function get_avail_disk_size().

I have problem: I can't restore connection to window in quit() if user abort to
quit.

### DIFF
--- a/gramps/gen/utils/file.py
+++ b/gramps/gen/utils/file.py
@@ -1,3 +1,5 @@
+#! /usr/bin/env python3
+# -*- coding: utf-8 -*-
 #
 # Gramps - a GTK+/GNOME based genealogy program
 #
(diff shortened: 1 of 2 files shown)


gramps/gen/utils/file.py
```diff
@@ -86,6 +88,27 @@ def get_new_filename(ext, folder='~/'):
...
+# TODO check this function, add 'try - exeption', add check "dbfolder are present?"
+def get_avail_disk_size(path_to_folder):
+    """
+    Check available disk space in MB
+    """
+    if win():                               # for windows:
+    #     import psutil
+    #     DISK = get disk from 'dbfolder'
+    #     freedisk = psutil.disk_usage(DISK).free/(1024*1024)
+    #     print(f"{freedisk:.4} Mb free on disk {DISK}")
```
prculley, line 100: https://docs.python.org/3/library/shutil.html should be used here to get the disk space free "shutil.disk_usage(path)" as it is supposed to work across platforms as of Python 3.3 (which we are now requiring).  And we already use the shutil library, so no new dependencies.

Nick-Hall:
@PushKK Does this replace #1212?

gramps/gen/utils/file.py
PushKK: Thank you, I вшвтэе know about this function.

PushKK:
I had problems with PR. All my commit join to last PR, @1212 created without .py file...

Nick-Hall:
I have converted this PR to draft because it is not ready for merging.

It doesn't work for Mac or Windows yet.

Nick-Hall:

PR #1221, opened 2021-05-24
Add the missing word

tiansworld:
Hello,
I think the word "missing" should be also in the plural form of the string.

### DIFF
--- a/gramps/plugins/test/reports_test.py
+++ b/gramps/plugins/test/reports_test.py
@@ -366,7 +366,7 @@ reports.addcli(TestDynamic, "tool_verify",
                "--options", "name=verify")
 
 txt_list = ["6 media objects were referenced, but not found",
-            "References to 6 media objects were kept"]
+            "References to 6 missing media objects were kept"]
 reports.addcli(TestDynamic, "tool_check",
                out_does_contain(txt_list),
                [None],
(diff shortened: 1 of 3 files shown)


Nick-Hall:
You will also need to change the unit tests [here](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/test/reports_test.py#L368) and [here](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/test/tools_test.py#L142).

tiansworld:
> You will also need to change the unit tests [here](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/test/reports_test.py#L368) and [here](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/test/tools_test.py#L142).

Thanks Nick.
I found the string when I was translating.  I didn't know that it also affects the unit tests.

But base on that string's content, I think [L369](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/test/reports_test.py#L369) and  [L143](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/test/tools_test.py#L143) should be changed(add "missing" before "media") respectively instead.

Nick-Hall:
Rebased on the _master_ branch.  This typo wasn't serious enough to justify a string change in the maintenance branch.

PR #1223, opened 2021-05-24
For Weblate good work: small fix

PushKK:
Tips.xml:
1. Old terms changed,
2. Add dots (.) to ends of rows.
Other files:
3. Delete whitespaces.
4. Capitalize first letters.

Commit #2
Birthdayreport.py:
1. Add space,
Other .py:
2. Add first row,
3. Change strings to code.

Less strings - less errors in Weblate!

### DIFF
--- a/gramps/gui/glade/editplace.glade
+++ b/gramps/gui/glade/editplace.glade
@@ -235,7 +235,7 @@ You can set these values via the Geography View by searching the place, or via a
                       <object class="ValidatableMaskedEntry" id="latlon_entry">
                         <property name="visible">True</property>
                         <property name="can_focus">True</property>
-                        <property name="tooltip_text" translatable="yes">Field used to paste info from a web page like google, openstreetmap, ... </property>
+                        <property name="tooltip_text" translatable="yes">Field used to paste info from a web page like OpenStreetMap, Google, ...</property>
                         <property name="hexpand">True</property>
                         <property name="invisible_char">●</property>
                       </object>
@@ -375,7 +375,7 @@ You can set these values via the Geography View by searching the place, or via a
                   <object class="GtkComboBox" id="place_type">
                     <property name="visible">True</property>
                     <property name="can_focus">False</property>
-                    <property name="tooltip_text" translatable="yes">What type of place this is. Eg 'Country', 'City', ... .</property>
+                    <property name="tooltip_text" translatable="yes">What type of place this is. Eg 'Country', 'City', ...</property>
                     <property name="has_entry">True</property>
                     <child internal-child="entry">
                       <object class="GtkEntry" id="combobox-entry2">
--- a/gramps/gui/glade/editplace.glade
+++ b/gramps/gui/glade/editplace.glade
@@ -235,7 +235,7 @@ You can set these values via the Geography View by searching the place, or via a
                       <object class="ValidatableMaskedEntry" id="latlon_entry">
                         <property name="visible">True</property>
                         <property name="can_focus">True</property>
-                        <property name="tooltip_text" translatable="yes">Field used to paste info from a web page like OpenStreetMap, Google, ...</property>
+                        <property name="tooltip_text" translatable="yes">Field used to paste info from a web page like google, openstreetmap, ... </property>
                         <property name="hexpand">True</property>
                         <property name="invisible_char">●</property>
                       </object>
--- a/gramps/plugins/textreport/descendreport.py
+++ b/gramps/plugins/textreport/descendreport.py
@@ -1,3 +1,4 @@
+#
 # Gramps - a GTK+/GNOME based genealogy program
 #
 # Copyright (C) 2000-2007  Donald N. Allingham
--- a/gramps/plugins/textreport/descendreport.py
+++ b/gramps/plugins/textreport/descendreport.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
 #
 # Gramps - a GTK+/GNOME based genealogy program
 #
--- a/gramps/plugins/textreport/descendreport.py
+++ b/gramps/plugins/textreport/descendreport.py
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
-#
 # Gramps - a GTK+/GNOME based genealogy program
 #
 # Copyright (C) 2000-2007  Donald N. Allingham
--- a/gramps/plugins/textreport/descendreport.py
+++ b/gramps/plugins/textreport/descendreport.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
 # Gramps - a GTK+/GNOME based genealogy program
--- a/gramps/plugins/textreport/detancestralreport.py
+++ b/gramps/plugins/textreport/detancestralreport.py
@@ -406,7 +406,7 @@ class DetAncestorReport(Report):
                 self.doc.start_paragraph('DAR-MoreDetails')
 
                 text = utils.get_address_str(addr)
-                self.doc.write_text(addr.get_schema()["title"] + ': ')
+                self.doc.write_text(_('{address}: ').format(address=addr.get_schema()["title"]))
 
                 if self.fulldate:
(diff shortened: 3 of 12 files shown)


tosky:
This change partially duplicates https://github.com/gramps-project/gramps/pull/1213. Please remove the duplicated changes, or wait for that change to be merged.

gramps/gui/glade/editplace.glade
```diff
@@ -235,7 +235,7 @@ You can set these values via the Geography View by searching the place, or via a
                       <object class="ValidatableMaskedEntry" id="latlon_entry">
                         <property name="visible">True</property>
                         <property name="can_focus">True</property>
-                        <property name="tooltip_text" translatable="yes">Field used to paste info from a web page like google, openstreetmap, ... </property>
+                        <property name="tooltip_text" translatable="yes">Field used to paste info from a web page like OpenStreetMap, Google, ...</property>
```
tosky, line 238: This is addressed in https://github.com/gramps-project/gramps/pull/1213

PushKK: :(
I fork project today and this fix not be present in code.

gramps/plugins/textreport/descendreport.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
```
Nick-Hall, line 1: There is no need to add this.

gramps/plugins/textreport/detancestralreport.py
```diff
@@ -405,7 +406,7 @@ def write_more_header(first, name):
                 self.doc.start_paragraph('DAR-MoreDetails')
 
                 text = utils.get_address_str(addr)
-                self.doc.write_text(self._('Address: '))
+                self.doc.write_text(addr.get_schema()["title"] + ': ')
```
Nick-Hall, line 409: Good idea, but I don't think that it will work with deferred translations in reports.  Also, you can't simply concatenate ': ' because the French will want a space before the colon.

gramps/plugins/textreport/detdescendantreport.py
```diff
@@ -927,13 +928,13 @@ def write_person_info(self, person):
                 self.doc.start_paragraph('DDR-MoreDetails')
 
                 text = utils.get_address_str(addr)
+                self.doc.write_text(addr.get_schema()["title"] + ': ')
```
Nick-Hall, line 931: As previous comment.

gramps/plugins/webreport/repository.py
```diff
@@ -170,7 +170,7 @@ def repositorylistpage(self, report, the_lang, the_title, repos_dict, keys):
                     Html("th", "&nbsp;", class_="ColumnRowLabel", inline=True),
                     Html("th", self._("Type"), class_="ColumnType",
                          inline=True),
-                    Html("th", self._("Name", "Repository "),
```
Nick-Hall, line 173: Yes.  There is also a similar trailing space in the `gramps/plugins/webreport/media.py` file.

Nick-Hall (changes requested):
As @tosky pointed out, there is some duplication.  I'll merge his PR first.

gramps/plugins/textreport/detancestralreport.py
PushKK: You say about this:
                self.doc.write_text(addr.get_schema()["title"] + _(': '))

gramps/plugins/webreport/repository.py
PushKK: This word insert to html code, where all extra whitespaces are deleted when display.
And we see one space between words all time.

gramps/plugins/textreport/descendreport.py
PushKK: 1. Some .py files in Gramps have:
#!/usr/bin/env python
Some files have:
# -*- coding: utf-8 -*-
Some files not have nothing in it's header.
Usually all source in project use one header for unification (l' uniformité).

2. Gramps can run on Linux, Windows, Mac.
As you know - in Russia OS Windows can use **CP-1251** for GUI and **cp866** for Command line. 

3. And on https://docs.python.org/3/howto/unicode.html and https://stackoverflow.com/questions/14083111/should-i-use-encoding-declaration-in-python-3 I see - 'this header need'.

gramps/plugins/textreport/detancestralreport.py
tosky: No, this is incorrect. This is called "a string puzzle". It doesn't work in RTL languages for  example.

Please check what is written here: it is from the KDE community, but the basic principles of the "Writing Good Texts" section applies to all translations, regardless of the syntax:
https://api.kde.org/frameworks/ki18n/html/prg_guide.html#good_text

PushKK: You're right. I forgot about RTL.
It's must be:
`                self.doc.write_text(_('{address}: ').format(address=addr.get_schema()["title"]))
`

gramps/plugins/textreport/detdescendantreport.py
PushKK: Change it too:
`
                self.doc.write_text(_('{address}: ').format(address=addr.get_schema()["title"]))
`

gramps/plugins/textreport/descendreport.py
jralls: @PushKK The coding comment is no longer needed for Python3 but as the SO article points out it's still useful for some editors. The format originated in emacs but there are other editors that understand it. The python comment is used to allow those files to be executed directly from a shell on Unix systems. Neither is anything that translators need to worry about.

No, Windows can not use cp1251--or any other antique code page--for the native GUI. All internal strings in Windows are encoded in UTF16. Windows still uses code pages for the CMD shell and Powershell, though they're deprecated in favor of UTF8 in the latest version of Powershell. Gramps however doesn't use the native GUI, it uses [Gtk](https://www.gtk.org) and that requires that strings be encoded in UTF8, so all translation (i.e. po) files need to be encoded in UTF8. You don't need to worry about that either as long as you use weblate because it will take care of the encoding for you.

PushKK: Thank you. My last work with Windows had many problems with codepages.
MS Windows 7 Pro can use Unicode, software from Germany use CP-1252, our customer's software use CP-1251.
Every update software can was causing problems.

jralls:
@PushKK Why all of the adds/deletes of shebangs? That has nothing to do with translation and should be in a separate PR, plus each change needs an explanation and a correct description. It's not an "update".

gramps/plugins/textreport/detancestralreport.py
Nick-Hall: This still doesn't work with deferred translations.  Test by selecting a different "Translation" in "Report Options (2)".

PushKK (commented):
> This still doesn't work with deferred translations. Test by selecting a different "Translation" in "Report Options (2)".
Yes, it is. I can not solve this problem.
Gettext can use translated strings for 'locale_opt' (locale 'ar') from database, but not use translated strings for locale 'ar' from schema. And use default locale only.

gramps/plugins/textreport/detancestralreport.py
PushKK: For work in other locale "Report Options (2)" must send additional locale to 'get_schema()':
```
                self.doc.write_text(self._('{address}: ').format(address=addr.get_schema(self._locale)["title"]))
```

And /address.py.get_schema() must have additional parameter and code:
```
    def get_schema(cls, s_locale):
...
        from .date import Date

        new_lang = s_locale.translation
        new_lang.install()
        _=new_lang.gettext

        return {
            "type": "object",
            "title": _("Address"),
            "properties": {
```
Many .py files in '/gramps/gen/lib/' have schemas. All this files can be used with additional locale for creating reports in any supported language. Now this schema can be used in default language only.

But it's must be one big PR with many changed files...
It's need?

Nick-Hall:
Merged in commit 3b59aa3a44598c9f36725aa3d95605659bb08e69.

Nick-Hall:
Merged in commit 3b59aa3a44598c9f36725aa3d95605659bb08e69.

PR #1225, opened 2021-05-25
narrativeweb.py language reworked

comradekingu:
Could "direct editing" be turned on in Weblate? Can leave suggestions for the stuff that isn't clear.
Also, if you make https://hosted.weblate.org/user/kingu/ a moderator there I can add context to the stuff it is possible to figure out.

### DIFF
--- a/gramps/plugins/webreport/narrativeweb.py
+++ b/gramps/plugins/webreport/narrativeweb.py
@@ -1585,8 +1585,6 @@ class NavWebReport(Report):
         The extension is added to the filename as well.
 
         Notice that we do NOT use os.path.join() because we're creating a URL.
-        Imagine we run Gramps on Windows (gods forbidding), we don't want to
-        see backslashes in the URL.
         """
         if not fname:
             return ""
@@ -2382,14 +2380,14 @@ class NavWebOptions(MenuReportOptions):
             _('Add a complete events list and relevant pages or not'))
         addopt("inc_events", inc_events)
 
-        inc_places = BooleanOption(_('Include places pages'), False)
+        inc_places = BooleanOption(_('Include place pages'), False)
         inc_places.set_help(
-            _('Whether or not to include the places pages.'))
+            _('Whether or not to include the place pages.'))
         addopt("inc_places", inc_places)
 
-        inc_sources = BooleanOption(_('Include sources pages'), False)
+        inc_sources = BooleanOption(_('Include source pages'), False)
         inc_sources.set_help(
-            _('Whether or not to include the sources pages.'))
+            _('Whether or not to include the source pages.'))
         addopt("inc_sources", inc_sources)
 
         inc_repository = BooleanOption(_('Include repository pages'), False)
--- a/gramps/plugins/webreport/narrativeweb.py
+++ b/gramps/plugins/webreport/narrativeweb.py
@@ -219,21 +219,21 @@ class NavWebReport(Report):
         self.inc_updates = self.opts['updates']
         self.create_unused_media = self.opts['unused']
 
-        # Do we need to include this in a cms ?
+        # Do we need to include this in a CMS?
         self.usecms = self.options['usecms']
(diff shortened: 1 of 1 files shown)


gramps/plugins/webreport/narrativeweb.py
```diff
@@ -2424,7 +2424,7 @@ def __add_place_map_options(self, menu):
 
         mapopts = [
             [_("OpenStreetMap"), "OpenStreetMap"],
-            [_("StamenMap"), "StamenMap"],
+            [_("StamenMap"), "Stamen Map"],
```
comradekingu, line 2427: Not sure about this

Nick-Hall:
> Could "direct editing" be turned on in Weblate?

I can't find this in the settings.  What does it do?

gramps/plugins/webreport/narrativeweb.py
```diff
@@ -1940,9 +1940,9 @@ def __add_report_options(self, menu):
         category_name = _("Report Options")
         addopt = partial(menu.add_option, category_name)
 
-        self.__archive = BooleanOption(_('Store web pages in .tar.gz archive'),
+        self.__archive = BooleanOption(_('Store webpages in .tar.gz archive'),
```
Nick-Hall, line 1943: Whilst the use of "website" seems to be more common than "web site", I'm not too sure about " webpages".  I think that it would be better the use "website" here.

gramps/plugins/webreport/narrativeweb.py
Nick-Hall: You could do this, but it is the first item in the list that is visible to the user, not the second.

gramps/plugins/webreport/narrativeweb.py
```diff
@@ -1585,7 +1585,7 @@ def build_url_fname(self, fname, subdir=None, uplink=False,
         The extension is added to the filename as well.
 
         Notice that we do NOT use os.path.join() because we're creating a URL.
-        Imagine we run gramps on Windows (heaven forbits), we don't want to
```
Nick-Hall, line 1588: The English this would be "heaven forbid!".  :)

Nick-Hall:
We should also think about changing the menu "Reports -> Web Pages -> Narrated Web Site...".

gramps/plugins/webreport/narrativeweb.py
comradekingu, line 2427:
```suggestion
            [_("Stamen Map"), "StamenMap"],
```

gramps/plugins/webreport/narrativeweb.py
comradekingu, line 1943:
```suggestion
        self.__archive = BooleanOption(_('Store websites in .tar.gz archive'),
```

gramps/plugins/webreport/narrativeweb.py
comradekingu: Yes, but that idea isn't chiefly British, nor English. What sort of ancestral outfit would this be were it to forward such absurdities, the mere thought of which beckons the mind. It is the very backslash in the domain that is European history and culture.

comradekingu:
@Nick-Hall Direct edits can be turned on from https://hosted.weblate.org/settings/gramps-project/gramps/#files under "edit base file". That way source strings can be edited by users.

Nick-Hall:
@comradekingu The "Edit base file" option only applies to monolingual language files.  We use the bilingual gettext format.

Our POT file is generated from the source code, so any changes in Weblate would be overwritten.

comradekingu:
@Nick-Hall
OK, but this seems good to go? Have some other PRs planned.

Nick-Hall:
@comradekingu I made a couple of small changes.

We don't use the ellipsis character anywhere in Gramps, so I decided not to include them here.  To keep a consistent interface it would be better to make all ellipsis changes at the same time.

PR #1228, opened 2021-05-26
Navweb: add place name to birth and death dates

SNoiraud:
when we use the toggle section.

Fixes #012316

### DIFF
--- a/gramps/plugins/webreport/person.py
+++ b/gramps/plugins/webreport/person.py
@@ -60,6 +60,7 @@ from gramps.gen.plug.report import utils
 from gramps.gen.utils.alive import probably_alive
 from gramps.gen.constfunc import win
 from gramps.gen.display.name import displayer as _nd
+from gramps.gen.display.place import displayer as _pd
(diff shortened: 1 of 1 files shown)


gramps/plugins/webreport/person.py
```diff
@@ -1705,19 +1715,26 @@ def display_ind_general(self):
...
+                    if p_birth:
+                        p_birth = " (" + p_birth + ")"
+                    if p_death:
+                        p_death = " (" + p_death + ")"
                     if birth_date and birth_date is not Date.EMPTY:
                         trow = Html("tr") + (
                             Html("td", self._("Birth date"),
                                  class_="ColumnAttribute", inline=True),
-                            Html("td", self.rlocale.get_date(birth_date),
+                            Html("td", self.rlocale.get_date(birth_date) +
```
Nick-Hall, line 1727: This may cause a problem with RTL languages such as Hebrew.

SNoiraud: Done

PR #1229, opened 2021-05-26
Update and improve Simplified Chinese translation

tiansworld:
Hello guys,
I made some new translation as well as improvements to the old ones.
Please review.


Nick-Hall:
Thanks.  Simplified Chinese is now 90.55% translated.

PR #1232, opened 2021-05-28
Improve and correct the Simplified Chinese translation

tiansworld:
Hello guys,
This update contains mainly improvements and corrections to the existing translation(Correct the typos, mistakes, mix of translatable English words, also rephrase some of the sentences to make them smooth in Chinese).
Please review it.

Also please have a look at #1221, I want to know if I am right or not, so I can change the affected files too.
Thanks a lot.

For the rest 10% of the untranslated strings, I think they will spend me more time. Most of them are not easy to understand and translate without seeing them in the GUI. While finding their positions in the GUI is another challenge. But I will try my best.

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -15,7 +15,7 @@ msgstr ""
 "Project-Id-Version: Gramps 4.2.0-dev\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2020-07-15 22:20+0200\n"
-"PO-Revision-Date: 2021-05-26 15:38+0800\n"
+"PO-Revision-Date: 2021-05-28 14:02+0800\n"
 "Last-Translator: Shixiong Tian <tiansworld@fedoraproject.org>\n"
 "Language-Team: Chinese (simplified)\n"
 "Language: zh_CN\n"
@@ -203,7 +203,7 @@ msgstr "犹太教的新年"
 
 #: ../data/holidays.xml.in.h:26
 msgid "Rosh Ha'Shana 2"
-msgstr ""
+msgstr "Rosh Ha'Shana 2"
 
 #: ../data/holidays.xml.in.h:27
 msgid "Yom Kippur"
@@ -211,31 +211,31 @@ msgstr "『犹太教』赎罪日"
 
 #: ../data/holidays.xml.in.h:28
 msgid "Sukot"
-msgstr ""
+msgstr "Sukot"
 
 #: ../data/holidays.xml.in.h:29
 msgid "2 of Sukot"
-msgstr ""
+msgstr "Sukot 第二天"
 
 #: ../data/holidays.xml.in.h:30
 msgid "3 of Sukot"
-msgstr ""
+msgstr "Sukot 第三天"
 
 #: ../data/holidays.xml.in.h:31
 msgid "4 of Sukot"
-msgstr ""
+msgstr "Sukot 第四天"
 
 #: ../data/holidays.xml.in.h:32
 msgid "5 of Sukot"
-msgstr ""
+msgstr "Sukot 第五天"
 
 #: ../data/holidays.xml.in.h:33
 msgid "6 of Sukot"
-msgstr ""
+msgstr "Sukot 第六天"
 
 #: ../data/holidays.xml.in.h:34
 msgid "7 of Sukot"
-msgstr ""
+msgstr "Sukot 第七天"
 
 #: ../data/holidays.xml.in.h:35
 msgid "Simhat Tora"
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Thanks.  Simplified Chinese is now 91.63% translated.

Nick-Hall:
> For the rest 10% of the untranslated strings, I think they will spend me more time. Most of them are not easy to understand and translate without seeing them in the GUI. While finding their positions in the GUI is another challenge. But I will try my best.

@tiansworld If you let me know the strings you are having problems with, I'll create some screenshots for you.

tiansworld:
> @tiansworld If you let me know the strings you are having problems with, I'll create some screenshots for you.

@Nick-Hall Thank you.

Take this for example, "Place %(id)s with %(id2)s", it's  in line 263 of gramps/plugins/importer/importxml.py. The comment of the source function says:"Extract info from obj about 'merge-candidate'...", so I guess it is from import and merge tool. But I couldn't find it from "Import and Merge Tool".
It's a little ambiguous to me, I can understand it as "a place ID number id, which contains the second id number id2?" or "A place, with id number id and id2?".
I need to know what is displaying around this string, or even the whole dialog, to translate it appropriately.

I will meet some other strings alike later, I will be much appreciate if you can teach me how to create a screenshot or find the desired string in GUI efficiently. I mean not the way I am currently using, by opening Gramps, then find and try through the menus/settings/views etc. So I can handle them by myself in the future.

Nick-Hall:
@tiansworld I have uploaded a screenshot to Weblate [here](https://hosted.weblate.org/translate/gramps-project/gramps/en/?checksum=d23f12e040d05e65).

In this case you would have to look at the code and realise that the [`ImportInfo`](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/importer/importxml.py#L187) class holds information about an import and the method [`info_text`](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/importer/importxml.py#L277) constructs the message.  Then you would look at the references for `info_text` to find that it is called by the [`DbLoader`](https://github.com/gramps-project/gramps/blob/9c29b8e2b528a71b70f84212e6bb0ae61033d422/gramps/gui/dbloader.py#L141) and then the [`ViewManager`](https://github.com/gramps-project/gramps/blob/9c29b8e2b528a71b70f84212e6bb0ae61033d422/gramps/gui/viewmanager.py#L947).  In the GitHub editor references can be found by just clicking on the method and looking in the "References" tab.

Unfortunately, this process is made easier with some coding experience and knowledge of the Gramps codebase.  Let me know if you need help with any other strings.

jralls:
> In this case you would have to look at the code and realise that the ImportInfo class holds information about an import and the method info_text constructs the message. Then you would look at the references for info_text to find that it is called by the DbLoader and then the ViewManager. In the GitHub editor references can be found by just clicking on the method and looking in the "References" tab.

That's too much to ask of translators, few of whom know much about programming. `xgettext` will extract comments if you tell it to, but we do so only for [glade](https://github.com/gramps-project/gramps/blob/master/po/update_po.py#L340) and bizarrely [C header](https://github.com/gramps-project/gramps/blob/master/po/update_po.py#L440) files, not [Python](https://github.com/gramps-project/gramps/blob/master/po/update_po.py#L426). Common practice is to pass something like [`--add-comments=Translators`](https://www.gnu.org/software/gettext/manual/gettext.html#index-_002d_002dadd_002dcomments_002c-xgettext-option) to xgettext and have comments like
```
# Translators: This is part of the import results report, it shows the number of places imported followed by the 
# number of references to those places in parentheses.
```
On any string that isn't obvious without examining the code or where an English word has multiple meanings and which one isn't apparent from the context of the msgid.

Nick-Hall:
> Common practice is to pass something like --add-comments=Translators to xgettext

We specify the `-c` flag rather than `--add-comments` for the python files.  In fact, I recently moved some comments in [this commit](https://github.com/gramps-project/gramps/commit/1423662e33d9468b36ee343498bd7d986ee4a13c) beacuse they were confusing for translators.  Specifying a tag would be much better.  I'll investigate the impact.

Nick-Hall:
A custom search in Weblate for "has:note and language:en and not note:translator" reveals that there are 335 source code comments in the po files which do not contain the word "translator".   Most of these are irrelevant to translators, but unfortunately some are intended for them.

I'll go through the search results in more detail tomorrow.

tiansworld:
> In this case you would have to look at the code and realise that the [`ImportInfo`](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/importer/importxml.py#L187) class holds information about an import and the method [`info_text`](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/importer/importxml.py#L277) constructs the message. Then you would look at the references for `info_text` to find that it is called by the [`DbLoader`](https://github.com/gramps-project/gramps/blob/9c29b8e2b528a71b70f84212e6bb0ae61033d422/gramps/gui/dbloader.py#L141) and then the [`ViewManager`](https://github.com/gramps-project/gramps/blob/9c29b8e2b528a71b70f84212e6bb0ae61033d422/gramps/gui/viewmanager.py#L947). In the GitHub editor references can be found by just clicking on the method and looking in the "References" tab.
> 

Thanks, I think you provide me a new way of tracking strings. My Python knowledge is more or less like my Deutsch level, I can read and understand a few of words(variables/functions), even can guess very simple sentences(classes/methods), of course reading a passage(block) of novel(whole source file) is impossible. So the process you suggested above is worth to try. Though it's a little hard at the moment.

tiansworld:
> In this case you would have to look at the code and realise that the [`ImportInfo`](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/importer/importxml.py#L187) class holds information about an import and the method [`info_text`](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/importer/importxml.py#L277) constructs the message. Then you would look at the references for `info_text` to find that it is called by the [`DbLoader`](https://github.com/gramps-project/gramps/blob/9c29b8e2b528a71b70f84212e6bb0ae61033d422/gramps/gui/dbloader.py#L141) and then the [`ViewManager`](https://github.com/gramps-project/gramps/blob/9c29b8e2b528a71b70f84212e6bb0ae61033d422/gramps/gui/viewmanager.py#L947). In the GitHub editor references can be found by just clicking on the method and looking in the "References" tab.
> 

Thanks, I think you provide me a new way of tracking strings. My Python knowledge is more or less like my Deutsch level, I can read and understand a few of words(variables/functions), even can guess very simple sentences(classes/methods), of course reading a passage(block) of novel(whole source file) is impossible. So the process you suggested above is worth to try. Though it's a little hard at the moment.

tiansworld:
> ```
> # Translators: This is part of the import results report, it shows the number of places imported followed by the 
> # number of references to those places in parentheses.
> ```
> 
> On any string that isn't obvious without examining the code or where an English word has multiple meanings and which one isn't apparent from the context of the msgid.

Yes, I agree.
But having these kind of comment is better than having nothing in the code. They will be the second clue for experienced translators.

For weblate single string mode, it's good to include these comment in the 'String information' on the right side of the page, but they are not visible in Zen mode.

Nick-Hall:
> Though it's a little hard at the moment.

As @jralls said, you are not expected to understand the code.  Let me know if you would like more screenshots.  They will be helpful to other translators as well.

PR #1233, opened 2021-05-29
Add round trip Ancestry.com _APID tag support

cdhorn:
This PR addresses feature request [#9925](https://gramps-project.org/bugs/view.php?id=9925).  

It adds support for saving the _APID tags used by Ancestry.com as attributes on citation and source objects during a Gedcom import and then including them in a Gedcom export. Note for citations these identify the specific item in the database being cited, for sources they merely identify the database. 

Tested an import of a raw Gedcom with 159097 _APID tags. On export 159089 were present.  A test upload to Ancestry.com worked as expected.  The 8 unaccounted for appear to have been missed due to unrelated structural parsing errors I did not take the time to investigate in detail.

### DIFF
--- a/gramps/plugins/export/exportgedcom.py
+++ b/gramps/plugins/export/exportgedcom.py
@@ -999,6 +999,11 @@ class GedcomWriter(UpdateCallback):
             self._note_references(source.get_note_list(), 1)
             self._change(source.get_change_time(), 1)
 
+            for srcattr in source.get_attribute_list():
+                if str(srcattr.type) == "_APID":
+                    self._writeln(1, "_APID", srcattr.value)
+                    break
+
     def _notes(self):
         """
         Write out the list of notes, sorting by Gramps ID.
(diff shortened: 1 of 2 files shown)


emyoulation:
Is there a way to use a GUID scheme instead?

Supporting a generic round-trip feature seems better than building a idiosyncratic Ancestry feature.

For instance, the tag for FamilySearch IDs in Rootsmagic, Legacy, Heredis, and MyHeritage is:  _FSFTID 
Family Historian v7 uses: _FSID tag

FTM2019 uses  FSID

%INDI._FSID%

Perhaps it could be a GUID attribute with a CSV of the ID tag and value?

cdhorn:
I think a more generic mechanism for handling various IDs like the above was what Paul's _UID PR was about.  From mailing list conversation my understanding is that needs to be reevaluated in light of Gedcom 7 and the EXID support. That seems the more proper way to handle those needs and where time is probably better spent.

Note with Ancestry the _APID tag represents a specific database or reference in a database and only applies to Citation and Source records. If you import a tree into Ancestry and have stripped out all the _APID records then there is no association to any of the Ancestry databases anymore, they're all treated as user generated citations that display under Other Sources and not Ancestry Sources.  And thus you loose the ability to view any of the associated images.

A related item of note is Gedcom 7 does not yet support EXID on SOURCE_CITATION records.

Nick-Hall:
Rebased.

Nick-Hall:
Rebased.

emyoulation:
Does this also address the issues in:
[0008332](https://gramps-project.org/bugs/view.php?id=8332): Provide an option to omit imported "3 _APID" Tags from Ancestry.com generated GEDCOMs

cdhorn:
I must have not seen that one.  No it does not provide an option to omit them. If I recall we do not have any import dialogs/options for adjusting what data will and will not be imported.

PR #1237, opened 2021-05-30
INSTALL: Bump required Python version to 3.5

PushKK:
1. Python version -> 3.5,
2. python -> Python,
3. gramps -> Gramps,
4. Reorder directories, delete duplicate,
5. Small fix.


Nick-Hall:
This all looks good.

PR #1239, opened 2021-05-30
README.md: update

PushKK:
1. Small fix,
2. I not sure about fontconfig - needed or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ to your language:
  Translation of GTK elements to your language, with
  xx your language code; e.g. for Dutch you need
  language-pack-gnome-nl. The translation of the
- Gramps strings is included with the gramps source.
+ Gramps strings is included with the Gramps source.
 
 
 The following packages are **STRONGLY RECOMMENDED** to be installed:
@@ -45,7 +45,7 @@ The following packages are **STRONGLY RECOMMENDED** to be installed:
  It may be osmgpsmap, osm-gps-map, or python-osmgpsmap,
  but the Python bindings for this must also be present, so gir1.2-osmgpsmap-1.0.
  Without this the GeoView will not be active, see
- https://gramps-project.org/wiki/index.php?title=Gramps_5.0_Wiki_Manual_-_Categories#Geography_Category
+ https://gramps-project.org/wiki/index.php?title=Gramps_5.1_Wiki_Manual_-_Categories#Geography_Category
 
 * **Graphviz**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,35 +63,35 @@ The following packages are **STRONGLY RECOMMENDED** to be installed:
...
  distributions. See http://pyicu.osafoundation.org/
- (These are Python bindings for the ICU package.
- https://pypi.python.org/pypi/PyICU/)
+ (These are Python bindings for the ICU package
+ https://pypi.python.org/pypi/PyICU/).
 
 * **Ghostscript**
 
-  Used by Graphviz reports to help create PDF's
+  Used by Graphviz reports to help create PDF's.
```
jralls, line 71: Since you're here please remove the apostrophe too.

PushKK:
What you say about using fontconfig?
It's need for Gramps?

P.S. For Linux Mint 20 and 20.1 'fontconfig' not work for Gramps,
'python-fontconfig' not present more.

P.P.S. Can you select special symbol (U+2628 or U+2626) in .po file for each language?
1 way. Add to gramps.ini this rows:
```
[symbols]
dead=☦
```
And select symbols in Preferences.
2 way. Use gettext for symbol select:
`_("☦")
`In .po file user select need symbol for his language.

SNoiraud:
See PR #1069

PushKK:
Sorry, but 
```
>gramps/gen/utils/symbols.py

class Symbols(object):
    # genealogical symbols
    SYMBOL_FEMALE = 0
    SYMBOL_MALE = 1
    SYMBOL_ASEXUAL_SEXLESS = 2 # Unknown
...
    SYMBOL_ILLEGITIM = 9
```
not relation to ISO/IEC-5218-2004:
```
Data elements Code
Not known       0 (zero)
Male                1 (one)
Female            2 (two)
Not applicable 9 (nine)
```

SNoiraud:
> not corresponds to ISO/IEC-5218-2004

Perhaps, but this is only for an internal use and  not to be exported for the moment.
The ISO norm you refer is really the  ISO/IEC DIS 5218 norm which is only in an inquiry stage.
So we can wait several years before it become a standard.

README.md
```diff
@@ -102,7 +102,8 @@ The following packages are optional:
 * **ttf-freefont**
 
  More font support in the reports
-
+ For use genealogic symbols you must have FreeSerif or DejaVu Sans fonts (or other) installed.
```
Nick-Hall, line 105: This looks wrong. It gets appended to the previous line which doesn't end with a full stop.

Also DejaVu Sans is not included in this package.

Nick-Hall (changes requested):
Apart from my one comment, everything else looks OK.

Nick-Hall (approved):

README.md
Nick-Hall: I changed this myself.

PushKK: 1. I check Gramps on fresh Linux system some years ago.
FreeSerif font had better MARRIAGE symbol - big and clear.
DejaVu - had this symbol too, but this symbol - small. 
2. Thank you.

PR #1240, opened 2021-05-30
Childreftype.py: adding context

PushKK:
For Russian language it need.


Nick-Hall:
See also #1049.  It is a similar PR, but not a duplicate.

PushKK:
In Russian language:
Birth (as Event) - 'Рождение'
Birth (as Relationship) - 'Родной'

PR #1241, opened 2021-05-30
Use GitHub Actions to run continuous integration checks

Nick-Hall:
First attempt.

Still need to consider:

* Test coverage
* Deployment of source distribution and python wheel


PushKK:
Say me please:
1. Can you create 'gramps.pot' and update all .po files before build Gramps?
This procedure can take some time but can found errors in i18n strings. And Gramps tests can use fresh .mo files.
2. After commit merge - new 'gramps.pot' file created and sending to Weblate automatically?

PR #1249, opened 2021-06-04
Update filtereditor.py

PushKK:
Add signs and '\n'.


Nick-Hall:
Perhaps use tabs in the formatting.  I've added a commit.

This PR causes problems in en_US because of the "|".  We still have some code that strips out the old translation context.  I'll fix this.

PR #1251, opened 2021-06-04
Localization: toolsdebug.gpr.py

PushKK:
1. Add localization,
2. Code formatted for PEP8.

PR #1253, opened 2021-06-05
Localization: toolsdebug.gpr.py

PushKK:
Add localization, for https://github.com/gramps-project/gramps/pull/1251


Nick-Hall:
Debugging tools should not be localised.

PR #1255, opened 2021-06-05
Localization: populatesources.py

PushKK:
1. Add localization,
2. Add localized values for all fields for Sources and Citations,
3. Change messages for user,
4. Change position for 'dialog.destroy()'. For Python 3.8 it work,
5. Delete 'try except', for me - it's not need (after change position of 'dialog.destroy()'),
6. Add ngettext(). I nor sure about this...


Nick-Hall:
Another development tool that we don't want to translate.

Nick-Hall:
Debugging tools should not be localised.

PR #1256, opened 2021-06-05
Localization: const.py

PushKK:
Sign 'COLON' can translated for RTL, French and other languages.
Combination ':' + " " can resolve problems, e.g. this: https://github.com/gramps-project/gramps/pull/1223#discussion_r638147788
For work with COLON need change all 'STRING:' to 
`_('STRING{colon}').format(colon=COLON)`

I add whitespace after colon and change comment only.

### DIFF
--- a/gramps/gen/const.py
+++ b/gramps/gen/const.py
@@ -262,7 +262,7 @@ NO_GIVEN = "(%s)" % _("none", "given-name")
 ARABIC_COMMA = "،"
 ARABIC_SEMICOLON = "؛"
 DOCGEN_OPTIONS = 'Docgen Options'
-COLON = _(':') # Translators: needed for French, ignore otherwise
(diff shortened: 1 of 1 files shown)


prculley:
The way 'COLON' is used, it looks to me like we should NOT add a trailing space, at least for LTR languages.  Most often it is used at the end of a label (stand alone text for the GUI), but in gramps\gui\aboutdialog.py it is prepending more text that includes its own leading space.

PushKK:
> in gramps\gui\aboutdialog.py
e.g. "GRAMPS" + COLON + " %s \n"

Now this string for 'en', 'de' and 'ru' languages are
`"GRAMPS: %s \n"`
for 'fr' language are
`"GRAMPS :  %s \n"` (2 whitespaces after colon)
for 'ar' and 'he' language are
`"GRAMPS: %s \n"`

Nick-Hall:
No change required.

PR #1257, opened 2021-06-05
Update _exportassistant.py

PushKK:
1. Split one big message to two small messages, small message better for use with Weblate,
2. Reorder buttons names in message (to real button order).
Снимок экрана_2021-06-05_14-35-34

### DIFF
--- a/gramps/gui/plug/export/_exportassistant.py
+++ b/gramps/gui/plug/export/_exportassistant.py
@@ -448,11 +448,12 @@ class ExportAssistant(ManagedWindow, Gtk.Assistant):
                 hasattr(self.option_box_instance, "no_fileselect")):
                 # No file selection
                 filename = ''
-                confirm_text = _(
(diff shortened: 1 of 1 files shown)


Nick-Hall:
> Split one big message to two small messages, small message better for use with Weblate

Isn't one big message better for translators?

> Reorder buttons names in message (to real button order).

The order of the buttons is probably dependant on the operating system.  However,  The order in Gnome is also Cancel, Back, Apply.  I have no real objections to changing the order.

PushKK:
> Isn't one big message better for translators?

String "Press Cancel to ..." can be use in other messages:
../gramps/plugins/tool/mediamanager.py:351
../gramps/gui/plug/export/_exportassistant.py:465
../gramps/gui/plug/export/_exportassistant.py:452

One phrase - one translate. We must use Weblate for best results.

PR #1262, opened 2021-06-05
Localization: dateparserdisplaytest.py

PushKK:
1. Add localization,
2. Add new function - receive date, return Calendar. It's needed for Russia - we use Gregorian calendar from 1918-02-01.
I not know - Gramps have this function?


romjerome:
Note, the "french republican/revolutionary" starts on 1792 (Year I). Using '1789' as year will test a date before the "french calendar". Sure, this is a valid test, but not always to expected one.

Nick-Hall:
Replaced by PR #1462.

PR #1263, opened 2021-06-10
Image Meta Viewer - expand support for IPTC, XMP and Named regions with Thumbnails

brucejackson:
- Expand support in image metadata viewer for XMP and IPTC metadata
- Display thumbnails of Named Regions in the metadata viewer


brucejackson:
One thing I would do is increase the row height for people tags when a person's thumbnail is displayed.  I don't like the face being cut off.  Any ideas, on how I can specify the row height in the view for specific rows?

image

emyoulation:
Perhaps you could download Chris Horn'# experimental LinkedView (Browse View) add-on and see how the Media viewer handles thumbnail allocation?

https://github.com/cdhorn/LinkedView

I plan to download your mods & see how this addon works in combination. 

(Hoping to eventually find something that will let me see marquee rectangles & a GUI to numerically tweak sizes & order. And something that looks at sidecar files plus possibly converting a sidecar to embedded data.)

Nick-Hall:
> Any ideas, on how I can specify the row height in the view for specific rows?

Check that the fixed-height-mode for the TreeView is set to False.

`treeview.set_fixed_height_mode(False)`

brucejackson:
I don't think reading sidecar data within Gramps will be possible.  I don't see any methods in GExiv2 (https://lazka.github.io/pgi-docs/GExiv2-0.10/classes/Metadata.html).  Though it does have the ability to save to sidecar files.  ExifTool will let you batch embed the data from sidecar files (https://exiftool.org/forum/index.php?topic=4659.msg22238#msg22238)

I was thinking that the rectangles (named regions) should maybe be added to the image preview tab as a separate feature.  

Reordering and resizing probably is probably possible.  It would fit better in the Edit Metadata Gramplet.  

I will check out the linked view.

emyoulation:
Chris put in a Configuration 3-state options for thumbnails: off, small & large. Come to think of it, there are left & right options too.

He separated options for People vs. Media thumbnails too.

brucejackson:
> treeview.set_fixed_height_mode(False)

Thanks, Nick.  That will get me started.   It works on the first image, but once the row height is expanded that row height remains expanded until I restart gramps.  I'll see what I can learn to help correct this.

image

emyoulation:
After swapping in the files to 5.1.4 (I expected it wouldn't work with 5.1.4 and that I needed to be using the master.  Just haven't gotten there yet.), the following error is reported.

```
2022-01-21 16:24:32.464: WARNING: dummydb.py: line 1428: database is closed
2022-01-21 16:24:41.689: WARNING: _manager.py: line 308: Plugin error (from 'editexifmetadata'): cannot import name 'format_datetime'
2022-01-21 16:24:41.689: WARNING: grampletpane.py: line 181: Error loading gramplet 'Edit Image Exif Metadata': skipping content
```

brucejackson:
Thanks, I've fixed what I broke with this commit 9700ac9. You should be able to get the latest version and be running again.

Nick-Hall:
Rebased onto _master_ branch.

gramps/plugins/lib/libmetadata.py
```diff
@@ -4,6 +4,8 @@
 #
 # Copyright (C) 2011      Nick Hall
 # Copyright (C) 2011      Rob G. Healey <robhealey1@gmail.com>
+# Copyright (C) 2014    Nick Hall
```
Nick-Hall, line 7: I'm not sure why another copyright line is added for me.

gramps/plugins/lib/libmetadata.py
```diff
@@ -97,159 +112,284 @@ def format_gps(raw_dms, nsew):
...
-        (GPS, 'Exif.GPSInfo.GPSSatellites', None, None),
-        (ADVANCED, 'Exif.Image.Software', None, None),
-        (ADVANCED, 'Exif.Photo.ImageUniqueID', None, None),
-        (ADVANCED, 'Exif.Image.CameraSerialNumber', None, None),
-        (ADVANCED, 'Exif.Photo.ExifVersion', None, None),
-        (ADVANCED, 'Exif.Photo.FlashpixVersion', None, None),
-        (ADVANCED, 'Exif.Image.ExifTag', None, None),
-        (ADVANCED, 'Exif.Image.GPSTag', None, None),
-        (ADVANCED, 'Exif.Image.BatteryLevel', None, None)]
+def get_NamedRegions(self, key, metadata, image_path, mypixbuf):
```
Nick-Hall, line 115: This should be lowercase.  Check the [PEP8](https://peps.python.org/pep-0008/#function-and-variable-names) naming conventions.

Nick-Hall (changes requested):
This generally looks good.  I made a few very minor comments.

I'll mark this as "Needs Testing".  Adding a file with metadata to the example family tree would be useful.

gramps/plugins/lib/libmetadata.py
brucejackson: Thanks for the feedback, I will work through it.   The credit to you is for the code that detects the faces that you posted in Mantis (https://gramps-project.org/bugs/view.php?id=7442).  In looking at it now, I see is in the context of the PhotoTagging Gramplet not libmetadata.py

I can merge the lines (Copyright (C) 2011,2014      Nick Hall) or leave them separate or omit 2014.   I leave it to you as to how you want t your contribution recognized.

Nick-Hall: Are you sure it was me?  It looks like Artem Glebov wrote that code.

brucejackson: I'm not positive.  I've assumed that Nick_H in Mantis is also Nick-Hall in Github.  I inferred that you added the face detection code because these two posts in Mantis:

- https://gramps-project.org/bugs/view.php?id=7442#c34616
- https://gramps-project.org/bugs/view.php?id=7442#c34651.   'I only attached the file I changed'

Nick-Hall: Yes.  Nick_H is me.  I'd totally forgotten that I'd written that patch.  Thanks for reminding me.

You can just add ", 2014" to my existing line to give me credit.  If you only used my code then there is no need to credit Artem.

gramps/plugins/lib/libmetadata.py
```diff
@@ -97,159 +112,284 @@ def format_gps(raw_dms, nsew):
...
+                    
                     self.model.tree.expand_all()
                     retval = self.model.count > 0
             except:
                 pass
 
         return retval
 
-    def __add_section(self, section):
+    def add_section(self, section):
```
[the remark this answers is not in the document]
brucejackson: When it was private the gramplet wouldn't display any metadata.  I suspect because `get_named_regions `is not part of the class.  When I added `get_named_regions `into the class the code failed here:
https://github.com/brucejackson/gramps/blob/d24998743fad23b5a3d111896025f68327c34003/gramps/plugins/lib/libmetadata.py#L232

I could not figure out how to call `get_named_regions` that would cause an error.  Any pointers?

gramps/plugins/lib/libmetadata.py
brucejackson: Done

gramps/plugins/lib/libmetadata.py
brucejackson: Done

gramps/plugins/lib/libmetadata.py
```diff
@@ -97,159 +112,284 @@ def format_gps(raw_dms, nsew):
...
+                    if func2 is not None:
+                        func2(self, metadata[key], metadata, image_path, mypixbuf)
                         continue
 
                     if func is not None:
                         if key2 is None:
                             human_value = func(metadata[key])
                         else:
-                            if key2 in metadata.get_exif_tags():
+                            if key2 in metadata.get_exif_tags() + metadata.get_xmp_tags() + metadata.get_iptc_tags():
```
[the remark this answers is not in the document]
brucejackson: Good suggestion.  Done

gramps/plugins/lib/libmetadata.py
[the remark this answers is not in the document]
Nick-Hall: I would convert `get_named_regions` into a method.  I would also get rid of the `func2` definitions.

Then in the loop evaluating the tags, add some code like:
```
if key == 'Xmp.mwg-rs.Regions':
    node = self.add_section(PEOPLE)
    for row in self.get_named_regions(self.pixbuf):
        self.model.add(row, node=node)
    break
```

I notice that you can also simplify the code for the rectangles.  You don't need to use the `SelectionWidget' method `proportional_to_real_rect`.  If you multiply the xmp value (a decimal between 0 an 1) by the width or height you will get the pixel value.

gramps/plugins/gramplet/metadataviewer.py
```diff
@@ -31,6 +31,7 @@
 from gramps.plugins.lib.libmetadata import MetadataView
 from gramps.gen.plug import Gramplet
 from gramps.gen.utils.file import media_path_full
+from gramps.gui.widgets import SelectionWidget, Region
```
Nick-Hall, line 34: These imports are not needed.

gramps/plugins/lib/libmetadata.py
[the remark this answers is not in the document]
brucejackson: I've moved the get_named_regions to __get_named_regions and did the same to the other functions.  I've hopefully improved things.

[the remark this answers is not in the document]
Nick-Hall: Thanks.  It looks good now.

Nick-Hall (approved):

brucejackson:
Thank you for the feedback and help!

Nick-Hall:
@brucejackson Can we have an option to display the user friendly tag names that the previous version had?

ExifMetadata

Do you also have time to add a region to one of our example files?  The "Yawn" image already has some Exif tags, but you could choose another image.

hklaufus:
Hi, sorry for dropping in late to the party.

I'm trying to find where in the `sqlite.db` _Corner 1 (X, Y)_ and _Corner 2 (X, Y)_ from the _Media Reference Editor_ are stored?

* From the data model (https://www.gramps-project.org/wiki/index.php/Gramps_Data_Model) I tend to believe it is __not__ stored in the db.
* From the `thumb` folder I tend to believe that a cropped copy of the original image is stored.

Are my assumptions correct?

prculley:
In the data model it is in the 'rect' tuple part of the MediaRef.  Unfortunately if looking at the sqlite db it is buried in a blob for one of the primary objects (like person, or event etc.) and thus is difficult to decode/view.  The MediaRef objects are in a 'media_list' of the media references of each of these primary objects.

hklaufus:
@prculley : super, thanks. Overlooked that one.

Nick-Hall:
> What do you think about showing the namespace and the user-friendly label rather than an option?

Yes, that's fine.  I didn't want to lose anything from the previous version.

> Might it make some sense to look for new images, licensed for Creative Commons, or whatever scheme is appropriate for the Gramps Project?

I totally agree.  We don't want to be associated with such a website.

Nick-Hall:
Testing should involve creating a test file with all the tags that can be displayed by this gramplet and then checking the output.

A copy of `gramps/images/splash.jpg` would be a good choice of test file.  This could then be added to our example family tree.

emyoulation:
https://github.com/gramps-project/gramps/blob/master/images/splash.jpg

Image is of Robert Allingham, great-grandfather of Gramps creator, Don Allingham. (Robert Allingham, Jr. 1886 Monorhamilton, Ireland - 1968 Pittsburgh, Pennsylvania, USA)

Reference: 
Gramps Developer maillist: reply by Don Allingham on 26 Feb 2009 in thread "re:GRAMPS website"

brucejackson:
I'm questioning if my proposed change should go forward, given the different approach (and better approach) that is being taken by @metadata in the DisplayMetadata plugin - https://gramps.discourse.group/t/0012219-expand-imagemetadata-gramplet-to-support-xmp-and-iptc-metadata-tags/1240/34?u=ebj075.   

If the packaging issues can be resolved, I think it is a better long-term approach to metadata.   Exiftool from what I see has more metadata support than exiv2.  The DisplayMetadata gramplet is reporting all metadata in a file rather than making somewhat educated guesses about which fields are important.

It also doesn't make sense to me to have 2 gramplets that are so similar.

Nick-Hall:
Rebased.

Nick-Hall:
> I'm questioning if my proposed change should go forward, given the different approach (and better approach) that is being taken by @metadata in the DisplayMetadata plugin.

There seems to be no progress made with the DisplayMetadata plugin.  For Gramps, exiv2 is more convenient than exiftool and we already package it.

I think that this is ready to merge as it offers a clear benefit over the existing plugin.  @brucejackson Are you happy for me to go ahead with this?  We plan to release v5.2 at the end of July.

PR #1264, opened 2021-06-10
Inherit month names from base class

Nick-Hall:
This should allow the German date handler to work with Austrian month names.

### DIFF
--- a/gramps/gen/datehandler/_date_de.py
+++ b/gramps/gen/datehandler/_date_de.py
@@ -269,13 +269,6 @@ class DateDisplayDE(DateDisplay):
     German language date display class.
     """
 
-    long_months = (  "", "Januar", "Februar", "März", "April", "Mai",
-                    "Juni", "Juli", "August", "September", "Oktober",
-                    "November", "Dezember" )
-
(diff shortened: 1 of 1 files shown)


cedecode:
Dear @Nick-Hall ,
thank you for this fix. I did some tests but it didn't work, maybe `datehandler/_date_de.py` was not the only place.

**For any report**: Regardless whether I choose translation `de` or `de_at` it always takes the translation from `de.po`. So commenting out the list in datehandler didn't change this behaviour.

I tried it with a test person who was born in January. Gramps displays "Jänner" (correct), but its reports persistently write "Januar".


Another thing - I'm sure whether this has the same causation. When I start Gramps it reports:
`WARNING: relationship.py: line 2318: Family relationship translator not available for 'de_AT'. Using 'english' instead.`

prculley:
The problem for the report issue is that the register_datehandler call at the bottom of the file is aliasing all the variations of German to de_DE.  What you want to do is to add a second call with only de_AT.
`register_datehandler(
    ('de_AT', ('%d.%m.%Y',)),
    DateParserDE, DateDisplayDE)`

cedecode:
I suggest closing this because I created a pull request for these problems in #1268.
Thank you @prculley!

PR #1268, opened 2021-06-17
datehandler and relationship calculator with de_AT

cedecode:
- This should finally solve the datehandler issue in #1264. Thank you @prculley and @Nick-Hall!
- I also found a solution for the error mentioned in the same thread (`WARNING: relationship.py: line 2318: Family relationship translator not available for 'de_AT'. Using 'english' instead.`). New `rel_de_AT.py` contains some minor differences between de and de_AT which are commented in line #32.
- added: the current de_AT.po from weblate, which is not mandatory for there problems here.

### DIFF
--- a/gramps/gen/datehandler/_date_de.py
+++ b/gramps/gen/datehandler/_date_de.py
@@ -269,13 +269,6 @@ class DateDisplayDE(DateDisplay):
     German language date display class.
     """
 
-    long_months = (  "", "Januar", "Februar", "März", "April", "Mai",
-                    "Juni", "Juli", "August", "September", "Oktober",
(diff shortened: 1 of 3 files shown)


Nick-Hall:
Thanks.

Please remove the po file.  Translation changes should be made in Weblate.

You also have some trailing whitespace.

cedecode:
I have re-pushed without the po file and removed some blanks.

cedecode:
Dear @Nick-Hall ,
could you please check if you can remove the bug label and merge?

Nick-Hall:
Thanks for the reminder. I don't see an issue with merging this now.

PR #1269, opened 2021-06-18
Display "[Missing surname]" for blank surname in peopletreemodel

Mattkmmr:
Persons where no surname was entered or the surname was entered into a wrong field, only an arrow is shown in the treeview person selector. This could confuse newbies, therefore a simple discription `[Missing surname]` would be helpful.

The problem came up in [#12333](https://gramps-project.org/bugs/view.php?id=12333)

Edit: Use `[Missing surname]` from preferences.no-surname-text instead of `[No surname]`

### DIFF
--- a/gramps/gui/views/treemodels/peoplemodel.py
+++ b/gramps/gui/views/treemodels/peoplemodel.py
@@ -84,6 +84,7 @@ COLUMN_TAGS = 18
 COLUMN_PRIV = 19
 
 invalid_date_format = config.get('preferences.invalid-date-format')
+no_surname = config.get("preferences.no-surname-text")
 
 #-------------------------------------------------------------------------
 #
@@ -614,7 +615,7 @@ class PersonTreeModel(PeopleBaseModel, TreeBaseModel):
         return [_('Group As'), _('Name')]
 
     def column_header(self, node):
-        return node.name
+        return node.name if node.name else no_surname
 
     def add_row(self, handle, data):
         """


PQYPLZXHGF:
Can it use the existing entry **Missing surname:**  on the "Text" preferences tab please?

> https://gramps-project.org/wiki/index.php/Gramps_5.1_Wiki_Manual_-_Settings#Text

EditPreferences-Text-tab-default-51

Or maybe add a new entry to cover this condition as I would like it to remain obviously empty, but can understand the need to help newbies 👍

gramps/gui/views/treemodels/peoplemodel.py
```diff
@@ -614,7 +614,7 @@ def get_tree_levels(self):
         return [_('Group As'), _('Name')]
 
     def column_header(self, node):
-        return node.name
+        return node.name if node.name != "" else "[%s]" % _("No surname")
```
Nick-Hall, line 617: You could just write `return node.name if node.name else no_surname` and define the `no_surname` variable at the top of the file with `no_surname = config.get("preferences.no-surname-text")`.

Mattkmmr:
> Counter point do you expand this to make it obvious that the **Given name** is missing also?

I think the given name is not that much of a problem, because even for a person without given name and surname, there are still the fields `gramps id` and `gender` displayed for each row/person.

lordemannd:
I like it.
One problem of the latest version with the  'config.get("preferences.no-surname-text")' change:
If the user edits the text string under Preferences -> Text -> Missing surname:
the changes wont take effect in this code until Gramps is restarted.

Is there some way to let peopletreemodel know the value has changed and needs to be updated?

kulath:
Where is the Default text for conditions actually used if it is not used in the treeview selector?

How will people know that a surname is missing if it is not shown as a blank?

PQYPLZXHGF:
> Where is the Default text for conditions actually used if it is not used in the treeview selector?

@kulath Git search of the code shows that **preferences.no-surname-text** appears [7 times ](https://github.com/gramps-project/gramps/search?q=preferences.no-surname-text)  ( 3 times in a Gramplet and 1 time in export for GenWeb ) 

Which is why I suggested adding a new entry covering this specfic condition as it is not a report or exporter and I would like it to remain empty in the treeview list?

> How will people know that a surname is missing if it is not shown as a blank?

An  alternative would be to have a number in brackets indicating how many blank surnames are listed ! 

Can tools tips be added that list where it is currently used?

By the way it has been 10 months since the last Gramps release any plans for a point update at least?

prculley:
> By the way it has been 10 months since the last Gramps release any plans for a point update at least?
Looks like about 33 real bug fixes and a lot more translations changes since 5.1.3.  Probably about time for 5.1.4.
Now if I could just find the time/energy...

Nick-Hall:
> Git search of the code shows that preferences.no-surname-text appears 7 times ( 3 times in a Gramplet and 1 time in export for GenWeb )

It is commented out in the exporter, and appears to be a bug in the Given Names Cloud gramplet.  It seems like an appropriate use for this PR.

Nick-Hall:
> Probably about time for 5.1.4.

I have time for a release.  I'll post on the _gramps-devel_ list to find a convenient date for everyone.

lordemannd:
aren't those text strings, set under Preferences -> Text, are used mostly in (3rd party?) reports and maybe in exports?

Nick-Hall:
_no-given-text_ and _no-record-text_ aren't used anywhere, although _no-given-text_ probably should be used in the Given Name Cloud gramplet.

_private-surname-text_ and _private-given-text_ are used in the living database proxy, so will be used widely in reports, exports and anything else that uses the privacy filter including addons.

_no-record-text_ is only used in the Descendants Lines third-party addon.

PR #1272, opened 2021-06-21
Narweb: Inconsistency between the name of the link

SNoiraud:
to a reference and the name of this reference for an individual.

Fixes #12336


SNoiraud:
We have now the following:
indiv-12

PR #1274, opened 2021-06-25
Fix for crash when changing views if part of toolbar is not shown

prculley:
because of a small screen when changing views.

Fixes #12048

Part of the Gtk code was deleting a toolbar object too soon; and another part of Gtk still had a reference.

This is an alternate (hopefully better) fix than https://github.com/gramps-project/gramps/pull/1165

### DIFF
--- a/gramps/gui/viewmanager.py
+++ b/gramps/gui/viewmanager.py
@@ -56,6 +56,7 @@ LOG = logging.getLogger(".")
 #-------------------------------------------------------------------------
 from gi.repository import Gtk
 from gi.repository import Gdk
+from gi.repository import GLib
 
 #-------------------------------------------------------------------------
 #
@@ -884,12 +885,17 @@ class ViewManager(CLIManager):
         while Gtk.events_pending():
             Gtk.main_iteration()
 
-        self.uimanager.update_menu()
+        # bug 12048 this avoids crash if part of toolbar in view is not shown
+        # because of a small screen when changing views.  Part of the Gtk code
+        # was deleting a toolbar object too soon; and another part of Gtk still
+        # had a reference.
+        def page_changer(self):
+            self.uimanager.update_menu()
+            self.active_page.change_page()
+            return False
 
-        while Gtk.events_pending():
-            Gtk.main_iteration()
-
-        self.active_page.change_page()
+        GLib.idle_add(page_changer, self,
+                      priority=GLib.PRIORITY_DEFAULT_IDLE - 10)
 
     def __delete_pages(self):
         """


jralls:
@prculley Works great and it's a lot cleaner. I'm still not keen on running the main loop in an event handler and it seems to work with that removed, i.e.
```
        def page_changer(self):
            self.uimanager.update_menu()
            self.active_page.change_page()
            return False
```

Nick-Hall:
Are we aiming to get this into the next release?

jralls:
It would be nice to, it's kinda dumb to do a release with a crash that we know how to fix.

Nick-Hall:
@prculley Should I make John's suggested change and merge?

emyoulation:
I have a question if this changes the sequence of operations when initializing a view. (Although this seems like a bug report, it is not. A report has been filed. I'd just like to know the order of operations ... so I can be a more effective observer of the GUI refresh.)

In the Windoze version, when changing view mode via the navigator, toolbar or menu, occasionally the contextual refresh of those GUI elements does not complete. (i.e., more than one button in the Navigator is highlighted, the toolbar bottom & menu items still have items appropriate to the previous view mode.) And once this happens, the user might as well restart Gramps... because the GUI contextual refresh becomes unusably sluggish. 

This happens FAR more frequently when processor intensive Gramplets are in (but not necessarily topmost) the sidebar or Bottombar. (I've noticed this most often with Deep Connections or Pedigree Gramplets active.)

It appears that the context of these GUI elements waits to refresh AFTER the view contents have been fully updated (tables read, sorted, filtered & displayed). 

Moreover, when the sluggishness occurs, the record list refresh can be affected. 

Specifically, I have observed the Family View focus not being updated after the Active Person has been changed (via direct selection in the Person View or by using ctrl-click on Session Log hotlinks in the Dashboard View). Yet an instance of the QuickView Gramplet (set to the Person biography addon quick report) in the Relationships bottombar DOES correctly show the updated Active Person.

What is the order of operations for the GUI?

jralls:
> What is the order of operations for the GUI?

@emyoulation it depends on the order of events queued on the GUI's event loop. Events come from user actions like key presses and mouse movements, from system requests like a redraw, or from programmatically queuing an event as in the case of this code. From the program's viewpoint events occur randomly there's no deterministic order of operations for the GUI as a whole.

Unfortunately it seems that Gramps's basic UI design doesn't update, it starts from scratch every time, rebuilding notebook pages from the GtkBuilder files. The bug this commit addresses is caused by doing that while handling an event on an element in the page so that the element is destroyed before the handler is finished with it, crashing Gramps when the handler tries to access the freed memory. This commit moves that rebuild outside of the handler by queuing a low-priority idle event for the redraw. My proposed change reduces the risk further by not iterating the mainloop in the handler so that there's no chance of conflicts.

 It's possible that what you've observed are manifestations of similar misbehavior and that this might help, but it's also possible that they're caused by something else. Only some quality time in the C debugger (i.e. gdb) will tell. Unfortunately MSYS2 strips the images and doesn't provide debug symbol patches so that's quite difficult on Windows.

lordemannd:
tested on Windows and debian/linux, seems working without causing problems,
BUT on debian, the down arrow leads to the extended menu doesn't appear until the toolbar gets a mouse over.
debian:
* Python 3.9.2 (Passed: version 3.3.0 or greater installed.)
* GTK+ 3.24.24 (Passed: version 3.12.0 or greater installed.)
* PyGObject 3.38.0 (Passed: version 3.12.0 or greater installed.)
* Cairo 1.16.0 (Passed: version 1.13.1 or greater installed.)
* Pycairo 1.16.2 (Passed: version 1.13.3 or greater installed.)
* Pango 1.46.2 (Passed: version 1.29.3 or greater installed.)
* PangoCairo 1.0

jralls:
@lordemannd does @prculley's version display the menu arrow without the mouseover? If it does can you try with just a single iteration, no `while` loop?

lordemannd:
@jralls, I just took the current contents of the changed file (gramps/gui/viewmanager.py) and used it to replace the one from the distribution. (I think that is prculley's). and the screen shot was taken on debian/linux.

on debian, I need to mouseover or bring another window in front of Gramps to get the down arrow to show up.

I'm not sure which while loop you mean, can you give a line number or screen shot?

lordemannd:
@jralls, you can send me a file to use, if that is easier.
To be clear, on debian, the down arrow shows up if I mouse over any location on the toolbar, or cover any part of the Gramps window (forcing it to be redrawn?).

lordemannd:
on more clarification: I also have PR1274 included, do you think it would have an effect here?

lordemannd:
on more clarification: I also have both PR1724 and PR1273 included, do you think it would have an effect here?

jralls:
@lordemannd wrote:
> I just took the current contents of the changed file (gramps/gui/viewmanager.py) and used it to replace the one from the distribution. 

Ah, OK, then it's with the loop in place. I'll fire up my Debian-10 VM and see if I can figure out why it needs a kick to draw the arrow.

jralls:
@lordemannd, can't reproduce:

https://user-images.githubusercontent.com/58910/126049777-7f0a1085-4970-4ca5-96b3-ee5b6cccb7d3.mov

I cloned gramps, checked out v5.1.3 and cherry-picked this PR's commit onto that, then built Gramps. Applied my proposed change, no difference. That change is
```diff
diff --git a/gramps/gui/viewmanager.py b/gramps/gui/viewmanager.py
index 6ba977ce3..ef906f85b 100644
--- a/gramps/gui/viewmanager.py
+++ b/gramps/gui/viewmanager.py
@@ -890,10 +890,6 @@ class ViewManager(CLIManager):
         # had a reference.
         def page_changer(self):
             self.uimanager.update_menu()
-
-            while Gtk.events_pending():
-                Gtk.main_iteration()
-
             self.active_page.change_page()
             return False
 ```

jralls:
@lordemannd Same result on Debian Sid. I tried to replicate your approach more exactly but copying viewmanager.py from this PR to /usr/lib/python3/dist-packages/gramps/gui/ produced an attribute error, 'DashboardView' has no attribute 'post_create'. Copying dashboardview.py didn't fix that, but ISTM you're not starting from quite 5.1.3.

I tried a different tack, making a patch from the commit and applying that to /usr/lib/python3/dist-packages/gramps/gui/viewmanager.py. That doesn't crash and it displays the toolbar menu arrow immediately.

lordemannd:
Screenshot_withOUT_arrow
Screenshot_with_arrow

I think you need to test the spirit of bug #0012048,
compress your screen so the toolbar has to be extended by the drop down menu. 

Once you can see the down arrow in the toolbar, then use the drop down menu to change to the alternate view:
either "Grouped People" or just "People"
For me when the view changes, the Arrow is absent until a mouseover or other redraw of the screen.

(I tried with your deleted lines, and it works the same)
BTW: my debian: uname -a produces "5.10.0-8-amd64 #1 SMP Debian 5.10.46-1 (2021-06-24) x86_64 GNU/Linux"

lordemannd:
I turned on the text titles, to force the issue on the overflow of the toolbar.

jralls:
@lordemannd Ah, I missed that step. Yes, now I can reproduce the missing arrow after using the menu--both with and without this PR, with git gramps51 HEAD, git v5.1.3, and with the stock 5.1.3 from Debian. It's a separate bug.

lordemannd:
I wonder, it isn't visible on Windows, can you see it on MacOS? or it is just Linux?

lordemannd:
It isn't visible on Windows, can you see it on MacOS? or it is just Linux?

Will you create a new bug report, or should I take a stab at it?

lordemannd:
It isn't visible on Windows, can you see it on MacOS? or it is just Linux?

Will you create a new bug report, or should I take a stab at it?

This could be related  to issue #9998, which was on Windows, and @prculley  created a Partial fix in https://github.com/gramps-project/gramps/pull/369. (applied in 5.0.0-alpha2)

jralls:
Happens on macOS too.

Go for it! ;-)

lordemannd:
I have created separate bug (#12358) for the minor "toolbar overflow indicator (down arrow) is not shown" issue. 
I agree it seems to be independent of this change and shouldn't hold up the changes covered in this PR.
The issue covered by this PR was only reported against macOS, 
but I have tested it on Windows and debian/linux didn't see any bad behavior.
(only the believed to be independent #12358)

prculley:
The toolbar overflow indicator issue has been around for a while.  I'm pretty sure it is independent of this PR issue.

I'm OK with John Ralls latest tweak, removing the Gtk.events pending.  I did some testing and it doesn't seem to be causing any problems (although I thought it was for a moment, only to discover another small issue).  I updated the PR to deal with this.

PR #1278, opened 2021-07-10
Place editor, copy and paste of lat and long text

SNoiraud:
no longer auto-populating latitude and longitude fields

Fixes #012349

### DIFF
--- a/gramps/gui/editors/editplace.py
+++ b/gramps/gui/editors/editplace.py
@@ -188,7 +188,7 @@ class EditPlace(EditPrimary):
     def set_latlongitude(self, value):
         try:
             # Bug 12349
-            parts = value.index(', ')
+            parts = value.split(', ')
             if len(parts) == 2:
                 longitude = parts[0].strip().replace(',', '.')
                 latitude = parts[1].strip().replace(',', '.')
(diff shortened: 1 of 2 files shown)


gramps/gui/editors/editplace.py
```diff
@@ -187,9 +187,14 @@ def _setup_fields(self):
...
     def set_latlongitude(self, value):
         try:
-            coma = value.index(', ')
-            longitude = value[coma+2:].strip().replace(',','.')
-            latitude = value[:coma].strip().replace(',','.')
+            # Bug 12349
+            parts = value.split(', ')
+            if len(parts) == 2:
+                longitude = parts[0].strip().replace(',', '.')
+                latitude = parts[1].strip().replace(',', '.')
```
zertrin, line 194: This is reversed! Latitude is the first field and longitude the second one according to the label above the text input.

See the result for a village in Belgium:

image

emyoulation: A patch has been posted. 

See 
https://gramps.discourse.group/t/5-1-4-bug-in-gps-location-dialog/1731

zertrin: Yes only saw the existing patch #1290 after submitting mine #1320 😄 

Sorry for the noise.

emyoulation: It is noise in the right place if it helps the next person looking here.

PR #1280, opened 2021-07-11
Narweb: some fixes to alphabet_navigation

kulath:
Fixes [#12350](https://gramps-project.org/bugs/view.php?id=12350) Narweb: some issues with alphabet_navigation

Ensure correct index letters according to normal indexing conventions by
using ICU AlphabeticIndex.

- Add ICU AlphabeticIndex
- Modify modules that use alphabet_navigation to use AlphabeticIndex

If Alphabetic index is not available, fix some other issues.

- Implement a local AlphabeticIndex for when ICU AlphabeticIndex is not
available, using existing code algorithms
- Fix Serbian contraction DŽ not being recognised
- Fix contractions not getting a separate entry in the
alphabet_navigation
- Ensure that import from either icu or PyICU works
- Fix primary_difference to set collation strength to PRIMARY
- Fix contractions lookup to use locale for current report
- Fix get_first_letter to pass current report locale


kulath:
Bugs in the current code are noted under the fixes for when ICU is not available. In addition in the current code, for example, if the only index entry starting with A is A with circle above, the this will always be indexed under A with circle above, even in languages where it should be indexed under A.

SNoiraud:
Did you try it with multiple languages ?
Did you test all the css files ?
I see some problems. 
For example, I have the following:
narrativeweb 1

If I make a diff between my version and yours, I have for a site with 6 languages 2 500 000 lines!
I look at this.

SNoiraud:
ignore what I said before. The snapshot was what we had before your patch in some languages.
It is now correct.
I have some differences in backward references:
narrativeweb 2
narrativeweb 3

SNoiraud:
OK. references are sorted by role.
Reed sarah and McCoy francis have a Primary role (0)
You have the family with a family role (1)
and you have Reed patrick who has another role (3)

This seems correct.

SNoiraud (approved):
It seems correct for me.

Nick-Hall:
@prculley Do you wish to review this?

prculley:
In my limited testing, I did not see any issues.  But I don't have real good test cases with lots of international characters and contractions to test with.
As I expect the 5.2 to be at least a beta release, I'd go ahead and merge.

kulath:
> In my limited testing, I did not see any issues. But I don't have real good test cases with lots of international characters and contractions to test with.

I did try it with other languages, but to accommodate names in foreign languages relying on the user's language is not much use. So the code cycles through the available scripts to create index entries for each character in any language. This showed that I needed to increase the number of possible buckets, which was what the code does.

The standard Gramps test data has names in multiple languages, so that shows the index entries for many languages. In addition, I created some test data with contractions etc., to check that worked too.

PR #1285, opened 2021-07-23
probably-alive incorrect if death without date

SNoiraud:
Fixes [#12364](https://gramps-project.org/bugs/view.php?id=12364)


lordemannd:
Tested and working on both Windows and debian.

I though that Windows was working without this change, but additional testing should that Windows does have the issue.

SNoiraud:
I tested with an invalid date like "unknown date".

Nick-Hall:
This looks like a reasonable fix.  It is impossible to estimate a birth date from a text-only death date.  People with a death event should be regarded as not alive.

As three people have tested this, I'll merge it ready for the next release.

PR #1287, opened 2021-07-26
Fix error when trying to close name editor during long name group mapping view rebuild.

prculley:
Fixes #12328

This problem occurs when saving a setting for a name group mapping in large dbs, when user tries to close the dialog while the person view is rebuilding.  With large dbs, this can take a while, and the user may try to interact with the dialog, since it remains present, even after the "save" or close buttons are pushed.  The GUI responds (due Gtk to progress bar wait for events), and will reentrently close the dialog again.  Since closing deletes some necessary variables, we get a crash when the original code continues executing after the view rebuild.

This fix corrects this specific issue by delaying the name group mapping and subsequent view rebuild, until the dialog is fully closed.

Note: rebuilding the view from within an editor like this has some other potential issues, so there may be other bugs lurking.

### DIFF
--- a/gramps/gui/editors/editname.py
+++ b/gramps/gui/editors/editname.py
@@ -366,6 +366,7 @@ class EditName(EditSecondary):
             5/     local set, not global set --> set (change local)
             6/     local set,     global set --> set (set to global if possible)
         """
+        ngm = False  # name group mapping setting
         closeit = True
         surname = self.obj.get_primary_surname().get_surname()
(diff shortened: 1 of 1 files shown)


lordemannd:
PR 1287 & PR 1288, work for me on debian/linux and Windows

lordemannd:
I notice that the progress bar in lower left seems present all the time. wasn't there in 5.1.3v2. I don't see how these PRs could cause it. I will try to isolate it.

PR #1288, opened 2021-07-26
Fix exception when removing a group name in Sqlite db when group name is already missing.

prculley:
Fixes #12367


lordemannd:
PR 1287 & PR 1288, work for me on debian/linux and Windows

PR #1291, opened 2021-08-07
Group As Override is ignored when the Surname is

SNoiraud:
given the Origin Patronymic or Matronymic

Fixes [#12395](https://gramps-project.org/bugs/view.php?id=12395)

discourse:
https://gramps.discourse.group/t/patronymic-and-matronymic-name/1684/5

### DIFF
--- a/gramps/gen/display/name.py
+++ b/gramps/gen/display/name.py
@@ -1007,8 +1007,16 @@ class NameDisplay:
         """
         if pn[_GROUP]:
             return pn[_GROUP]
-        return db.get_name_group_mapping(_raw_primary_surname_only(
+        name = pn[_GROUP]
+        if name == "":
+            # if we have no prinary surname, perhaps we have a
+            # patronymic/matronynic name ?
+            name = db.get_name_group_mapping(_raw_patro_surname_only(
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Have you tried checking the "Consider single pa/matronymic as surname" option in the "Display" tab in the preferences?

gramps/gen/display/name.py
```diff
@@ -1007,8 +1007,16 @@ def name_grouping_data(self, db, pn):
         """
         if pn[_GROUP]:
             return pn[_GROUP]
-        return db.get_name_group_mapping(_raw_primary_surname_only(
+        name = pn[_GROUP]
```
Nick-Hall, line 1010: If you are modifying the `name_grouping_data` method you should probably make similar changes to the  `name_grouping_name` method.

SNoiraud: If I look at where the name_grouping_name method is used, I only see it's called by the name_grouping method.
This last method is never called either in gramps or in addons.
Where could we call this method?
I think the two method name_grouping_name and name_grouping could be removed.

PR #1296, opened 2021-09-04
Solve InterpolationSyntaxError if "%" in a string

SNoiraud:
The grampletpane module save data in a config file for all the gramplets added in the dashboard.
The python configparser module doesn't like if we have a "%" character in a string.
    
Fixes #012423


QuLogic:
There are a lot of extraneous changes here, making it difficult to determine what the fix actually is.

wroldwiedbwe:
@QuLogic Check the first commit https://github.com/SNoiraud/gramps/commit/e8799a7fea1a5bd3276589393ad8db11bbb26c64

Nick-Hall:
@SNoiraud This causes a unit test to fail.

PR #1298, opened 2021-09-18
Strange behavior for the scrollbar in the bottombar

SNoiraud:
Fixes [#12438](https://gramps-project.org/bugs/view.php?id=12438)

### DIFF
--- a/gramps/gui/views/pageview.py
+++ b/gramps/gui/views/pageview.py
@@ -165,7 +165,7 @@ class PageView(DbGUIElement, metaclass=ABCMeta):
         self.widget.show_all()
         self.widget.set_name('view')
         self.vpane.pack1(self.widget, resize=True, shrink=False)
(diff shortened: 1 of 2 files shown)


Nick-Hall:
Why remove the default size?  When starting Gramps from a fresh installation, this will result in a very small bottombar beacuse all the default gramplets have scrollbars.

Nick-Hall:
OK.  I have just been testing this.  The problem is that the widget never resizes below its default size, so when the slider reduces the size of the pane, the widget is cropped rather than displaying the scrollbar.

Removing the default size will result in a bottombar that is too small for new users, but this is probably better than leaving the bug unfixed.  Ideally we should use the natural size of the child widget as we did for the sidebar in the [set_page_slider](https://github.com/gramps-project/gramps/blob/master/gramps/gui/views/pageview.py#L175) method.

You other change setting `shrink=False` is unrelated, but I have no objections.  It makes the bottombar behaviour consistent with the sidebar.

PR #1300, opened 2021-10-07
Add new Indonesian translation

Nick-Hall:
I have already added an Indonesian translation on Weblate in response to a request.  We can merge this once the translation has been started.

@jralls @prculley How do I find the correct locale code for Indonesian?  Should it be "Indonesian" or "Indonesian_Indonesia"?

### DIFF
--- a/gramps/gen/utils/grampslocale.py
+++ b/gramps/gen/utils/grampslocale.py
@@ -87,6 +87,7 @@ _LOCALE_NAMES = {
     'he': ('Hebrew_Israel', '1255', _("Hebrew")),
     'hr': ('Croatian_Croatia', '1250', _("Croatian")),
     'hu': ('Hungarian_Hungary', '1250', _("Hungarian")),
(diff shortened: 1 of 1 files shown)


Nick-Hall:
We have one translator working on this, but the progress is slow.  Only about 1% of strings have been translated so far.

I'll put an entry in the `_LOCALE_NAMES` list, but won't make the translation active yet.

emyoulation:
Can you make the Irish inactive too? We don't have a translator list and it's been sitting at zero for quite a long while.

Nick-Hall:
@emyoulation Irish is already inactive.

emyoulation:
That's odd. It still shows on the Weblate bar chart progress gadget. 


Is there an extra step that is required to stop Weblate from reporting it there?

Nick-Hall:
Weblate widgets include all languages with translations.  Read the [widgets](https://hosted.weblate.org/widgets/gramps-project/) page for details.

PR #1301, opened 2021-10-10
Remove rounding when formatting a span

DaveSch-gramps:
Removes the +1 when the Age precision is set to 1 (Years) and the month is >= 6

Bug [#12422](https://gramps-project.org/bugs/view.php?id=12422)

### DIFF
--- a/gramps/gen/lib/date.py
+++ b/gramps/gen/lib/date.py
@@ -429,11 +429,6 @@ class Span:
                               ).format(number_of=diff_tuple[0])
             detail += 1
         if self.precision == detail:
-            if diff_tuple[1] >= 6: # round up years
-                # Translators: leave all/any {...} untranslated
-                retval = ngettext("{number_of} year", "{number_of} years",
-                                  diff_tuple[0] + 1
-                                 ).format(number_of=diff_tuple[0] + 1)
             return retval
         if diff_tuple[1] != 0:
             if retval != "":


Nick-Hall:
This proposed fix reverts the change made for bug [#8765](https://gramps-project.org/bugs/view.php?id=8765).

The issue is that when calculating the length of a span, "about 1900" is taken as "1 Jan 1900" instead of a mid-point.  This leads to the span "May 1900" to "about 1901" being displayed as "0 years" instead of "1 year" which may appear more reasonable.

DaveSch-gramps:
I just did a test

Birth = "May 1900" 
Death = "about 1901"

about defined with default 50 years
PR 1301 NOT applied

Years = about 8 months
Years, Months = about 8 months
Years, Months, Days = about 8 months


PR 1301 applied

Years = about 8 months
Years, Months = about 8 months
Years, Months, Days = about 8 months

With PR 1301 still applied

Birth = "May 1900" 
Death = "about 1950"

Years = about 49 years

All correct calculations

But going back to to my exact date scenario, with PR1301 not applied

Date of Birth: 1909-12-12
Date of Death: 1993-11-03

Age: 83y, 10m, 22d
Age: 83y, 10m
Age: 84y

My relative was not 84 when she died.

Nick-Hall:
The argument made in the original bug report ([#8765](https://gramps-project.org/bugs/view.php?id=8765)) was that an incomplete date such as "1900" was on average more likely to represent an event happening nearer to 1 Jul 1900 than 1 Jan 1900.

The fix caused other issues though.  With two exact dates the age is rounded up which is unexpected in some countries.  In England, using the example above, we would refer to a person as 83 years old even if they had lived for 83y, 10m, 22d.

However, this may not be the case in all countries, so perhaps we should make this an option as @SNoiraud suggests in bug report [#12422](https://gramps-project.org/bugs/view.php?id=12422).

DaveSch-gramps:
Personally, I use Year, Month, Day as my setting but I did notice it was rounding up when I first migrated to Gramps. Odd that the setting in Preferences is "Age display precision:"

I have no strong opinion that the code needs to be changed. If you (and/or others) feel this pull should be closed, I will not be disappointed.

Nick-Hall:
In general, rounding a span seems like a good idea.  When the span represents an age, then rounding may not be appropriate and perhaps should be made optional.

I'll leave this open for now and think about it further.

DaveSch-gramps:
Closing PR in favor of @SNoiraud 's PR

https://github.com/gramps-project/gramps/pull/1615

PR #1302, opened 2021-10-15
Make all the tabs columns resizable

SNoiraud:
Fixes #08767

This was related to https://gramps.discourse.group/t/editor-dialogs-not-remembering-table-spacing/1871


DaveSch-gramps:
Did some test and most is working as expected.

Could not test the locationembedlist. This is a legacy tab that only displays after the place upgrade. (Had to search to find out what it was.)

The two LDS embedlists are linked together. Change the person lds and it shows for the family lds tab. and the other way, change family and it changes the person.

The person attribute works fine. None of the other attribute tabs retain any change in the column widths.

The person Source/Citation works fine. None of the other Source/Citation tabs are remembering any changes in the column widths. 

Note: I know this is not part of this pull, but with the new columns in citationembedlist, column 2 is selected to display

line 124
    def column_order(self):
        """
        Return the column order of the columns in the display tab.
        """
        return ((1, 0), (1, 1), (1, 5), (1, 2), (1, 3), (1, 6), (1, 4), (1, 7))

But no information is defined for 2. Should be Date ?? (line 82)

And the Privacy lock which is normally displayed as far left item is now far right.

SNoiraud:
> But no information is defined for 2. Should be Date ?? (line 82)

I don't care.

> And the Privacy lock which is normally displayed as far left item is now far right.

I didn't change the order of the columns. If you're talking about the event tab, it's always been on the right side.

> The two LDS embedlists are linked together. Change the person lds and it shows for the family lds tab. and the other way, change family and it changes the person.

Do you want they have different sizes ? They use the same model. So it will be difficult to change. I look at this.

DaveSch-gramps:
Like the two LDS tabs, all attribute column settings are referencing attrembedlist so are using the embeddedlist.attr setting regardless of where the Attribute tab is located.  Each tab cannot have their own column settings.

SNoiraud:
This PR will be removed.

Nick-Hall:
@SNoiraud Why was this closed?

Nick-Hall:
OK.  Thanks for closing this.  It will stop me looking at it by mistake.

SNoiraud:
If I remember, this was a first try. The real PR was the #1306.
perhaps I made a mistake. I look at this tomorrow.
...
This is OK. #1306 is the good PR. This PR is useless.

PR #1305, opened 2021-10-19
Update citationembedlist.py

DaveSch-gramps:
Fix date column and move Privacy lock to first Item


DaveSch-gramps:
Fixes columns from https://github.com/gramps-project/gramps/pull/1180

PR #1306, opened 2021-10-30
Make the tabs columns resizable: ResizableTreeView

SNoiraud:
Fixes [#8767](https://gramps-project.org/bugs/view.php?id=8767)

This was related to https://gramps.discourse.group/t/editor-dialogs-not-remembering-table-spacing/1871
and replace the following PR: #1302

### DIFF
--- a/gramps/gui/editors/displaytabs/embeddedlist.py
+++ b/gramps/gui/editors/displaytabs/embeddedlist.py
@@ -43,6 +43,7 @@ from gi.repository import Pango
 #
 #-------------------------------------------------------------------------
 from ...widgets.cellrenderertextedit import CellRendererTextEdit
+from ...widgets.persistenttreeview import PersistentTreeView
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
 from ...utils import is_right_click
@@ -388,7 +389,8 @@ class EmbeddedList(ButtonTab):
         # create the tree, turn on rule hinting and connect the
         # button press to the double click function.
 
-        self.tree = Gtk.TreeView()
+        _name = self.get_config_name()
+        self.tree = PersistentTreeView(self.uistate, _name)
         self.tree.set_reorderable(True)
         self.tree.connect('button_press_event', self.double_click)
         self.tree.connect('key_press_event', self.key_pressed)
@@ -533,6 +535,12 @@ class EmbeddedList(ButtonTab):
             self.columns.append(column)
             self.tree.append_column(column)
         self.track_ref_for_deletion("columns")
+        self.tree.restore_column_size()
+
+    def get_config_name(self):
+        """ used to associate the selector config name to the PersistentTreeView
+        """
+        assert False, "Must be defined in the subclass"
 
     def icon_func(self, column, renderer, model, iter_, col_num):
         '''
--- a/gramps/gui/editors/displaytabs/eventattrembedlist.py
+++ b/gramps/gui/editors/displaytabs/eventattrembedlist.py
@@ -41,3 +41,6 @@ class EventAttrEmbedList(AttrEmbedList):
 
     def get_user_values(self):
         return self.dbstate.db.get_event_attribute_types()
+
+    def get_config_name(self):
+        return __name__
--- a/gramps/gui/clipboard.py
+++ b/gramps/gui/clipboard.py
@@ -59,6 +59,7 @@ from .ddtargets import DdTargets
 from .makefilter import make_filter
 from .utils import is_right_click, no_match_primary_mask
 from gramps.gen.const import GRAMPS_LOCALE as glocale
+from gramps.gui.widgets.persistenttreeview import PersistentTreeView
 _ = glocale.translation.sgettext
 
 #-------------------------------------------------------------------------
@@ -1412,6 +1413,7 @@ class ClipboardWindow(ManagedWindow):
         self.db.connect('database-changed',
                         lambda x: ClipboardWindow.otree.clear())
 
+        mtv.restore_column_size()
         self.show()
 
     def build_menu_names(self, obj):
@@ -1453,7 +1455,7 @@ class ClipboardWindow(ManagedWindow):
 # MultiTreeView class
 #
 #-------------------------------------------------------------------------
-class MultiTreeView(Gtk.TreeView):
+class MultiTreeView(PersistentTreeView):
     '''
     TreeView that captures mouse events to make drag and drop work properly
     '''
@@ -1461,7 +1463,7 @@ class MultiTreeView(Gtk.TreeView):
         self.dbstate = dbstate
         self.uistate = uistate
         self.title = title if title else _("Clipboard")
-        Gtk.TreeView.__init__(self)
+        PersistentTreeView.__init__(self, self.uistate, "clipboard")
         self.connect('button_press_event', self.on_button_press)
         self.connect('button_release_event', self.on_button_release)
         self.connect('drag-end', self.on_drag_end)
--- a/gramps/gui/columnorder.py
+++ b/gramps/gui/columnorder.py
@@ -63,7 +63,8 @@ class ColumnOrder(Gtk.Box):
     Column ordering selection widget
     """
 
-    def __init__(self, config, column_names, widths, on_apply, tree=False):
+    def __init__(self, config, column_names, widths, on_apply, tree=False,
+                 resizable=None):
         """
         Create the Column Ordering widget based on config
 
@@ -80,6 +81,7 @@ class ColumnOrder(Gtk.Box):
         self.colnames = column_names
         self.config = config
         self.on_apply = on_apply
+        self.resizable = resizable
 
         self.pack_start(Gtk.Label(label=' '), False, False, 0)
 
@@ -137,7 +139,23 @@ class ColumnOrder(Gtk.Box):
 
         #obtain the columns from config file
         self.oldorder = self.config.get('columns.rank')
-        self.oldsize = self.config.get('columns.size')
+        oldsize = self.config.get('columns.size')
+        # try to use the current columns size.
+        if resizable:
+            self.oldsize = []
+            newsize = self.resizable.get_columns_size()
+            nbc = 0
+            for size in oldsize:
+                if nbc < len(newsize):
+                    if newsize[nbc] == 0:
+                        self.oldsize.append(size)
+                    else:
+                        self.oldsize.append(newsize[nbc])
+                else:
+                    self.oldsize.append(size)
+                nbc += 1
+        else:
+            self.oldsize = oldsize
         self.oldvis = self.config.get('columns.visible')
         colord = []
         index = 0
@@ -185,6 +203,7 @@ class ColumnOrder(Gtk.Box):
             self.config.set('columns.visible', newvis)
             self.config.save()
             self.on_apply()
+        self.resizable.save_column_info()
 
 def toggled(cell, path, model):
     """
--- a/gramps/gui/dbman.py
+++ b/gramps/gui/dbman.py
@@ -69,6 +69,7 @@ from .listmodel import ListModel
 from gramps.gen.constfunc import win
 from gramps.gen.plug import BasePluginManager
 from gramps.gen.const import GRAMPS_LOCALE as glocale
+from gramps.gui.widgets.persistenttreeview import PersistentTreeView
 _ = glocale.translation.gettext
 
 #-------------------------------------------------------------------------
@@ -170,7 +171,7 @@ class DbManager(CLIDbManager, ManagedWindow):
         self.viewmanager = viewmanager
 
         for attr in ['connect_btn', 'cancel_btn', 'new_btn', 'remove_btn',
-                     'info_btn', 'dblist', 'rename_btn', 'convert_btn',
+                     'info_btn', 'rename_btn', 'convert_btn',
                      'repair_btn', 'rcs_btn', 'msg', 'close_btn']:
             setattr(self, attr, self.glade.get_object(attr))
 
@@ -179,6 +180,11 @@ class DbManager(CLIDbManager, ManagedWindow):
         self.lock_file = None
         self.data_to_delete = None
 
+        objectlist = self.glade.get_object('dblist')
+        self.dblist = PersistentTreeView(uistate, "dbman")
+        scrolledwindow = self.glade.get_object('scrolledwindow88')
+        scrolledwindow.remove(objectlist)
+        scrolledwindow.add(self.dblist)
         self.selection = self.dblist.get_selection()
 
         # For already loaded database:
@@ -410,6 +416,7 @@ class DbManager(CLIDbManager, ManagedWindow):
         column = Gtk.TreeViewColumn(_('Last accessed'), render, text=DATE_COL)
         column.set_sort_column_id(DSORT_COL)
         self.dblist.append_column(column)
+        self.dblist.restore_column_size()
 
     def __populate(self):
         """
--- a/gramps/gui/editors/displaytabs/addrembedlist.py
+++ b/gramps/gui/editors/displaytabs/addrembedlist.py
@@ -148,3 +148,6 @@ class AddrEmbedList(EmbeddedList):
         Called to update the screen when the address changes
         """
         self.rebuild()
+
+    def get_config_name(self):
+        return __name__
--- a/gramps/gui/editors/displaytabs/attrembedlist.py
+++ b/gramps/gui/editors/displaytabs/attrembedlist.py
@@ -123,3 +123,6 @@ class AttrEmbedList(EmbeddedList):
     def edit_callback(self, name):
         self.changed = True
         self.rebuild()
+
+    def get_config_name(self):
+        return __name__
--- a/gramps/gui/editors/displaytabs/backreflist.py
+++ b/gramps/gui/editors/displaytabs/backreflist.py
@@ -143,3 +143,6 @@ class BackRefList(EmbeddedList):
     def edit_button_clicked(self, obj):
         (reftype, ref) = self.find_node()
         edit_object(self.dbstate, self.uistate, reftype, ref)
+
+    def get_config_name(self):
+        return __name__
--- a/gramps/gui/editors/displaytabs/citationembedlist.py
+++ b/gramps/gui/editors/displaytabs/citationembedlist.py
@@ -283,3 +283,6 @@ class CitationEmbedList(EmbeddedList, DbGUIElement):
                                   parent=self.uistate.window)
             else:
                 raise ValueError("selection must be either source or citation")
+
+    def get_config_name(self):
+        return __name__
--- a/gramps/gui/editors/displaytabs/eventembedlist.py
+++ b/gramps/gui/editors/displaytabs/eventembedlist.py
@@ -419,3 +419,6 @@ class EventEmbedList(DbGUIElement, GroupEmbeddedList):
         self.tree.expand_all()
         if prebuildpath is not None:
             self.selection.select_path(prebuildpath)
+
+    def get_config_name(self):
+        return __name__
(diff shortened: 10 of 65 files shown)


Nick-Hall:
Yes, this seems like a good design.

Perhaps in the future we could store other information such as the column order and sort column.  If we want to go in this direction then `PersistentTreeView` may be a better name than `ResizableTreeView`.

Have you thought about passing a config object and key name in the constructor?

gramps/gui/widgets/resizabletreeview.py
```diff
@@ -0,0 +1,97 @@
...
+
+    __gtype_name__ = 'ResizableTreeView'
+
+    def __init__(self):
+        """
+        Create a TreeView widgets with column size saving
+        """
+        Gtk.TreeView.__init__(self)
+        self.config_name = "undefined"
+        self.connect("destroy", self.save_column_info)
```
Nick-Hall, line 50: Is this equivalent to using the `__del__` method?

gramps/gui/widgets/resizabletreeview.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def set_config_name(self, name):
+        """
+        parameter: name: The associated config name string for the treeview
+        This parameter must be unique as it allow the tab columns size saving
+        """
+        self.config_name = name
+
+    def save_column_info(self, tree=None):
```
Nick-Hall, line 59: The `tree` parameter isn't used.

emyoulation:
While working on a User problem with surname not grouping correctly, they posted a [screen capture to Facebook](https://www.facebook.com/groups/370692133138445/permalink/1820094838198160/).

In the Netherlands translation of the Person Editor, note that the "Personal 1" section label in the Events table is cropped with an ellipsis.

Will this modification allow foreign language users to customize to compensate?

gramps/gui/widgets/resizabletreeview.py
SNoiraud: Yes, I know, but when the destroy signal is send, we call that method with the TreeView as parameter.

gramps/gui/widgets/resizabletreeview.py
SNoiraud: This doesn't always work. I have some cases for which the __del__ method is never called.
I have this case for all selectors. Is it related to glade ?
For the embedded list, it works.
For the selectors, the __del__ method is called when we quit gramps. It is too late
For the listviews, this method is never called.

SNoiraud:
> Have you thought about passing a config object and key name in the constructor?

Yes, but I don't know how to do that with glade.

SNoiraud:
> Will this modification allow foreign language users to customize to compensate?

Yes, this is the primary purpose af such this implementation.
Once you resize the column, this new size will be keeped between gramps session.

emyoulation:
> Will this modification allow foreign language users to customize to compensate?\n\nYes, this is the primary purpose as such this implementation.\nOnce you resize the column, this new size will be kept between gramps session.

With 56 files being changed, is there a recommendation for how to encourage translator beta testers to try this design and give feedback? I doubt many are GitHub savvy. (Although maybe the lead developers know of a few multiple language developers who could be enlisted to test it?)

Polyglots might find some issues... Like how to share default layout widths for each of their installed languages. Or if something odd happens when reconfiguring the language selection.

SNoiraud:
> With 56 files being changed, is there a recommendation for how to encourage translator beta testers to try this design and give feedback?

These changes only affect the size of the columns. Just test only in a language other than English. I did my tests in French so I know that I haven't encountered any problems so far.

emyoulation:
Glad to hear that!  

Did you test changing Theme settings? Are the customized column scaled when you choose a larger font point size? Or does it scale the default width? And what happens on a font family change?

SNoiraud:
> Are the customized column scaled when you choose a larger font point size?

The width of the columns width is independent of the font.
If we want to associate the size of the column to the font size, it is a much more complex problem.

Do your font size very different ?
The better would be to set your columns size with your larger font and then use your smaller font.

emyoulation:
Some things seems to scale dynamically with the font. But that be a misperception because the row height definitely scales.

I probably won't be using this feature because it would be too many steps to reset before doing screen captures to illustrate the wiki. 

(I should start using the PortableGramps for the wiki screen captures. There are too many features that I avoid because of that.)

SNoiraud:
If we change the font or the font size in the Themes plugin, the column width is adapted to this change.

DaveSch-gramps:
Been putting the new pull through its paces. One issue has shown itself.

Setting the shared media attribute works. When using the referenced media attribute, it first takes on the column width set by shared. But after leaving the referenced attribute tab it reverts to column widths 1 and resets the shared attribute tab columns.

It seems to be isolated to this tab only in the media records.

SNoiraud:
I can't reproduce this. Can you attach a snapshot to see exactly what you get ?

DaveSch-gramps:
Steps to reproduce

Step 1: add media

mediaattr1
ia

Step 2 Attach to Family gallery

mediaattr2

Step 3 share with individual, tag individual. Add attribute

mediaattr3

Step 4 After saving, notice changed spacing in both attribute tabs

mediaattr4a

mediaattr4b

Hope this helps

SNoiraud:
I can't reproduce your problem. Can you attach your gramps.ini configuration file or copy only the [spacing] section ?

DaveSch-gramps:
[gramps.txt](https://github.com/gramps-project/gramps/files/7491177/gramps.txt)

DaveSch-gramps:
The Surname columns (Name edit window) are not being saved once the person's record is saved.

SNoiraud:
Effectively, you have the following:
...
emb_mediaattr=[1, 1]
...
emb_surname=[1, 1, 1, 1]
...

If you remove them, can you reproduce your problem.
I look at if I can reproduce that.

DaveSch-gramps:
Manually reset to default settings
```
;;emb_mediaattr=[]
;;emb_surname=[]
```
The media attribute immediately changed to 

`emb_mediaattr=[1, 1]`

after accessing the person's referenced attribute tab.

For the surname list in the name edit window,  the column spacing was normal when initially accessed and normal again when re-accessing the edit window. 

`emb_surname=[10, 32, 7, 10]`

Once the person's record was saved and a new person edit window and name edit was opened it returned this spacing.

`emb_surname=[1, 46, 1, 10]`

Once Gramps was closed, it returned this spacing.

`emb_surname=[1, 1 1, 1]`

Relaunching Gramps and going back to a name edit window the spacing reverts to 

`emb_surname=[1, 46, 1, 10]`

Note: at no time during these test did I attempt to adjust column widths.

Resizing the surname columns sets the column spacing until a new edit windows are launched which sets them to 

`emb_surname=[1, 1 1, 1]`

DaveSch-gramps:
Did some more testing.

I made a clean install of GrampsAIO-5.1.4-1_win64 to C;\ (to keep previous install with my modifications.)

Deleted the gramps51 version folder from the user directory.

Made the changes (manually) from PR1306 to the newly installed files.

NOTE: there was no `gui\glade\catalog\` folder and its files. I added them.

I am still getting the same results as reported above for media attributes and the surname list in the name edit window.

SNoiraud:
@Nick-Hall  I have a big problem with git
One file disappeared from my repository : gramps/gui/widgets/resizabletreeview.py

I saw nothing in the log. What can have happened ?

I saw I forgotten the filter tabs. This is how I discovered the problem.

SNoiraud:
I think there is nothing more to do.

emyoulation:
Could you create a companion Database Repair tool for this new feature?

If there is an unanticipated failure with these preferences, it would be nice to be able to purge all these values from where they are stored ( gramps.ini ?) without having to manually inspect every line. 

There were already instances where remembered customizations have made features of Gramps inaccessible until the customization was ripped out. 

Here is a "stored Preferences" example (unrelated to the new stored column width) but representative of the "law of unintended consequences." It was related to remembered window position & size.  Values could become invalid when the virtual screen size had reduced since the preferences had been stored. They could also become invalid if the preferences called for using portion of the virtual screen that was inaccessible in a multiple screen configuration. This occurred when removing/repositioning a 2ndry screen or by reducing the screen size. (See [bug 11831](https://gramps-project.org/bugs/view.php?id=11831):
[GTK]virtual screen size changes can cause inaccessible windows )

Several changes in virtual screen size could potentially make a column width (previously valid with a different screen/dialog layout) become invalid. Until it gets into wide beta, we won't know what weirdness user configurations can create. Nor how such weirdness will interact with the feature.

The gramps.ini has grown to include too many critical configurations to simply delete it when there is a problem. And it has grown dense enough that selectively cleaning it with Notepad++ is a significant task. It may have grown to the point where Preference dialog controlled choices and GUI drag customizations should be in separate files. 

It is unclear if changes committed with an external text editor will be applied during a live Gramps session. If you have to close Gramps to commit external repairs, those repair options become much less accessible. In that case, a Internal Gramps repair tool becomes more desirable.

So it would be useful to be able to have either a smart sanity check tool (or quick&dirty dumb purge tool) for the values stored for this feature in the preferences that is effective without restarting Gramps. 

A "smart sanity check" might have a bunch of rules. Perhaps it could compare widths against defaults: when a stored value was <25% or more than 300% of the default, that tab width would likely be an insane value; if the width is more than 50% of the overall screen or dialog width, that tab width would likely be an insane value. A smart tool could have the option to purge potentially insane values but keep the rest. This could simplify recovery from a minor corruption.

If such a tool could be designed for expansion, other hard-to-collectively-manage elements could added. Options could allow them to be selectively purgeable. Initially with a quick&dirty dumb purge option and evolving to include options to selectively clean too.

Expansions to such a repair tool could purge: any Gramps window width (alternately or in combination with height) that exceeds the Virtual Screen; any off-screen window origins; non-primary screen locations; any values associated with an no-longer installed add-on; any excessive Sidebar/Bottombar widths/heights; any Gramplet collections for Sidebar/Bottombar.

SNoiraud:
I don't save widths for columns. I convert the width to characters. It is easier to resize the columns when you change the font size.
I can't know the window size. The problem already existed before this PR.
The only thing you can do is either to rename or remove the gramps.ini either to remove the "[spacing]" section in gramps.ini

For standard usage, this should never happen.

emyoulation:
Thank you

SNoiraud:
I now begin the code cleanup (pylint).

emyoulation:
Could you also implement this change in the add-on "Plugin Manager Enhanced"?

Once Paul accepts the PR, this would give a broad beta testing opportunity before 5.2 goes to general release.

And documenting this particular evolution would allow other add-on developers to use the PR as reference for adapting their own code.

Nick-Hall:
Rebased and removed pylint changes.  We can format the code with _black_ later.

gramps/gen/config.py
```diff
@@ -325,6 +325,67 @@ def emit(key):
 register('utf8.cremated-symbol', "⚱")
 register('utf8.killed-symbol', "x")
 
+register('spacing.lst_person', [])
```
Nick-Hall, line 328: If we register these in the `__init__` method of the `PersistentTreeView` then it would make using this class easier.

gramps/gui/editors/displaytabs/embeddedlist.py
```diff
@@ -533,6 +534,13 @@ def build_columns(self):
             self.columns.append(column)
             self.tree.append_column(column)
         self.track_ref_for_deletion("columns")
+        # associate the selector config name to the PersistentTreeView
+        self.tree.set_config_name(self.get_config_name())
+        self.uistate.connect('font-changed', self.tree.restore_column_size)
```
Nick-Hall, line 539: This belongs in the `PersistentTreeView` class.  We will always need it.

gramps/gui/editors/displaytabs/eventattrembedlist.py
```diff
@@ -41,3 +41,6 @@ def get_editor(self):
 
     def get_user_values(self):
         return self.dbstate.db.get_event_attribute_types()
+
+    def get_config_name(self):
```
Nick-Hall, line 45: Can we use the module `__name__` instead?

Nick-Hall (changes requested):

Nick-Hall:
This is almost ready.  However, I would like to see a few changes to make it easier to use in the future.

If we pass the _uistate_, a config object and key when instantiating a `PersistentTreeView`, then very little code will be required outside of the class.

gramps/gui/editors/displaytabs/embeddedlist.py
SNoiraud: Yes, I know but the PersistentTreeView doesn't know the self.uistate and the self.tree.
It is a Gtk widget subclass with new methods.

gramps/gen/config.py
SNoiraud: OK.

SNoiraud:
I think I broke my branch. How could I solve my problem ?
I did a git pull and tried to merge and solve the problem. after that, I was not able to start gramps. The PluginWindows cannot Cbe imported.
So I made a : git reset --hard HEAD~1
It was OK but I think I loose your modifications.
I made my new modifications. Tested it. But I can't push my last commit.
To github.com:SNoiraud/gramps.git
 ! [rejected]            FR8767-bis -> FR8767-bis (non-fast-forward)
error: failed to push some refs to 'git@github.com:SNoiraud/gramps.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.

Is there an easy way to solve my problem or I need to create a new branch ?

SNoiraud:
> If we pass the uistate, a config object and key when instantiating

it's OK for an embededlist, but not the baseselector which is created with a glade file. I don't know how to add self.uistate from a glade file.
This is why I didn't add that in the PersistentTreeView. Do we need to add optional args ?

SNoiraud:
I have a new working branch: FR8767-ter

Nick-Hall:
> It was OK but I think I loose your modifications.

I didn't modify anything.  I just dropped the merge commit and the pylint changes.

Your new branch `FR8767-ter` looks good.  Just force push it to `FR8767-bis` to update this PR.  Glad to see that you rebased rather than merged.

I'd still like to see the `PersistentTreeView` as just a drop in replacement for `Gtk.TreeView`.  All related code should be within the class.

In cases where the TreeView is within a glade file, this will cause a slight problem.  I'll have a look at the baseselector code tomorrow.

SNoiraud:
I had a problem with my local repo. why ? network problem ?
I cloned again my repo and now it works.

SNoiraud:
I saw nothing about glade. The only solution could be to pass the uistate in the set_config_name method.

Nick-Hall:
> The only solution could be to pass the uistate in the set_config_name method.

Or you could create another method and do something like:

```
ptv = self.top.get_object('view')
ptv.set_uistate(self.uistate)
ptv.set_config(config, 'key-name')
```
You could still pass these in the constructor:

```
ptv = PersistentTreeView(self.uistate, config, 'key-name')
```
The `__init__` method would call the `set_uistate` and `set_config` methods.

In the case of the baseselector, you could remove the `PersistentTreeView` from the glade file and add it later.

```
sw = self.top.get_object('scrolledwindow33')
sw.add(ptv)
```
Although, your approach of keeping the `PersistentTreeView` in the glade file also looks good.

SNoiraud:
> In the case of the baseselector, you could remove the PersistentTreeView from the glade file and add it later.

If I do that, I need to modify clipboard.glade, dbman.glade, rule.glade and baseselector.glade and the associated python files clipboard.py, dbman.py, rule.py and baseselector.py.
Am I doing it ?

Since all PR merged I see a new problem with some columns: They have a minimum size which can't be reduce. I look at this.

SNoiraud:
I have another solution without modifying the glade file. I saw it's already used in the clipboard.
`        objectlist = self.top.get_object('objectlist')
        mtv = MultiTreeView(self.dbstate, self.uistate, _("Clipboard"))
        scrolledwindow = self.top.get_object('scrolledwindow86')
        scrolledwindow.remove(objectlist)
        scrolledwindow.add(mtv)
`

Nick-Hall:
> If I do that, I need to modify clipboard.glade, dbman.glade, rule.glade and baseselector.glade and the associated  python files clipboard.py, dbman.py, rule.py and baseselector.py.
Am I doing it ?

I don't think you gain much.  Setting the uistate and config after creation is OK in these cases.

> I have another solution without modifying the glade file. I saw it's already used in the clipboard.

Yes.  That's similar to my last suggestion.

SNoiraud:
It's OK for me. I think I tested all cases.

moorob:
I had to make a change to persistenttreeview.py to get it to work with the add source icon to editor dialogs mod #11372. Had to change get_column_size to below.  After this everything works fine.

def get_columns_size(self):
        """
        Get all the columns size
        """
        columns = self.get_columns()
        newsize = []
        nbc = 0
        context = Gtk.Label().get_pango_context()
        font_desc = context.get_font_description()
        char_width = 1.2 * (font_desc.get_size() / 1000)
        for column in columns:
            if nbc < len(columns) - 1 or len(columns) == 1:
                # Don't save the last column size, it's wrong
                # except if we have only one column (i.e. EditRule)
                child = column.get_widget()
                if isinstance(child, Gtk.Image):
                    # Don't resize the icons
                    size = 3
                else:
                    size = int(column.get_width() / char_width) + 1
                    size = 3 if size < 3 else size
                newsize.append(size)
            nbc += 1
        return newsize

SNoiraud:
> # Don't resize the icons
> size = 3

I think this is not the solution. It's too wide for me on linux.
On what OS do you test this ?
What is the icon size you use ?

I think we must convert the icon size into characters.

SNoiraud:
Can you test the following:
replace 
```
        size = 3
```
by
```
        size = child.get_allocation().width / char_width + 1
```

moorob:
Sorry been away getting my heart worked on.  I'm using Manjaro KDE.  Used it on linux and windows versions.  I'll give that a try.

Nick-Hall:
Rebased.

Nick-Hall (approved):

emyoulation:
I just set up a 5.2 on a Fedora box.

Upon 1st start, I had to resize all the columns of the list style Categories because they were so narrow. That seems wrong.

Now that column resizing can be remembered, are the default column widths initialized on first start?
It seems to auto-fit ... to the text size of the header label since there are no records. Is that true?

And if there is an autofit already built, it would seem a shame if it was only used on initial startup.  It would be useful to autofit again after changing font size ... or doing something like Reordering IDs to a larger ID format... or changing languages.

emyoulation:
Every time is run 'Find' with a filter gramplet, the columns resize to the default width that is just large enough to show the title label. 

Tested in People ungrouped view mode and Citation view.  Column re-orderings and Configuration of extra visible columns were retained but also reset widths. 

Column widths are NOT remembered

GRAMPS: 5.2.0-b5e88c13f 
Python: 3.11.4 (main, Jun  7 2023, 00:00:00) [GC... 
BSDDB: 6.2.9 (5, 3, 28) 
sqlite: 3.40.0 (2.6.0)
LANG: en_US.UTF-8
OS: Linux
Distribution: 6.3.12-100.fc37.x86_64

Nick-Hall:
@emyoulation Is there a bug report for this one?

emyoulation:
Report filed: (But accidentally filed as "Feature Request" and I don't have rights to change to Gramps bug report and set to 5.2beta1)

https://gramps-project.org/bugs/view.php?id=12943

emyoulation:
Raising another bug report that is related to saved Column Widths:
   0013473: View Configure re-ordering of columns does not reorder saved column widths


@Nick-Hall  said in https://github.com/gramps-project/gramps/pull/1306#issuecomment-955806546
> Yes, this seems like a good design.
>
> Perhaps in the future we could store other information such as the column order and sort column. If we want to go in this direction then PersistentTreeView may be a better name than ResizableTreeView.
>
> Have you thought about passing a config object and key name in the constructor? 

Before configuration allows changing column order in more GUI areas than View Modes, Column width needs to begin moving with column order changes in View Modes. And when column order control is extended into table Tabs, column width needs to be included

PR #1310, opened 2021-12-01
Added fliptext feature to Fan Chart Report

albertoriella:
Added flip_text option to Graphical Report Fan Chart.
This option flips the names in generation 2, 3 and 4 that are at the bottom of the fan.

Original Fan Chart Report
Original_FanChart_Report

FlippedText Fan Chart Report
FlippedText_FanChart_Report

Checkbox in menu
New_FanChart_Report_Menu

### DIFF
--- a/gramps/plugins/drawreport/fanchart.py
+++ b/gramps/plugins/drawreport/fanchart.py
@@ -168,6 +168,7 @@ class FanChart(Report):
         radial       - Print radial texts roundabout or as upright as possible.
         draw_empty   - draw background when there is no information
         same_style   - use the same style for all generation
+        flip_text    - flip names that are on the bottom half of the fan
         incl_private - Whether to include private data
         living_people - How to handle living people
         years_past_death - Consider as living this many years after death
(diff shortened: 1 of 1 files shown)


emyoulation:
This is a nice option. It would be nice to have it for any Gramps fan chart.

But the Fan Charting code is spread throughout multiple modules. Is there a possibility of consolidating the charting features so that the Reports & Views (and possibly the Gramplets) are expanded to support the same configuration options? Or would consolidating it in the core inhibit the ability for Fan charts to evolve?

And perhaps in the future, have a configuration handoff... where once the user interactively configures a chart to their satisfaction in the chart view, all the settings can be conveyed to the Fan Chart Report? Then the user need only configure the print specific options. 

And vice versa, an option import settings to the view from the report. (Or maybe from a Book.) Since users can set a complex report configuration with the Book feature, various chart layouts could be reused not only for reports, but also for Views. (Like an age color spectrum that highlights ancestors who exceeded their full biblical "three score years and ten" & another highlighting premature deaths.)

Here are the places Fan Charts are spread across Gramps:


Gramplets:

Fan Chart (ancestor)

2-way Fan Chart (ancestor and descendant)

Descendant Fan Chart

Charts category view modes:

Fan Chart (ancestor)

2-way Fan (ancestor and descendant)

Descendant Fan

Reports : Graphical reports:

Fan Chart... (ancestor

SNoiraud:
If you do this on gramps/gui/widgets/fanchart.py, the whole fancharts will be impacted

SNoiraud:
Did you see PR #1131 ?
Could it generate some problems ?

Nick-Hall:
Rebased.

Nick-Hall:
This PR is only relevant to the report.  I've made some minor changes to the text, otherwise this looks OK.

@albertoriella Thank you.

PR #1312, opened 2021-12-11
Create new Addon Manager dialog

Nick-Hall:
A proof of concept for a new addon manager.

Interface design inspired by the [addons listings](https://mattkmmr.github.io/gramps-addons-listings/) web page created by @Mattkmmr.

The idea is to provide a simple dialog for users to explore and install third-party addons.  The main features are:

* Access to the [Isotammi addons](https://github.com/Taapeli/isotammi-addons).
* Filter by project and type.
* Quick access to wiki pages.

Next steps:

* Add a configuration page for projects.
* Move automatic update settings from the preferences dialog.
* Possibly add an "Uninstall" button for installed addons.
* Update the addon listings to include a help link.
* Enhance the addon status to include a beta test status in addition to stable and unstable.


emyoulation:
Could you add a couple other things to this addon text file?

1) A comment line identifying the file as an add-on module catalog for Gramps genealogical database application (version defined 'g': parameter) in ConfigParser {or whatever is the recommended parser for this format?} format.

2) add-on placeholders (for future expansion) for 3 extra data items in the addon listing text file? URIs for a thumbnail image, target audience, and the 'status' Nick mentioned above.


That only leaves off the Contact info needed to replicate the [Add-ons list](https://www.gramps-project.org/wiki/index.php/5.1_Addons#Addon_List) on the wiki page.

With the above, we should be able to come up with scheduled tools that regenerate the lists/tables daily. Then those chunks can be transcluded that wiki page. (The transclusion chunks  could created for each dialect discovered in the GitHib/SourceForge directory containing the add lists file. It would be less fragile if we host the generated transclusion chunks on the same server as the WikiMedia CMS.)

emyoulation:
Since I only speak English, this next part is guesswork. Some users have mentioned that is useful to see the add-on's English name too. 

Sometimes the English equivalent is needed for Online forum discussions, googling & seeking documentation that might be more current but not yet translated.

cdhorn:
Instead of a hard coded repository list would it make sense to have the core Gramps one as the default and then pull down a repositories.txt with a list of others although there is only the Isotammi one at this point?

daleathan:
Tried to email this to you on the list not sure if you received it?

-------------------------------
Hi,

Hope you don't mind me mentioning that I saw something about that the other day on sam888's wiki user page where he mentions the **"depends_on**" option can be used; so Gramps already has it built in and you just have to use it!

Dale

copy pasted from his page:
https://gramps-project.org/wiki/index.php/User:Sam888#Notes_on_User_manual_update


..

[] Gramps CLI info "gramps -v" needs a section that list third-party addons with version numbers and if prerequisites have been met.

...
.................................................................................................

[] for addon prerequistes have a look at converting them to use "**depends_on =** "  like in the following addons

https://github.com/gramps-project/addons-source/search?utf8=%E2%9C%93&q=depends_on&type=

https://github.com/gramps-project/gramps/search?utf8=%E2%9C%93&q=depends_on&type=

which can be a comma separated list for multiple dependencies/prerequisites (so that addons that share prerequisites import the same library?) (maybe move spell check into an addon etc? tip of the day... possibilities...)

`depends_on = ["libwebconnect"]`

mentioned in the following
https://sourceforge.net/p/gramps/mailman/message/27070037/
_''In your gpr file, you can have a line like:

depends_on = ["libwebconnect"]

which is a list of id's from other gpr files. This example will ensure
that libwebconnect is loaded before your plugin. If it can't be found,
or you have a cycle, then your plugin won't be loaded.

If it is a common function, perhaps libhtml is the right place to put
it, and you can put a depends_on on that.

-Doug''_


example output is on terminal only:

```
Cannot resolve the following plugin dependencies:
   Plugin 'UK Web Connect Pack' requires: ['libwebconnect']

```
[](feature)  For plugins/addons if prerequisites not available display a page stating why/what is missing that includes the addon description and a url to the support page.

emyoulation:
Thank you. 

I saw the post with a general inquiry on the subject of dependencies support by John Ralls & immediate reply by Nick on the Developer mail-list. 

But GitHub has been silent about your contribution to the conversation. 

-Brian 
 
  On Sat, Dec 11, 2021 at 20:03, Dale ***@***.***> wrote:   


Tried to email this to you on the list not sure if you received it?

Hi,

Hope you don't mind me mentioning that I saw something about that the other day on sam888's wiki user page where he mentions the "depends_on" option can be used; so Gramps already has it built in and you just have to use it!

Dale

copy pasted from his page:
https://gramps-project.org/wiki/index.php/User:Sam888#Notes_on_User_manual_update

..

[] Gramps CLI info "gramps -v" needs a section that list third-party addons with version numbers and if prerequisites have been met.

...
.................................................................................................

[] for addon prerequistes have a look at converting them to use "depends_on = " like in the following addons

https://github.com/gramps-project/addons-source/search?utf8=%E2%9C%93&q=depends_on&type=

https://github.com/gramps-project/gramps/search?utf8=%E2%9C%93&q=depends_on&type=

which can be a comma separated list for multiple dependencies/prerequisites (so that addons that share prerequisites import the same library?) (maybe move spell check into an addon etc? tip of the day... possibilities...)

depends_on = ["libwebconnect"]

mentioned in the following
https://sourceforge.net/p/gramps/mailman/message/27070037/
_''In your gpr file, you can have a line like:

depends_on = ["libwebconnect"]

which is a list of id's from other gpr files. This example will ensure
that libwebconnect is loaded before your plugin. If it can't be found,
or you have a cycle, then your plugin won't be loaded.

If it is a common function, perhaps libhtml is the right place to put
it, and you can put a depends_on on that.

-Doug''_

example output is on terminal only:
Cannot resolve the following plugin dependencies:
   Plugin 'UK Web Connect Pack' requires: ['libwebconnect']


 For plugins/addons if prerequisites not available display a page stating why/what is missing that includes the addon description and a url to the support page.

Mattkmmr:
> A proof of concept for a new addon manager.
> Interface design inspired by the addons listings web page created by @Mattkmmr.

@Nick-Hall I'm glad to hear that my web page inspired you to work on a new addon manager. 😄 

---

> Add a configuration page for projects.

What is planed here? Do you mean that users can add different addon source links like @cdhorn mentioned?

---

@daleathan I think the **depends_on** option is for addons depending on other Gramps addons, while in the mailing list and the forum people also asked for a way to install all kinds of python modules on which Gramps addons can depend on. Right now for the windows installer you'd have to bundle them all.

wroldwiedbwe:
>  forum people also **asked for a way to install all kinds of python modules** on which Gramps addons can depend on. Right now for the windows installer you'd have to bundle them all.

@Mattkmmr You mean like how the [Life Line Chart View addon](https://www.gramps-project.org/wiki/index.php/Life_Line_Chart_View#Prerequisites) installs [python modules from pypi](https://github.com/gramps-project/addons-source/blob/41c096c65ba1b2eb9bf1bde1aa632f9e5890af93/LifeLineChartView/lifelinechartview.gpr.py#L68) already (see class ModuleProvider) maybe split that code off into its own gramps addon?

Nick-Hall:
> Instead of a hard coded repository list would it make sense to have the core Gramps one as the default and then pull down a repositories.txt with a list of others although there is only the Isotammi one at this point?

I was thinking of allowing the user to enter the urls for addon repositories and store them in the ini file.   The listings file could be extended to include the name and description of the repository.

Nick-Hall:
> You mean like how the Life Line Chart View addon installs python modules from pypi already (see class ModuleProvider) maybe split that code off into its own gramps addon?

Will installing modules from PyPI actually work with the Mac and Windows AIO?  Perhaps @jralls and @prculley could comment here.  If so, we could move the ModuleProvider code into core Gramps and introduce a list of required modules into the gpr file.

emyoulation:
From my IT admin days, I liked to be able to curate my own add-on collections. I restrict users to installing only certain pre-approved extensions. 

If I was to start a Gramps Users Ring with a few of my less technically adept relatives, I'd probably want to have those users only download add-ons I'm willing to do do tech support on.

Likewise, if Gramps was adopted in a classroom environment (whether as a framework for teaching genealogy, history or programming), the teacher would want to control what students were allowed to install... and when. 
e.g., the add-on available for lesson 2 is at a certain stage of development, the teacher uploads a new add-ons.txt making a different stage available for the following lesson.

emyoulation:
What other things can you not add later? Probably some could be done with a file manager add-on. (Like a Gramps Internet Store for: Custom Filter, themes, book designs, form definition collections). 

But don't the prerequisites tend to go in the core? Where file permissions are more of an issue than the User Directory. Self-modifying code in application/admin areas is highly suspicious. 

Core changes: 
  1) language support (translation, dictionary, grammar) 
  2) critical patches

jralls:
> Will installing modules from PyPI actually work with the Mac and Windows AIO? 

Pure python modules should be OK for Windows, provided Paul bundles `pip3`. It could be made to work on macOS with a little fiddling of `PYTHONPATH` and `PYTHONUSERBASE`, though I'd have to add a `pip3` executable similar to `gramps-launcher`.

Anything involving shared libraries will be more difficult. On windows the shared libraries have to get installed somewhere on the path in a way that the python module can find them. On macOS they'll have to be configured as plugins somehow and the plugin entitlement added to the bundle's code signature.  I think  they'd also have to be code-signed with an Apple Developer certificate, something I doubt that PyPI is set up for.

BTW, using PyPI is [not without risk](https://arstechnica.com/information-technology/2021/11/malware-downloaded-from-pypi-41000-times-was-surprisingly-stealthy/). Lots of our users are in @emyoulation's "less technically adept" category. We should be careful on their behalf.

Nick-Hall:
> From my IT admin days, I liked to be able to curate my own add-on collections.

We will allow personal archives, not just Gramps and Isotammi.

Nick-Hall:
@jralls Thanks.  I suspected that would be the case.  I'll just add a check for required modules for now.

emyoulation:
> Instead of a hard coded repository list would it make sense to have the core Gramps one as the default and then pull down a repositories.txt with a list of others although there is only the Isotammi one at this point?

Perhaps, in a future evolution, this core default path could combined in a single text file with the other paths that lock Gramps into a single purpose, OS, User Directory location?

So an Animal Pedigree fork or a DNA fork of Gramps might want totally different default repositories. (Although users would still be able to add custom overrides to get to other Repositories.) Of course, anyone capable of maintaining an entirely separate fork would be capable of tweaking the default add-on repository. But flexible is nice.

The ability of the PortableApps to drag'n'drop on Windows machines with relative paths makes it easy to isolate instances of Gramps. The combination of relative & absolute paths is empowering.

emyoulation:
Not a request but...

MediaWiki has an apparently AWFUL add-on installer but the user documentation system for Extensions has a few good things. 

I that we can eventually adapt their [Extensions template](https://www.mediawiki.org/wiki/Template:Extension) to start giving a more consistent and less clumsy look to our extension documentation. MediaWiki Extension documentation Examples in use: [ImageMap](https://www.mediawiki.org/wiki/Extension:ImageMap) and (with a thumbnail) [TreeAndMenu](Extension:TreeAndMenu)

GitHub's readme.md is somewhere we WikiContributors can start when composing user docs. (Right now, we've been missing a couple important aspects on the individual pages: giving developer credits & versioning related documentation.) But it is probably a good plan to avoid raising documentation burden on developers. Some developers hate (or are ill-suited for) creating user documentation. Extra requirements there might be a disincentive for sharing.

MediaWiki have some good things from which we can pick & choose.  Are there other interesting systems (code submission, documentation, distribution, etc.) for extension management that could be inspirational?

emyoulation:
> I hope that we can eventually adapt their Extensions template to start giving a more consistent and less clumsy look to our extension documentation.\n\nCreating our own template should be reasonably easy.

Adapting that particular template has a MediaWiki prerequisite: Lua 
(Personally, I'm not certain the risks/burden of adding a language capability to our website domain doesn't outweigh the benefits. Like JavaScript, it would seem to invite attack.)

I'm sure we can do a template. That one was just an example with a clean look and broad features.

emyoulation:
Can this be converted to an Add-On for testing & feedback? 

Just replacing the modified 5.1.4 files didn't work. Or is it dependent on other changes found only in the 5.2 Master?

Nick-Hall:
I will need to include the help url and updated plugin status in the listings file, so this feature will not be available for v5.1.

I started to add all the help urls to the gpr files, but it became a long job.  I'll probably get going again shortly.

The template idea for the wiki works well though.  I have a simple version based upon the MediaWiki link you gave in an earlier post.

emyoulation:
> I started to add all the help urls to the gpr files, but it became a long job. I'll probably get going again shortly.

How can I add help-url lines to the non-gramplet add-ons in 5.1 `.gpr.py` files that will NOT break the parser.  

Can I just prefix the line with a double semi-colon in the same way as the `.ini` files are commented? If so, we can add the URLs in 5.1 and take that off your plate.

Nick-Hall:
Here is a screenshot of my progress so far:

Addon Manager

It is possible to filter by:

1. Type:  View, Gramplet, Tool, Report etc...
2. Project:  Gramps or Isotammi.
3. Audience:  Everyone, Expert or Developer.
4.  Status: Unstable, Experimental, Beta or Stable.

The next step will be to add a configuration tab.  We will need to way to add external projects and private repositories.  I'll probably also move the "Addons" tab in the Preferences dialog.

Mattkmmr:
@Nick-Hall Some questions and ideas:
1. Gramps addon script [make.py](https://github.com/gramps-project/addons-source/blob/master/make.py) is currently generating dictionary like text strings for addon listing data. Could we use JSON instead? The main problem is that the script is mixing single and double quotes which violates JSON.
2. How about using three status labels `Development`, `BetaTesting` and `Stable`.
3. Maybe we should also include the project name or an abbreviation as a prefix to the addon ID to avoid problems in case two projects have addons with the same ID. So the ID only needs to be unique within one project.

emyoulation:
Please consider the migration path for add-on development.

When a developer gets to a point where they want to share a experimental or conceptual add-on to a select group of alpha testers, this might be done with a link to a file in their GitHub repository. The link might be to a single add-on or a collection of add-ons in various states of development.

When entering public beta, they might want to submit that link to their personal GitHub repository into a general Gramps-project that has requires no approval.

How would they submit their release candidate into peer review? Could they would move the personal repository and the link file into a Submissions folder in the Gramps-project repository. Perhaps via another Pull Request workflow? (This is probably the 1st point where Weblate translators might be able to contribute.)

When approved via Peer Review, the add-on would be moved to the normal add-ons source.

Perhaps there could be a tool which, when given a URL/URI to a gramps .gpr.py file, validates the Gramps Plugin Registration file and generate the addonlist.txt (or addonlist.json , as Matt suggested) that is processable by the addon manager?

In such an imaginary system, a particular addon might simultaneously exist in ALL states: a soon-to-be-replaced version in full release, a version in Peer Review, a version in Public Beta, a version in Private Beta, a version in Conceptualization Alpha. And possibly forks in 5.2 and 5.1 too?

Nick-Hall:
> Gramps addon script make.py is currently generating dictionary like text strings for addon listing data. Could we use JSON instead? The main problem is that the script is mixing single and double quotes which violates JSON.

I'll investigate.  JSON would be preferable to text.

> How about using three status labels Development, BetaTesting and Stable.

I have "Unstable", "Experimental", "Beta" and "Stable" at the moment.  Previously we only had "Unstable" and "Stable".  Do you want "Development" in addition to these?

> Maybe we should also include the project name or an abbreviation as a prefix to the addon ID to avoid problems in case two projects have addons with the same ID. So the ID only needs to be unique within one project.

How would we assign the project prefix?  Perhaps using a UUID would be better?

Nick-Hall:
> The link might be to a single add-on or a collection of add-ons in various states of development.

Developers will have the ability to share their repository.

Mattkmmr:
> I have "Unstable", "Experimental", "Beta" and "Stable" at the moment. Previously we only had "Unstable" and "Stable". Do you want "Development" in addition to these?

No I meant I'm unsure about the difference between Unstable and Experimental. Maybe using three status levels are already enough. Audience and status could also be combined:
1. `Development` = new features added, addon is unstable (Audience:  Developer & Experts)
2. `Beta` = beta testing & bugfixing (Audience:  Experts)
3. `Stable` = stable addon (Audience: Everyone)

Nick-Hall:
I've added an "Updates" tab for the settings:

Settings

I've also converted the listings file to use JSON.

Nick-Hall:
> Audience and status could also be combined

Possibly, but there could be stable addons intended just for developers or expert users.

Nick-Hall:
I've added a list of addon repositories to the "Updates" tab.

Settings2

The idea was to create a simple dialog for the user to enter a repository name and url.  Perhaps the listings file could contain some repository metadata such as name, description, maintainer, last updated etc...  Then the user would only have to enter an url.

Mattkmmr:
> I've added a list of addon repositories to the "Updates" tab.

The interface change looks great, but I wonder if the space isn't too small for the repository list. Especially when the number of addon repositories grows in future. Perhaps the list should get its own tab.

> Perhaps the listings file could contain some repository metadata such as name, description, maintainer, last updated etc... Then the user would only have to enter an url.

Repository metadata sounds like a good idea.

> Possibly, but there could be stable addons intended just for developers or expert users.

Yes, you are right. It's better to keep it seperated.

emyoulation:
Could the add-on Manager also download a Gramps Version list that the About Box (and possibly the PrerequisitesChecker addon) could use to compare the currently running version? 

Definitely NOT suggesting a Nag upon startup feature but that the About box indicate a new 'x.y.z' version has been available since mmm yyyy. And that when the Addon Manager checks the Addon List, it notes that since a newer version is in release, not newest available add-ons may be listed. (The feature won't make any difference for 5.2 but could for 5.3 and newer.)

emyoulation:
> I've added a list of addon repositories to the "Updates" tab.

If the release addon repository moved to the A2 hosting where the Gramps WordPress, MediaWiki and MantisBT tools reside, maybe the reins on the GitHub repository could be loosened?  If users choose to download from GitHub, (which would be listed as a deselected option by default, it could be implicit that those add-ons are not curated as closely as the ones on the `Gramps-Project.org` web server.

Mattkmmr:
> Could the add-on Manager also download a Gramps Version list that the About Box (and possibly the PrerequisitesChecker addon) could use to compare the currently running version?

Maybe there could be several places where users are notified for newer Gramps versions. 
* Addon Manager
* About Dialog
* Gramps Settings
* Prerequisites Checker

emyoulation:
Was just looking back through the dialogs because of a a Tech Support question on Discourse.

How will the 'check for updates' handle "hidden" add-ons. Will there be an override in case the user "hid" an add-on due to bugs (or hid unintentionally) and is looking for updates for the hidden add-on.

I would like to permanent stop be offered the Cairo Clock Gramplet but don't want it installed on my system. The same with the database backends that I will never use. 

On the flip side, having a couple perennial not-yet-installed add-ons means a dialog always appears after checking for updates. This inspires more confidence that the add-on list was found than the dialog that says no new updates were found. 

Perhaps that 'no updates found' could give a more informative status message:
"No new or updated Gramps add-ons found. This server lists xx of the zzz installed add-ons.  The list was last revised on dd-mmm-yyyy."

And could this status dialog (whether is a 'no updates' or a hotlinked list of updates that were installed) generate to a New Note instead of a dialog? (Although this might be problematic if no Tree is loaded.)

If it was a New Note, the user could save it as a ToDo type reminder of updates they have not yet explored... or as a "log"

(Writing reports to Notes would have many uses. For instance, the preferences option to create a New Source during import might have an 'error log' of type 'Source Note' with links to the bad objects & another 'Import Statistics' note. Both of which do double duty is they are shown as post-process feedback.)

emyoulation:
@Nick-Hall 

The help file and translation is going to need a LOT of beta testing, help URL list collation, wiki preparation, and internationalization evolution. 

For instance what happens if Isotammi's initial default GUI and `help_url` has Finnish, not English, content. Should that `help_url` line be `help_url-fi` ?

Maybe having a English dialect `https://raw.githubusercontent.com/gramps-project/addons/master/gramps52/listings/addons-en-bz.txt` list be the dummy file would also give English speakers a change to learn how to do a real Weblate submission? (With a chance to do a functional review of the wiki page documented the use of the translation tools.) That might let us compile a GOOD set of URLs via Weblate.

Could you make a dummy 5.1 add-on version of this (which ignores the version and reads just the  5.2 repository addonlist) but will not actually INSTALL, HIDE or UNINSTALL anything. 

Perhaps you could add a feature that links to an auto-generated `.gpr.py` placeholder `help_url`  (like "`https://gramps-project.org/wiki/index.php/Addon:<addon_name>.draft`").  We can start building out the destination wiki pages from the README.md files... or submit a better `help_url` for updating the `.gpr.py` file.  

Maybe there could be an automated script for an en masse update to insert a commented 
`# help_url=_"Addon:Sample_Addon"` 
to all the 5.1 .gpr.py addon files and 
`# help_url=_"Plugin:Sample_Plugin"` 
builtin/core Gramps Plugin Registration Python files in the 5.1 and Master branch that are not Gramplet types that already have a `help_url` line? 

@Mattkmmr might be able to experiment with making his Plugin Browsing website leverage commented `help-url` lines for his Help buttons or look at your evolving 5.2 addonlist file format.

The dummy Addon Manager  might use an English dialect addonlist that the current automated routines will not overwrite.  


## References:

* [Writing a plugin](https://www.gramps-project.org/wiki/index.php/Writing_a_plugin#Internationalize_it_.283.2_and_later.29)
* [Addons development](https://www.gramps-project.org/wiki/index.php/Addons_development#Get_translators_to_translate_your_addon_into_multiple_languages)

English | en
-- | --
English (Australia) | en-au
English (Belize) | en-bz
English (Canada) | en-ca
English (Ireland) | en-ie
English (Jamaica) | en-jm
English (New Zealand) | en-nz
English (South Africa) | en-za
English (Trinidad) | en-tt
English (United Kingdom) | en-gb
English (United States) | en-us


language codes from `https://www.andiamo.co.uk/resources/iso-language-codes/`

Nick-Hall:
I will look at this on Wednesday.  It is probably best to put the enhancements to the plugin data into the master branch.

emyoulation:
A sample set of Gramps plugins first being developed in a language other than English:

* An esperanto " Importo de FamilySearch datoj" (Import of FamilySearch dates) addon consisting of 2 addons: a scraper Gramplet and import Tool (still in alpha test)
   https://github.com/jmichault/PersonFS

cdhorn:
@Nick-Hall maybe the text should also be center aligned?  It feels like a lot of white space, and potentially wasted space too with only a single column. This is also the older code I guess, as it is missing much from your screenshots.
Screenshot from 2023-03-04 22-26-34

cdhorn:
@Nick-Hall a problem I seem to have is that it takes forever for the addon manager to load. This is with the old one as well though lately, but I don't recall it always being so.  I got curious and timed it tonight and found the fetch of the addons-en.txt file was taking 40 seconds and same for the Isotammi. It seems urllib is trying both ipv6 and ipv4, but it looks up and tries ipv6 first and I have to wait 10 seconds for each address to time out and there are four of them before it tries an ipv4 address and goes through. I think this has been an issue since I switched my ISP. I imagine some other users must have run into it as well but maybe thought it was normal.

There is a technique to monkey patch it to force it to ipv4 documented at https://stackoverflow.com/questions/2014534/force-python-mechanize-urllib2-to-only-use-a-requests which is ugly but is one way to solve the problem. It seems to be working well for me, but there are probably other approaches. Thoughts on how best to handle this?

emyoulation:
> * Update the addon listings to include a help link.

URL placeholder targets have been created as a workaround for the Help buttons in the Report option dialogs trying to find Addon report documentation in the Built-in Reports wiki manual page.

see
https://gramps-project.org/bugs/view.php?id=12901

Nick-Hall:
I've made some changes following your feedback.
Addon Manager 1
It looks a bit neater and has a search facility.
Addon Manager 2
There is now a setting to allow Gramps to attempt to install required python modules.  I'm going to need some help testing this!
Addon Manager 3
The projects are now on a separate tab.  Isotammi addons cannot be tested yet.  The new listing files need to be in json format.
Addon Manager 4
The requirements for an addon can be listed with an indication of which need to be installed.

Comments welcome.

emyoulation:
Please check the dialogs in a dark mode theme.  The Addons tab was actually MORE readable in the color blindness simulator than with normal vision. (see https://pilestone.com/pages/color-blindness-simulator-1)  

Also could you generate a Installation report (with links to Readme or wiki help for each installed/updaetd addon) so users can follow-up installation with exploration?  Ideally, the reports would generate a new unsaved Note of "To Do" type since hotlinks do not survive copy and paste.

emyoulation:
a consistency hobgoblin issue.  The "Addons" tab buttons being variable in number positions increases the amount of time needed to assess your options.

Could it have the same buttons (Install, Update, Requirements, Wiki) but have any button that is not viable be dimmed?

emyoulation:
Ummm.... can users only install 1 add-on at a time now?????

Nick-Hall:
> Please check the dialogs in a dark mode theme. The Addons tab was actually MORE readable in the color blindness simulator than with normal vision.

They look OK using "Adwaita-dark".  What do you want changed?

> Also could you generate a Installation report (with links to Readme or wiki help for each installed/updated addon) so users can follow-up installation with exploration? Ideally, the reports would generate a new unsaved Note of "To Do" type since hotlinks do not survive copy and paste.

I could generate a report, but I doubt that many people would want a note created in their family tree.  Perhaps an option to filter on installed addons would be better?

> The "Addons" tab buttons being variable in number positions increases the amount of time needed to assess your options.

I'll move these to more consistent positions.

> can users only install 1 add-on at a time now?????

The new Addon Manager installs one addon at a time.  I considered allowing the users to select some addons and then click an "Install selected" button.  However, this approach is slower than just browsing the list and clicking "Install" buttons directly.

An "Install all" option also isn't very appealing.  Ordinary users would end up installing "developer" and "expert" addons together with experimental code.  An option to install all addons marked for "everyone" and "stable" might make some sense.

emyoulation:
I have not got a 5.2 set up yet so it was just a question... a note about something often overlooked. (I don't use Dark mode but when doing beta test on a couple add-ons, I tried it and we discovered the add-ons were unusable.)

A Note is the only way people could manage explorations of dozens of add-ons without an external tool. That overload means exploration will NOT be completed in a single session. With a Note, they can delete/annotate items as the exploration is complete and the hotlinks will be accessible in the next session. And, in an editable format, the order can be reorganized (again, without losing the hotlinks) so it can be used to design a workflow or exploration plan. (Say, the order of 4 tools to clean data annually, 6 filter rules to build a special validation, 5 reports to put in a Book.) 

If the output is to an 'unsaved' Note Editor, the people who don't want it in the tree can close without saving. 

The only other output format with similar functionality (persistence of live links to wiki/readme, editability) would be an .ods spreadsheet. 

Not having a way to multiple/mass select is a HUGE step backwards. When installing fresh (or when a large number of related addons are released), it is typically easier to select all (or select all of a type) and then de-select the few that are unwanted. (For instance, I will install all but the extra database backends, the clock, the foreign language webConnect tools and the addons with prerequisites. ) Installing/selecting one-by-one would be utterly tedious and error-prone. We'd lose users!

Nick-Hall:
> Not having a way to multiple/mass select is a HUGE step backwards.

You can still use the existing functionality if you want to install multiple addons at once.  I haven't removed it.

Nick-Hall:
This new interface is based heavily on the web page created by @Mattkmmr.  I think the design is friendly for new users trying to find an addon they might like.

We could try another design, but I'd like to try this first.  It would be quite easy to allow users to select rows and move the install buttons into a button bar, but this would change the look and feel of the dialog.  I'm happy to make such a change if we get enough feedback though.

Nick-Hall:
Moved buttons and captured a dark mode screenshot:

Addon Manager 1 - Dark

emyoulation:
Do you think some sort of delineation between listings would be helpful? Like the panels in CDHorn's Card view? Or alternating background shadings.  (The way people used to design to print in 3-line groupings ... to take advantage of the old greenbar line printer paper layouts.)

When scrolling, it is not instantly intuitive whether the boldface text is a header or footer for the other content. From experience, you may know it is a header from the initial screen. But the brain doesn't instantly associate the grouped lines... it takes a bit of extra effort.

Nick-Hall:
I could add borders.  The following example shows a 1 pixel grey border with rounded corners and a small margin.
Addon Manager Borders
An alternative would be to use separators.  Zebra striping is difficult with filtered lists.

emyoulation:
That looks **_much_** easier to grasp. Thank you.

But I'm  biased.  What do others think?

cdhorn:
Agree the borders look better. 

After installing or updating I'm guessing without checking code that the plugin is not being loaded or reloaded as the list is not updated, and the refresh button doesn't update it to reflect the change either. Neither does closing and reopening the dialog. I do see the new versions after shutting down Gramps and restarting it.

I also get the following on startup when check for addon updates always is enabled:
```
Exception in thread Thread-1:
  File "/usr/lib64/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/home/gramps/gramps/gramps/gui/utils.py", line 687, in run
    self.addon_update_list = available_updates()
                             ^^^^^^^^^^^^^^^^^^^
  File "/home/gramps/.gramps/gramps52/plugins/PluginManager/PluginManager.py", line 779, in available_updates
    (config.get("behavior.addons-url"), lang))
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gramps/gramps/gramps/gen/utils/configmanager.py", line 381, in get
    raise AttributeError("No such config setting name: '%s.%s'" %
AttributeError: No such config setting name: 'behavior.addons-url'
```

emyoulation:
The hint text at the top of the "Check for Updated addons now" dialog has the 5.1.x menu path in Edit -> Preferences.  Flag it for updating. 

image

emyoulation:
Resizing the "Edit Project" dialog does not make the fields larger. They just slide to the right.

image

emyoulation:
The lists of addon are less "Projects" than a Collection, an Archive, or a Library  (Storehouse, Depository, Depot ) . It could be called a Repository but that word already has meaning in the Genealogy database.

Please use a different label than "Project"


image

Nick-Hall:
> It could be called a Repository but that word already has meaning in the Genealogy database.

Yes.  I avoided terms like Repository, Collection, Archive, and Library for this reason.

I'm not too keen on Storehouse, Depository or Depot either.  Any other suggestions.

emyoulation:
Annex?

jralls:
Set, Site, Distribution?

emyoulation:
How do I add a Feature Request in MantisBT that is resolved by this Pull to the 5.2 roadmap?

https://gramps-project.org/bugs/view.php?id=12363

wroldwiedbwe:
1.Accidentally deleted the default "addon manager > project" and the only way I could bring it back was to delete gramps.ini. so maybe have a restore default button that also keeps any additional project I've added ?  

2. On "addon manager > settings" selecting either of the checkbox and close then viewing that tab again shows the setting have not been retained.

3. no helps buttons!

Nick-Hall:
@wroldwiedbwe I'll implement the "Reset" and "Help" buttons.

Changes to the settings should be retained.  I'll investigate.

wroldwiedbwe:
@Nick-Hall Thanks.

Can I also suggest that you have a enable/disable option for each project line, which should be also help when you you don't want to see anything from that project and also when you use [gramps safemode](https://gramps-project.org/wiki/index.php/Gramps_5.1_Wiki_Manual_-_Command_Line#Safe_mode) all but the default should be ignored?

Nick-Hall:
@wroldwiedbwe I've created PR #1523 to address these and other issues resulting from user feedback.

Each project line already has a checkbox where you can enable or disable it.

PR #1314, opened 2021-12-17
CSV: possibility to select the dialect.

SNoiraud:
You can also select a delimiter in a list

The current possible values are , : ; | and \t (tab)

### DIFF
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -158,6 +158,9 @@ register('behavior.welcome', 100)
 register('behavior.web-search-url', 'http://google.com/#&q=%(text)s')
 register('behavior.addons-url', "https://raw.githubusercontent.com/gramps-project/addons/master/gramps52")
 
+register('csv.dialect', 'excel')
+register('csv.delimiter', ',')
+
 register('database.backend', 'sqlite')
 register('database.compress-backup', True)
 register('database.backup-path', USER_HOME)
--- a/gramps/gen/utils/docgen/csvtab.py
+++ b/gramps/gen/utils/docgen/csvtab.py
@@ -30,8 +30,13 @@ import csv
 # gramps modules
 #
 #-------------------------------------------------------------------------
+from gramps.gen.const import GRAMPS_LOCALE as glocale
+from gramps.gen.config import config
 from .tabbeddoc import *
 from ...constfunc import win
+
+_ = glocale.translation.gettext
+
 
 class CSVTab(TabbedDoc):
 
@@ -51,7 +56,12 @@ class CSVTab(TabbedDoc):
 
         self.f = open(self.filename, "w", newline='',
                       encoding='utf_8_sig' if win() else 'utf_8')
-        self.writer = csv.writer(self.f)
+        my_dialect = config.get('csv.dialect')
+        my_delimiter = config.get('csv.delimiter')
+        if my_dialect == _("Custom"):
+            self.writer = csv.writer(self.f, delimiter=my_delimiter)
+        else:
+            self.writer = csv.writer(self.f, dialect=my_dialect)
 
     def close(self):
         assert(self.f)
--- a/gramps/gui/csvdialect.py
+++ b/gramps/gui/csvdialect.py
@@ -1,0 +1,156 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2021-      Serge Noiraud
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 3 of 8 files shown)


SNoiraud:
When you try to configure a view (people, family, place, ...) you'll get the first screenshot:
csv 1

If you select the CSV dialect tab, you'll get the second screenshot:
csv 2

Nick-Hall:
* Need to add `gramps/gui/csvdialect.py` to `POTFILES.in`.
* Better to display "Tab" to users rather than "\t".
* Neater to store actual delimeter in config file.

I've added a commit.

SNoiraud:
It works for me.

emyoulation:
Maybe it should be cloned & renamed?  This expands it into Delimited Text, not Comma Separated Values.

However disappearing the CSV into a broader category would remove an important feature checklist item: CSV.

So cloning and perhaps downsizing the CSV side to eliminate confusing options could double the value of this expansion. (CSV might still need an option to choose the Text Qualifier & escaping character.)

Nick-Hall:
@emyoulation You can see all the options available for [dialects](https://docs.python.org/3/library/csv.html#dialects-and-formatting-parameters) in the python documentation.

emyoulation:
You & Serge have the programming background to know that.  And I've good enough years to grasp it too.

But the "Aunt Martha", "newbie" & "reviewer" segments find smaller scope subsets easier to grasp and for ticking boxes their checklists.  

Since an import of a GEDCOM or CSV is likely to be among the first tasks attempted, it is also better if documentation for those features can be covered in 2 screens of scrolling text too. So eliminating the concept of delimiters would make a digestible "CSV for Dummies" section much more achievable. The Swiss army knife Delimited variant would be an advanced/power user feature. (Suitable to be kept as an add-on rather while the CSV stays a built-in too. It's likely to need tweaking.)

emyoulation:
The 
[Line 63 of `csvdialect.py`](https://github.com/SNoiraud/gramps/blob/3b2d845992669dfca8da6e2b3797429e20031b9f/gramps/gui/csvdialect.py#L63) is causing a console error when the 7 changed files are updated in Gramps 5.1 version 
`'\t': _('Tab', 'character'),`

gettext expects 2 and finds 3. The error is gone after deleting the ", 'character'" from the list.

[Line 24 of `export.gpr.py`](https://github.com/SNoiraud/gramps/blob/3b2d845992669dfca8da6e2b3797429e20031b9f/gramps/plugins/export/export.gpr.py#L24) causes a version mismatch when copied into version 5.1 core. Changing to version 5.1 eliminates the error and the CSV delimiter selection feature still seems to work.
`MODULE_VERSION="5.2"`

emyoulation:
The CSV Dialect configuration dialog has a pull-down menu control that is incorrectly sized. The menu selector that goes with the "Delimiter:" label is the full width of the dialog which splits it to the next line.  
image

In the default format "excel" the custom delimited is ',' (not 'comma') which looks like screen dirt. The menu is not selectable unless the "Custom" radio button is selected. (Shouldn't the menu be active and selecting a different delimiter simultaneously select the "Custom" option?)  

The combination of the inactive full-width menu and almost invisible default menu value makes it appear as though the bottom of the dialog is some sort of informational box.
image

emyoulation:
The CSV Dialect selection uses the Microsoft trademark name 'excel' with the wrong capitalization and not following Microsoft trademark guidelines. ('Excel' is not allowed be used generically to mean spreadsheet.)  We CAN use "Excel as an adjective: "Excel spreadsheet" without a ® or the "trademark of Microsoft Corp."
https://www.microsoft.com/en-us/legal/intellectualproperty/trademarks


Does it make sense to use  "Comma Separated Values text (.csv)" as the default and "OpenDocument Spreadsheet (.ods)" instead? Can we keep trademarks from being a consideration?


UNIX® is a registered trademark of The Open Group. Our capitalization is incorrect and does not meet their trademark guidelines.
What is meant by the 'unix' dialect anyway? Is it referring ASCII End Of Line markers for  Windows (CR LF) vs UNIX (LF)
If so, is there only CSV with UNIX LF? How would you get TSV with a UNIX LF ?
image

 (The following is not definitive but it IS neatly formatted)
> from https://confluence.qps.nl/fledermaus7/questions-answers/other/differences-in-end-of-line-characters-mac-windows-and-linux
>
> Here is how the different platforms end lines in ASCII files. 
> 
> Windows - Lines end with both a <CR> followed by a <LF> character
> Linux  - Lines end with only a <LF> character
> Macintosh (Mac OSX) - Lines end with only a <LF> character
> Macintosh (old) - Lines end with only a <CR> character

emyoulation:
Thank you for this enhancement.

It makes using the Text Import gramplet in concert with Excel extremely easy.  (The copy'n'paste workflow is FAR more simple than saving a CSV file from Excel, opening the CSV in a text editor, copy'n'paste, and finally -- deleting the extraneous CSV file.)

PR #1317, opened 2021-12-26
Birthday on February 29 only taken into account every 4 years

SNoiraud:
Fixes #012511
    
Birthday on February 29th only taken into account every 4 years
in the calendars
    
Added an option in the preferences display tab
If checked, we display the birthday on march 1 for non-leap years
If not checked, we display the birthday on february 28 for non-leap years
When we have a leap year, display the date on february 29


SNoiraud:
Problem with another PR.

PR #1318, opened 2021-12-26
Display leap day anniversaries every year on calendars

SNoiraud:
Fixes [#12511](https://gramps-project.org/bugs/view.php?id=12511)

Birthday on February 29th only taken into account every 4 years
in the calendars

Added an option in the preferences display tab
If checked, we display the birthday on march 1 for non-leap years
If not checked, we display the birthday on february 28 for non-leap years
When we have a leap year, display the date on february 29

### DIFF
--- a/gramps/gui/configure.py
+++ b/gramps/gui/configure.py
@@ -1387,6 +1387,23 @@ class GrampsPreferences(ConfigureDialog):
         grid.attach(obox, 2, row, 2, 1)
 
         row += 1
+        # Birthday on february 29
+        feb29 = Gtk.ComboBoxText()
+        show_on = [_("on the previous day"),
+                   _("on the next day"),
+                   _("only on leap years")]
+        list(map(feb29.append_text, show_on))
+        active = config.get('preferences.february-29')
+        feb29.set_active(active)
+        feb29.connect('changed', self.date_february_29_display_on)
+        ttip = _("For non leap years, anniversaries are displayed on either "
+                 "February 28, March 1 or not at all in Gregorian calendars")
+        feb29.set_tooltip_text(ttip)
+        lwidget = BasicLabel(_("Show leap day anniversaries"))
+        grid.attach(lwidget, 1, row, 1, 1)
+        grid.attach(feb29, 2, row, 2, 1)
+
+        row += 1
         # Status bar:
         obox = Gtk.ComboBoxText()
         formats = [_("Active person's name and ID"),
@@ -1571,6 +1588,12 @@ class GrampsPreferences(ConfigureDialog):
         Save "Date calendar for input" option.
         """
         config.set('preferences.calendar-format-input', obj.get_active())
+
+    def date_february_29_display_on(self, obj):
+        """
+        Save "February 29 display on " option.
+        """
+        config.set('preferences.february-29', obj.get_active())
 
     def autobackup_changed(self, obj):
         """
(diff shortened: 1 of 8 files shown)


emyoulation:
Ummm. Kinda curious if it handles leap centuries... since we just had one in 2000.

SNoiraud:
> Kinda curious if it handles leap centuries... since we just had one in 2000.

I use the isleap method from the calendar library. I expect the standard function is correct about centuries and multiple of 400 years.

gramps/plugins/drawreport/calendarreport.py
```diff
@@ -335,6 +337,16 @@ def collect_data(self):
 
                     prob_alive_date = Date(self.year, month, day)
 
+                    # if birth on february, 29, we must choose
```
Nick-Hall, line 340: This code is used 4 times, which indicates that it should be made into a utility function.

gramps/gui/configure.py
```diff
@@ -1324,6 +1324,15 @@ def add_data_panel(self, configdialog):
         grid.attach(lwidget, 1, row, 1, 1)
         grid.attach(obox, 2, row, 2, 1)
 
+        row += 1
+        # Birthday on february 29
```
Nick-Hall, line 1328: Do we need a third option of not displaying the birthday?

Nick-Hall (changes requested):

gramps/gui/configure.py
SNoiraud: > Do we need a third option of not displaying the birthday?

This is only to allow the user to display the birthday on February 28 or March 1 for non leap years

gramps/plugins/drawreport/calendarreport.py
SNoiraud: Do I add this in gramps/gen/lib/date.py ?

Nick-Hall: Yes.  I was thinking something along the lines of:

`day, month = birth_date.anniversary(this_year)`

We can probably code just for the Gregorian calendar to start with.  Other calendars can be added later.

gramps/gui/configure.py
Nick-Hall: Perhaps some people would prefer the existing functionality of only displaying the birthday every 4 years.  It does seem unlikely though.

jralls: [Like this guy?](https://www.youtube.com/watch?v=W7lEunJvRUU)

SNoiraud: Instead of a binary choice, I could set a 0228, 0301 and 0229 choice

SNoiraud: This is not only for birth but also for death and marriage in calendars.

Nick-Hall:
@SNoiraud The duplicate code needs to be put in a utility function. There is also a conflict to resolve.

Nick-Hall:
@SNoiraud I suggest a new `anniversary(year)` date method, to return the day and month of the anniversary of a date in a given year.

For example:  `Date(1904,2,29).anniversary(1910)` would return (28, 2), (1, 3) or (29, 2) depending on the preference settings.

I have added a commit with the function and associated unit tests.  Run them with:

`GRAMPS_RESOURCES=. python -m unittest gramps.gen.lib.test.date_test.AnniversaryDateTest`

Let me know if this is OK.

Nick-Hall (approved):

Nick-Hall:
Rebased.

PR #1319, opened 2021-12-28
Calendar gramplet: add the name of the days.

SNoiraud:
The first day is define according on the LANG

https://gramps.discourse.group/t/built-in-calendar-gramplet/2080

Fixes [#12515](https://gramps-project.org/bugs/view.php?id=12515)

### DIFF
--- a/gramps/plugins/gramplet/calendargramplet.py
+++ b/gramps/plugins/gramplet/calendargramplet.py
@@ -50,6 +50,7 @@ class CalendarGramplet(Gramplet):
         self.gui.calendar.connect('day-selected-double-click',
                                   self.double_click)
         self.gui.calendar.set_display_options(
+            Gtk.CalendarDisplayOptions.SHOW_DAY_NAMES |
             Gtk.CalendarDisplayOptions.SHOW_HEADING)
         self.gui.get_container_widget().remove(self.gui.textview)
         vbox = Gtk.Box(orientation=Gtk.Orientation.VERTICAL)


SNoiraud:
In france, the first day is monday:
FR12515 fr_FR 1

in the US, it is sunday:
FR12515 en_US 2

in portugal:
FR12515 pt_PT 4

in brazil:
FR12515 pt_BR 3

SNoiraud:
And for an arabic language (ar_AE):
FR12515 ar_AE 5


The first day of the week is set in Gtk po file:
#. Translate to calendar:week_start:0 if you want Sunday to be the
#. * first day of the week to calendar:week_start:1 if you want Monday
#. * to be the first day of the week, and so on.
#.
#: gtk/gtkcalendar.c:853
msgid "calendar:week_start:0"
msgstr "calendar:week_start:1"

example: for ar language we have 
msgstr "calendar:week_start:7"

SNoiraud:
And for an arabic language (ar_AE):
FR12515 ar_AE 5

To choose the first week day, you have 3 methods:

- import locale then locale.first_weekday
- If the first method doesn't work: import calendar then calendar.getfirstweekday()
- Then the Gtk method :

The first day of the week is set in the Gtk po file:
#. Translate to calendar:week_start:0 if you want Sunday to be the
#. * first day of the week to calendar:week_start:1 if you want Monday
#. * to be the first day of the week, and so on.
#.
#: gtk/gtkcalendar.c:853
msgid "calendar:week_start:0"
msgstr "calendar:week_start:1"

example: for ar language we have 
msgstr "calendar:week_start:7"

emyoulation:
This is excellent information.  Thank you.

A question that affects when I should update the documentation.  Does the `msgid "calendar:week_start:0"` set in Gtk po file affect the column order of Gtk.Calendar widget in the Calendar.py gramplet without the `SHOW_DAY_NAMES` modification?

If so, I should add a notation to the 5.1 Calendar Gramplet docs.  If not, it should wait for the 5.2 release. Either way, should a Translator notice should go out to check that the `calendar:week_start:` setting has been set for their language?

emyoulation:
Patrice verified that the patch adapts in 5.1.4 for the Monday week start when using French.  The wiki has been updated to point out the patch availability.

PR #1290, opened 2021-07-28
Place editor, lat and long text are swapped

SNoiraud:
Fixes #012374

PR #1320, opened 2021-12-28
Fix wrong order for lat/lon in set_latlongitude

zertrin:
Bug was introduced in pull request !1278

Before the fix this resulted in this:

image


zertrin:
Ah snap, this is a duplicate pull request of #1290 

Closing this one in favor of the other one.

PR #1322, opened 2022-01-06
Family timeline error when dates are not Gregorian

SNoiraud:
Fixes [#12525](https://gramps-project.org/bugs/view.php?id=12525)

### DIFF
--- a/gramps/gui/editors/displaytabs/eventrefmodel.py
+++ b/gramps/gui/editors/displaytabs/eventrefmodel.py
@@ -180,6 +180,11 @@ class EventRefModel(Gtk.TreeStore):
                              event.get_type() == EventType.DEATH)):
                 return ""
             else:
+                # bug with french calendar events. Convert if not gregorian
+                if date.calendar != Date.CAL_GREGORIAN:
+                    date = date.to_calendar("gregorian")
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Both arguments of a `Span` are already converted to the Gregorian calendar.  See [line 80](https://github.com/gramps-project/gramps/blob/f3e536da6c301b2eab24f133066add18524644a6/gramps/gen/lib/date.py#L80) in the date code.  I'm not sure why this fix necessary.

SNoiraud:
> Both arguments of a `Span` are already converted to the Gregorian calendar. See [line 80](https://github.com/gramps-project/gramps/blob/f3e536da6c301b2eab24f133066add18524644a6/gramps/gen/lib/date.py#L80) in the date code. I'm not sure why this fix necessary.

Yes, I saw that but when we have a french republican date in one event, this is not done. I think we have the same problem when we have different calendars in the event list.

Nick-Hall:
@SNoiraud The actual bug was with the way the dates were being swapped for negative spans.  The swapped dates were taken from the unconverted parameters.  I have created PR #1332 with an alternative fix.  Please test if you have time, but it works for me.

PR #1324, opened 2022-01-17
Fix bug 12535: "Descendants" is misspelled

hgohel:
Fix bug [12535](https://gramps-project.org/bugs/view.php?id=12535)


Nick-Hall:
Rebased.

Nick-Hall:
@hgohel Thanks you.

PR #1329, opened 2022-01-29
Adding Turkish Holidays

BorATICI:


BorATICI:
Adding Turkish Holidays

Nick-Hall (changes requested):
I notice that this is causing a few unit tests to fail.

The first line should read `<country name="Turkey">`.  The country name will be flagged for translation.  We used "_name" here in v5.1, but this has changed to "name".

Elsewhere you also need to use "name" rather than "_name".

(I tried to comment in the code, but GitGub failed to save my comments).

BorATICI:
I sent new PR. #1330

PR #1332, opened 2022-01-29
Fix negative Span when dates are not Gregorian

Nick-Hall:
A replacement fix for PR #1322.

Fixes [#12525](https://gramps-project.org/bugs/view.php?id=12525).


SNoiraud:
OK. This works for me. I didn't see it was a negative value.
I'll close my PR.

PR #1333, opened 2022-02-04
Fix place object element order in DTD and RNG schemas

romjerome:
Element placeobj content does not follow the DTD and RNG, expecting (ptitle? , pname+)

### DIFF
--- a/data/grampsxml.dtd
+++ b/data/grampsxml.dtd
@@ -251,7 +251,7 @@ PLACES
 
 <!ELEMENT places (placeobj)*>
 
-<!ELEMENT placeobj (ptitle?, pname+, code?, coord?, placeref*, location*,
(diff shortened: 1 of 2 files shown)


romjerome:
Have a look at gramps/plugins/export/exportxml.py, under def write_place_obj() section.
Exporting 'code' before 'pname', schemas are listing 'pname' before 'code'.

Gramps xml file with code value on Place Object will be invalid according to current schema and DTD (documentation, no real validation). Anyway, a bug report has been filled:
https://gramps-project.org/bugs/view.php?id=12500

romjerome:
@Nick-Hall 
Can I commit the change? (squash and merge)

Nick-Hall:
The fix looks good.  I merged it so it can get in the next release.

PR #1337, opened 2022-02-12
Fix month&day for French date format

khrys63:
In French, Day & month are on 2 characters
01 for January
01 for the first day of the month

### DIFF
--- a/gramps/gen/datehandler/_date_fr.py
+++ b/gramps/gen/datehandler/_date_fr.py
@@ -252,9 +252,6 @@ class DateDisplayFR(DateDisplay):
                "MOI Jour, Année",)  # 7
         # this definition must agree with its "_display_gregorian" method
 
-    def right(self, s, amount):
-        return s[-amount:]
-
     def _display_gregorian(self, date_val, **kwargs):
         """
         display gregorian calendar date in different format
@@ -276,8 +273,8 @@ class DateDisplayFR(DateDisplay):
                 if date_val[0] == date_val[1] == 0:
                     value = str(date_val[2])
                 else:
-                    value = self.dhformat.replace('%m', self.right('0'+str(date_val[1]),2))
-                    value = value.replace('%d', self.right('0'+str(date_val[0]),2))
+                    value = self.dhformat.replace('%m', str(date_val[1]).zfill(2))
+                    value = value.replace('%d', str(date_val[0]).zfill(2))
 
                     # base_display :
                     # value = value.replace('%Y', str(abs(date_val[2])))
--- a/gramps/gen/datehandler/_date_fr.py
+++ b/gramps/gen/datehandler/_date_fr.py
@@ -254,7 +254,7 @@ class DateDisplayFR(DateDisplay):
 
     def right(self, s, amount):
         return s[-amount:]
-    
+
     def _display_gregorian(self, date_val, **kwargs):
         """
         display gregorian calendar date in different format
--- a/gramps/gen/datehandler/_date_fr.py
+++ b/gramps/gen/datehandler/_date_fr.py
@@ -5,6 +5,7 @@
 #
 # Copyright (C) 2004-2006  Donald N. Allingham
 # Copyright (C) 2012       Mathieu MD
+# Copyright (C) 2022       Christophe aka khrys63
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
--- a/gramps/gen/datehandler/_date_fr.py
+++ b/gramps/gen/datehandler/_date_fr.py
@@ -251,6 +251,9 @@ class DateDisplayFR(DateDisplay):
                "MOI Jour, Année",)  # 7
         # this definition must agree with its "_display_gregorian" method
 
+    def right(self, s, amount):
+        return s[-amount:]
+    
     def _display_gregorian(self, date_val, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


khrys63:
With the fix, FR date are correct
image

Nick-Hall (commented):
You can use the python function `zfill` to pad with leading zeros.  For example, `str(date_val[0]).zfill(2)`.

Alternatively, it may be neater to use a format string for the whole expression. Something like `value = '{:02d}/{:02d}/{:d}'.format(date_val[0], date_val[1], date_val[2])`

Do other French speakers agree with this change?

SNoiraud:
I am french and I never use that. 6/5/2021 is a good date for me. Where did you find this rule ? is it an iso rule ?

If you want to do that there are better ways to do that. You can use %02d.

personally, I don't agree with that.

romjerome:
True, maybe a consistency issue, as the format use a custom display "J/M/A" related to system ("Défaut système"). About the way to do change on display,
either via 'zfill' or '%02d', I just have no idea... Sorry!
 ```
        # do have an explicit meaning: "System default (format)"
        example = self.dhformat
        example = example.replace('%d', "J")
        example = example.replace('%m', "M")
        example = example.replace('%Y', "A")
```

Nick-Hall:
Is _dd/mm/yyyy_ common in Canada, Belgium or Switzerland?

We could always add an extra format if it is useful.

emyoulation:
Is this reformatting of the day (to add leading zero) a workaround? Specifically, so that sorting works with the formatted Text string representation of a date? 


There have been recurrent "row order sorting" glitches in various sections of Gramps. Mostly due related to table designs based on the formatted date string instead of the value. Because the default date format in yyyy-mm-dd (as long as they are anno Domini dates) in in the same order as the value, this often not noticed with superficial testing. 


https://gramps-project.org/bugs/view.php?id=10893


We just tried a hack changing the Citation selector from the Tree model to List model & a descending Modification Date sort order. And that also did not sort the Modification Date (nor the added Citation Date) columns on value. 


It looks like the citation tree & base models had a MUCH more extensive ordering map than the list model.


https://github.com/gramps-project/gramps/blob/master/gramps/gui/views/treemodels/citationlistmodel.py 


Is there a snippit that specifically gives examples of compensating for sorting being different than Formatting? 


Some common special case are very easy to neglect: case insensitive sorting, formatted values (dates, GPS coordinates, name field concatenations), tree subsorts (name display preferences conflicting with Grouped and Grouped-As subsorts), whitespace trimming before sorting.

khrys63:
> Is _dd/mm/yyyy_ common in Canada, Belgium or Switzerland?
> 
> We could always add an extra format if it is useful.

yes, good idea

khrys63:
An exemple on an official document
phpCDwGn0

romjerome:
> Is this reformatting of the day (to add leading zero) a workaround? Specifically, so that sorting works with the formatted Text string representation of a date? 

I do not remember well (have a look at previous .sh script for localized date handlers), but the format with the dot after the day is for switzerland (cultural issue: some locales on east of the Rhine). I did not test all locales using "french" date formats. Tests passed after these customizations (set system/locale as dependency).

The extra parsing for the day was an old bug (+/- 10 days by switching between calendars).
Maybe fixed on last versions (gramps, os, python, system). 
The extra parsing for months is for generating a date object instead of a textual value.

romjerome:
> Mostly due related to table designs based on the formatted date string instead of the value.
> ..
> We just tried a hack changing the Citation selector from the Tree model to List model 

Does change column really look at date handlers? There is no gramps' date object, here.

romjerome:
> We just tried a hack changing the Citation selector from the Tree model to List model & a descending Modification Date sort order

If useful, I can try to compare/update/upgrade https://github.com/romjerome/gramps_cryptic_rebase/tree/gep-041-new-selector ?

emyoulation:
> > We just tried a hack changing the Citation selector from the Tree model to List model & a descending Modification Date sort order
> 
> If useful, I can try to compare/update/upgrade https://github.com/romjerome/gramps_cryptic_rebase/tree/gep-041-new-selector ?
> 
> 

The discussion was in [Discourse](https://gramps.discourse.group/t/sorting-dialogs-by-last-update-date-keeping-last-used-sort-order/2238/8)

The modified date column was sorting incorrectly.  But so I could locate a Periodical/Serial more easily, I added a publication date. It makes discriminating between newspaper birth announcements, wedding announcements, obits & funeral notices easier.

<code>
def get_column_titles(self): return [ (_('Source: Title'), 150, BaseSelector.TEXT, 7), (_('Source: Author'), 150, BaseSelector.TEXT, 9), (_('Date'), 150, BaseSelector.TEXT, 2), (_('Source: Title or Citation: Volume/Page'), 350, BaseSelector.TEXT, 0), (_('ID'), 75, BaseSelector.TEXT, 1), (_('Last Change'), 150, BaseSelector.TEXT, 6), </code>

Nick-Hall:
@SNoiraud It seems that we don't want to change the default format.  Would adding an additional French date format be acceptable?

SNoiraud:
@Nick-Hall Yes, I think it would be acceptable and certainly for other countries

Nick-Hall:
I'm closing this now as it is not acceptable to the French speakers who commented.

@khrys63 Please feel free to create a new PR which adds an additional French date format as suggested.

PR #1340, opened 2022-02-15
Add xdg-utils to no longer needed list in README

Nick-Hall:
The dependency was removed in PR #1077.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,9 @@ Prerequistes required for the following Addons to work:
 
 No longer needed:
 -----------------
+* Since Gramps 5.2:
+   **xdg-utils**
+
 * Since Gramps 4.2:
    **gir-webkit**
 


lordemannd:
does prerequisites checker also need changes for this?

wroldwiedbwe:
> does prerequisites checker also need changes for this?

@lordemannd  Probably you can see it mentioned in the code [here](https://github.com/gramps-project/addons-source/blob/f3f3845e7879e090208723d652efdd3461ad9185/PrerequisitesCheckerGramplet/PrerequisitesCheckerGramplet.py#L655)

emyoulation:
If someone works on the Prerequisites Checker gramplet, it could use some tweaks to eliminate some ambiguity.

Some of the lines include a "Passed: version x.xx or greater installed".
Others have a "________ : not found" or "________ : not installed" (which be more symmetrical if these were preceded with a "Failed: ")

But the "Optional packages required by Third-party Addons" is difficult for causal users to interpret the pass/fail status. Particularly difficult are those that have multiple requirements.

But

PQYPLZXHGF:
@Nick-Hall .  👍😊

PR #1341, opened 2022-02-16
Crash in narrative web if invalid latitude/longitude

SNoiraud:
We add a message after the latitude/longitude to show the value and the syntax.
For translators, I used the same message as in editplace.py and editplaceref.py, so no need to translate new strings.

Fixes [#12565](https://gramps-project.org/bugs/view.php?id=12565)

### DIFF
--- a/gramps/plugins/webreport/basepage.py
+++ b/gramps/plugins/webreport/basepage.py
@@ -2878,6 +2878,16 @@ class BasePage:
             tbody += trow
 
         data = place.get_latitude()
+        v_lat, v_lon = conv_lat_lon(data, "0.0", "D.D8")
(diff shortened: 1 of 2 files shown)


Nick-Hall:
@SNoiraud There were conflicts with the pylint improvements so I dropped that commit.

I suggest that we use [black](https://black.readthedocs.io/en/stable/index.html) as a code formatter.  From the look of your code, you may already use it.  I'll discuss this on the mailing list and maybe we can format our code in bulk.

Is this ready to merge?  I don't use the narrative web report myself.

SNoiraud:
@Nick-Hall Is this ready to merge?
Yes

PR #1342, opened 2022-02-19
Common calendar on Statistics chart

romjerome:


romjerome:
We can have a [mixup of dates with different calendars set on Statistics Chart.](https://gramps-project.org/bugs/view.php?id=12571)

Nick-Hall:
Rebased.

I get no charts in the report if I set the output calendar in the preferences to French Republican or Hebrew.

PR #1350, opened 2022-02-25
Skip some tests requiring an English locale on non-English systems

umlaeute:
Closes: https://gramps-project.org/bugs/view.php?id=12578

(the full reasoning is explained in the bugreport)

The check whether the tests should be skipped is a bit simplistic:

~~~python3
from ....utils.grampslocale import GrampsLocale
if not GrampsLocale().language[0].startswith("en"):
    self.skipTest("skipping test for non-English language %s" % (GrampsLocale().language,))
~~~

possible improvements:
- catch `IndexError`
- what if `GrampsLocale().language` contains multiple languages? is it sufficient if *any* of them is English? all? just the first?


Nick-Hall:
Alternatively we could modify the tests to work with non-English locales.   See PR #1352.

Nick-Hall:
OK.  I'll close this and merge PR #1352.

PR #1359, opened 2022-03-07
Cannot show spouse's name in Records Report

SNoiraud:
Fixes [#12391](https://gramps-project.org/bugs/view.php?id=12391)

### DIFF
--- a/gramps/plugins/lib/librecords.py
+++ b/gramps/plugins/lib/librecords.py
@@ -314,13 +314,16 @@ def find_records(db, filter, top_size, callname,
         if mother is None:
             continue
 
-        name = StyledText(trans_text("%(father)s and %(mother)s")) % {
-                'father': _get_styled_primary_name(father, callname,
-                                                   trans_text=trans_text,
-                                                   name_format=name_format),
-                'mother': _get_styled_primary_name(mother, callname,
-                                                   trans_text=trans_text,
-                                                   name_format=name_format)}
+        father_name = _get_styled_primary_name(father, callname,
+                                               trans_text=trans_text,
+                                               name_format=name_format)
+        mother_name = _get_styled_primary_name(mother, callname,
+                                               trans_text=trans_text,
+                                               name_format=name_format)
+
+        name = StyledText(trans_text("%(father)s and %(mother)s"))
+        name = name.replace('%(father)s', father_name)
+        name = name.replace('%(mother)s', mother_name)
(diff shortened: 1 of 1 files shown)


lordemannd:
works for me, tested and working with Gramps 5.1.5 using single character parent surnames (chinese and latin)

Nick-Hall:
Added a patch that retains underlined call names.

lordemannd:
just a note:
not very important and  probably unrelated to these changes, I do see a strangeness in the Call name appearance in PDF.
[record_report-03.pdf](https://github.com/gramps-project/gramps/files/8200617/record_report-03.pdf)

notice the 1st listed person, surname "B", given name "ff".
in the given name, both "f"s are underlined. 
(I used "ff" as given, and "f" as the call name in the DB "issue-12391-v4.gramps" from the mantis bug report)

is some smart software converting the "ff" into a ligature (that is, making it a single character "ﬀ") before the underline is applied?

Nick-Hall:
@lordemannd Interesting.  This seems to be a Pango / Cairo issue.

I wrote a simple test script:

```
import cairo
import gi
gi.require_version('Pango', '1.0')
gi.require_version('PangoCairo', '1.0')
from gi.repository import Pango, PangoCairo

surface = cairo.PDFSurface('test.pdf', 100, 100)
context = cairo.Context(surface)
layout = PangoCairo.create_layout(context)
layout.set_markup('<u>f</u>f')
PangoCairo.show_layout(context, layout)
surface.show_page()
```

SNoiraud:
This works correctly only for PRINT, HTML, PDF and POSTSCRIPT output format. 
underlining doesn't work for all other formats

Nick-Hall:
The ODF, RTF and LaTeX document generators do not implement the [`start_underline()`](https://github.com/gramps-project/gramps/blob/master/gramps/gen/plug/docgen/textdoc.py#L295) and [`stop_underline()`](https://github.com/gramps-project/gramps/blob/master/gramps/gen/plug/docgen/textdoc.py#L302) methods.

Nick-Hall:
We should be able to use the `add_markup_from_styled` method.  The `OdfBackend` defines supported markup.

I've only briefly tested this with the Records report.

SNoiraud:
It's OK for me.

Nick-Hall:
Split the ODF backend changes into PR #1369.

PR #1362, opened 2022-03-09
Test: Use temporary directory to test bsddb

a-andre:


Nick-Hall:
We are dropping support for BSDDB in this release.

a-andre:
The main motivation for this PR is to get rid of `HOME_DIR`, in preparation of XDG Base Directory Specification support.
So maybe it can be merged anyway, if removing BSDDB code is delayed.

PR #1363, opened 2022-03-09
Add support for thumbnailer plugins

Nick-Hall:


Nick-Hall:
@a-andre Thanks for the review.

I decided to use `GLib.get_user_data_dir()` as you suggested, and also `GLib.get_system_data_dirs()`.

A few other changes:

* Improved error checking in the Gnome thumbnailer.
* Added the "order" parameter to thumbnail plugin registration.  The Image thumbnailer now runs first.
* If a thumbnailer fails, other thumbnailers are run if available.

PR #1364, opened 2022-03-13
Spelling: "vCard" in importvcard.py

comradekingu:
As per https://datatracker.ietf.org/doc/html/rfc6350#section-6.1.4

Also 
>LOG = logging.getLogger(".ImportVCard")

But that might cause problems?


Nick-Hall:
The name of the logger isn't particularly important.  It is only used for debugging purposes.

PR #1368, opened 2022-03-17
Support XDG base directory specification

a-andre:
Fixes [#8025](https://gramps-project.org/bugs/view.php?id=8025)

### DIFF
--- a/gramps/cli/argparser.py
+++ b/gramps/cli/argparser.py
@@ -48,7 +48,8 @@ from glob import glob
 #
 #-------------------------------------------------------------------------
 from gramps.gen.const import (LONGOPTS, SHORTOPTS, USER_PLUGINS, VERSION_DIR,
-                              HOME_DIR, THUMB_DIR, USER_CSS)
+                              USER_CACHE, USER_CONFIG, USER_DATA, THUMB_DIR,
+                              USER_CSS)
 from gramps.gen.utils.cast import get_type_converter
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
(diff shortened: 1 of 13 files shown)


Nick-Hall:
Are `USER_DATA`, `USER_CONFIG` and `USER_CACHE` set to sensible paths on Windows?

QuLogic:
Since this is using the GLib functions:

data:
> On Windows it follows XDG Base Directory Specification if `XDG_DATA_HOME` is defined. If `XDG_DATA_HOME` is undefined, the folder to use for local (as opposed to roaming) application data is used instead. See the [documentation for `FOLDERID_LocalAppData`](https://docs.microsoft.com/en-us/windows/win32/shell/knownfolderid). Note that in this case on Windows it will be the same as what `g_get_user_config_dir()` returns.

config:
> On Windows it follows XDG Base Directory Specification if `XDG_CONFIG_HOME` is defined. If `XDG_CONFIG_HOME` is undefined, the folder to use for local (as opposed to roaming) application data is used instead. See the [documentation for `FOLDERID_LocalAppData`](https://docs.microsoft.com/en-us/windows/win32/shell/knownfolderid). Note that in this case on Windows it will be the same as what `g_get_user_data_dir()` returns.

cache:
> On Windows it follows XDG Base Directory Specification if `XDG_CACHE_HOME` is defined. If `XDG_CACHE_HOME` is undefined, the directory that serves as a common repository for temporary Internet files is used instead. A typical path is `C:\Documents and Settings\username\Local Settings\Temporary Internet Files`. See the [documentation for `FOLDERID_InternetCache`](https://docs.microsoft.com/en-us/windows/win32/shell/knownfolderid).

which all seem reasonable?

Nick-Hall:
Yes, they look reasonable.

emyoulation:
There is a discussion about the drawbacks of using %localappdata% instead of %appdata% in started in Discourse during the Beta Test cycle:
https://gramps.discourse.group/t/aio-move-of-user-directory-from-appdata-to-localappdata/4156/9

PR #1371, opened 2022-03-19
Use ObjEntry to select a person as an association

Nick-Hall:
Uses existing code to make the user interface consistent with
similar object selections.

Adds person add and remove buttons.

Replaces PR #1177.

### DIFF
--- a/gramps/gui/editors/editpersonref.py
+++ b/gramps/gui/editors/editpersonref.py
@@ -48,9 +48,9 @@ from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.sgettext
 from gramps.gen.display.name import displayer as name_displayer
 from .editsecondary import EditSecondary
+from .objectentries import PersonEntry
 from gramps.gen.lib import NoteType
 from ..widgets import MonitoredEntry, PrivacyButton
-from ..selectors import SelectorFactory
 from .displaytabs import CitationEmbedList, NoteTab
 from ..glade import Glade
 from ..ddtargets import DdTargets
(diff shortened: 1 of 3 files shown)


Nick-Hall:
Testing needs to check the functionality of selecting a person in the Associations dialog.

Selecting a person in this dialog should now work like any other object selector.  Check all the buttons and also drag & drop selection.

emyoulation:
How many of the following Bug/Feature reports can be partially or fully addressed with this change:

related to 	[0007050](https://gramps-project.org/bugs/view.php?id=0007050)	confirmed	Associations not shown at both participants  
related to 	[0011102](https://gramps-project.org/bugs/view.php?id=0011102)	confirmed	[Associations editor] Change the Association textbox to an editable list  


related to 	[0005709](https://gramps-project.org/bugs/view.php?id=0005709)	confirmed	No easy way to go to an associated person  
related to 	[0008141](https://gramps-project.org/bugs/view.php?id=0008141)	confirmed	Usability of Associations of Person  
related to 	[0002378](https://gramps-project.org/bugs/view.php?id=0002378)	confirmed	[Associations]Improve the linkage of Witnesses to a marriage event  
related to 	[0011883](https://gramps-project.org/bugs/view.php?id=0011883)	confirmed	[Person Reference Editor] Should remember the last Association entered.  

related to 	[0008090](https://gramps-project.org/bugs/view.php?id=0008090)	confirmed	add associations to the not related tool  
related to 	[0007567](https://gramps-project.org/bugs/view.php?id=0007567)	confirmed	save more settings to config file, as default in gramps startup  (related to 	[0011883](https://gramps-project.org/bugs/view.php?id=0011883))

Features:
* Add drag'n'drop target for Person section (should be full width of the row! Not just the text witch of an existing person.  include the "Person label and the Select/Share button. A small target in the middle of a dialog is hard to expose with spawned dialogs.) 
* add Edit icon to next to 'Share" icon on Person selection
* Make the Association type a standard "Custom Type" gui element with pop-up list
* Add optional Symmetrical/Reciprocal creation of Association  

The v5.1.5-1 Windows  version of the dialog looks like:
image

Since an Association is a Relationship, the dialog could be  based on the Relationship Editor dialog (but where the Partner2 position is based on the Person of the parent dialog that spawned the Person Reference/Association Editor and cannot be changed:. It is only displayed as a reference and as an access point for the Reciprocal/Symetrical (or non-existent opposite Association) And which gives the option to create one or to open the complementary Person Reference Editor. 

(The "Association Information" section of the header corresponds to the Family Editor. But it contained things that don't exist yet.  It might be a future expansion when an ID as assigned to Associations and should a feature is added to auto-create/remove complementary Association.) 

image

emyoulation:
The tooltips for the Association suggest using the Role in a shared Event rather than an Association wherever possible.  Perhaps we should start using the more generic "Role" label for the Association value?

The default "Godfather" is probably a bad Association example. It could be a gender-neutral "Godparent" Role in a shared Christening Event. (The secular incarnation is a "Guardian", "Conservator" and is a Role assigned during Probate event and tentatively designated in a will/living will).

Namesake and Eponym may be a better default example.  e.g., "Brig. Gen. Cornelius Vanderbilt III (1873-1942)" is a namesake of his eponymous great-grandfather, "Cdre. Cornelius Vanderbilt (1794-1877)"

possible revised tooltip for the Glade file:
C:\Program Files\GrampsAIO64-5.1.5\gramps\gui\glade\editpersonref.glade
<property name="tooltip_text" translatable="yes">Description of the association, e.g., Namesake, Coworker, Friend, Surrogate ...
Note: Where possible, use Roles in shared Events instead of this feature. Associations have Roles but Events also connect the Role to specific time frames, places or occasions.</property>

URl linked to Help button:
https://gramps-project.org/wiki/index.php?title=Gramps_5.1_Wiki_Manual_-_Entering_and_editing_data%3A_detailed_-_part_1#Person_Reference_Editor

cdhorn:
@Nick-Hall this worked fine for me, I think you should merge.

Nick-Hall:
@cdhorn Thanks for reminding me.  I have a couple of weeks where I can work on Gramps now.

PR #1372, opened 2022-03-30
SrcAttribute object has no attribute has_citations

SNoiraud:
Fixes [#12599](https://gramps-project.org/bugs/view.php?id=12599)


moorob:
Wasn't this added as part of #11372?  I remember having to add this line.

SNoiraud:
> Wasn't this added as part of #11372? I remember having to add this line.

Yes, you added this line, but Attributes have no sources citations. This is why we get this error.

moorob:
Since not using has_citations, I removed the second bool from

Gtk.ListStore.__init__(self, str, str, bool, bool, object)

Haven't seen any problems since.

PR #1375, opened 2022-06-08
Improvement: Add stock_category_icon support

cdhorn:
This adds support for views that create new non-default categories to provide an icon for the category by setting the "stock_category_icon" attribute in the plugin registry file.

### DIFF
--- a/gramps/gen/plug/_pluginreg.py
+++ b/gramps/gen/plug/_pluginreg.py
@@ -373,6 +373,8 @@ class PluginData:
     .. attribute:: viewclass
        A class of type ViewCreator that holds the needed info of the
        view to be created: icon, viewclass that derives from pageview, ...
+    .. attribute:: stock_category_icon
+       The icon used for the view category if there is none
     .. attribute:: stock_icon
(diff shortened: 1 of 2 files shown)


emyoulation:
Is there an opportunity to change icon sets with Theme controls?

Nick-Hall:
We already have the [`get_stock()`](https://github.com/gramps-project/gramps/blob/master/gramps/gui/views/pageview.py#L391) method in the `PageView` class for this.  I'm not sure why we have both mechanisms.

Nick-Hall:
>Is there an opportunity to change icon sets with Theme controls?

I'm not sure if a theme icon would be used in preference to a Gramps icon with the same name.

cdhorn:
If I recall the categories are added and their icons initialized when the view plugin is loaded prior to view initialization. So when I added the tag list and card views I could not set the icon for the new category.

Nick-Hall:
Yes.  The views are registered, but not loaded at the point the navigation bar is constructed.

We definitely want addons to be able to create a new category if necessary.

PR #1376, opened 2022-06-08
Add Proof Argument Note Type

cdhorn:
Adds a new note type for Proof Arguments so they can easily be identified in a standardized manner


cdhorn:
Stumped on the failing test at the moment which seems totally unrelated to such a trivial change. Oddly my machine fails faster and harder than this run on Ubuntu 18.04 does. On my machine it hits cleanup_missing_photos, cycles through, and when it returns from the call it appears to exit immediately and never calls the cleanup_deleted_name_formats step. I would guess the illegal characters the test injects are causing the issue. Fedora 35 Python 3.10 at the moment as I've not updated to Fedora 36 yet.

PR #1377, opened 2022-06-08
Check that view exists before calling post_create method

SNoiraud:
Avoids 'NoneType' object has no attribute 'post_create' error.

Fixes [#12638](https://gramps-project.org/bugs/view.php?id=12638).

This patch can be applied on 5.1 or 5.2


lordemannd:
Tested on Gramps 5.1.5 on Windows, using GUI. 
fix is working, was not able to reproduce the issue following the steps in the problem report.
https://gramps-project.org/bugs/view.php?id=12638#c63996

PR #1378, opened 2022-06-09
Improvement: Add optional support for comments section at top of ini in the save method

cdhorn:
This adds a comments= keyword to the ConfigManager save method that can be set to a list of strings to be written out to the init file after the initial two header comment lines and before the data sections.


cdhorn:
Update should bump pylint score from 7.88 to 9.76 along with some other minor optimizations.

Nick-Hall:
This seems like a reasonable enhancement.

PR #1380, opened 2022-06-15
Fix a wrong operator bug in the web calendar report

LostBenjamin:
Hi,

This pull request is a fix to a potential wrong operator bug at `gramps/plugins/webreport/webcal.py`. Please check the changes.

Best,
Jingxuan


SNoiraud:
You could do the same in the calendar.py file at line 958

Nick-Hall:
Yes.  The code should check to see if `birth_date` is empty.

Comparing with an empty date using `!=` will work, but `is not` will fail because one empty date will not be the same instance as another.

Alternatively, `not birth_date.is_empty()` will also work.

PR #1385, opened 2022-07-08
FIX: _LEVEL_NAME not defined in get_plural_relationship_string

ennoborg:
I found this when experimenting with the Consanguinity Gramplet, and it seems like a bit of code that is not tested, and is almost never hit by any other code, like the Deep Connections Gramplet.

A quick glance suggests that the error is exclusive to the NL version of this plugin.

### DIFF
--- a/gramps/gen/datehandler/_dateparser.py
+++ b/gramps/gen/datehandler/_dateparser.py
@@ -627,6 +627,10 @@ class DateParser:
             y = self._get_int(groups[0])
             m = self._get_int(groups[3])
             d = self._get_int(groups[4])
+            if d > 99:
+                yy = d
+                d = y
+                y = yy
             if groups[2] and julian_valid((d, m, y + 1)):
                 return (d, m, y + 1, True) # slash year
             if check is None or check((d, m, y)):
--- a/gramps/gui/listmodel.py
+++ b/gramps/gui/listmodel.py
@@ -233,7 +233,7 @@ class ListModel:
         new_value = not self.model[path][col]
         self.model[path][col] = new_value
         if col in self.function:
-            self.function[col](int(path), new_value)
+            self.function[col](path, new_value)
 
     def __edited_cb(self, cell, path, new_text, col):
         """
@@ -241,7 +241,7 @@ class ListModel:
         """
         self.model[path][col] = new_text
         if col in self.function:
-            self.function[col](int(path), new_text)
+            self.function[col](path, new_text)
 
     def unselect(self):
         """
--- a/gramps/gui/views/treemodels/eventmodel.py
+++ b/gramps/gui/views/treemodels/eventmodel.py
@@ -160,6 +160,8 @@ class EventModel(FlatBaseModel):
             date_str = get_date(event)
             if date_str != "":
                 retval = escape(date_str)
+            else:
+                retval = ''
             if not get_date_valid(event):
                 return INVALID_DATE_FORMAT % retval
             else:
--- a/gramps/plugins/export/exportgedcom.py
+++ b/gramps/plugins/export/exportgedcom.py
@@ -1499,32 +1499,6 @@ class GedcomWriter(UpdateCallback):
             self._writeln(level + 2, 'LATI', latitude)
             self._writeln(level + 2, 'LONG', longitude)
 
-        # The Gedcom standard shows that an optional address structure can
-        # be written out in the event detail.
-        # http://homepages.rootsweb.com/~pmcbride/gedcom/55gcch2.htm#EVENT_DETAIL
-        location = get_main_location(self.dbase, place)
-        street = location.get(PlaceType.STREET)
-        locality = location.get(PlaceType.LOCALITY)
-        city = location.get(PlaceType.CITY)
-        state = location.get(PlaceType.STATE)
-        country = location.get(PlaceType.COUNTRY)
-        postal_code = place.get_code()
-
-        if street or locality or city or state or postal_code or country:
-            self._writeln(level, "ADDR", street)
-            if street:
-                self._writeln(level + 1, 'ADR1', street)
-            if locality:
-                self._writeln(level + 1, 'ADR2', locality)
-            if city:
-                self._writeln(level + 1, 'CITY', city)
-            if state:
-                self._writeln(level + 1, 'STAE', state)
-            if postal_code:
-                self._writeln(level + 1, 'POST', postal_code)
-            if country:
-                self._writeln(level + 1, 'CTRY', country)
-
         self._note_references(place.get_note_list(), level + 1)
 
     def __write_addr(self, level, addr):
--- a/gramps/plugins/lib/libgedcom.py
+++ b/gramps/plugins/lib/libgedcom.py
@@ -3304,7 +3304,7 @@ class GedcomParser(UpdateCallback):
                         family_handle not in father.get_family_handle_list():
                     father.add_family_handle(family_handle)
                     self.dbase.commit_person(father, self.trans)
-                    self.__add_msg("Error: family '%(family)s' (input as"
+                    self.__add_msg("Warning: family '%(family)s' (input as"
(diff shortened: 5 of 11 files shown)


PQYPLZXHGF:
@Nick-Hall Personally I would not accept any PR request from @ennoborg as they have indicated they are using Chatgpt to create code on your [forum](https://gramps.discourse.group/t/introducing-uids-for-new-persons-and-families/3507/12) beware that sourcecode could possibly be from a non GPL licensed source fragment of code however well intended!

ennoborg:
I have better things to do than reacting to false accusations, especially for code that I pushed before I used ChatGPT.

ennoborg:
The code produced by ChatGPT has an MIT license.

Nick-Hall:
> @Nick-Hall Personally I would not accept any PR request from @ennoborg as they have indicated they are using Chatgpt to create code on your [forum](https://gramps.discourse.group/t/introducing-uids-for-new-persons-and-families/3507/12) beware that sourcecode could possibly be from a non GPL licensed source fragment of code however well intended!

Thanks for bringing this to my attention.  I have discussed this with other members of the team and can confirm that code generated by an AI will not be accepted.

I have updated the [Howto: Contribute to Gramps](https://gramps-project.org/wiki/index.php/Howto:_Contribute_to_Gramps) page in the wiki to clarify the situation.  In particular, a contributor should hold the copyright to their code, license it under GPLv2 or later, and must not use AI generators.

ennoborg:
Please note that this particular PR was created 8 months before I first used ChatGPT, so for this PR, the comment does not apply.

So @Nick-Hall are you're going to accept this, or not? If not, I will close this PR right away.

Nick-Hall:
@ennoborg If the code is original and you hold the copyright, then I don't see a problem.

I don't really understand the PR though, it seems to contain code unrelated to the fix described.

ennoborg:
The code in the commit, replacing _LEVEL_NAME with _ordinal_level, is my own. It fixes the Dutch relationship plug-in for an error that's always been there (_LEVEL_NAME is used, but not defined), but is not hit by the Deep Donnections Gramplet. It is triggered by the Consanguinity Gramplet however, so other Dutch speaking users of that Gramplet may be hit by it.

ennoborg:
It looks like I can better close this, because the PR has other commits on it too, including the lines generated by ChatGPT. I can refile it, by creating a branch for this specific error, if that helps.

ennoborg:
So @Nick-Hall , to bother you again. Can you advise on this? Should I have created a PR for a special branch that only contains this specific fix? I'm asking, because GitHub now advises you to delete this branch (in my repo!) and or even my whole fork! And I really don't want that.

Nick-Hall:
You don't have to delete your existing branch.

Create a new branch called `fix-level-name` or something similar, and then put your fix in it.  Then create a new PR for that branch.

emyoulation:
What will the policy be on AI proofreading code and the wiki?

Much like when calculators and word-processers availability affected industry, there were prohibitions that were implemented and which eroded over time.

(And remember when libraries of code were once prohibited as well? All code had to be written in-house. The OS wars blew down a lot of those ideas. Looking at the performance differences between Gramps on Windoze & Linux, it is arguable that humanity was enslaved in those battles.)

But we need the guidance to be clear.

ennoborg:
For the first, I don't trust AI, but I can't forbid anyone to let AI do some proofreading. It may find something, just like pylint, but it may also miss a few things, like pylint did last time I tried that.

For the wiki, I really have no clue at all. I'd love to have some AI to navigate that, but that's another issue.

ennoborg:
Please note that I just unsubscribed from this thread, because this PR is quite useless. I should never have made it, because there are too many commits on this branch, and only a few have a value for the community.

cdhorn:
This is an interesting topic.

Being a public project it is highly likely all the Gramps related repositories are in the training data and already contributed to the development of these models.

It is important to keep in mind these models are pure mathematical constructs. The input data does not exist in the model in that sense of the word at all.  And when it generates output it is not copying it from someplace else in the traditional sense at all either.

In some way these models can be thought of as being like a lossy compression algorithm because they can not always reproduce the original information that was input into them 100% accurately though they can often produce a reasonable facsimile.

So consider the questions around them then from an information theory perspective.  

A developer answering a question on stack exchange or anywhere for that matter may have worked on a lot of commercial and Open Source code over a period of time.

How does that differ from the training phase a model like ChatGPT this goes through?

All that information and experience is in their memory somehow, and in some manner that knowledge combines to produce the answers they provide to questions you may ask.  You have no idea of every line of code they ever read and what the licensing was around it or what particular peices of that information in their memory combined and contributed to their answer.

So if you make use of the solution they provided how does it differ from an answer generated from ChatGPT?

No answers needed, just thoughts in my head about it I wanted to share.

ennoborg:
I have the same thoughts, and like to add that part of my training comes from books, with copyright, and sample code.

jralls:
There's a difference between our training as engineers including the code that we've learned from over the years and the training of generative AI models: The code we used taught us how the language works to instruct the computer. While there are certainly idiomatic ways to implement any particular algorithm so that you and I would probably write a simple function exactly the same way, we'd combine those simple functions or algorithms differently to construct a complex function, and we'd combine complex functions differently to make a module.

Generative AI models don't learn in the human sense. If you ask one to write a complex function for you it won't generate an original expression, it will find a function from its training set that context tells it will answer your question and regurgitate it unchanged. [Getty vs. Stable Diffusion](https://www.theverge.com/2023/2/6/23587393/ai-art-copyright-lawsuit-getty-images-stable-diffusion) alleges that Stable Diffusion's supposedly generative model lifted whole images, copyright watermarks and all, and attempted to pass them off as original work.

For code the copyright issue is interesting because it's reasonable to assume that all of the code in the model's training set is from open-source repositories and is therefore free to use subject to the author's licensing. The problem for the programmer using that code is that the AI model won't tell you what license that was nor the original author so there's no way to include the author in the module's copyright statement nor to comply with the licensing terms if they've used a BSD license or derivative that requires acknowledging the original copyright, nor for non-GPL projects to know that thanks to the AI model they've turned their proprietary program into a GPL one and that they must now publish all of their source code.

cdhorn:
> Generative AI models don't learn in the human sense. 

At the present time yes I suppose I have to agree.  I wonder if we really understand enough about how human learning really works to say that will never be the case though?  And I suppose how you view the whole situation depends on whether you think there is merit in the computational theory of mind or not.

An interesting experiment would be to train a dozen ChatGPT like models but with variations in the input data set, and with some of the overlapping data feed in but in different times or in different sequence, perhaps building up from base concepts to more complicated ones the way a human mind forms. And then examining how their responses would be similar or might differ to various questions.

> [Getty vs. Stable Diffusion](https://www.theverge.com/2023/2/6/23587393/ai-art-copyright-lawsuit-getty-images-stable-diffusion) alleges that Stable Diffusion's supposedly generative model lifted whole images, copyright watermarks and all, and attempted to pass them off as original work.

Yes I heard about this, if I understand they were not perfect replicas hence my lossy compression analogy.

> thanks to the AI model they've turned their proprietary program into a GPL one and that they must now publish all of their source code.

Now there is an interesting thought. Can a legal argument be made that these models are fundamentally no different than a computer program, and that if trained on GPL code they themselves are a derivative work and thus need to be made available under the terms of the GPL?

cdhorn:
BTW I think I mostly agree with the current choice. Taking code it outright generates may be problematic with the copyright and other questions but I am unsure on the other use case.  If one wants to use it like a linter and gives it a piece of code and asks for help locating a bug in it or for suggestions on how to improve it is that considered fair use?  I tend to think yes. And that if it provides a sample of refactored code with the fix or improvements that seems possibly okay to me as well. I say possibly as if it is just reswizzling the existing code that is okay, if it adds something new that was not there before then you have the concerns as to where that came from. So I hope at some point the current policy is revisited perhaps after some legal dust settles on how to interpret the type of data the models output.

PR #1386, opened 2022-07-11
Avoid application crash on invalid user input

jpschewe:
This is a fix for bug [0012658](https://gramps-project.org/bugs/view.php?id=12658). If the user inputs an invalid date this
change keeps the application from crashing when such a date is found.

### DIFF
--- a/gramps/plugins/webreport/basepage.py
+++ b/gramps/plugins/webreport/basepage.py
@@ -2873,9 +2873,17 @@ class BasePage: # pylint: disable=C1001
                 if self.reference_sort:
                     role = ""
                 elif role[1:2] == ':':
+                    # format of role is role_type:ISO date string
+                    if role.count(':') > 1:
+                        print("Invalid date :", role[2:], " for individual with ID:", gid,
+                              ". Please, use the 'verify the data' tool to correct this.")
+                        cal, role = role.split(':', 1)
+                    else:
+                        cal, role = role.split(':')
+
                     # cal is the original calendar
-                    cal, role = role.split(':')
-                    # conver ISO date to Date for translation.
+
+                    # convert ISO date to Date for translation.
                     # all modifiers are in english, so convert them
                     # to the local language
                     if len(role.split(' - ')) > 1:


SNoiraud:
I don't like this. You must verify your data is correct. Ans this is your problem.

SNoiraud:
Your patch doesn't work when you have spanned date.
Many reports will not work if you don't want to correct the dates.

I propose the following patch for the narrativeweb:

```
@@ -2873,8 +2873,13 @@ class BasePage: # pylint: disable=C1001
                 if self.reference_sort:
                     role = ""
                 elif role[1:2] == ':':
-                    # cal is the original calendar
-                    cal, role = role.split(':')
+                    if len(role.split(':')) > 2:
+                        print("Invalid date :", role[2:], " for :", gid,
+                              ". Please, use the 'verify the data' tool to correct this.")
+                        cal, role = role.split(':', 1)
+                    else:
+                        # cal is the original calendar
+                        cal, role = role.split(':')
                     # conver ISO date to Date for translation.
                     # all modifiers are in english, so convert them
                     # to the local language
```

jpschewe:
That does seem better, however where does the output go? Is there a way to report this to the user?

If there isn't a direct way to send a message raising an exception with the information would give the user a chance to figure out what is broken.

                        raise RuntimeError("Invalid date :", role[2:], " for individual with ID:", gid,
                              ". Please, use the 'verify the data' tool to correct this.")

SNoiraud:
> however where does the output go?

All the results go in the gramps.log if you have one, or on the terminal where you start gramps.

We can show a warning window for this. But this case is so rare (you are the first in 15 years) that I think the print is sufficient.

jpschewe:
OK, if I'm the first to ask, then the print sounds fine. Would you like me to update this PR or just close it out?

SNoiraud:
Yes please. This PR don't need to be closed.

jpschewe:
OK, I just pushed changes that will output the warning to the log or stdout. I just tested and I saw the output on stdout.

Nick-Hall:
@SNoiraud Is this ready to be merged?

PR #1387, opened 2022-08-25
The latest version of openlayers doesn't work

SNoiraud:
Every years, openlayers change its "latest" version.
If this no longer works in the browser we need to change "latest"
by another version.
v6.15.1 is the latest working version and is available several years.
If this no longer the case for this version, we need to change this version
in the gramps.ini file.

The section will be:
...
[external_modules]
;;openlayers_version=‘v6.15.1’
…

You'll need to remove the two commas before openlayers_version and
change v6.15.1 to the working version.
Available versions are provided in the OLDER SECTION of https://openlayers.org/

Fixes #012680


SNoiraud:
From Issue https://github.com/openlayers/openlayers/issues/14055
I adapt the location of js and css files depending on the version

PR #1390, opened 2022-09-05
Improvement: Add bookmark-list-changed support

cdhorn:
This minor enhancement adds support for emitting a bookmark-list-changed signal to the DbBookmarks class.

### DIFF
--- a/gramps/gen/db/bookmarks.py
+++ b/gramps/gen/db/bookmarks.py
@@ -20,35 +20,86 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
-#-------------------------------------------------------------------------
-#
-# class DbBookmarks
-#
-#-------------------------------------------------------------------------
-class DbBookmarks:
-    def __init__(self, default=[]):
-        self.bookmarks = list(default) # want a copy (not an alias)
+"""
+DbBookmarks Class
+"""
 
-    def set(self, new_list):
+from ..utils.callback import Callback
+
+
+# -------------------------------------------------------------------------
+#
+# DbBookmarks class
+#
+# -------------------------------------------------------------------------
+class DbBookmarks(Callback):
+    """
+    Simple class for tracking a list of bookmarked objects
+    """
+
+    __signals__ = {"bookmark-list-changed": None}
+
+    def __init__(self, default=None):
+        super().__init__()
+        self.bookmarks = default or []
+
+    def load(self, new_list):
+        """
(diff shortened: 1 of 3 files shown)


gramps/gen/db/bookmarks.py
```diff
@@ -20,35 +20,50 @@
...
+
+    __signals__ = {"bookmark-list-changed": (str,)}
+
+    def __init__(self, default=[], cls_type="Person"):
+        super().__init__()
+        self.cls_type = cls_type
+        self.bookmarks = list(default)  # want a copy (not an alias)
 
-    def set(self, new_list):
+    def set(self, new_list, quiet=False):
```
Nick-Hall, line 39: Rather than having a quiet mode, the usual way to do this would be to use the `disable_signals()` and `enable_signals()` methods.

gramps/gen/db/bookmarks.py
```diff
@@ -20,35 +20,50 @@
...
         self.bookmarks = list(new_list)
+        if not quiet:
+            self.emit("bookmark-list-changed", (self.cls_type,))
```
Nick-Hall, line 42: The `cls_type` isn't necessary here since we know the class type when we connect to the signal.  A common handler could be used as follows:
```
self.db.event_bookmarks.connect('bookmark-list-changed', handler_func, 'Event')
```

Nick-Hall (commented):
Did you consider emitting the signal from the database class instead of the individual bookmark classes?

cdhorn:
I need to review this one, it has been a while.

cdhorn:
> Did you consider emitting the signal from the database class instead of the individual bookmark classes?

Nothing in the database class is called to change a bookmark, it merely reads them in on an open and writes them out on a close.

gramps/gen/db/bookmarks.py
cdhorn: So if I follow your thinking that would necessitate three lines for each bookmark type when they are loaded by the database class instead of one.  I think this is cleaner., although I suppose a keyword like skip_signal would make intent clearer.

gramps/gen/db/bookmarks.py
cdhorn: Yes it could be that way.

cdhorn: Cleaned up.

gramps/gen/db/bookmarks.py
cdhorn: Choose to add a load method that doesn't emit and call that from the db class as seems cleaner than this.

PR #1391, opened 2022-09-05
Improvement: Add citation_list to event references

cdhorn:
This takes a pass at extending eventref.py to add citation_list support since the event reference contains the role of the participant and not the event itself.

### DIFF
--- a/data/grampsxml.dtd
+++ b/data/grampsxml.dtd
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<!-- 
+<!--
 # Gramps - a GTK+/GNOME based genealogy program
 #
 # Copyright (C) 2001       Graham J. Williams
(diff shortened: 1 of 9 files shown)


Nick-Hall:
Adding a citation list to event references seems like a reasonable request to me.  It would give a mechanism to cite evidence for a person's participation in an event, rather than just for the event itself.

Nick-Hall:
The database schema files `grampsxml.rng` and `grampsxml.dtd` will also need updating.

We should also update the examples files with some relevant data.

cdhorn:
Ah yes, okay.

Is this something you want to fold into 5.2?  I am not sure it makes sense to do a database update for something this minor if it is the only thing, maybe it should wait for the next release?

Nick-Hall:
> Is this something you want to fold into 5.2?

I don't see a problem with putting this in.  Sometimes we don't want to make a schema change when we are making other large changes.  When we upgrade from Gtk3 to Gtk4 we probably won't make schema changes.

Nick-Hall:
Rebased.

Nick-Hall:
Updated the schema definitions and examples file.

cdhorn:
Thank you Nick!

PR #1393, opened 2022-09-27
Fix IndexError that sometimes occurs when changing view

SNoiraud:
This occours when restarting gramps.
This should solve bugs:
[#12636](https://gramps-project.org/bugs/view.php?id=12636), [#12304](https://gramps-project.org/bugs/view.php?id=12304), [#12429](https://gramps-project.org/bugs/view.php?id=12429), [#12623](https://gramps-project.org/bugs/view.php?id=12623), [#12695](https://gramps-project.org/bugs/view.php?id=12695)

Fixes #012636, #012304, #012429, #012623, #012695

### DIFF
--- a/gramps/gui/viewmanager.py
+++ b/gramps/gui/viewmanager.py
@@ -786,7 +786,12 @@ class ViewManager(CLIManager):
             self.__create_page(page_def[0], page_def[1])
 
         self.notebook.set_current_page(page_num)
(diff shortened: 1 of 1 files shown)


cdhorn:
@Nick-Hall this works fine for me.  I had a portion of this patch applied by hand to my 5.1 as well for some time without issues.

PR #591, opened 2018-03-29
Enclose tree report image path and file name in braces

Nick-Hall:
Fixes #10495.

PR #1395, opened 2022-10-08
Revert "Enclose tree report image path and file name in braces"

hgohel:
Fixes #12437 by reverting commit 75921ceaf40f3ced597d99c43794b98f81e49957 due to reports of regression where processing of the generated TeX file fails due to bad path specification for image files.

Change was originally introduced as a bug fix for #10495. Bugs #12437 and #12697 reported the regression and confirmed that reverting the change fixes the regression.

This fix has been tested with tree reports on Windows 10 with Gramps 5.1.5 AIO using gramps sample tree. Verified that images in the generated TeX file were processed with pdflualatex without errors and shown in the resulting PDF.

**Needs testing on**
- [X]  Linux
- [ ] MacOS

FYI @azrdev @Nick-Hall 

My hunch is that improved handling of spaces in file paths in LaTeX since the original problem was reported in 2018 may have resolved the problem. Reference: [LaTeX News Issue 30, October 2019](https://www.latex-project.org/news/latex2e-news/ltnews30.pdf):

> Improving file name handling in pdfTEX
> A related change is that file names used as part of \input, \includegraphics, etc., commands can now contain any Unicode characters allowed by the file system in use, including spaces.

### DIFF
--- a/gramps/gen/plug/docgen/treedoc.py
+++ b/gramps/gen/plug/docgen/treedoc.py
@@ -460,8 +460,7 @@ class TreeDocBase(BaseDoc, TreeDoc):
             if os.path.isfile(path):
                 if win():
                     path = path.replace('\\', '/')
-                self.write(level+1, 'image = {{%s}%s},\n' %
-                           os.path.splitext(path))
+                self.write(level+1, 'image = {%s},\n' % path)
                 break # first image only
         self.write(level, '}\n')
 


Nick-Hall:
> My hunch is that improved handling of spaces in file paths in LaTeX since the original problem was reported in 2018 may have resolved the problem.

Yes.  The preferred solution is to use the [grffile](https://ctan.org/pkg/grffile) package.  This is now a legacy package and its documentation states "The base LaTeX code now supports multiple dots and spaces".

I'll wait in case anyone else would like to make a comment, but otherwise I think that this can be merged.

emyoulation:
Your Latex related PR looks similar to the problem the [Graph View print](https://www.gramps-project.org/wiki/index.php/Addon:Graph_View#On_main_Gramps_Toolbar) (which actually a ‘save’).

The embedded paths for gallery images in SVG files [fail with Firefox & Inkscape](https://www.reddit.com/r/gramps/comments/jhnj9w/gramps_print_svg_does_not_show_photo/) but work with Adobe Illustrator. I saw similar problems with their .gv format.

hgohel:
> Your LaTex related PR looks similar to the problem the [Graph View print](https://www.gramps-project.org/wiki/index.php/Addon:Graph_View#On_main_Gramps_Toolbar) (which actually a ‘save’).

@emyoulation Took a look at that briefly. Seems to be a different problem which may involve multiple issues both related to gramps, and other 3P software such as Firefox and Inkscape. One thing's for sure, this bug fix will not resolve it.

emyoulation:
Thanks for looking! Do you mind if I add that note to the associated bug report?

hgohel:
> Thanks for looking! Do you mind if I add that note to the associated bug report?

Oh, if there is a bug report for that issue, sure. It deserves to be investigated.

emyoulation:
Hmmm...

Looks like there isn't a bug report. It may have been another that failed on 'submit' and the 'back' didn't restore the cached form data. It was involved enough that this would've been aggravating.  I would've had to walk away to avoid breaking something.

My guess is that the .gv and .svg problem is that the Relative media path wasn't resolved. Or maybe a Linux developer found another Windows compatibility problem in making a valid fully qualified filepath from the Relative path and filename.

cdhorn:
Tested fine on Linux.

azrdev:
Sorry for coming back late to this, but indeed on gramps 5.1.5 code is generated like `image = {{/path/to/image}.ext}`, i.e. between https://github.com/gramps-project/gramps/pull/591 and this #1395. This code fails to render to pdf with file not found. Removing the inner `{}` manually makes this error go away, even with spaces and non-ascii chars in the path -- instead I get `Cannot determine size of graphic in <path> (no BoundingBox).`

So I think I can also confirm this patch is the right thing to do, but maybe there'd be more in addition.

hgohel:
@azrdev Thanks for taking the time to confirm the patch, which has been included in the recent [5.1.6 release](https://github.com/gramps-project/gramps/releases/tag/v5.1.6).

PR #1399, opened 2022-10-22
Add option to hide ages for events after death

Nick-Hall:
This PR adds the option to hide the age in the person editor for events after death.

Read the [Calculating age for burials](https://gramps.discourse.group/t/calculating-age-for-burials) topic for further details.

### DIFF
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -282,6 +282,7 @@ register('preferences.last-view', '')
 register('preferences.last-views', [])
 register('preferences.family-relation-type', 3) # UNKNOWN
 register('preferences.age-display-precision', 1)
+register('preferences.age-after-death', True)
 
 register('colors.scheme', 0)
 register('colors.male-alive', ['#b8cee6', '#1f344a'])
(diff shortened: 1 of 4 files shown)


DaveSch-gramps:
Installed the PR and works great.

I am wondering if the default should be "True" which is the current scenario and expected by users.

emyoulation:
Please identify the label text of the option for those of use who have not installed. Is the 'True' for a "Hide" option or its inverse, a "Show" option.

Either way, a "show" should be the default. Otherwise newbies may never find it.

Nick-Hall:
The option is "Display ages for events after death".  I'll change the default to True so that it matches the current behaviour.

emyoulation:
The other 2 options (in the "display" Preferences tab) are labeled "Show", not "Display"

Maybe the code starts with "Display" but the label is consistent in the GUI with "Show"?

DaveSch-gramps:
> The other 2 options (in the "display" Preferences tab) are labeled "Show", not "Display"
> 
> Maybe the code starts with "Display" but the GUI label is consistent in the rest in the tab with "Show"?

The display age after death option will not become available until 5.2 and will display on the Data tab of a totally reworked configure.py file.

cdhorn:
This one worked fine for me too.

PR #1400, opened 2022-10-23
Add new gender option of "Other"

Nick-Hall:


gramps/plugins/export/exportgedcom.py
```diff
@@ -513,14 +513,16 @@ def _gender(self, person):
         Write out the gender of the person to the file.
 
         If the gender is not male or female, simply do not output anything.
-        The only valid values are M (male) or F (female). So if the geneder is
-        unknown, we output nothing.
+        The only valid values are M (male), F (female) and X (other). So if
```
cdhorn, line 516: X is only valid for Gedcom >= 7 so should this be handled differently here?

Nick-Hall: Possibly.  I'm not sure which version of Gedcom we want to support in the export. We could just export "X" as "U".

cdhorn: I think that might make sense for 5.5.1 and attach an auto generated FACT SEX attribute with the Other setting so at least the data doesn't get lost?

cdhorn:
It is worth reviewing Tamura's recent assessment:

https://www.tamurajones.net/GenderInGenealogySoftware.xhtml

Nick-Hall:
This is failing the unit test.  I need to investigate.

liquidsunshine:
@Nick-Hall - I proposed [a way forward](https://gramps.discourse.group/t/additional-gender-options/2321/97) of handling the data around gender and sex in the forum thread.  I'm happy to help with development and/or testing if that would be helpful.

Nick-Hall:
@liquidsunshine Thanks for your offer to help.

This PR adds an existing gender option of "Other".  Using "Non-binary" may be better, but I was trying to avoid any term which may offend people.  By "Other" I mean someone whose gender identity cannot be described as either "Male" or "Female".

Recording sex and gender information from sources is probably best done using attributes.  We are currently discussing adding dates to attributes in PR #1427.

As far as I am concerned this PR is ready to be tested and merged.

emyoulation:
For future cross-reference (no action needed)
See the following Discourse topic (thread) for beta testing discussion of "Other" gender aware features. The 1st posting also has links to related topics.  
https://gramps.discourse.group/t/add-missing-awareness-of-other-gender/6034

PR #1402, opened 2022-10-24
Add support for CITE plugins

Nick-Hall:
Experimental code to allow citations in reports to be formatted using plugins.

The "Default" plugin just implements the current functionality.

The idea is to investigate improvements such as templates or external formatting.  CSL formatting using [citeproc-py](https://github.com/brechtm/citeproc-py) is one idea.

### DIFF
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -289,6 +289,7 @@ register('preferences.last-views', [])
 register('preferences.family-relation-type', 3) # UNKNOWN
 register('preferences.age-display-precision', 1)
 register('preferences.age-after-death', True)
+register('preferences.cite-plugin', 'cite-default')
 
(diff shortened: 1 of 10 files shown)


cdhorn:
My initial thought looking at this is that maybe all Citation and Source key/value pairs should be stored in attributes, converting the existing fields along the way, so that things are done in a consistent manner. 
The Citation editor would allow entering of any of the standard CSL locators, the Source editor of any of the standard CSL variables. And then the formatters work off those, defaulting to the legacy formatter and using the legacy CSL field values that are currently hard coded entry fields. Gedcom exports would require some work to extract and reformat the attributes to shoehorn things into the existing fields obviously. This likely breaks and requires reworking a lot of existing code though.

PR #1403, opened 2022-10-26
Tree metadata tool and related changes

cdhorn:
This PR implements the following changes:

- Add a Tree object to encapsulate information about the tree and save it in the database similar to the current Researcher entity. Properties of a Tree are uuid, name, description, copyright, license, and contributors. The tree uuid represents the dataset that comprises the logical tree.
- Maintain a change timestamp for changes to the Tree object as a whole.
- Provide a Tree Editor Tool for managing the tree related data.
- Generate a unique uuid to represent the current backing/persistent store for the tree data.
- Maintain a last transaction timestamp for the last database transaction.
- Maintain a change timestamp for changes to the Researcher object as a whole.
- Add support for associating the Researcher object with a Person in the tree.
- Update the Owner Editor to enable adding/removing the Researcher <> Person association.
- Update the Gramps Xml export header to include the new information being captured. A sample is as follows:
```
  <header>
    <created date="2022-10-25" version="5.2.0"/>
    <storage uuid="b2be8956-4c3c-4ac5-8f01-692034026c2d" change="1666738821"/>
    <tree uuid="c7db6c8a-a300-4218-93d6-039b71c9decd" change="1666736955">
      <name>Tree Testing</name>
      <copyright>Copyright 2022 Christopher Horn</copyright>
      <license>None</license>
      <description>A Gramps example tree used for testing the tree metadata editor and other changes.</description>
      <contributors>None</contributors>
    </tree>
    <researcher change="1666739578" handle="_f28adbe86b15cd123e7bd84e043">
      <resname>Alex Roitman</resname>
    </researcher>
    <mediapath>{GRAMPS_RESOURCES}/doc/gramps/example/gramps</mediapath>
  </header>
```
- Add support for including a user export note or comment in the Gramps Xml header. Note the exporter GUI has not been updated for the user to use this yet.
- Update the Gramps Xml import to import some of the new information when it is an import for an empty tree.
- Update the Gedcom export to use the Tree copyright information if it is available.
- Add get_addon_metadata and set_addon_metadata methods so add ons can choose to save state in the database. This state is not backed up or exported at this time and values must be str type only.
- Other additional methods include get_database_uuid, get_last_transaction_time, get_tree, set_tree, set_tree_change, set_researcher_change, get_researcher_handle, get_researcher_person, set_researcher_handle, and get_tree_metadata
- Update the DTD and RNG files and bump the Gramps Xml version to 1.7.2

### DIFF
--- a/data/gramps_canonicalize.xsl
+++ b/data/gramps_canonicalize.xsl
@@ -22,7 +22,7 @@
 -->
 <xsl:stylesheet version="1.0"
     xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
-    xmlns:g="http://gramps-project.org/xml/1.4.0/">
+    xmlns:g="http://gramps-project.org/xml/1.7.2/">
 
 <!--
     Transform a Gramps XML file into "canonical form", that is strip the
@@ -33,8 +33,7 @@
 
 <xsl:output method="xml"/>
 
-<xsl:param name="replace_handles"/>
-<xsl:key name="primary_obj" match="g:person|g:family|g:event|g:placeobj|g:source|g:repository|g:object|g:note|g:tag" use="@handle"/>
+<xsl:key name="primary_obj" match="g:person|g:family|g:event|g:placeobj|g:source|g:repository|g:object|g:note|g:tag|g:researcher" use="@handle"/>
 
 <xsl:template match="*|@*|text()">
     <xsl:copy>
@@ -45,7 +44,7 @@
 <xsl:template match="@change">
 </xsl:template>
 
-<xsl:template match="g:researcher">
+<xsl:template match="*[local-name()='provenance' or local-name()='tree' or local-name()='researcher']">
     <xsl:copy/>
 </xsl:template>
 
@@ -57,6 +56,7 @@
     </xsl:copy>
 </xsl:template>
 
+<xsl:param name="replace_handles"/>
 <xsl:template match="@handle">
     <xsl:choose>
         <xsl:when test="$replace_handles='ID'">
--- a/data/grampsxml.dtd
+++ b/data/grampsxml.dtd
@@ -67,12 +67,11 @@ DATABASE
 <!--    ************************************************************
 HEADER
 
-    A <header> consists of <created> (information about this
-    genealogical database) and <researcher> (information about the
-    person who created this genealogical database)
+    A <header> consists of <created> and several additional sections
+    with high level information about this genealogical database.
 -->
 
-<!ELEMENT header (created, researcher?, mediapath?)>
+<!ELEMENT header (created, origin?, comment?, tree?, researcher?, mediapath?)>
 
 <!ELEMENT created EMPTY>
 <!ATTLIST created
@@ -80,8 +79,40 @@ HEADER
         version  CDATA #REQUIRED
 >
 
+<!--
+PROVENANCE identifies the database used to generate this export.
+-->
+
+<!ELEMENT provenance (database-id, last-transaction-timestamp, comment>
+<!ELEMENT database-id                 (#PCDATA)>
+<!ELEMENT last-transaction-timestamp  (#PCDATA)>
+<!ELEMENT export-note                 (#PCDATA)>
+
+<!--
+TREE contains information about the tree.
+-->
+
+<!ELEMENT tree (name, copyright?, license?, description?,
+                contributors?)>
+<!ATTLIST tree
+        change   CDATA #REQUIRED
+>
+<!ELEMENT name           (#PCDATA)>
+<!ELEMENT copyright      (#PCDATA)>
+<!ELEMENT license        (#PCDATA)>
+<!ELEMENT description    (#PCDATA)>
+<!ELEMENT contributors   (#PCDATA)>
+
+<!--
+RESEARCHER
+-->
+
 <!ELEMENT researcher (resname?, resaddr?, reslocality?, rescity?, resstate?,
                       rescountry?, respostal?, resphone?, resemail?)>
+<!ATTLIST researcher
+        change      CDATA #REQUIRED
+        handle      IDREF #IMPLIED
+>
 <!ELEMENT resname        (#PCDATA)>
 <!ELEMENT resaddr        (#PCDATA)>
 <!ELEMENT reslocality    (#PCDATA)>
--- a/data/grampsxml.rng
+++ b/data/grampsxml.rng
@@ -41,14 +41,25 @@
             <attribute name="date"><data type="date"/></attribute>
             <attribute name="version"><text/></attribute>
         </element>
-        <element name="researcher">
-            <optional>
-            <ref name="researcher-content"/>
-            </optional>
-        </element>
         <optional>
-           <element name="mediapath"><text/>
-           </element>
+            <element name="provenance">
+                <ref name="provenance-content"/>
+            </element>
+            <element name="tree">
+                <attribute name="change"><text/></attribute>
+                <ref name="tree-content"/>
+            </element>
+            <element name="researcher">
+                <attribute name="change"><text/></attribute>
+                <optional><attribute name="handle">
+                    <data type="IDREF"/>
+                </attribute></optional>
+                <optional>
+                    <ref name="researcher-content"/>
+                </optional>
+            </element>
+            <element name="mediapath"><text/>
+            </element>
         </optional>
       </element>
 
@@ -134,6 +145,22 @@
           </element></zeroOrMore>
       </element></optional>
   </element></start>
+
+  <define name="provenance-content">
+     <element name="database-id"><text/></element>
+     <element name="last-transaction-timestamp"><text/></element>
+     <element name="export-note"><text/></element>
+  </define>
+
+  <define name="tree-content">
+      <element name="name"><text/></element>
+      <optional>
+          <element name="copyright"><text/></element>
+          <element name="license"><text/></element>
+          <element name="description"><text/></element>
+          <element name="contributors"><text/></element>
+      </optional>
+  </define>
 
   <define name="researcher-content">
       <element name="resname"><text/></element>
--- a/example/gramps/example.gramps
+++ b/example/gramps/example.gramps
@@ -1,10 +1,21 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE database PUBLIC "-//Gramps//DTD Gramps XML 1.7.1//EN"
-"http://gramps-project.org/xml/1.7.1/grampsxml.dtd">
-<database xmlns="http://gramps-project.org/xml/1.7.1/">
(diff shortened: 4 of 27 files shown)


cdhorn:
Need to look closer at the failing tests yet I guess.

DavidMStraub:
This looks great. Is anything missing apart from fixing the tests?

emyoulation:
Is the 'owner' similar to the 'Researcher' ? Where they can be associated with a Person in the Tree? 

Are the Address and Internet Gramps XML data structures in the Tree the only way to store Tree contact data: Phone, postal address, email, website.

Users could make that info accessible at the metadata level as a user entered comment. And it would be less likely to be accidentally included. (Although the "probably alive" export filters are likely to invalidate the Researcher handle more often than not.)

cdhorn:
> Is anything missing apart from fixing the tests?

I don't think so, but being the one who wrote it not sure if I overlooked something.

I got stuck trying to track down the issue with the tests and set it aside and of course got side tracked and have not revisited.

> Is the 'owner' similar to the 'Researcher' ? Where they can be associated with a Person in the Tree?

Yes, you have Researcher in preferences and you also have a Researcher object stored in the database in the Researcher entity.  The one in the database you can edit with the Database Owner Editor under Tools -> Family Tree Processing and the editor lets you sync with the one in the preferences.  The one in the database is the one included in XML exports. So I added something so you can associate the database owner/researcher stored in the database with a Person in the tree as it seemed logical to do so.

emyoulation:
> So I added something so you can associate the database owner/researcher stored in the database with a Person in the tree as it seemed logical to do so.

I agree completely that associating is logical.

But my question is if there is a SEPARATE association for owner. 

Rather than OnePlace Studies, I have about four different OneSource studies. These revolve around creating Trees to transport each fact presented within a single Reference Book. (An OCR''d PDF of the Reference Book is attached as a Media Object.) 

The researcher for the 1924 booklet (which incidentally roped me into this obsession) died in 1948. He is in the Tree and is indisputedly the Researcher. His research only comes as close to me as a paternal great-grandmother. 

If I distribute the Tree, I would be the owner. Yet my existence as a 'Person' connected to that Tree is only with an pair of Associations (to my great-grandmother, and to the 6th great-grandfather who is the progenitor in the research) & there's a Note detailing my direct descendancy from the progenitor. 

We don't want to taint the Tree built from his original research with other sources. (Yes, there are significant errors.)

And I'm talking with a 6th cousin (who built a website around the book) is connected to his closest ancestors in that book with 2 Associations and a Note. He is considering distribution of that Tree via the website. So happily, HE might become stuck as the owner soon.

The problem is that discovering either of the 2 possible owners via browsing the Tree is virtually impossible ... unless there is a separate GrampsID/handle for Owner from that of the Researcher.

cdhorn:
> But my question is if there is a SEPARATE association for owner.

Oh I am sorry I did not understand.  No there is not.

I think ideally you would eliminate the concept of a singleton researcher.  I'd just model it as a list of contacts and the tree owners and researchers/collaborators are just different types of contacts. I use plural because with Gramps.js everyone should probably be thinking multi-user in anything done going forward.

cdhorn:
@DavidMStraub Nick asked that I pull out the db and tree uuids so just confirming you were not looking to use them before I refactor this.

DavidMStraub:
One question, I noticed the `tree_change` (which I'm particularly looking forward to as it will make caching for Web API much easier) is set as `int(time.time())`. Is there any reason to drop sub-second precision? In a multi-user/server context, there might be multiple transactions per second. Actually, even on a Desktop system, there might be cases of an add-on or batch script running through a loop and processing a transaction in every iteration. So I think it would be much safer to store the timestamp float directly without the `int()`.

cdhorn:
@Nick-Hall @DavidMStraub this has been refactored now.  I pulled out the uuids as well as some other things to try to make it a little more focused. Please let me know if you see any other changes you think are needed.

Nick-Hall:
@cdhorn Thank you.  I'll look at this soon.  I've just arrived back from travelling where I had limited internet access.

DavidMStraub:
Concerning my previous comment, do I understand correctly that it's not actually `tree_change` that is relevant but `db_last_transaction` and this is stored as float?

cdhorn:
> Concerning my previous comment, do I understand correctly that it's not actually `tree_change` that is relevant but `db_last_transaction` and this is stored as float?

Correct.

The tree_change tracks when the Tree object with the description, copyright etc was last changed.  I also added owner_change for the Researcher object.

emyoulation:
The OS may important for the Media path and archive restoration. 

Media restores fail terminally when destination paths are not compatible. (I've seen overwrite permission problems when the destination was the same machine too. And the Tree data was not imported either in that case.)  Will originating OS be important for resolving (or crosswalking path structures for) such problems?

emyoulation:
To make the header dates and media paths be more parsable by humans, should the date format CCYY-MM-DD and OS be explicit?

cdhorn:
> To make the header dates and media paths be more parsable by humans, should the date format CCYY-MM-DD and OS be explicit?

Unsure I follow this question. The media path attribute is untouched by this.  The created date is already in that format.

No information about the OS the export was generated on or database engine it was generated from is captured, though that could be added to the `<provenance>` section.

Currently this adds the following, as an example, to the `<header>` in addition to change timestamp for the `<researcher>` object:

```
    <provenance>
      <database-id>5d73241c</database-id>
      <last-transaction-timestamp>1690640156.5227873</last-transaction-timestamp>
      <export-note>An optional export note</export-note>
    </provenance>
    <tree change="1690638919">
      <name>Family Tree 2</name>
      <copyright>2001-2006 Donald N. Allingham, 2007-2023 Gramps Project</copyright>
      <license>Creative Commons Attribution-ShareAlike 2.5</license>
      <description>Example Gramps database</description>
      <contributors>Gramps Developers</contributors>
    </tree>
```

The change timestamp on the `<tree>` object is not the change time stamp for the database. That is in the `<provenance>` section at the moment as you can see above.

The support for adding the `<export-note>` was added but I did not update the exporter dialog to support it, that would be a separate item at some point. 

I just realized I need to also include an update to `example.gramps` with the above as well now. 

I also just noticed that the `example.gramps` under `data/tests` directory differs slightly from the one under the `example/gramps` directory so they seem to have diverged. I am unsure if that was deliberate.

@Nick-Hall when you have a few cycles if you could look this over again and let me know of any further changes you would like to see here.

cdhorn:
| I looked at the sample header posted. Notice it had an August 8th date.

2017-08-08 is the date in the current `example.gramps` not the example I posted when I opened the PR which was 2022-10-25. The date format is always explicit.

Path conversion is already handled in the existing code and OS information not needed in the XML for that. 

The created element identifies the version of Gramps, the XML schema version is in the document type declaration itself.

cdhorn:
Rebased.

cdhorn:
@Nick-Hall should I rebase or close this? Not sure you're thoughts.

Nick-Hall:
I'd like to include this in the next release.  It was added to the roadmap.

This Isotammi project adds a `<comment>` which is similar to your `<export-note>`.  They also have a field that indicates the type of export.  From what I understand this is used when the export is intended to be imported into another system, rather than just a backup.  I'll try to find out more details.

cdhorn:
I'll review what they added before rebasing this and adjust this to accomodate the type use case.

kulath:
Isn’t Researcher a GEDCOM element, so there can be only one. I’m not sure whether some earlier comments implied that there could be more than one Researcher?Sent from my iPhoneOn 18 May 2024, at 23:26, Christopher Horn ***@***.***> wrote:﻿
I'll review what they added before rebasing this and adjust this to accomodate the type use case.

cdhorn:
> Isn’t Researcher a GEDCOM element, so there can be only one. I’m not sure whether some earlier comments implied that there could be more than one Researcher?

I'd argue Researcher should be a top level entity, which it is in Gedcom X for example where it is represented as "Agent" and in DeadEnds in a partial sense by "Entity". It makes perfect sense as a top level object in the data model as soon as you have a tree that is shared online that you allow other people to modify as it is desireable for the owner to know who contributed or edited various bits of information in it.

Anyway, this PR only adds a last changed timestamp for it.

cdhorn:
@kulath regarding Researcher, in Gedcom 7 "Submitter" is a separate top level record type. The submitter of the Gedcom is assumed to have provided all the information, but specific person and family records can reference a different submitter. This further suggests Researcher should be promoted to a top level object.

emyoulation:
Could we reconsider these items systemically instead of in isolation?

The Researcher, Owner and "About the Database" Note are similar to the Home Person and Active Record... their data all fits within an existing record type used for general purposes in Gramps. Why have a separate record type that leads to Harmonization errors and redundant data entry?

The Researcher and Owner records are Person records that only use the "Given Surname Suffix" Name Format, an Address secondary object and an E-mail Internet secondary object.  (I am baffled by the choice to not have the Address fields use a Place hierarchy reference to auto-fill some of the data.)  

The Database Note is suggested elsewhere (by CDHorn and Isotammi, I believe) for inclusion as a Text note in Backups and exports. It is to be visible in the Manage Family Tree, import verification (and possible linking as a Import Source note), possibly for alternative content for a Dashboard gramplet. And possibly in other features that I do not recall elsewhere.

cdhorn:
It is important to remember that the submitter/researcher/contributor may be an organization, not necessarily a person. 
From the Gedcom 7 specification:
```
The submitter record identifies an individual or organization that contributed information contained in
the dataset. All records in the document are assumed to be contributed by the submitter referenced in
the HEAD (p.74) , unless a SUBM (p.85) structure inside a specific record points at a different submitter
record.
```

cdhorn:
@Nick-Hall I have let this sit for some time and did not rebase it as most of my time has been occupied elsewhere.

Looking it over, maybe it would have been better off as a couple smaller, more focused PRs.

Given all the work @dsblank is doing for 6.0 not sure this would make the cut for that either. So I am going to close this for now and revisit things when it seems more appropriate.

DavidMStraub:
Among the great things in this PR, the easiest and most useful to me would be the global change timestamp.

Do we have a release roadmap for 6.0 yet? This should be doable.

cdhorn:
@Nick-Hall would some subset or any of this be considered for 6.0? There are so many other PR's also sitting out there waiting and it is no longer clear to me what will or will not get accepted and merged for the 6.0 release which is why I figured best to revisit it after.

Nick-Hall:
> would some subset or any of this be considered for 6.0?

No.  I won't be accepting any schema changes for 6.0.  I was just about to suggest postponing this PR anyway.

We could consider an early 6.1 though.  I promise to make time to work on your PR when it is re-submitted.

cdhorn:
> No. I won't be accepting any schema changes for 6.0.

Thanks, this makes sense. I'll revisit this once 6.0 is released.

PR #1407, opened 2022-12-16
Make "gramps -v" consistent with Prerequisites Checker Gramplet

SNoiraud:
This patch is related to Bug report [#12770](https://gramps-project.org/bugs/view.php?id=12770).
When a problem occurs, we ask to the user the result of "gramps -v" and not the result of the PrerequisitesCheckerGramplet

### DIFF
--- a/gramps/grampsapp.py
+++ b/gramps/grampsapp.py
@@ -270,6 +270,13 @@ def show_settings():
         cairover_str = 'not found'
 
     try:
+        gi.require_version('PangoCairo', '1.0')
+        from gi.repository import PangoCairo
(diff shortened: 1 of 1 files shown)


SNoiraud:
This PR is related to https://github.com/gramps-project/addons-source/pull/554

emyoulation:
It appears as though line 326 has a 'tense mismatch' typo: "can't found it"? 


"cannot find it" would be more correct. But that doesn't seem to match the phrasing elsewhere in the Gramplet either.

Nick-Hall:
To make things easier for our translators, we should probably stick to strings like "not found" and "unknown version".

PR #1409, opened 2023-01-19
Implement the GEDCOM tag "_RUFNAME"

smurfix:
RUFNAME is the German word for "call name", i.e. the part of the given name commonly used to address somebody. Some official documents mark it as such.

The GEDCOM standards 5.5.1 and 5.5.5 mention this tag.

### DIFF
--- a/data/tests/exp_sample.gramps
+++ b/data/tests/exp_sample.gramps
@@ -774,6 +774,7 @@
       <name type="Birth Name">
         <first>Amber Marie</first>
         <surname>Smith</surname>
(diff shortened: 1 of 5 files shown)


Nick-Hall:
The `_RUFNAME` tag is commonly used by Gernam software.  See:

http://wiki-alt.genealogy.net/Name_in_Gedcom

Nick-Hall:
Rebased.

Nick-Hall:
@romjerome The fix that you remember is probably PR #684, which has already been merged.

PR #1411, opened 2023-02-06
Replace deprecated imp by importlib

DavidMStraub:
`imp` is deprecated and will be removed in Python 3.12, to be released this year.

The code I added will work with Python 3.4 and up, which is enough four Gramps 5.1 and up.

Could also be cherry-picked into the 5.1 maintenance branch.

### DIFF
--- a/gramps/gen/plug/_manager.py
+++ b/gramps/gen/plug/_manager.py
@@ -372,13 +372,9 @@ class BasePluginManager:
         Reloads modules that might not be in the path.
         """
         try:
-            import imp
-            fp, pathname, description = imp.find_module(pdata.mod_name, [pdata.fpath])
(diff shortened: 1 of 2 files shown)


DavidMStraub:
@Nick-Hall, as part of this PR, I replaced the Ubuntu 18.04 image for the unit tests with 20.04, as 18.04 is no longer supported by Github:
image

Nick-Hall:
Thanks.  I see that 18.04 was removed on 12th Jan.

I'll probably cherry-pick that commit immediately.

Nick-Hall:
Code added to Gramps 5.1 is required to work with python 3.3, so I'll merge this into the _master_ branch only.

PR #1413, opened 2023-02-19
Narrative web: Solves some problems

SNoiraud:
Fixes [#12821](https://gramps-project.org/bugs/view.php?id=12821)

Force place name to be "Full" names.
Force the country field to be the last field of the place name Force the State/province to be the penultimate field of the place name

We use place names that stop in the country, but if we have place names such as: city, province, country, continent, planet with the continent and planet as custom place type, the State/Province field will be the name of the place continent... (should never happen, but...)


Nick-Hall:
@SNoiraud The unit test for the narrative web report is failing.

PR #1414, opened 2023-02-26
Fix spanish translation when GExiv2 module not found

diegomeretta:


Nick-Hall:
Thank you for the translation update.  However we now use [Weblate](https://hosted.weblate.org/projects/gramps-project/gramps/es/) for our translations.

The string that needs correcting can be found [here](https://hosted.weblate.org/translate/gramps-project/gramps/es/?checksum=0ae15485d1575df5).

romjerome:
Proposal (for Spanish) is also included in pull request #1419 (Weblate update)

PR #1415, opened 2023-03-01
Missing icon for Tags in the main navigation bar

olivierberten:


cdhorn:
Note this is also included in PR #1420 should it be accepted.

PR #1416, opened 2023-03-04
Backup: Add a popup to show the gramps state

SNoiraud:
Fixes [#12846](https://gramps-project.org/bugs/view.php?id=12846), [#12475](https://gramps-project.org/bugs/view.php?id=12475), [#12538](https://gramps-project.org/bugs/view.php?id=12538)

### DIFF
--- a/gramps/gui/utils.py
+++ b/gramps/gui/utils.py
@@ -101,6 +101,61 @@ class CLIDialog:
 
 #-------------------------------------------------------------------------
 #
+#  Popup class
+#
+#-------------------------------------------------------------------------
+
+class Popup:
+    """
+    Popup a message for Gramps.
+
+    The Popup is used to display a message
+    This class must be closed when no longer needed.
+
+    """
+
+    def __init__(self, title, message, parent=None):
+        """
+        Specify the title and the message to display.
+        """
+        from gi.repository import Gtk
+        self.__popup = None
+        if has_display():
+            self.__popup = Gtk.Dialog()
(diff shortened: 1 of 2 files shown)


lordemannd:
I adapted for 5.1.5, and it worked for me

emyoulation:
The new dialog (“Please, wait before closing gramps”) should have different text.
image

    Gramps is not capitalized
    the comma is in the wrong place, and
    this is invoked after closing by the Backup on Exit feature.
    image

How about:
"Please wait for backup in progress. Gramps will close when complete." 

Notes:  It is also REALLY disturbing another non-cancellable Exit activity is being layered on.

You already could not abort a backup. And that is scary when you are in an emergency shutdown situation. (Out of power, the boss/spouse is coming, an OS backup is starting, et cetera)

lordemannd:
Both manual backup and timed backups could be aborted in 5.1.5 (and 5.1.6) on Windows.
It is true that it needs a db big enough that the backup takes more than a few seconds.
 - for the manual backup, clicking the X gets you to the ErrorReportingWizard as in #12846
 - if timed backup is on going, clicking the X closes Gramps leaving a partial/useless backup file. (remains unaddressed)

Is it less disturbing to simply allow the backup to be aborted in both cases, with a 10 second pop-up message that says something like "backup aborted by user command, backup file deleted."?

The most disturbing issue for me is the partial backup file existing, giving the user the thought that they have a good backup.

PR #1417, opened 2023-03-04
Add additional note types

cdhorn:
Adds several additional note types to allow for finer differentiation and categorization of note content.


Nick-Hall:
I think that we should consider implementing the Gedcom TRAN tag structure instead of creating note types for TRANSLATION,   TRANSLITERATION and STRUCTURED_DATA.

For RESEARCH_PLAN we already have TODO which is intended to be used for this purpose. Likewise for RESEARCH_REPORT we already have RESEARCH to record the outcome of research.

Perhaps you could describe the following in more detail:

+ STORY
+ BIOGRAPHY
+ NAME_ORIGIN
+ EVIDENCE_ANALYSIS
+ PROOF_ARGUMENT

We will have to describe how they should be used in our documentation.

cdhorn:
> I think that we should consider implementing the Gedcom TRAN tag structure instead of creating note types for TRANSLATION, TRANSLITERATION and STRUCTURED_DATA.

Looking at how it is defined in the 7.0 specification it is not clear to me that it really applies to STRUCTURED_DATA. 

You raise an interesting point with TRANSLATION and TRANSLITERATION.  In some sense it might be better to model a NOTE as a list of secondary objects to pair the text with corresponding language to clearly keep all the associated data together. It is probably safe to assume when annotating a citation they're all pertaining to the same data, but not elsewhere.

> Perhaps you could describe the following in more detail:

They are kind of self explanatory I thought. 

STORY contains a short story about one or more people.
BIOGRAPHY contains a biographical sketch of the life of a person.
NAME_ORIGIN contains information about the origin of a name.
EVIDENCE_ANALYSIS contains reasoning behind a conclusion.
PROOF_ARGUMENT would be a more formal written argument about a major conclusion.

Having enumerated types allows data/content to be clearly identified for programmatic reasons.

emyoulation:
If a TRAN was directly supported within the Gramps data model, how would it be available in the GUI?  If N0001 in example.gramps is in English ("Names can notes, too. This is a note for the alternate name of Louse Garner for Lewis Anderson Garner.") and we wanted a Japanese Kanji version too, how would the equivalent be linked? And one asssumes that the preferences would be for the lang corresponding to the OS environment with fallback to English?

Nick-Hall:
You would have something like the following:

```
0 @N0001@ SNOTE Names can notes, too. This is a note for the alternate name of Louse Garner for Lewis Anderson Garner.
1 LANG en
1 TRAN Namae mo memo dekimasu. Kore wa, Ruisu andāson gānā no betsumei Rausu gānā ni kansuru memodesu.
2 LANG ja-Latn
1 TRAN 名前もメモできます。これは、ルイス・アンダーソン・ガーナーの別名ラウス・ガーナーに関するメモです。
2 LANG ja-Hani
```

The LANG tags contain a combination of ISO 639 and ISO 15924 codes.

All versions are contained in the same note.  I would probably suggest separate tabs within the note editor for the GUI implementation.  This needs more thought and a proper design though.

emyoulation:
> All versions are contained in the same note. I would probably suggest separate tabs within the note editor for the GUI implementation. This needs more thought and a proper design though.

The description for Events would be another complication for a multi-lingual GUI.

Nick-Hall:
It seems that discussions are ongoing about enhancing the Gedcom standard in this area.

cdhorn:
> An example would be helpful.

Consider a prototype for an experimental add on that perhaps leverages notes to stash yaml or json data for serialized objects.  At least that was my initial thinking for a potential use case. I can pull it if you desire.

cdhorn:
> STORY, BIOGRAPHY and NAME_ORIGIN seem to be just a narrative attached to different object types. Can we use a single NARRATIVE type for these?

We could but then you can not discriminate between them if you wanted to pull up a list of all the biographical sketches for example as the other types of data are mixed in.

Nick-Hall:
> Consider a prototype for an experimental add on that perhaps leverages notes to stash yaml or json data for serialized objects. At least that was my initial thinking for a potential use case. I can pull it if you desire.

Yes.  Use a custom type for prototypes and experimental code.

> We could but then you can not discriminate between them if you wanted to pull up a list of all the biographical sketches for example as the other types of data are mixed in.

It seems that REPORT_TEXT should be used for biographical sketches.  This feature is confusing and poorly documented.

The idea was that this type of notes should be included in reports.  Of course, any type of note can be included in reports depending on the report.  A research report is may well contain RESEARCH and TODO types.

If the REPORT_TEXT is attached to a person then it will be biographical, if attached to an event it will be a story or anecdote.

A NAME_ORIGIN type could be interesting if notes had a title field.  It could possibly be used for surname origins where the surname was stored in the title field.  There have been discussions on the forum asking for something like this.  More thought is needed on this one though.

Nick-Hall:
Replaced by PR #1474.

I was happy to add "Analysis" as it is clearly missing and used in other standards such as the GedcomX [DocumentType](http://www.gedcomx.org/types/DocumentType.html).

We can consider other types in the future.

emyoulation:
Why don't these additional notes show up in the history for example.gramps ?

https://github.com/gramps-project/gramps/commits/ebc213f0accf12392925b7197d67e6e42b03bd78/example/gramps/example.gramps

emyoulation:
Kari added a functionality to the SuperTool to save/load Scripts to/from Notes of type "SuperTool Script"

See https://gramps.discourse.group/t/new-update-to-supertool-available/4634

PR #1420, opened 2023-03-11
Add tag list view, enable clipboard for tags, enable basic tag drag and drop

cdhorn:


emyoulation:
Serge just added support for the Ctrl-Jump to the Families category in  [Implement the "<CTRL>J" for the family view #1432](https://github.com/gramps-project/gramps/pull/1432)"

And Riccardo is exploring the addition of a similar feature for Select Object dialogs in [In selection dialogs, allow to directly input a Gramps ID #1431 ](https://github.com/gramps-project/gramps/pull/1431). 

But I think Nick will probably want the 2 features harmonized for Views and dialogs.

Since [Oliver's PR](https://github.com/gramps-project/gramps/pull/1415) shows the Tag View is gaining visibility, should a "Jump to" capability be added to your Tags view?  

Issues: Tags don't have Gramps IDs --- which complicates things for this feature in a view. (Do you think Nick will grace Tags with IDs for 5.2?) 
And the Clipboard context menu's "**Make the _&lt;object type>_ active**" is more explicit and more grammatically appropriate than the Jump's "**Jump to by Gramps ID**" pidgin.

Nick-Hall:
> But I think Nick will probably want the 2 features harmonized for Views and dialogs.

Tags were never intended to store genealogical information.  This is why tags are not primary objects.

To quote Benny in a [post](https://sourceforge.net/p/gramps/mailman/message/26391276/) from the original design discussions:

> We don't want to have people using tags for genealogical data, and then
requesting they show in reports, or that special action is taken for extra
support in Gramps. We want tags to be something of which you take a group of
objects, then assign the same tag to it for organizational purposes.

This is why tags don't have their own category, view or Gramps IDs.

See also:

+ [GEPS 011](https://gramps-project.org/wiki/index.php/GEPS_011:_Tagging)
+ Feature request [#5806](https://gramps-project.org/bugs/view.php?id=5806)

However, tags were implemented over 10 years ago so we may wish to reassess how people are using them.  My initial reaction is that we don't want to enhance tags in this way.

cdhorn:
@Nick-Hall this came about because I had a tag card view and that required a new category so someone asked for it as they saw the matching list view as being missing.  If it needs to live outside as something separate that is okay, I can close the PR. Please let me know when you decide.

emyoulation:
It also allows drag'n'drop Tagging... This very intuitive interface fits the temporary organizational objectives. 

The 3 primary (multi-step) methods of tagging suit using Tags as permanent genealogical data instead: from the Edit Object dialogs, the select in list views and using the Edit -> Tag (by direct tag menu selection or using Organize Tags), or the Add/Remove Tag Tool addon. And it changes the object's 'last changed' datestamp too. This very deliberate UI and data impact is more like one would expect permanent Attributes to be stored, not temporary organization groupings.

As a new category (and a List view mode at that), it is dramatically different from the UI of the other CardView view modes. It should probably be a separate add-on download. 

Several users indicate that they have large sets of Tags. And that the pop-up menu has become unusable.

Does the Tags view require core changes too to support being spawned as an add-on? (I'm guessing that basic function does not since it is part of the 5.1 CardView download. But were there 5.2 extras? Like clipboarding?)

Maybe the view's functionalities could be adapted into an updated Organize Tags dialog? (On Windoze, that was a semi-modal dialog but on Linux, it is fully modal. I cannot even move the Organize Tags to see what is selected in the view below. That flexibility would make the clipboard much preferable.

cdhorn:
It adds drag 'n drop tagging but only enabled drop on the tag widget in the editors. And it supports the clipboard.

Those are among the things that require core code changes to support in the current design, so this is a little further along than the previous iteration.

If rethinking tags note that Ancestry now exports tag data in their Gedcoms. Stock tags are enumerated and export a RIN along with their name. Custom tags have a description exported as a NOTE in addition to their name.

olivierberten:
Not totally related but since you're working on tags: as far as I can see the Tag Organizer isn't accessible from the Tags view so it would be great to have a button or a menu item in that view, rather than going to another object and then Edit->Tags->Organize tags

emyoulation:
That could be interesting! 

But just looked at the Organize Tags dialog... if users can change ( up & down) the Priority of Tags in the Tag view then all the other functionality (add, remove, edit, list of names & colors) of the Organize dialog I'll already exist. The extra access would be more of a "security blanket", reassuring users that they can get to all functionality. 

https://www.gramps-project.org/wiki/images/9/99/MenuEditTag-OrganizeTags-dialog-example-50.png


However, similar to the clipboard's context menu that creates a quick custom filter for the selected objects of a single type, a button that creates a Custom Filter would be new functionality.  Particularly if extended selection of Tags was allowed in the Tag category view and the filter .

...but that leads to thoughts of a Filter gramplet for the Tag category. And then to Kari Kujansuu's Filter2 gramplet variants (they scrape the Filter parameters from the form into a Custom Filter definition).

cdhorn:
| My initial reaction is that we don't want to enhance tags in this way.

@Nick-Hall let me know if I should withdraw this one.

cdhorn:
Closing this based on other discussions, view itself will continue to exist as an add on for those who may want it.

PR #1421, opened 2023-03-12
Save/restore navigation history and mru state to database

cdhorn:
This adds set_navigation_state and get_navigation_state methods to the database class for saving the history and mru lists, and updates the History class to call utilitze them.  This removes the need for something like the RestoreHist add on.

Included is a cleanup of displaystate to bring the pylint score from 7.65 up to 9.29, and I reformated it with black.  A few pylint related cleanups to the db classes to bump them up a little but nothing as significant.

Needs more testing, I have also not checked to see how it might interact with the RestoreHist addon if it is installed.

### DIFF
--- a/gramps/gen/db/base.py
+++ b/gramps/gen/db/base.py
@@ -30,9 +30,6 @@ from this class.
 # Python libraries
 #
 #-------------------------------------------------------------------------
-import re
-import time
-from operator import itemgetter
 import logging
 
 #-------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


cdhorn:
Yeah this needs more work.

Nick-Hall:
Navigation history and MRU lists don't belong in the database.

cdhorn:
Would a file in the database directory be acceptible then as they're database specific and not global in nature?

PR #1422, opened 2023-03-13
Lossless Image Optimization

tiagofilipesilva:
This pull request contains losslessly optimized versions for a majority of the image files used in the Gramps project. By reducing the file sizes without compromising on quality, this optimization improves the app's overall performance and load times. This pull request will greatly benefit users who use Gramps from slow or limited-bandwidth devices, making their experience smoother and more enjoyable.


emyoulation:
Could you identify the tool & setting used to optimize the images?

It is likely that the greater majority of the images are from themes. If the master theme is updated, those optimized files will probably be overwritten when Geamps is harmonized. And it would probably be easier to re-optimize the updated theme than to check each image for differences.

And maybe the optimization process might be good fodder for a tutorial about reducing media object sizes in Gramps Trees too.

tiagofilipesilva:
- The used tool was [Papa's Best Optimizer](https://papas-best.com/optimizer_en) - which is a well-done Windows wrapper for many tools.
- The used settings used were metadata removal and no removal of transparent pixels - due to issues on non-Windows and legacy systems)  which is what I always use for these. Full lossless optimization is enabled by default.
- This is a Windows tool so it's not portable and full lossless optimization is enabled, so it's very slow.
- There could be some Python wrappers for other optimization tools that can be parametrized, but I do not know for sure.

PR #1423, opened 2023-03-13
[Feature Request]: relational graph with circle

pacles82:
Goodmorning,

is it possible generate rel graph with person as a circle, like this attach?
circle_person

Thank you

### DIFF
--- a/.github/workflows/gramps-ci.yml
+++ b/.github/workflows/gramps-ci.yml
@@ -28,7 +28,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


emyoulation:
The oval generic Male & Female Avatars would be easy to implement with no code changes. Just use the custom option. 
See https://github.com/gramps-project/addons-source/pull/459

(If you make new Avatars, please share them with the whole community.)

The oval photo thumbnails might be more of a challenge. The thumbnails are created by scaling the full Media Object or a user defined rectangular marquee in a media object.  So a ellipsoid marquee (selection area) would have to be added to the Photo Tagging add-on and  recognition of that shape would have to be added to the Thumbnail Generator.

https://gramps-project.org/wiki/index.php/Addon:Photo_Tagging_Gramplet

Nick-Hall:
Please use the [bug tracker](https://gramps-project.org/bugs) to file feature requests.

We also have a [discourse group](https://gramps.discourse.group) if you have any questions.

PR #1427, opened 2023-04-09
Attribute Date support

cdhorn:


emyoulation:
I am curious.

Tamura Jones commented on the phrase feature of Date values in GEDCOM 5.5 at https://www.tamurajones.net/GEDCOMDatePhrases.xhtml


He notes that parenthesis can be used to cue date parser that the value is a non-parsable "phrase"


Does the Gramps parser leverage that?

cdhorn:
Closing this since numerous items raised on mailing list in discussion around it a while back were left open ended and nothing seems to have been decided. As a note of reference here:

1. Nick noted date doesn't really seem to apply to attributes for events, which suggests need to introduce a third attribute class alongside the two existing ones, which broadens scope. 

2. If concerned about model consistency in that manner then instead of a single attribute type shared across multiple primary objects there should probably be different types broken out by primary object type, which broadens scope even more.

3. An argument made and some discussion that perhaps attributes should not be used for recording genealogical facts/characteristics about something at all, which suggests time should not be spent enhancing attributes as they should only be used for metadata, in which case date is not needed among other things.

PR #1428, opened 2023-04-11
Add SQLite undo log

DavidMStraub:
Triggered by the discussion in https://github.com/gramps-project/gramps-webapi/issues/320, I implemented an SQLite-based undo log by simply modifying the existing methods of `gramps.gen.db.DbGenericUndo`. That way, no other parts of the codebase need to be modified as the interface of the class stays the same.

Some more details:

### Existing implementation

An instance of `DbGenericUndo` has the properties `undoq` (of type `collections.deque`), `redoq` and `undodb` (of type list). When a transaction is commited to the database, the transaction object (`DbTxn`) is appended to `undoq` and the commits (1 or more) are appended to the `undodb` list.

When a transaction is undone, the transaction is moved from the `undoq` to the `redoq` and the commits (identified by two integers `first` and `last` in the transaction) are applied inversely and in opposite order to the database.

When a transaction is redone, it is moved again from the `redoq` to the `undoq` and the commits in `undodb` are applied again in the original order.

### New implementation

The steps outlined above are not modified, but the `undodb` list is removed and the list-like interface of `DbGenericUndo` is instead implemented by a SQLite table in a new database `undo.db` (the constant for this filename was already present in the API of the class, but unused, so it looks like this was already foreseen).

Instead of just a single table of commits, there are actually three tables to allow storing the history from multiple "sessions" (opening and closing the application). It is implemented in a way that even simultaneous sessions would record their transactions and commits in a consistent way. This would be relevant e.g. for Web API, where each HTTP request opens a new "session".

The first table is called "sessions" and just adds an ID and a timestamp for each new session. It only adds a row if the session edits the database (if transactions/commits are done).

grafik

The second table is called "transactions" and contains one row for every transaction. This is essentially the data stored in `undoq`, but persisted to the database, with one important addition: it also adds a row for any undo/redo transaction. This allows to fully reconsstruct the edit history of the database, except in the case of batch transactions like the GEDCOM import below, which is not linked to any traceable commits (`first` and `last` are `NULL`):

grafik

Finally, the actual commit data that was previously held by `undodb` is in the `commits` table. It is partially expanded out into columns to potentially allow filtering etc., but the interface of the class is still exactly the same as before. I also added an as yet empty JSON column to potentially replace the `old_data`/`new_data` BLOBs by a JSON patch object in the future (as suggested by @Nick-Hall) without the need to change the database schema.

grafik

### Final comments

This change does not yet enable the user to undo actions performed in a previous session of Gramps desktop, but this could be added in a future PR with relatively small changes. (Basically, it just requires loading part of the `transactions` table into `undoq` on startup.)

There is also no mechanism to truncate or VACUUM the undo.db, so it will keep growing; however, it can be safely deleted anytime for a fresh start.

### DIFF
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -80,10 +80,11 @@ class DbGenericUndo(DbUndo):
     table_txn = "transactions"
     table_undo = "commits"
 
-    def __init__(self, grampsdb, path) -> None:
+    def __init__(self, grampsdb: DbWriteBase, path: Optional[str] = None) -> None:
         super(DbGenericUndo, self).__init__(grampsdb)
         self.path = path
         self._session_id: Optional[int] = None
+        self.undodb: List[bytes] = []
 
     @property
     def session_id(self) -> int:
@@ -100,8 +101,9 @@ class DbGenericUndo(DbUndo):
         """
         Open the backing storage.
         """
-        self._create_tables()
-    
+        if self.path:
+            self._create_tables()
+
     def _create_tables(self) -> None:
         """Create the tables if they don't exist yet."""
         with self._connect() as connection:
@@ -110,13 +112,13 @@ class DbGenericUndo(DbUndo):
                 f"""CREATE TABLE IF NOT EXISTS {self.table_undo} (
                     session INTEGER,
                     id INTEGER,
-                    obj_class STRING,
+                    obj_class TEXT,
                     trans_type INTEGER,
-                    obj_handle STRING,
-                    ref_handle STRING,
+                    obj_handle TEXT,
+                    ref_handle TEXT,
                     old_data BLOB,
                     new_data BLOB,
-                    json STRING,
+                    json TEXT,
                     timestamp INTEGER,
                     PRIMARY KEY (session, id)
                 )
@@ -133,7 +135,7 @@ class DbGenericUndo(DbUndo):
                 f"""CREATE TABLE IF NOT EXISTS {self.table_txn} (
                     id INTEGER PRIMARY KEY,
                     session INTEGER,
-                    description STRING,
+                    description TEXT,
                     timestamp INTEGER,
                     first INTEGER,
                     last INTEGER,
@@ -159,11 +161,14 @@ class DbGenericUndo(DbUndo):
         Close the backing storage.
         """
         pass
-        
+
     def append(self, value) -> None:
         """
         Add a new entry on the end.
         """
+        if not self.path:
+            self.undodb.append(value)
+            return
         (obj_type, trans_type, handle, old_data, new_data) = pickle.loads(value)
         if isinstance(handle, tuple):
             obj_handle, ref_handle = handle
@@ -201,6 +206,8 @@ class DbGenericUndo(DbUndo):
         self, transaction: DbTxn, undo: bool = False, redo: bool = False
         ) -> None:
         """Post-transaction commit processing."""
+        if not self.path:
+            return
         msg = transaction.get_description()
         if redo:
             msg = _("_Redo %s") % msg
@@ -231,6 +238,8 @@ class DbGenericUndo(DbUndo):
         """
         Returns an entry by index number.
         """
+        if not self.path:
+            return self.undodb[index]
         with self._connect() as connection:
             cursor = connection.cursor()
             cursor.execute(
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Thanks for your comments, let me try to defend the implementation a bit :slightly_smiling_face: 

> I'd like to see the undo queue in the main database. The reason for this is to allow multiple front-ends to access a single remote database. We really need to what transactions another user has made. Recording the username of the user making the changes would also be useful.

I'm biased, but I don't see this as a compelling use case. If two users want to do occasional collaborative editing, they can just put their Gramps database directory in Dropbox and it will work perfectly fine, also with this implementation.

So the scenario you describe would only be relevant when using a remote database like PostgreSQL, but there is so much that's missing in current Gramps that makes it anyway impossible to do simultaneous collaborative editing:

- The GUI doesn't get notified or updated on remote changes
- The lock file is not synchronized
- There is no user management

I also considered implementing it in the main database, but here is why I decided not to:

- Simplicity: it is very natural to just extend the current `DbUndo` class to have a different storage mechanism.
- Does not break plugins: implementing new tables in the main database would require changing all existing database plugins. Since this implementation uses Python standard lib `sqlite3`, it does not break any plugin.
- Ease of truncation: it happened to me in various apps that the history got too large (maybe some plugins running multiple transaction and generating lots of log entries). Keeping it separate from the (much more valuable) genealogical data means that you can simply delete the transaction log at any point in time without losing any relevant genealogical data
- Ease of backup of genealogical data: I like my genealogical data to be separate and not meddled with less relevant log data, so my backups are smaller and easier to handle.
- Much easier to improve: this is perhaps the most important point in my opinion. This implementation allows to add the feature now with releatively small changes, but also allows to change it in the future. For instance, it currently stores blobs, but we could easily switch to JSON in the future - we simply inform the user that upgrading will empty the transaction log (which is anyway done on every restart currently). If we implement this as feature in the main database, we will have to worry about migrations, versioning, and all that - much more work.

Nick-Hall:
Allowing multiple front-ends to access a single remote database has been a design goal for some time.  We often get asked about sharing a database stored on a file server.  I have previously raised the issue of the GUI not being kept up-to-date, but the opinion was that such users were happy to refresh the interface.  I expect that the user management provided by the database would also be sufficient to start with.

There are two main issues preventing multi-user access at the moment:  the undo mechanism and the editors.  All other issues are fairly minor.  I already have a prototype for the editors.

Can we come up with a design that will satisfy both use cases?

Perhaps we could have an `undo_truncate` method that could truncate the logs to a given date or given number of transactions?

I'm happy for a database backup to include the undo logs, but I wouldn't include them in an XML export.  I also don't see a problem clearing the logs when performing a database upgrade.  This all seems reasonable.

We only have one database backend at the moment.  SqLite and PostgreSql are both DBAPI, BSDDB is being retired, and the MongoDB is only experimental.  Perhaps we could provide a `can_undo` method to allow a backend to choose to not implement the undo functionality?

Nick-Hall:
I have been looking into this further.  There was a similar discussion back in 2015 resulting in [GEPS 038](https://gramps-project.org/wiki/index.php/GEPS_038:_Enhanced_Transaction_Log): Enhanced Transaction Log.

People still ask for multi-user functionality on a fairly regular basis.  It would be short-sighted of us not to factor this into our design.

I suggest the following:

1. Keep the `DbGenericUndo` class unchanged to be used as a fallback.
2. In `DbGeneric`, write a `_create_undodb` method to instantiate this class.
3. In `DBAPI`, override this method to instantiate your new undo class instead.
4. Modify your class to use the main database instead of creating a new one.

The GUI also uses the `DbUndo` class.  To get the [`UndoHistory`](https://github.com/gramps-project/gramps/blob/master/gramps/gui/undohistory.py#L64) window to work you will need to implement the `undo_count` and `redo_count` properties.  There is a "Clear" button to which calls the `clear()` method to truncate the logs.  We may want to enhance this so that the user can select the amount of history to remove.

Let me know what you think.  We can always continue to refine this further.

Nick-Hall:
@DavidMStraub I have created PR #1479 which will allow you to implement this functionality in a plugin.

DavidMStraub:
Cool, will have a look.

DavidMStraub:
Will close this for now and reimplement as plugin using #1479.

PR #1433, opened 2023-05-12
Fix get_change_display

olivierberten:
Not sure it's the right fix but the current function raises a TypeError:
```  File "/usr/lib/python3/dist-packages/gramps/gen/lib/tableobj.py", line 128, in get_change_display
    return str(time.strftime('%x %X', time.localtime(self.change)),
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: decoding str is not supported
```
AFAIK this function isn't used anywhere in the code but as I wanted to use it, I realized it's broken.

### DIFF
--- a/gramps/plugins/webreport/basepage.py
+++ b/gramps/plugins/webreport/basepage.py
@@ -881,7 +881,7 @@ class BasePage:
         found = any(data[3] == place_handle and data[4] == event_date
                     for data in place_lat_long)
         if not found:
-            placetitle = _pd.display(self.r_db, place)
+            placetitle = _pd.display(self.r_db, place, fmt=0)
             latitude = place.get_latitude()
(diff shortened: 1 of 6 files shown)


gramps/gen/lib/tableobj.py
```diff
@@ -125,7 +125,7 @@ def get_change_display(self):
 
         """
         if self.change:
-            return str(time.strftime('%x %X', time.localtime(self.change)),
```
Nick-Hall, line 128: I think that the following should work:

`return time.strftime('%x %X', time.localtime(self.change))`

Nick-Hall:
Well spotted.  I think we should fix this.

PR #1434, opened 2023-05-13
Narrative web: problem with small pictures.

SNoiraud:
Fixes [#12884](https://gramps-project.org/bugs/view.php?id=12884)

### DIFF
--- a/gramps/plugins/webreport/media.py
+++ b/gramps/plugins/webreport/media.py
@@ -472,7 +472,13 @@ class MediaPages(BasePage):
                                 if orig_image_path != newpath:
                                     url = self.report.build_url_fname(
                                         newpath, None, self.uplink)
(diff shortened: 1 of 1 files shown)


lordemannd:
It works well for me on Windows:

simple demonstration:
- before adding PR 1434
- create & load new empty DB
- add a person
- add image (scanned_microfilm-from-gramp-example-db.png) from the gramps example tree.
- add the user supplied image from the bug report (1910c_ElizabethEleanorKleine.jpg)
- generate a nar web report "nar-web-before"

add the code from PR 1434
- RESTART Gramps
- load the test DB
- generate a nar web report "nar-web-after"

compare the before and after,
the image "1910c_ElizabethEleanorKleine.jpg" is now more viewable size
the image "scanned_microfilm-from-gramp-example-db.png" is no longer destorted

PR #1435, opened 2023-05-13
Fix export where private citations are excluded

Nick-Hall:
Fixes bug [#12886](https://gramps-project.org/bugs/view.php?id=12886).


lordemannd:
works for me with Gramps 5.1.5 on Windows

Nick-Hall:
Merged in commit bfa7c0c9f4fcd91f866710c0f68b215b52907eb2

PR #1437, opened 2023-05-19
Fix htmldoc image filenames

sbeyer:
Filenames of images generated by `HtmlDoc` did not contain the crop value. This had the effect that photos of several people from the same source image (but with different crops) were overwritten, so all images of those people showed the same person.

The issue is fixed by adding a crop suffix.


Nick-Hall:
This looks like a duplicate of #1412 which I just merged in the _gramps51_ maintenance branch.

sbeyer:
It is a duplicate. I used master and the bug was still there, also I checked if there's another (non-completed) pull request. I didn't know about custom branches.

PR #1439, opened 2023-05-27
add a splash screen on startup

jmichault:
The delay between double-clicking the application and the icon appearing can be long.
This pull request add a splash screen as soon as possible, allowing the user to wait.

### DIFF
--- a/gramps/gui/grampsgui.py
+++ b/gramps/gui/grampsgui.py
@@ -38,7 +38,7 @@ LOG = logging.getLogger(".grampsgui")
 #
 #-------------------------------------------------------------------------
 from gramps.gen.config import config
-from gramps.gen.const import DATA_DIR, IMAGE_DIR, GTK_GETTEXT_DOMAIN
+from gramps.gen.const import DATA_DIR, IMAGE_DIR, GTK_GETTEXT_DOMAIN, SPLASH
 from gramps.gen.constfunc import has_display, lin
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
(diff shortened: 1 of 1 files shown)


PQYPLZXHGF:
@jmichault How about a screenshot please?  Is it like the [portable app version of Gramps](https://portableapps.com/apps/education/gramps_portable) ?

emyoulation:
See Discourse forum discussion [loader progress window](https://gramps.discourse.group/t/loader-progress-window-can-a-splash-screen-be-added-to-the-windows-version-of-gramps/806/6) for previous discussions and capture of the PortableApps launcher/splashscreen
image

A big question, how fast is this splash screen for the initial install launching of Gramps?  

Does it have the near half-minute delay while the Gtk libraries and their dependencies are cached before starting?

jmichault:
> @jmichault How about a screenshot please? Is it like the [portable app version of Gramps](https://portableapps.com/apps/education/gramps_portable) ?

Capture du 2023-05-27 17-29-49

It's only the flash.jpg picture found in images directory. No progress bar.

emyoulation:
FYI, that is a photo of Robert "Bert" Allingham, Jr.; the Gramps of the original author of GRAMPS 1.0, Don Allingham. Which is from the About dialog.

We might be better served by:

https://upload.wikimedia.org/wikipedia/commons/b/ba/GRAMPS_logo.png

Or a variant of the website's WordPress decoration (WITHOUT the OS mentions - due to Trademark hassles)

https://gramps-project.org/blog/wp-content/uploads/2016/01/Gramps-Downloads-Main-2015-01.jpg

SNoiraud:
I don't like that and this splash screen must be an option. I don't want to have it.

emyoulation:
How about if the Splash only appeared from the GUI launcher, not the console?

PR #1440, opened 2023-05-27
Add substitution key for maiden name

Gnarflord:
### Why
I wanted to add both the married name as well as the maiden name to our family tree. For this I'm using a new substitution key `h`, which tries to find out a person's maiden name (`h` was chosen at random because it was still free, I'm open to other suggestions).

### Example
For example one could use the pattern `$n {born "$n(h)" }`

For a person who has a preferred surname which is different from their birth surname this results in:
- Mary Foo born Bar

For a person who did not adopt a new surname, `h` remains empty. This results in:
- Mary Bar

### How
The approach for this is:
1. Search for birth names and married names
2. If there's both a birth name and a married name AND they differ: Return the birth day
```python
    def get_maiden_name(self):
        married_name = None
        birth_name = None
        for name in [self.get_primary_name()] + self.get_alternate_names():
            if name.get_type() == NameType.BIRTH and name.get_surname():
                birth_name   = name.get_surname()
            elif name.get_type() == NameType.MARRIED and name.get_surname():
                married_name = name.get_surname()
        if birth_name and married_name and birth_name != married_name:
            return birth_name
        return None
```

### Caveats
The pattern only woks when the preferred name is the married name (if the preferred name is the birth name it would result in "Mary Bar born Bar"). I was thinking about an additional substitution key to find out the married name, but this is more complicated because there can be multiple married names. I guess one could search for the most recent?

### DIFF
--- a/gramps/gen/lib/person.py
+++ b/gramps/gen/lib/person.py
@@ -40,6 +40,7 @@ from .ldsordbase import LdsOrdBase
 from .urlbase import UrlBase
 from .tagbase import TagBase
 from .name import Name
+from .nametype import NameType
 from .eventref import EventRef
 from .personref import PersonRef
 from .attrtype import AttributeType
@@ -634,6 +635,18 @@ class Person(CitationBase, NoteBase, AttributeBase, MediaBase,
             if int(attr.type) == AttributeType.NICKNAME:
                 return attr.get_value()
         return ''
+
+    def get_maiden_name(self):
+        married_name = None
+        birth_name = None
+        for name in [self.get_primary_name()] + self.get_alternate_names():
+            if name.get_type() == NameType.BIRTH and name.get_surname():
+                birth_name   = name.get_surname()
+            elif name.get_type() == NameType.MARRIED and name.get_surname():
+                married_name = name.get_surname()
+        if birth_name and married_name and birth_name != married_name:
+            return birth_name
+        return None
 
     def set_gender(self, gender):
         """
--- a/gramps/plugins/lib/libsubstkeyword.py
+++ b/gramps/plugins/lib/libsubstkeyword.py
@@ -186,7 +186,7 @@ class NameFormat(GenericFormat):
                     common,                    # x
                     name.get_suffix,           # s
                     name.get_surname,          # l
-                    name.get_family_nick_name, # g
+                    name.get_family_nick_name  # g
                    ]
 
         return self.generic_format(name, code, upper, function)
@@ -823,7 +823,7 @@ class VariableParse:
     def is_a(self):
         """ check """
         return self._in.this == "$" and self._in.next is not None and \
-                              "näsijbBdDmMvVauetTpPG".find(self._in.next) != -1
+                              "nhsijbBdDmMvVauetTpPG".find(self._in.next) != -1
 
     def get_event_by_type(self, marriage, e_type):
         """ get an event from a type """
@@ -946,7 +946,8 @@ class VariableParse:
         if next_char == "n":
             #Person's name
             return self.__parse_name(self.friend.person, attrib_parse)
-        elif next_char == "ä":
+
+        elif next_char == "h":
             if self.empty_item(self.friend.person):
                 return
 #            print(self.friend.person.get_maiden_name())
--- a/gramps/plugins/lib/libsubstkeyword.py
+++ b/gramps/plugins/lib/libsubstkeyword.py
@@ -186,7 +186,7 @@ class NameFormat(GenericFormat):
                     common,                    # x
                     name.get_suffix,           # s
                     name.get_surname,          # l
-                    name.get_family_nick_name  # g
+                    name.get_family_nick_name, # g
                    ]
 
         return self.generic_format(name, code, upper, function)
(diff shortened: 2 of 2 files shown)


Kirusi:
Definition of maiden name is quite vague. A girl could be born and registered with father's last name, but after a few months  parents are divorced, and girl's last name is changed to mother's last name, which becomes her maiden name at the time of marriage. Another relatively common case is that a girl is born, but then gets adopted and her last name is changed to adoptive family's last name.

Examples above are edge cases for regular life, but gramps is a genealogical software. It should be handling cases related to complex relationships.

In my personal trees I, personally, prefer to report person's preferred last name and other last names that they used during their lives (it's not a single last name, but a list of names). It's not only about girls, occasionally guys also change last names during their lives.

Gnarflord:
I see what you mean. Out of curiosity, how do you generate this list of last names? Is it also based on substitution keys or do you have some custom procedure in place?

I absolutely agree with you that gramps needs to be able to handle such edge cases. It's just that I've found multiple requests to list both maiden and married names in the past with no satisfactory answer so far. Do you have a better idea?

emyoulation:
I recall seeing a Report that had an option to show the last multiple surname. But cannot dig it up on the spur of the moment.

But this 2010 Blog entry is interesting:
https://gramps-project.org/blog/2010/11/final-multiple-surnames/

Kirusi:
> I see what you mean. Out of curiosity, how do you generate this list of last names? Is it also based on substitution keys or do you have some custom procedure in place?
> 
> I absolutely agree with you that gramps needs to be able to handle such edge cases. It's just that I've found multiple requests to list both maiden and married names in the past with no satisfactory answer so far. Do you have a better idea?

@Gnarflord To suit my needs I just created a custom version of the report. For the general case, I would probably consider a modifier on the substitution keys. Regular version reports the preferred name, substitution key with modifier reports preferred name and all remaining names  in parentheses. For example:
$n(g f) produces Sarah Smith
$n(g f!) produces Sarah Smith (Black, Jones, Miller)
where "Smith (Black, Jones, Miller)" is produced by a single substitution key - "f!". The first name in the list could be the preferred one. The rest would be sorted alphabetically (to keep things simple).

Another format could be:
$n(g f/) which could produce Sarah Smith/Black/Jones/Miller

Modifiers could be applied to any substitution keys that represent parts of the name.

I'm Russian, so the problem of multiple names applies not only to last names. Some people in my tree have multiple patronymic names. Some of them were adopted, hence the change of the father and of the patronymic. In some cases multiple patronymic names are produced from the same father's name, e.g. "Egorovich" and "Georgievich", because the name "Egor" is consider a synonym to "Georgiy". Hence the same physical person had documents with different patronymics.

my 2c

Kirusi:
Additionally, one needs to consider  cultures where people have patrilineal and matrilineal last names, like in Spanish speaking countries. A cousin whose maiden name was Beatriz Diaz Garcia, after marriage becomes Beatriz Perez Diaz (matrilineal last name is removed, husband's patrilineal last name is added). In this case combining multiple last names gets cumbersome.

Nick-Hall:
I'm not sure if there was any agreement here.

emyoulation:
We may need a tool to begin to create consistent surname Origin data for some statistical analysis.

Most of my Tree's "Birth Name" and "Married Name" records  do not have the "Origin" or date range defined.  (And the date range is kind of ugly since a bride's name date ranges are event date based.  In the US patriarchal society following couverture doctrine, a female's Birth name is likely a 'Patrilineal' origin from "'Birth' to 'Marriage'" and her married surname is likely 'Taken' origin from "'Marriage' to 'Death'". Redundant-entry of the date in the Marriage event and the 2 name date ranges seems both excessive and easily overlooked. References to the event as a date placeholder would be lower maintenance for the genealogist.)

The SuperTool script created as an [Example in a Discourse tutorial](https://gramps.discourse.group/t/tree-vivisection-experiments-with-the-isotammi-supertool/1621) shows a rule that could be one option in a Name harmonization tool.  It checks if Primary surname(s) matches the Father's Primary surname(s) and, if so, marks the null origins as "Patrilineal". (Looking at this years later, some logic holes are apparent.  Shouldn't it be preferring to working with Birth names rather than just Primary?)  

But without consistent date ranges and origins, is there enough information to extrapolate cultural bias?

dsblank:
For me, I just added a "Married Name" for each woman that is married but have a name different from cultural norms (used in the Calendar and Birthday/Anniversary Reports).

Can we close this PR, or move to "draft" status?

Nick-Hall:
I can see why someone would want to display a name as "Mary Smith née Jones" where "Mary Smith" is the married name and "Mary Jones" is the birth name.

At the moment the name displayer doesn't compose a name format from different names.  More thought is needed.  I don't think that we have a good solution yet.

Perhaps as a workaround a new name combining the married name and birth name could be created.

emyoulation:
Maybe if the Origin attribute was more consistently entered then the logic would have more to work with.

The [SuperTool script to set Patrilineal and Taken origins for Married Names](https://gramps.discourse.group/t/tree-vivisection-experiments-with-the-isotammi-supertool/1621/11?u=emyoulation) might be a starting point. (And making [Surname Guessing](https://gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Settings#Input_Options) also guess Patrilineal origin for children/fathers added while using "Father's surname" default.)

There may also be some conflict over whether to use the "Mary Smith née Jones" or "Mary (née Jones) Smith" format for Married Names. The latter works nicely with the couples formats: "Dr. Thomas and Prof. Mary (née Jones) Smith"

Nick-Hall:

PR #1443, opened 2023-06-09
allow build on windows

jmichault:
Build of development branch with `python3 setup.py build` fails under windows, because os.system run cmd, not sh.

This pull request forces run of sh.


Nick-Hall:
We should really be using `subprocess.call` rather than `os.system`.  Please test my update on Windows.

PR #1447, opened 2023-06-30
Allow web-accessible file references in media objects

Nick-Hall:
Gedcom 7.0 allows an URL with scheme ftp, http, or https in the FILE tag.

See:

https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#FILE

### DIFF
--- a/gramps/gen/const.py
+++ b/gramps/gen/const.py
@@ -288,6 +288,7 @@ THUMBSCALE = 96.0
 THUMBSCALE_LARGE = 180.0
 SIZE_NORMAL = 0
 SIZE_LARGE = 1
+REMOTE_MIME = "application/http"
 XMLFILE = "data.gramps"
 NO_SURNAME = "(%s)" % _("none", "surname")
 NO_GIVEN = "(%s)" % _("none", "given-name")
--- a/gramps/gen/mime/_pythonmime.py
+++ b/gramps/gen/mime/_pythonmime.py
@@ -20,7 +20,7 @@
 # gen/mime/_pythonmime.py
 
 import mimetypes
-from ..const import GRAMPS_LOCALE as glocale
+from ..const import GRAMPS_LOCALE as glocale, REMOTE_MIME
 
 _ = glocale.translation.gettext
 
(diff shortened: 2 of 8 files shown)


DavidMStraub:
An alternative to implenting URL handling in the GUI code would be implementing all file handling centrally in `gramps.gen`. This would make it easier for addons to also support web accessible files (using standard library functions). It would also eventually allow things like web accessible media base dirs natively, as used currently in Web API (e.g., `s3://some-bucket` as media base dir). And we could then forbid all direct use of `open()`, which is a big security headache in shared environments.

janskarvall:
I made a quick test making a web report from a simple tree with one person having a reference to a media object with a web url.
The media shows up in the report as a clickable web icon under the heading "Other media: vidéos, pdfs....". Clicking on the icon shows a page where it is stated that the file has been moved or deleted. Maybe clicking on the web icon under the heading "Other media: vidéos, pdfs...." could open that webpage instead of opening the page that now says that the file has been moved or deleted. Also the heading "Other media: vidéos, pdfs...." should mention web accessible media somehow.

janskarvall:
An html page may have a favicon (See "https://en.wikipedia.org/wiki/Favicon") that could be used as a "thumbnail". See for example the MIT licensed "https://pypi.org/project/favicon/". The first frame in a video could too. I used ffmpeg to extract it on Linux, but there may be other ways to extract it that is more portable.

Nick-Hall:
@janskarvall Thanks. I'll try to look at this again soon.

janskarvall:
Maybe a bit off topic, but I am thinking that showing the Media reference editor when adding web accessible media to a gallery is superfluous as long as the media reference has nothing more to add than cropping information. Looking closer, the media reference can override the title, which seems usable when e.g. the picture shows several people and it is cropped down to a singel person. But for media that can not be cropped, there is not likely any need to override the title. The same reasoning I think is valid for any non-image media. According to GEDCOM 7 MULTIMEDIA_LINK:

<pre>
The optional CROP (p.72) substructure indicates that a subregion of an image represents
or applies to the superstructure.
The optional TITL (p.91) substructure supersedes any OBJE.FILE.TITL substructures
included in the MULTIMEDIA_RECORD .
</pre>

The Media reference editor in gramps does not support overriding the title. Maybe it should.

emyoulation:
This PR has been listed in the proposed Roadmap for 5.3 as "Allow web-accessible files in media paths."
https://gramps.discourse.group/t/roadmap-for-v5-3/5203

Can it be committed yet?

PR #1450, opened 2023-07-06
Port from GtkSpell to Gspell

a-andre:
I could only test this on Linux, so additional changes might be necessary for MacOS and Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,11 +75,9 @@ The following packages are **STRONGLY RECOMMENDED** to be installed:
 
 The following packages are optional:
 ------------------------------------
-* **gtkspell**
+* **gspell**
 
- Enable spell checking in the notes. Gtkspell depends on
- enchant. A version of gtkspell with gobject introspection
- is needed, so minimally version 3.0.0.
+ Enable spell checking in the notes.
 
 * **rcs**
 
--- a/docs/coregui/gui_widgets.rst
+++ b/docs/coregui/gui_widgets.rst
@@ -302,7 +302,7 @@ Styled Text Buffer
    :members:
    :undoc-members:
    :show-inheritance:
-.. autoclass:: GtkSpellState
+.. autoclass:: GspellState
    :members:
    :undoc-members:
    :show-inheritance:
--- a/gramps/gen/utils/grampslocale.py
+++ b/gramps/gen/utils/grampslocale.py
@@ -790,6 +790,8 @@ class GrampsLocale:
         """
         return self.language
 
+    def locale_code(self):
+        return self.lang
 
     def get_addon_translator(self, filename, domain="addon",
                              languages=None):
--- a/gramps/grampsapp.py
+++ b/gramps/grampsapp.py
@@ -363,53 +363,12 @@ def show_settings():
     def verstr(nums):
         return '.'.join(str(num) for num in nums)
 
-    #GTKSPELL_MIN_VER = (3, 0)
-    #gtkspell_min_ver_str = verstr(GTKSPELL_MIN_VER)
-    # ENCHANT_MIN_VER = (0, 0)  # TODO ?
-    gtkspell_ver_tp = (0, 0)
-    # Attempting to import gtkspell gives an error dialog if gtkspell is
-    # not available so test first and log just a warning to the console
-    # instead.
     try:
-        from gi import Repository
-        repository = Repository.get_default()
-        gtkspell_ver = _("not found")
-        if repository.enumerate_versions("GtkSpell"):
-            try:
-                gi.require_version('GtkSpell', '3.0')
-                from gi.repository import GtkSpell as Gtkspell
-                gtkspell_ver = str(Gtkspell._version)
-                aaa = Gtkspell._version.split(".")
-                v1 = int(aaa[0])
-                v2 = int(aaa[1])
-                gtkspell_ver_tp = (v1, v2)
-                # print("gtkspell_ver " + gtkspell_ver)
-            except Exception:
-                gtkspell_ver = _("not found")
-        elif repository.enumerate_versions("Gtkspell"):
-            try:
-                gi.require_version('Gtkspell', '3.0')
-                from gi.repository import Gtkspell
-                gtkspell_ver = str(Gtkspell._version)
-                gtkspell_ver_tp = Gtkspell._version
-                # print("gtkspell_ver " + gtkspell_ver)
-            except Exception:
-                gtkspell_ver = _("not found")
+        gi.require_version('Gspell', '1')
+        from gi.repository import Gspell
+        gspell_ver = str(Gspell._version)
     except Exception:
-        gtkspell_ver = _("not found")
-
-    try:
-        import enchant
-        enchant_result = enchant.get_enchant_version()
-    except Exception:
-        from ctypes import cdll, c_char_p
-        try:
-            enchant = cdll.LoadLibrary("libenchant")
-        except FileNotFoundError:
-            enchant = cdll.LoadLibrary("libenchant-2")
-        enchant_ver_call = enchant.enchant_get_version
-        enchant_ver_call.restype = c_char_p
-        enchant_result = enchant_ver_call().decode("utf-8")
+        gspell_ver = _("not found")
 
     #RCS_MIN_VER = (5, 9, 4)
     #rcs_ver_str = verstr(RCS_MIN_VER)
@@ -539,8 +498,7 @@ def show_settings():
     print('')
     print("Optional:")
     print("---------")
-    print(' Gtkspell   :', gtkspell_ver)
-    print(' Enchant    :', enchant_result)
+    print(' Gspell     :', gspell_ver)
     print(' RCS        :', rcs_ver)
     print(' PILLOW     :', pil_ver)
     print(' GExiv2     : %s' % gexiv2_str)
--- a/gramps/gui/configure.py
+++ b/gramps/gui/configure.py
@@ -71,7 +71,7 @@ from .display import display_help
 from gramps.gen.plug.utils import available_updates
 from .plug import PluginWindows
 #from gramps.gen.errors import WindowActiveError
-from .spell import HAVE_GTKSPELL
+from .spell import HAVE_GSPELL
 from gramps.gen.constfunc import win
 _ = glocale.translation.gettext
 from gramps.gen.utils.symbols import Symbols
@@ -1654,14 +1654,14 @@ class GrampsPreferences(ConfigureDialog):
                                 row, 'behavior.spellcheck', start=1, stop=3,
                                 tooltip=_("Enable the spelling checker"
                                           " for notes."))
-        if not HAVE_GTKSPELL:
+        if not HAVE_GSPELL:
             obj.set_sensitive(False)
             spell_dict = {'gramps_wiki_build_spell_url':
                           URL_WIKISTRING +
                           "GEPS_029:_GTK3-GObject_introspection"
                           "_Conversion#Spell_Check_Install"}
             obj.set_tooltip_text(
-                _("GtkSpell not loaded. "
+                _("Gspell not loaded. "
                   "Spell checking will not be available.\n"
                   "To build it for Gramps see "
                   "%(gramps_wiki_build_spell_url)s") % spell_dict)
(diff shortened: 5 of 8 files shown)


Nick-Hall:
We still need to override the `LANGUAGE` variable if it is set to "en:fr".

Try running:

`LANGUAGE=en:fr python Gramps.py`

The menus are in French!

Perhaps we should only update the `LANGUAGE` variable if it is not empty.

In testing this I noticed a bug in the existing code if I set `LANGUAGE=en_GB:fr`.  @jralls - perhaps you could look at this if you have time.

The rest of this PR seems fine.

jralls:
@Nick-Hall 
> In testing this I noticed a bug in the existing code if I set LANGUAGE=en_GB:fr. @jralls - perhaps you could look at this if you have time.

Are you going to tell me what the bug is?

> Gspell.language_get_default() needs xx_XX to detect the right language, but if LANGUAGE gets overriden (with xx), only xx is provided to Gspell and language_get_default() fails.

I don't think that's true. `Gspell.language_get_default` calls [`g_get_language_names`](https://gitlab.gnome.org/GNOME/gspell/-/blame/master/gspell/gspell-language.c#L146) and that [is documented to return two-character language codes](https://docs.gtk.org/glib/func.get_language_names.html) and `gspell_lookup_language` [does a second match check](https://gitlab.gnome.org/GNOME/gspell/-/blame/master/gspell/gspell-language.c#L222) looking only at the passed-in code's length.

I suspect that it's not some xx that's causing trouble, it's "C". @a-andre can you explain in more detail what language codes tripped up `Gspell.language_get_default`?

Nick-Hall:
> Are you going to tell me what the bug is?

Sorry I wasn't very clear.

When I run `LANGUAGE=en_GB:fr python Gramps.py` in the _master_ branch, the main menu is in French.  I expect it in English.

a-andre:
`gramps/gen/utils/grampslocale.py` hasn't seen major changes for more than 10 years. A lot of workarounds included there might no longer be necessary, but cause issues now. So they should be removed.

jralls:
>>  Are you going to tell me what the bug is?

> Sorry I wasn't very clear.

> When I run LANGUAGE=en_GB:fr python Gramps.py in the master branch, the main menu is in French. I expect it in English.

Ah. I understood that to mean that was with this PR merged.

> gramps/gen/utils/grampslocale.py hasn't seen major changes for more than 10 years. A lot of workarounds included there might no longer be necessary, but cause issues now. So they should be removed.

Not quite. They should be checked and removed if no longer needed. As for there being no major changes in 10 years, the libc localization, derived from POSIX, hasn't had any substantial changes in 20 years, and I don't see anything in [Python's gettext history](https://github.com/python/cpython/commits/main/Lib/gettext.py) suggesting that's changed in a way that would affect grampslocale.py.

Nick-Hall:
There seems to be a feature of [gspell_language_lookup](https://gitlab.gnome.org/GNOME/gspell/-/blob/master/gspell/gspell-language.c#L222).  When finding the default language, this function is called with each item in the LANGUAGE list. If no exact match is found, a closest match is found by only matching to the length of the language of the available dictionary.  This works if the dictionary is shorter than the LANGUAGE item.  e.g. 'he_IL' matches 'he', but it doesn't work the other way around - 'fr' doesn't match 'fr_FR'.

So if a user sets LANGUAGE to 'fr' then they will not match a dictionary and 'en_US' is used as the next fallback.  This is probably the intended functionality since we don't know which French dictionary is required.

My suggestion is to set `os.environ["LANGUAGE"] = self.lang.split('.')[0]`.  This seems to work in most cases.

Nick-Hall:
Added some unit tests.

I'm getting the error message:  `ValueError: Namespace Gspell not available`

jralls:
> If no exact match is found, a closest match is found by only matching to the length of the language of the available dictionary. This works if the dictionary is shorter than the LANGUAGE item. e.g. 'he_IL' matches 'he', but it doesn't work the other way around - 'fr' doesn't match 'fr_FR'.

You have it backwards. Here's the code:
```
	available_languages = gspell_language_get_available ();

	for (l = available_languages; l != NULL; l = l->next)
	{
		const GspellLanguage *language = l->data;
		const gchar *code = language->code;
		gsize length = strlen (code);

		if (g_ascii_strcasecmp (language_code, code) == 0)
		{
			return language;
		}

		if (g_ascii_strncasecmp (language_code, code, length) == 0)
		{
			closest_match = language;
		}
	}

````

`language` is one of the languages that GSpell supports, `code` is the passed-in code. `length` is the length of *code* and `g_ascii_strncasecmp` limits the comparison to `length` characters. So if you pass in a 2-character `code` and `language` is 5 characters&mdash;e.g. `code = "fr"` and `language = "fr_FR"`&mdash;then it will match, but if you reverse that, i.e. pass in `code = "fr_FR"` while GSpell has only `"fr"` in its language list, it will fail. *That's a bug*, what they want to do is get the lengths of both `code` and `language` and pass the `MIN` to `g_ascii_strncasecmp`. Or, since they're presumably not dealing with MSWindows locale/language names, they could just hard-code 2 characters.

There are two ways to work around that: Setting the GSpell language ourselves instead of using `gspell.language_get_default()` or, considerably less work, catching 5-character codes and echoing the 2-character equivalent afterwards, e.g. if `LANGUAGE` has "en_GB" we replace it with "en_GB:en", except that we should not do that for Chinese, though maybe we should echo "zh_HK" as  "zh_HK:zh_TW" ([Oracle](https://docs.oracle.com/cd/E19683-01/806-6642/new-tbl-77/index.html) says that traditional characters are used in Hong Kong, though I suppose the new management is going to change that if they haven't already). The second is much simpler.

**Comment by a-andre** [comment]
> can you explain in more detail what language codes tripped up Gspell.language_get_default?

```
$ env | grep LANG
LANG=de_DE.UTF-8
```

Before overwriting `LANGUAGE` `GLib.get_language_names()` returns
 > `['de_DE.UTF-8', 'de_DE', 'de.UTF-8', 'de', 'C']`

and `Gspell.language_get_default().get_code()` returns
> `de_DE`

With `os.environ["LANGUAGE"] = ':'.join(self.language)` `GLib.get_language_names()` returns
 > `['de', 'C']`

and `Gspell.language_get_default().get_code()` returns
> `en_US`

With `os.environ["LANGUAGE"] = self.lang.split('.')[0]` `GLib.get_language_names()` returns
> `['de_DE', 'de', 'C']`

and `Gspell.language_get_default().get_code()` returns
> `de_DE`

So using  `os.environ["LANGUAGE"] = self.lang.split('.')[0]` works for me.

**Comment by jralls** [comment]
>  Or, since they're presumably not dealing with MSWindows locale/language names, they could just hard-code 2 characters

Following up on that I noticed a merge request to implement exactly that, and pulling the string found that GSpell [uses ICU to handle MSWindows](https://gitlab.gnome.org/GNOME/gspell/-/issues/14), creating an ICU dependency.

**Comment by jralls** [comment]
> So using  os.environ["LANGUAGE"] = self.lang.split('.')[0] works for me.

And breaks Gramps for everyone who set $LANGUAGE differently from $LANG.

>With os.environ["LANGUAGE"] = ':'.join(self.language) GLib.get_language_names() returns

>> ['de', 'C']

> and Gspell.language_get_default().get_code() returns

>> en_US

> [You have it backwards. Here's the code:](https://github.com/gramps-project/gramps/pull/1450#issuecomment-1626354931)

Crud, *I* had it backwards. Sorry, @Nick-Hall.

> `code` is the passed-in code.

Wrong, `language_code` is the passed in code, `code` is `language`'s code, so `length` is the length of the stored language code and if $LANGUAGE contains only the two-letter version (as the gettext manual [says it should](https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html)) then it won't match. 

@a-andre What OS/distro are you using and how did you install your spelling dictionary?

**Comment by a-andre** [comment]
> What OS/distro are you using and how did you install your spelling dictionary?

I'm on Arch Linux using the hunspell dictionaries [en_us](https://archlinux.org/packages/extra/any/hunspell-en_us/) and [de](https://archlinux.org/packages/extra/any/hunspell-de/). (Under "Package Contents" you can find the list of files these packages provide.)

**Comment by jralls** [comment]
Thanks.

I checked the latest Windows AIO as well and it's using hunspell too, as does macOS, so it seems we need worry only about that. Their dictionaries are all xx_XX.dic so that's what we need to pass to gspell.

We need xx for all but a handful of translation files, so ISTM $LANGUAGE and `GSpell.language_get_default` isn't the way there.  A clean way would be to add
```
locale_code():
    return self.lang
```
to GrampsLocale's public functions and replace
```
GSpell.Checker.new(GSpell.language_get_default())
```
with
```
locale_code = glocale.locale_code()
GSpell.Checker.new(locale_code is not None ? locale_code()[:5] : None)
```
Note that explicitly calling `GSpell.language_get_default()` is redundant: [GSpell.Checker.new()](https://lazka.github.io/pgi-docs/Gspell-1/classes/Checker.html#Gspell.Checker.new) does that by default.

We might consider adding a [GSpell Chooser Button](https://lazka.github.io/pgi-docs/Gspell-1/classes/LanguageChooserButton.html) to the preferences dialog next to and enabled by the enable spell check check box and let that override the locale value for users who might want spell check in a different language or region from their locale to apply in GtkEntries.

**Comment by Nick-Hall** [comment]
I tried a couple of backends listed in the [Enchant](https://abiword.github.io/enchant/) documentation.

For French and English (US and GB), I used hunspell.  I confirm that the dictionaries use xx_XX codes.

For Hebrew I used HSpell.  It gives a  single dictionary listing of `('he', <Enchant: Hspell Provider>)`.

I didn't try voikko or zemberek because they didn't look easy to install.

**Comment by jralls** [comment]
Humph. So I guess we need to be a bit more elaborate:
```
locale_code = glocale.locale_code()
gspell_language = None
if locale_code is not None:
    gspell_language = GSpell.language_lookup(locale_code:[5])
    if gspell_language is None:
        gspell_language= GSpell.language_lookup(locale_code:[2])
checker = GSpell.Checker.new(gspell_language)
```

Nick-Hall:
I'm sometimes getting a mismatch of UI languages now.  If I start Gramps with `LANGUAGE=fr_FR LANG=en_GB.utf8`, I get the UI in French but the dictionary list is in English with the US (not GB) dictionary as the default.  So it seems that the UI language for the dictionary list is taken from the default dictionary.

With `LANGUAGE=fr_FR LANG=fr_FR.utf8` everything is fine, however with `LANGUAGE=fr_BE LANG=fr_BE.utf8` the UI is correct but it defaults to the US dictionary not the Belgian or even French one.

jralls:
@Nick-Hall 
The first part isn't too surprising, except that it's not in English English since GSpell does have [an en_GB.po](https://gitlab.gnome.org/GNOME/gspell/-/blob/master/po/en_GB.po). Maybe it doesn't have msgstrs for the msgids you're looking at.

Do you have an fr_BE dictionary installed? GSpell's locale selection isn't smart enough to find a different French dictionary that you do have installed.

I don't think there's anything to be done for changing GSpell to using $LANGUAGE correctly except rewrite their localization for them and submit an MR. I don't have time. We'd also have to do our own dictionary matching to get it to fall back on fr_FR if you set LANG=fr_BE but have only an fr_FR dictionary installed.

Nick-Hall:
@jralls The `LANGUAGE=fr_BE LANG=fr_BE.utf8` case works OK.  I forgot to run `locale-gen` after adding the locale.

Where Gspell defaults to the US dictionary with `LANGUAGE=fr_FR LANG=en_GB.utf8` is probably not worth worrying about.  The main case I can see happening is where someone runs Gramps in a language that we don't translate and specifies some fallback in the LANGUAGE variable.  e.g. `LANGUAGE=ka:fr LANG=ka_GE.utf8` used by a Georgian specifying French as a fallback.  In the case, they will get a French UI with the spell checker dictionary listings in Georgian.

I'll merge this shortly if there are no objections.

jralls:
>  In the case, they will get a French UI with the spell checker dictionary listings in Georgian.

Which is probably what they actually want since they'll be filling in forms and writing Notes in Georgian, not French.

> I'll merge this shortly if there are no objections.

Fire away. I'm waiting for it so that I don't make conflicts when I fix the deprecations you noticed.

PR #1452, opened 2023-07-10
grampslocale: Remove unused member variables

a-andre:


Nick-Hall:
@jralls Are these useful?  They don't appear to be used in core Gramps or by any addon in our _addons-source_ repository.

jralls:
@Nick-Hall They're just placeholders so that the GrampsLocale structure has the same components that libc locale does. I can't think of a good use-case for Gramps to be formatting number or money fields, so I don't think there's any need to keep them.

PR #1455, opened 2023-07-12
Fix first name format for genealogytree docgen

stevestock:
If no "call" name, first name should be inside `pref{}` per [genealogytree docs](https://ctan.math.illinois.edu/macros/latex/contrib/genealogytree/genealogytree.pdf)

Also removed spaces between `\pref{}` and `\surn{}` blocks, because Genealogytree appears to render the spaces  if present

### DIFF
--- a/gramps/gen/plug/docgen/treedoc.py
+++ b/gramps/gen/plug/docgen/treedoc.py
@@ -65,7 +65,13 @@ _DETAIL = [{'name': _("Full"), 'value': "full"},
            {'name': _("Short"), 'value': "short"}]
 
 _NAME_FORMAT = [{'name': _("Given Nickname Surname"), 'value': "1"},
-                {'name': _("Surname Given Nickname"), 'value': "2"}]
+                {'name': _("Surname Given Nickname"), 'value': "2"},
+                {'name': _("Surname, Given Nickname"), 'value': "3"},
+                {'name': _("GivenNicknameSurname"), 'value': "4"}]
+
+_PREFERRED_NAME = [{'name': _("Auto"), 'value': "1"},
+                   {'name': _("Call Name Only"), 'value': "2"},
+                   {'name': _("Whole Given Name"), 'value': "3"}]
 
 _MARRIAGE = [{'name': _("Default"), 'value': ""},
              {'name': _("Above"), 'value': "marriage above"},
@@ -132,6 +138,24 @@ class TreeOptions:
     Defines all of the controls necessary
     to configure the genealogy tree reports.
     """
+    def __init__(self):
+        #Node Options
+        self.detail = None
+        self.marriage = None
+        self.nodesize = None
+        self.levelsize = None
+        self.nodecolor = None
+
+        #Tree Options
+        self.timeflow = None
+        self.edges = None
+        self.leveldist = None
+
+        #Note
+        self.note = None
+        self.noteloc = None
+        self.notesize = None
+
     def add_menu_options(self, menu):
         """
         Add all graph related options to the menu.
@@ -149,6 +173,7 @@ class TreeOptions:
             detail.add_item(item["value"], item["name"])
         detail.set_help(_("Detail of information to be shown in a node."))
         menu.add_option(category, "detail", detail)
+        self.detail = detail
 
         name_format = EnumeratedListOption(_("Name Format"), "1")
         for item in _NAME_FORMAT:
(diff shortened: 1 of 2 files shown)


Nick-Hall:
Running the ancestor tree report with the example database, I get all first names without a call name underlined.  Is this what we want?

PR 1455

The home person has a call name of "Anderson" which is underlined.  This is correct.  However the three other Garners in my screenshot don't have call names.

emyoulation:
The Name format editor says that when the call name is used, the 1st word of the Given name will become the Call name if no one is not explicitly defined. That does NOT happen in the views... it shows the whole Given Name string, not just the1stword.

This incorrect underlining looks like another example where the Call name doesn't do the "first word" fallback in the manner described.

Nick-Hall:
The genealogy tree reports don't use name formats.

emyoulation:
But it sure looks like that's what they are trying to do. So maybe they cloned the code that doesn't work in the name formatying?

Given name (with the call portion underlined) and Surname in [small] caps. This simulates the classic way genealogy societies formatted their listing for introducing a new person in a generation-by-generation descendants list.

Nick-Hall:
No.  We can only format a name with `\pref{}`, `\nick{}` and `\surn{}`. For example:

`name = {\pref{Elizabeth} \nick{Liz} \surn{Smith}}`

See section 7.6 on page 184 of the specification.

stevestock:
gramps has:
Given, Nick, Surname, Call

genealogy tree has:
`\pref{}`, `\nick{}`, `\surn{}`

All the database examples in the [genealogytree docs](https://ctan.math.illinois.edu/macros/latex/contrib/genealogytree/genealogytree.pdf) imply that pref = given if there's no call. Only a few examples in the entire manual have name text outside of `\pre{}`, `\nick{}`, or `\surn{}` at all. 

In my personal use cases:
(1) I don't have call names. I pull my tree directly from Ancestry.com
(2) I pull out and customize name components individually, which is easier if the given name is in \pref{} like the genealogytree doc examples. Especially because [T-F-S/genealogytree](https://github.com/T-F-S/genealogytree) does not expose `\surn` or `\pref` directly.

I also realized that the spaces between name components is only an issue in my specific use cases. I'm thinking these would all be better as options. 

I will rework the PR accordingly

stevestock:
@Nick-Hall 

I reworked all my changes as options. Two new options for name format:
new format options

Three options for preferred name format:
new pref options

A few examples. 
- [given nick surn]
- [Call Name Only]
given nick surn call name only

- [surn, given nick]
- [Auto] (meaning Call name if it exists, otherwise full given name)
surn, given nick auto

- [givennicksurn]
- [Whole Given Name]
givennicksurn whole given name


Bonus:

Greyed out irrelevant options when using "graph" output
Screenshot 2023-07-23 at 12 57 44 AM

stevestock:
I made "Auto" pref = **_whole_** given name if no call name defined. Maybe I should have made it:

 "**_1st word_** of the Given name will be considered the Call name if one is not explicitly defined." per @emyoulation

emyoulation:
I dont quite understand why this menu wouldn't also offer all the Name formats defined in the Display Name Editor? You expand on the set of preferred Display Names with those underline options. But couldn't the report also leverage the list in the GUI?


There's a 2005 discussion of Callname in the mail archive and countries where underlining the Callname is a standard in official documents: 

https://sourceforge.net/p/gramps/mailman/gramps-users/thread/1131822659.6685.19.camel%40otto.ehbuehl.net/#msg8989135

https://www.gramps-project.org/wiki/index.php/Names_in_Gramps#Call_Name

https://www.gramps-project.org/wiki/index.php/Gramps_5.1_Wiki_Manual_-_Settings#Display_Name_Editor

https://www.gramps-project.org/wiki/images/7/77/EditPreferences-Display-tab-DisplayNameEditor-dialog-example-50.png

Nick-Hall:
> I dont quite understand why this menu wouldn't also offer all the Name formats defined in the Display Name Editor? 

The name formatter produces a formatted name.  In this case LaTeX does the formatting, we just have to tag some parts of the name first.

stevestock:
@Nick-Hall are you OK with the updated approach?

@emyoulation I’m happy to add to the list if you have a specific need

emyoulation:
>  @emyoulation I’m happy to add to the list if you have a specific need

I do not have a specific need. But Gramps was written with an Anglicized bias. 

The point was raised while trying to consider cultures that may not fit that model. The complexity of Russian naming conventions always leaves me dazed even rendered in English. 

And I have no clue what happens for name formatting in non-latin alphabets across the world.

gramps/gen/plug/docgen/treedoc.py
```diff
@@ -552,17 +614,16 @@ def format_given_names(self, name):
...
                 where = first.index(call)
                 return '{before}\\pref{{{call}}}{after}'.format(
                     before=escape(first[:where]),
                     call=escape(call),
                     after=escape(first[where+len(call):]))
-            else:
-                # ignore erroneous call name
-                return escape(first)
-        else:
+        elif (self.preferred_name == "1") or (self.preferred_name == "3"):
```
Nick-Hall, line 624: You can write this line as:
`elif self.preferred_name in ("1", "3"):`

Nick-Hall:
@stevestock Yes, this looks good.  Just one minor comment.

stevestock:
Thanks, updated!

Also note my genealogytree addon PR gramps-project/addons-source#563

PR #1458, opened 2023-07-12
Close standard streams on exit

Nick-Hall:
This removes a warning on exit.


QuLogic:
Curious what warning there is on exit?

Nick-Hall:
Running `python -Wall Gramps.py` the warning on exit was:

`sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name=1 mode='w' encoding='utf-8'>`

PR #1475, opened 2023-07-27
Fix bug in editors that occurred when source icons were added

Nick-Hall:
Source attributes do not have citations which means that they cannot share the same attribute model.


emyoulation:
Something to test related to changing Sources to have Citation

Does this affect what happens when using the Clipboard to add Sources.
   In the 5.1.6 version and earlier, dropping a clipboarded Source to an object editor dialog pops a NEW Edit Citation dialog with that Source.

Does a Citation being found changes that behavior to just Share the Source? If so, it would kill a quick citing method.

PR #1479, opened 2023-07-29
Allow custom undo managers in database plugins

Nick-Hall:
Add a _create_undo_manager method in DbGeneric which can be overridden to supply a custom undo manager rather than the default DbGenericUndo.


DavidMStraub:
@Nick-Hall I'd like to start looking into this as a new plugin – just to make sure I understand correctly: I would just introduce new database backends that inherit from an existing backend and add an undo database, like `sqlite+undo`, `postgresql+undo`, etc., right?

PR #1483, opened 2023-07-31
Refactor and pylint cleanups for alive.py

cdhorn:
This is a refactor and cleanup of gramps/gen/utils/alive.py that brings the pylint score from 4.92 to 9.82.  It does not phase out the ugly globals in use as I am a bit unsure why they are there and why it was setup for possible standalone use without the Gramps config manager.


cdhorn:
Rebased.

cdhorn:
Closed in lieu of #1790

PR #1484, opened 2023-07-31
Migrate code style to Black

Nick-Hall:


jralls:
This seems pointless without something in place to enforce it going forward. I don't see anything in the changeset that does that. The Black docs have an example [precommit hook](https://black.readthedocs.io/en/stable/integrations/source_version_control.html) that might do the trick, but you also need setup.py to check for and install Black from PyPi and then hope that it's never the target of a supply chain attack.

cdhorn:
Black default line length is 88 characters, not 80 characters.  Gramps `.pylintrc` currently uses 80 characters.  The reformat used the black default of 88 characters.

Should either raise the `.pylintrc` requirement to 88 like gramps-webapi did or take a second pass through with `black -l 80`

I suspect a lot of thought went into the black default so would lean toward updating the `.pylintrc`

May also want to look at the precommit hook gramps-webapi uses.  

If adopting `isort` in such a hook take note that with the current Gramps coding standard of using `# ---` blocks to denote separate Python, GTK and Gramps import sections that `isort` often chokes and makes a mess of things.  It doesn't handle comments intermixed with imports very well.

emyoulation:
Will the reformat to 88 characters crwte new strings for Weblate and play havoc with the translators workload?

jralls:
> I was thinking of adding the [Github Actions integration](https://black.readthedocs.io/en/stable/integrations/github_actions.html).

Well, that's better than nothing, but if that's how we enforce it we should also turn on [pull request before merge](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule) on the active branches so that everything gets that check before merging. 

> We could also create a [.git-blame-ignore-revs](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame) file.

+1

PR #1488, opened 2023-08-02
Fix incorrect relationship symbols

Nick-Hall:
Fix incorrect relationship symbols in relationship view.

Fixes [#12937](https://gramps-project.org/bugs/view.php?id=12937).


Emmanuel-Vallois:
This change introduced a potential bug: the “constant” SYMBOL_TRANSGENDER is defined two times, the second assignment with value 8 overwriting the previous value 3.

PR #1496, opened 2023-08-06
Add support for Hebrew prefixes

Nick-Hall:
* Double the Vav if not already double
* Remove the leading He
* Prefix a maqaf for non-Hebrew words and numbers

### DIFF
--- a/gramps/plugins/lib/libnarrate.py
+++ b/gramps/plugins/lib/libnarrate.py
@@ -73,6 +73,24 @@ def _get_empty_endnote_numbers(obj):
     return ""
 
 
+def convert_prefix(word):
+    """
+    Convert Hebrew grammar for prefixes
+    """
+    if not word or len(word) < 2:
+        return word
+    if word[0] == "ו" and word[1] != "ו":
+        # Double the Vav if not already double
+        word = "ו" + word
+    if word[0] == "ה":
+        # Remove the leading He
+        word = word[1:]
+    if word[0] < "א" or word[0] > "ת":
+        # Prefix a maqaf for non-Hebrew words and numbers
+        word = "־" + word
+    return word
+
+
 # avoid normal translation!
 # enable deferred translations
 # (these days this is done elsewhere as _T_ but it was done here first)
@@ -1781,6 +1799,10 @@ class Narrator:
                 bdate_full = bdate_obj and bdate_obj.get_day_valid()
                 bdate_mod = bdate_obj and bdate_obj.get_modifier() != Date.MOD_NONE
 
+        if self._locale.locale_code() == "he":
+            bdate = convert_prefix(bdate)
+            bplace = convert_prefix(bplace)
+
         value_map = {
             "name": self.__first_name,
             "male_name": self.__first_name,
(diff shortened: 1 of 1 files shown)


romjerome:
Does it cover some words/dates into others locales too?
I did not check on code, but in french we might have at least two specific  issues on detailed sentences (libnarrate.py).
e.g., when date object has no day (month and year), we cannot write "né le janvier 2023", should be something like "né en janvier 2023". Also, when place object starts with a number or street, we cannot write "à 123 rue...", should be something like "au 123 rue" or only the street name. etc.

emyoulation:
These items seem to be creating sentence fragments (or 'asides') rather than stilted narrative sentences that would be grammatically correct. So, in English, rules of grammar do not apply.

More properly in English, the event is 'on' a day and, when approximated,  'in'  month or year.

And while English lacks a masculine/feminine difference for 'born'. We would have to insert a pronoun to be equivalent... which would raise arguments about the "preferred pronoun".

And it would require a modified verb. In this case, that would be: 'was', the (first and third person) singular past tense of 'be'.

So if we were writing for a grammatically correct narrative in English, the examples would generate either specifically:
"He was born on January 4th of 1923." (Even more formally, the abbreviated '4th' should be 'fourth')
or approximately: 
"She was born in January of 1923."

But this example is narrative is generating an outline of facts (or 'asides') with informal sentence fragments instead.

Are sentence fragments allowable in French?

romjerome:
> Are sentence fragments allowable in French?
Oh, with the "latin" flexibility, maybe everything is allowed! You can push/put two words on a paper, and the reader may (or not) understand the sense. So, allowed maybe not by the 'academie-francaise' or the 'office québécois de la langue française', but like with recents spelling reforms, the rules are only strict, in theory. So, if the common langage is understandable, one can blame you and someone else will be just keep the sense, the idea, behind sentence fragments!

Nick-Hall:
> Does it cover some words/dates into others locales too?

It doesn't at the moment, but it could do.  It looks like for French we may need to pass in a Date or Place object in addition to the substitution text.

romjerome:
@Nick-Hall , there is also a possible issue on a options sequence under textual detailed reports.

User choice like:

> #: ../gramps/plugins/textreport/detancestralreport.py:881
> #: ../gramps/plugins/textreport/detdescendantreport.py:1070
> msgid "Use full dates instead of only the year" 

can force to use a specific sentence... 

In this case (year only), I suppose this will display something like "in %(year)s" (for english), whatever date format set or locale issue. So, this may already pass over into some strings!
e.g., "%(female_name)s died on %(death_date)s in %(death_place)s." could look like  "%(female_name)s died on %(year)s in %(death_place)s." as 'year' might overpass 'death_date' variable.

romjerome:
Looking on internet ("['In' vs. 'On' followed by a year"](https://ell.stackexchange.com/questions/11041/in-vs-on-followed-by-a-year), found such answer:

> 
> Any time you are being less specific than the day itself, use “in” to indicate that the event happened at a time within the boundaries of that century, decade, year, or month.
> 
>     In the 20th century...
> 
>     In the late '80s...
> 
>     In 1989...
> 
>     In June, 1989
> 
> When the day is known, you can say “On June 20, 1989...” or “On that day in June...”.
> 
> When the hour is known, you can say “At 3:05 pm on June 20, 1989...”.
> 
> For any of these levels, estimates or approximations would be preceded by “around” or some similar term. (“Around 1990...”, “At approximately 3 pm...”, “At 3 pm or so...”, etc.)

yarons:
It looks amazing so far, I'll test it in the UI after the approval to make sure it covers all cases, thank you so much!

PR #1502, opened 2023-08-08
Fix issue with PrivateProxyDb stripping checksum

DavidMStraub:
Debugging a serious issue in Gramps Web (images not loading for users without authorization to view private records), I found that Gramps' PrivateProxyDb actually strips the checksum when sanitizing a media object. I don't think this is a feature.

@Nick-Hall would be great if this fix would make it into 5.2.

I'm now thinking of a way to patch this in Web API in the meantime :unamused:


Nick-Hall:
Rebased.

PR #1516, opened 2023-08-15
Fix glocale problems on Windows

prculley:
Found a coouple of issues when debugging the glocal issue.  Should fix the glocal.language getting set to a list of language codes, and the Windows should now work without setting LANG=whatever

### DIFF
--- a/gramps/gen/utils/grampslocale.py
+++ b/gramps/gen/utils/grampslocale.py
@@ -731,17 +731,17 @@ class GrampsLocale:
         # is cached so we don't have to query the file system every
         # time this function is called.
         if not hasattr(self, "languages"):
-            self.language = self.get_available_translations()
+            self.languages = self.get_available_translations()
 
         if not loc:
             return None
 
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Thanks.  Unless @jralls has any objections I'll go ahead and merge this.

prculley:
I think this is still required; it prevents the following "gramps -v" subset for bad LANGUAGE environment settings even though LANG was set to "en_US.utf8".  LANGUAGE was set to "xx" and had @jralls "Fix system locale detection on MS Windows" commit installed for this test.

```
Environment settings:
---------------------
 LANG      : ar_AA.UTF-8
 LANGUAGE  : ar:bg:ca:cs:da:de:de_AT:el:en
 GRAMPSI18N: not set
 GRAMPSHOME: not set
 GRAMPSDIR : not set
 GRAMPS_RESOURCES : D:\Users\PRC\Documents\Gramps
```

jralls:
> @jralls Is this still required?

Yes

PR #1519, opened 2023-08-16
Fixes #11744, #12626, #12629, #12963 delete surname

prculley:
These bugs occur when deleting one of multiple surnames under specific movements of users cursor and order of actions.
See [~0063983](https://gramps-project.org/bugs/view.php?id=12629#c63983) for a good description of at least one way to get this bug.

What is happening is that the cell renderer thinks it is in the middle of an edit operation, when the delete for that row of cells comes long.  The row disappears, and then the cell gets a signal to end editing and store results in the row that no longer exists.
Most of the time the user will cause a focus out signal that causes the edit to end before he tries to delete the row and everything works.

To fix I store the 'path' for the row when a cell edit starts, and then, on delete, check if  the path matches the row under edit.  If so, I invalidate the stored path.  When the end edit signal comes along, I check that the path is still valid before attempting to update the model, if it is not, we skip it as the row is gone anyway.

### DIFF
--- a/gramps/gui/editors/displaytabs/surnametab.py
+++ b/gramps/gui/editors/displaytabs/surnametab.py
@@ -248,8 +248,11 @@ class SurnameTab(EmbeddedList):
         """
         Delete button is clicked. Remove from the model
         """
-        (model, node) = self.selection.get_selected()
+        (_model, node) = self.selection.get_selected()
         if node:
+            path = self.model.get_path(node).to_string()
+            if path == self.curr_path:
+                self.curr_path = None
             self.model.remove(node)
             self.update()
 
@@ -260,6 +263,7 @@ class SurnameTab(EmbeddedList):
         self.curr_col = colnr
         self.curr_cellr = cellr
         self.curr_celle = celle
+        self.curr_path = path
 
     def on_edit_start_cmb(self, cellr, celle, path, colnr):
         """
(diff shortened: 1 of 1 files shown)


lordemannd:
I retested, using steps in [0012629:0063983](https://gramps-project.org/bugs/view.php?id=12629#c63983), with 5.2 (AIO64-5.2.0-r1-50f3ad3) and it looks like issue solved.

then tested steps in [0012629:0063979](https://gramps-project.org/bugs/view.php?id=12629#c63979), and it is failing again. In [0012963:0065363](https://gramps-project.org/bugs/view.php?id=12963#c65363), [@SNoiraud](https://gramps-project.org/bugs/view_user_page.php?id=55), reported that it was fixed in 5.1.6, and I tested it to be fixed,
is it possible that the 5.1.6 fix did not get into 5.2?

prculley:
I used the 2nd sequence described in the bug report on both 5.1.6 and 5.2 with latest commits.  Both failed as described.  I believe that this PR will correct the situation for 5.2, but did not test it for 5.1.6, although the patch is simple enough to try it there as well.  Wasn't sure if we will ever have another 5.1.x release.

I suspect that in the 1st sequence described in the bug report is not quite detailed enough for me to get it right.  This issue depends very much which surname editor (there are two) and how the cursor is clicked and moved and how the 'Delete' is activated.

It took a LOT of tries to get a consistently failing set of steps, and then to debug. At one point I realized that stopping at a breakpoint in the debugger was 'fixing' the issue.  I ended up adding print statements to see the sequence of GTK signals...

Nick-Hall:
> Wasn't sure if we will ever have another 5.1.x release.

No.  Version 5.1.6 was the last.

PR #1498, opened 2023-08-08
Fix 12945: Track all tab settings separately

cdhorn:
Addresses issue 12945 by adding a mandatory config_key to all embedded list tabs and the note tab and updates editors to set keys specific to the editor so column changes persist on a per tab per editor basis. 

```
address_editor_citations=[2, 16, 17, 9, 15, 17, 7]
address_editor_notes=[2, 6]
childref_editor_citations=[2, 29, 17, 9, 15, 17, 7]
childref_editor_notes=[2, 9]
citation_editor_attributes=[2, 10]
citation_editor_notes=[2, 6]
citation_editor_references=[9, 7]
event_editor_attributes=[2, 2, 14]
event_editor_citations=[2, 21, 17, 9, 15, 17, 7]
event_editor_notes=[2, 9]
event_editor_references=[9, 15]
eventref_editor_ref_attributes=[2, 2, 8]
eventref_editor_ref_citations=[2, 8, 17, 9, 15, 17, 7]
eventref_editor_ref_notes=[2, 9]
eventref_editor_references=[6, 13]
eventref_editor_shared_attributes=[2, 2, 21]
eventref_editor_shared_citations=[2, 37, 17, 9, 15, 17, 7]
eventref_editor_shared_notes=[2, 5]
family_editor_attributes=[2, 2, 21]
family_editor_childrefs=[2, 2, 3, 5, 21, 7, 9, 9]
family_editor_citations=[2, 29, 17, 9, 15, 17, 7]
family_editor_events=[10, 20, 13, 20, 20, 2, 2, 7, 5]
family_editor_ldsord=[2, 2, 13, 13, 7, 17]
family_editor_notes=[2, 9]
ldsord_editor_citations=[2, 22, 17, 9, 15, 17, 7]
ldsord_editor_notes=[2, 16]
media_editor_attributes=[2, 2, 21]
media_editor_citations=[2, 29, 17, 9, 15, 17, 7]
media_editor_notes=[2, 9]
media_editor_references=[9, 7]
mediaref_editor_ref_attributes=[2, 2, 14]
mediaref_editor_ref_citations=[2, 16, 17, 9, 15, 17, 7]
mediaref_editor_ref_notes=[2, 13]
mediaref_editor_shared_attributes=[2, 2, 21]
mediaref_editor_shared_citations=[2, 29, 17, 9, 15, 17, 7]
mediaref_editor_shared_notes=[2, 9]
mediaref_editor_shared_references=[9, 7]
[...]

PR #1520, opened 2023-08-17
Fixes #12970, Multiple Surname tab doesn't save column settings

prculley:
Fixes [#12970](https://gramps-project.org/bugs/view.php?id=12970)  Multiple Surname tab doesn't save its column settings.

The problem appears to be that when editing a person there are actually two SurnameTab instances active at the same time. One is started by the EditPerson, and a second is started by EditName. Attempting to change the EditName surname columns works as long as the enclosing EditPerson dialog doesn't close. If you close the EditPerson, its copy of SurnameTab overrides the config settings for the EditName instance.

Fixed by providing a different name for the EditPerson version of the SurnameTab, so the config settings of the two instances don't collide.

Note, a better fix might be to save the config settings as the user adjusts them, and not when the PersistentTreeView closes.  This would mean that in this particular case both SurnameTab instances would end up with the same settings.  It would also work if a user had multiple dialogs of the same type open (like EditPerson) and made changes in one dialog only to lose them when the second closed.  But it would be a lot harder and more pervasive to code up.

### DIFF
--- a/gramps/gui/editors/displaytabs/surnametab.py
+++ b/gramps/gui/editors/displaytabs/surnametab.py
@@ -91,12 +91,14 @@ class SurnameTab(EmbeddedList):
         name,
         on_change=None,
         top_label="<b>%s</b>" % _("Multiple Surnames"),
+        surname_override = None
     ):
         self.obj = name
         self.on_change = on_change
         self.curr_col = -1
         self.curr_cellr = None
         self.curr_celle = None
+        self.surname_override = surname_override
 
         EmbeddedList.__init__(
             self,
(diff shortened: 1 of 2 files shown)


cdhorn:
I'd need to double check but I think I addressed this in PR #1498

prculley:
I should have done my research more carefully https://github.com/gramps-project/gramps/pull/1498 indeed does a more comprehensive job of fixing this.

prculley:
@Nick-Hall I've changed my dev environment to use Black, instead of the pylint.  
I'm not sure what was wrong with my commit message, I think it does adhere to the committing policy (Bug referance, less than 70 chars, description of what was fixed).

Nick-Hall:
The bug reference should be at the end of the commit message separated by a blank line, not in the title.

PR #1521, opened 2023-08-18
Make the narrative web ltr and rtl.

SNoiraud:
Fixes [#12960](https://gramps-project.org/bugs/view.php?id=12960)

It should works for all css


SNoiraud:
Need to modify all css files...
The language selection on mobile does not work

SNoiraud:
I suppress this PR after a rebase problem.
The PR https://github.com/gramps-project/gramps/pull/1551 replace this PR

PR #1524, opened 2023-08-18
Add Farsi to Microsoft Windows Locale list.

jralls:
It's been added (as Persian) to weblate. It doesn't have any translations yet, but we might as well be ready.


Nick-Hall:
I also added Galician.  Both Farsi and Galician have been requested more than once, so I added them to Weblate.

Nick-Hall:
@jralls Your commit message mentions Farsi and Galician, but I can only see Farsi.

jralls:
@Nick-Hall Sorry, I had win32locale.py in different directories open and edited the wrong one. Fixed now.

PR #1527, opened 2023-08-19
Clean up some Gdk deprecations

prculley:
In particular: Gdk.Screen.get_width() and height deprecated as of 3.22 (new code works as of 3.22)
Gdk.Keymap.get_default() deprecated as of 3.22 (new code works as of 2.2)

Should be OK since we now have Gtk 3.24 as minimum version.

### DIFF
--- a/gramps/gen/utils/image.py
+++ b/gramps/gen/utils/image.py
@@ -150,10 +150,12 @@ def image_dpi(source):
     try:
         from gi.repository import Gdk
 
-        s = Gdk.Display.get_default().get_default_screen()
+        mon = Gdk.Display.get_default().get_primary_monitor()
+        mon_geom = mon.get_geometry()
+        scale = mon.get_scale_factor() * MM_PER_INCH
         dpi = (
-            s.get_width() * MM_PER_INCH / s.get_width_mm(),
-            s.get_height() * MM_PER_INCH / s.get_height_mm(),
+            mon_geom.width * scale / mon.get_width_mm(),
+            mon_geom.height * scale / mon.get_height_mm(),
         )
     except:
         dpi = (96.0, 96.0)  # LibOO 3.6 assumes this if image contains no DPI info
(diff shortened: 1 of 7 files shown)


Nick-Hall:
@prculley Why not pass a display into `get_screen_size` rather than a screen?  You are calculating the size based on the monitors associated with the display.  The screen is not needed.

prculley:
The main reason is that I never found a way to get Display directly for Gtk.Window.  Now this may be un-necessary for our application, but we seemed to have get screen_for_window this in a lot of places.  If we can be satisfied with Gdk.Display.get_default, we can simplify this a bit more.

I welcome your thoughts, as I suspect some of this code is pretty legacy and may represent Gtk/Gdk ideas that are no longer used.

BTW, I found another place that needs attention, gen.utils.image has the following:
`        s = Gdk.Display.get_default().get_default_screen()
        dpi = (
            s.get_width() * MM_PER_INCH / s.get_width_mm(),
            s.get_height() * MM_PER_INCH / s.get_height_mm(),
        )
`
This one is even more problematic, in that even on my system the dpi is different on each monitor.

jralls:
[GdkScreen](https://docs.gtk.org/gdk3/class.Screen.html) is deprecated in gtk3 and gone in gtk4. The new model is [GdkDisplay](https://docs.gtk.org/gdk3/class.Display.html) with one or more [GdkMonitor](https://docs.gtk.org/gdk3/class.Monitor.html)s. 

You can use `gdk_display_get_monitor_at_window` or `gdk_display_get_monitor_at_point` to get the monitor, but computing the dpi of the monitor takes some work: `gdk_monitor_get_geometry` gets you the size (height and width) in "application pixels", `gdk_monitor_get_scale_factor` gets the number to multiply that by to get physical pixels (only `1` and `2` are supported, even on Windows), and `gdk_monitor_get_height_mm` or`gdk_monitor_get_width_mm` to get the physical size. But for most purposes applications don't need to do that; just query `gdk_window_get_scale_factor` or `gtk_widget_get_scale_factor` to decide whether to present the HiDPI version of an image or graphic and let the toolkit sweat the details.

PR #1529, opened 2023-08-19
Pylint cleanups for gen/db/*.py

cdhorn:


gramps/gen/db/generic.py
```diff
@@ -311,27 +325,15 @@ def __next__(self):
     def __exit__(self, *args, **kwargs):
         pass
 
-    def iter(self):
```
Nick-Hall, line 314: I don't think that we should remove the `iter`, `first` and `next` methods.  They form part of the public API and third-party code may be using them.  If we want to remove these in the future then they should be deprecated first.

Nick-Hall (changes requested):
The only problem I can see with this is the removal of a few `Cursor` methods.

I generally don't mind the abbreviations "trans" for "transaction" or "dir" for "directory", but in the database code maybe the full form is clearer.  Keep the changes  you have made to these.

There are also a couple a pylint comments.

gramps/gen/db/generic.py
cdhorn: Restored them.

cdhorn:
Removed the pylint hints as well, think this is ready.

Nick-Hall (approved):

PR #1533, opened 2023-08-21
Some filter rule type related cleanups

cdhorn:
Some pylint related cleanups for a few filter rules, and slight refactor for a number of type related rules where found the type for the comparison was being initialized in the apply method for every handle instead of in the prepare method.


cdhorn:
@Nick-Hall rebased and checked that does not overlap improvements in #1721

Nick-Hall:
Thanks.  I should have time to merge this tomorrow.

Nick-Hall (approved):

PR #1538, opened 2023-08-22
Revert "Add fanchart option with overhang"

Nick-Hall:
The title problem reported in bug report #12926 is not easy to fix.  Putting the title at the bottom of the page causes problems because we don't have a accurate way to determine font height, and the centre_text method vertically aligns text to the top.

See PR #1133 and bug report [#12926](https://gramps-project.org/bugs/view.php?id=12926).

### DIFF
--- a/gramps/plugins/drawreport/fanchart.py
+++ b/gramps/plugins/drawreport/fanchart.py
@@ -30,7 +30,7 @@
 # python modules
 #
 # ------------------------------------------------------------------------
-from math import pi, cos, sin, log10, acos, asin, degrees, radians
+from math import pi, cos, sin, log10, acos
 
 
 def log2(val):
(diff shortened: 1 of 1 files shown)


emyoulation:
All the other "type of Graph" options (full circle, half circle, quarter circle) put the title as a header. This is the only format that tries to set it as an inset caption.

It should be a header like the others.

A header somewhat defeats the purpose of better paper use. But the Graph is more appealing than the other {single, double, full} quadrant-based renderings. So the original designer can come back and add a caption option that actually works in a future enhancement. A 2-line word-wrap caption in a rectangular box that falls entirely inside the blank area left by the wings would work.

Actually, a future enhancement would be useful to be able to have an checkbox option to turn off the titlebox anyway. Particularly since there is no option to change the content of the titlebox. People can count and the center person text identifies the focus. So that the title text is redundant.

Nick-Hall:
Replaced by PR #1541.

PR #1550, opened 2023-08-26
verify tool: convert the ListStore to a TreeStore

OlliL:
This change converts the ListStore to a TreeStore in the [Verify Data Tool](https://gramps-project.org/wiki/index.php/Gramps_5.1_Wiki_Manual_-_Tools#Verify_the_Data).
- select/unselect the root node populates it automatically to its child nodes
- unselecting a child not automatically unselects the root node
- selecting all child nodes automatically selects the root node as well
- mark, unmark, invert buttons work
- hide selected rows work
- double click on a root node expands or collaps it
- the shown root node message gets synced with the correct number of (visible) child nodes when "Hide marked" is activated as well as while its active and a selection is made
- when switching between "Hide marked" and "Show all"
  - all expanded (and visible) root nodes stay expanded 
  - the sorting gets reset - but that was always the case and I didn't found an easy way to fix it
- additionally I fixed a small amount of easy pylint issues (Try and leave this world a little better than you found it :smile:)
- CLI mode does still work too
- no idea why always the Gramps icon instead of the person or family icon is shown - but this was the case in the original version of verify.py too.

Please review & provide some feedback/guidance :face_in_clouds: 

Peek 2023-08-27 01-18

### DIFF
--- a/gramps/plugins/tool/verify.py
+++ b/gramps/plugins/tool/verify.py
@@ -152,8 +152,7 @@ def get_date_from_event_handle(db, event_handle, estimate=False):
         if not estimate and (date_obj.get_day() == 0 or date_obj.get_month() == 0):
             return 0
         return date_obj.get_sort_value()
-    else:
-        return 0
+    return 0
 
 
 def get_date_from_event_type(db, person, event_type, estimate=False):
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Rebased onto gramps52 branch.

PR #1551, opened 2023-08-27
Make the narrative web ltr and rtl.

SNoiraud:
Fixes [#12960](https://gramps-project.org/bugs/view.php?id=12960)

It should works for all css

This PR replace PR https://github.com/gramps-project/gramps/pull/1521

### DIFF
--- a/gramps/plugins/webreport/basepage.py
+++ b/gramps/plugins/webreport/basepage.py
@@ -222,6 +222,7 @@ class BasePage:
             self.rlocale = report.set_locale(the_lang)
         else:
             self.rlocale = report.set_locale(report.options["trans"])
+        self.dir = "rtl" if self.rlocale.rtl_locale else "ltr"
         self._ = self.rlocale.translation.sgettext
         self.colon = self._(":")  # Translators: needed for French, else ignore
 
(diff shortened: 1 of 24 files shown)


Nick-Hall:
@SNoiraud You now need to run [black](https://black.readthedocs.io/en/stable/getting_started.html) on your changes.

`> black gramps/plugins/webreport`

SNoiraud:
@Nick-Hall : I think black does not work correctly:
Before the last patch:
$ /usr/bin/black gramps/plugins/webreport
reformatted gramps/plugins/webreport/media.py
reformatted gramps/plugins/webreport/source.py
reformatted gramps/plugins/webreport/basepage.py
All done! ✨ 🍰 ✨
3 files reformatted, 27 files left unchanged.
$
and after that this is always incorrect.

Nick-Hall:
Our workflow is running on ubuntu-latest which is 22.04 currently.  However it loads black version 23.7.0 which is up-to-date.

You might get it to work using the `--preview` option.  This lets you format with the changes in the next major release.

Alternatively you could obtain the latest version with `pip install black`.

gramps/plugins/webreport/basepage.py
```diff
@@ -222,6 +222,9 @@ def __init__(self, report, the_lang, the_title, gid=None):
             self.rlocale = report.set_locale(the_lang)
         else:
             self.rlocale = report.set_locale(report.options["trans"])
+        self.dir = "ltr"
```
Nick-Hall, line 225: A list of [RTL locales](https://github.com/gramps-project/gramps/blob/master/gramps/gen/utils/grampslocale.py#L85) is maintained in the GrampsLocale class.

We could access this list with code like:

`self.dir = "rtl" if self.rlocale.rtl_locale else "ltr"`

SNoiraud: Thanks for that information. Applied.

PR #1564, opened 2023-09-01
Fix auto install of python modules

Nick-Hall:


gramps/gui/plug/_windows.py
```diff
@@ -283,12 +284,8 @@ def __on_install_clicked(self, button, addon):
         """
         # Install required modules
         for package in self.req.install(addon):
-            try:
-                subprocess.check_output(
-                    [sys.executable, "-m", "pip", "install", package],
-                    stderr=subprocess.STDOUT,
-                )
-            except subprocess.CalledProcessError as err:
+            result = pip.main(["install", "--target", LIB_PATH, package])
```
prculley, line 287: https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program
They have even removed "pip.main" in favor of "pip._internal.main" in attempt to discourage this.

Nick-Hall: Does this actually fail on Windows?  It works for me on Arch Linux.

prculley: I have not really tested, but surfing around discovered that the pip devs made the change back in 2018 at version 10.x.

From their NEWS file:`Move all of pip's APIs into the pip._internal package, properly reflecting the fact that pip does not currently have any public APIs. (#4696, #4700)`

Judging by some other comments, they may have moved the entry again more recently.  I don't know what the distro might be using (or patching).

prculley: We may want to try out https://github.com/di/pip-api

prculley: never mind about pip-api, it ends up calling python -m pip.py

Nick-Hall: I don't like the look of [this](https://github.com/di/pip-api/blob/master/pip_api/_call.py#L10) line.

Nick-Hall: I had a quick look at the pip source code to see if we could use some of the download code.  It is quite sophisticated, which is why I wanted to use it if possible.

I've got nothing against using the code in the Life Lines Chart addon.  In fact, I really wanted to make it easier for someone else to use it , rather than everyone doing a cut and paste.  It seemed like running it in core with a list of modules in the gpr was the way to go.

I was just a little bit concerned that the code in the addon will might not work in certain circumstances.  Does it support all file formats?  It also downloads a fixed version rather than the latest.

prculley: I have some new thoughts I want to look into, so if you would like to defer this a couple of days, I might be able to help come up with something that at least partly works.

Nick-Hall: OK.  I'll leave this PR closed, but if you come up with something then we can definitely include it.  An automatic download facility was part of the specification.

PR #1565, opened 2023-09-01
Fix positioning of gramplet bar popup

Nick-Hall:
Fixes [#13008](https://gramps-project.org/bugs/view.php?id=13008).


emyoulation:
That worked very nicely.  Handles the right edge of the screen well even when the right edge of the left screen in a dual screen setup.  If there is room to the right, the popup goes to the right, otherwise it goes to the left.  Nice.

Nick-Hall:
@emyoulation I've also replaced some deprecated popup method calls, which gets rid of some of the warnings you were seeing in the console.

PR #1444, opened 2023-06-10
Sources for windows AIO

jmichault:
Currently, the gramps repository lacks sources to create the Windows AIO installer.

This pull request adds:
* all necessary files in the aio directory, including a script that automates all steps.
* a workflow to verify that AIO can be built.

PR #1569, opened 2023-09-04
Make the Addon Manager available without a tree loaded

Nick-Hall:


emyoulation:
"availability of Addon Manager without tree loaded" was requested in feedback for PR144
https://github.com/gramps-project/gramps/pull/1444#issuecomment-1702841243

PR #1574, opened 2023-09-06
Verification Tool - 8 more rules + performance enhancements

OlliL:
- Add 8 additional rules:
   - test if the person has events in wrong order
   - test if the family has events in wrong order
   - test if the person has events with role Unknown
   - test if the family has events with role Unknown
   - test if the persons families are ordered incorrectly
   - test if children are ordered incorrectly within a family
   - test if a person's burial date is too late
   - test if a person's baptism date is too late considering family tradition
- Enhance performance by optimizing the caching-logic and change the order how people and families are process to maximize cache hits.

### DIFF
--- a/gramps/plugins/tool/verify.py
+++ b/gramps/plugins/tool/verify.py
@@ -43,6 +43,7 @@ options are defined (and read in) is not done the way it would be now.
 
 import os
 import pickle
+import statistics
 from hashlib import md5
 
 # ------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Fixed the remaining garbage collection issues.

The `get_message` method of four family rules was being set to `father_message` or `mother_message`.  This was preventing garbage collection when the rules failed.  I have set `self.get_message = None` to avoid this.

When fixing this, I noticed that the rules were being recreated for each person/family.  This appears inefficient and may be worth investigating in the future.

There was a similar garbage collection issue with the `UpdateCallback` class used to update the progress bar.  I set the `update` method to `None` after the rules are run to fix this.

PR #1587, opened 2023-09-10
Improve AIO build script to run without modifications

hgohel:
When used locally, the build script required modification before first use to set the full path to the local aio directory. With this change, by invoking the script from the aio directory, that local modification is no longer necessary. Internally the script uses pushd/popd to return to the aio directory.

Second change is to use the `-N` option in wget to avoid downloading a file if the timestamp hasn't changed on the server. This should reduce build time on second and subsequent builds if the source package hasn't changed. It also resolves a lurking error where an updated file on the server would not be used because wget appends a counter to the file names whereas the script was always extracting based on the original filename.

Third change is to avoid the script from asking for interactive input on second and subsequent runs when attempting to unzip dict.zip by passing the `-o` to unzip to allow overwrites.

### DIFF
--- a/aio/build.sh
+++ b/aio/build.sh
@@ -1,10 +1,14 @@
+#!/usr/bin/env bash
+#
+# Assumption: script is executed from the 'aio' directory
+#
 # install prerequisites
 ## prerequisites in msys packages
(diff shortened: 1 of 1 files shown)


hgohel:
@prculley @Nick-Hall I see a GitHub Actions workflow for windows-aio. I believe this change won't break the workflow, but please let me know if it does.

hgohel:
@ennoborg This PR should address your comment [here](https://github.com/gramps-project/gramps/pull/1444#issuecomment-1684328466). Local builds can be run no matter where the source directory is located, as long as the script is started from the aio directory.
@jralls posted a good suggestions in his [comment](https://github.com/gramps-project/gramps/pull/1444#issuecomment-1684352595) as well, which haven't been implemented in this PR.

Nick-Hall (approved):
This still runs as a GitHub action without error.

PR #1589, opened 2023-09-11
Add a new name displayer to provide brief given names

gburto01:
Family trees and charts become rather wide when displaying families with lots of children which each have multiple given names. Gramps already has a name displayer which initials all given names and this can be used to make charts more compact, but initialing all given names makes it more difficult to identify each person.
This PR adds a new name displayer for given names which shows the first given name in full, which aids identification, and for brevity any subsequent given names are shown initialed. For example, "Mary Ann Elizabeth Smith" would have her second and third given names initialed and her name would appear as "Mary A. E. Smith".

### DIFF
--- a/gramps/gen/display/name.py
+++ b/gramps/gen/display/name.py
@@ -30,6 +30,7 @@ Specific symbols for parts of a name are defined:
     ======  ===============================================================
     't'     title
     'f'     given (first names)
+    'b'     brief given name (first name plus initials)
     'l'     full surname (lastname)
     'c'     callname
     'x'     nick name, call, or otherwise first first name (common name)
@@ -628,6 +629,7 @@ class NameDisplay:
         Specific symbols for parts of a name are defined (keywords given):
         't' : title = title
         'f' : given = given (first names)
+        'b' : brief = given name (first name plus initials)
         'l' : surname = full surname (lastname)
         'c' : call = callname
         'x' : common = nick name, call, otherwise first first name (common name)
@@ -657,6 +659,11 @@ class NameDisplay:
         d = {
             "t": ("raw_data[_TITLE]", "title", _("title", "Person")),
             "f": ("raw_data[_FIRSTNAME]", "given", _("given")),
+            "b": ("raw_data[_FIRSTNAME].split(' ')[0] + ' ' + " +
+                "''.join([word[0] +'.' for word in ('. ' +" +
+                " raw_data[_FIRSTNAME]).split()][2:])",
+                "brief",
+                _("brief")),
             "l": (
                 "_raw_full_surname(raw_data[_SURNAME_LIST])",
                 "surname",
@@ -760,6 +767,7 @@ class NameDisplay:
         Specific symbols for parts of a name are defined (keywords given):
         't' : title = title
         'f' : given = given (first names)
+        'b' : brief = given name (first name plus initials)
         'l' : surname = full surname (lastname)
         'c' : call = callname
         'x' : common = nick name, call, or otherwise first first name (common name)
@@ -788,6 +796,10 @@ class NameDisplay:
         d = {
             "t": ("title", "title", _("title", "Person")),
             "f": ("first", "given", _("given")),
+            "b": ("first.split(' ')[0] + ' ' + ''.join([word[0] +'.' " +
+                "for word in ('. ' + first).split()][2:])",
+                "brief",
+                _("brief")),
             "l": ("_raw_full_surname(raw_surname_list)", "surname", _("surname")),
             "s": ("suffix", "suffix", _("suffix")),
             "c": ("call", "call", _("call", "Name")),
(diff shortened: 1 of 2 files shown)


pezball:
Does this take into account the "Call name" field?

If "Mary Ann Elizabeth Smith" is legally known by the call name "Elizabeth", as she could be in many countries, it would be preferable to show "M. A. Elizabeth Smith" instead.  I would even be slightly confused about who my own parents were if only their first names were shown in full.   The first name should only be the last choice, where there is no name match with the call field.

gburto01:
I have never used or even needed call name before so I didn't consider it. I will see what I can do to include it.

emyoulation:
So this would be a new Display Name Editor modifier to the "Initials"? Like the "[pre]" "[pre]" "[sur]" "[Sur]" "[con]" "[Con]" (prefix, surname, connector" keyword modifiers for "Primary"?

The "Call", "Common" and "Given" keywords would be the obvious modifier choices. 
image

Note that the person name in the Display Name Editor was added to the 5.2 example.gramps file. This was so that users could use him in a View or Report as a test run. It let's them verify the feature was using the Display Name format as expected.

emyoulation:
Note that Call Names has a reported bug that might affect impact this implementation.

https://gramps-project.org/bugs/view.php?id=8966

Gramps will not validate double-barreled Call Names that use a space instead of a hyphen.  (e.g., for Given Name "Mary Rose Marie", Gramps will not validate a "Rose Marie" Call Name.)

gburto01:
I have reworked the PR to use the call name if one is present. A new keyword is provided called "Brief".

image

With a name displayer using this new keyword Mary Ann Elizabeth Smith is displayed as Mary A. E. Smith.

image

If Elizabeth is selected as the call name, the name is displayed as M. A. Elizabeth Smith.

image

emyoulation:
To be consistent with other Display Name (3-letter) modifiers, shouldn't this be invoked as : 
 &nbsp; &nbsp; `Initials[Cal] Surname`

Does the new feature support the various capitalization options?    
where "initials[CAL] surname"  would return  "m. a. ELIZABETH smith"

gburto01:
OK, implementing this as modifiers on the existing Initials keyword is another approach rather than a defining a new keyword. I will take another look at it tomorrow:

Initials[cal] - First letters of given name except call name
Initials[fir] - First letters of given name except the first given name
Initials[com] - First letters of given name except the first given name or call name, if provided

For me the "fir" modifier is the one I would use because I have no use for call names. The other two modifiers could be useful for abbreviating given names where call names are used extensively or partially. The Initials keyword operates on given names so I shall not include nicknames because they are separate from given names.

emyoulation:
You probably don't need to do [fir] because the "Call" display option fallsback to the First when the field is undefined (null).

I have no idea what anyone would expect for the Common as a modifier for 'initials' if a Nickname value exists... of if both a Callname and Nickname values exist.

gburto01:
Rather than three modifiers, how about a single modifier to Initials called [Exc] which provides an exception for one name not to be initialed? [Exc] would implement the same logic I have already coded. i.e. don't initial the first given name, or call name if one is provided.

emyoulation:
A single modifier is exactly right. But instead of adding a new term (Exclude), use the same term for the same functionality elsewhere in this feature. 

Use the term 'Call' again. It means exactly what you stated: "the first given name, or call name if one is provided." This makes documentation (and user learning) less convoluted.

If using Call instead of a 3-letter form like the other modifiers lets you leverage existing code instead of having redundant code doing the same string operations to determine the which word remains unabbreviated and which Capitalization rules to apply, then do that. 

But it would avoid future maintenance woes if the same functionality isn't coded twice at different line numbers.

emyoulation:
Let's take a step back and review Initials as a whole concept and where it might go in the future.


To start, a new exclusionary term might be reasonable. But it could be variant:
`initials[-call]`
Which would would abbreviate to initial(s) of all words of the Givenname EXCEPT for leaving the Callname fully unabbreviated.

<hr />

Arguably, the original initials has a default modifier of:

`initials[given] `

Assume initials creation might eventually expand to allow:
* `initials[call]` would give JUST the initial(s) of the Callname
* `initials[common]` would give JUST the initial(s) of the Commonname
* `initials[nickname]` would give JUST the initial(s) of the nickname
* `initials[surname]` would give JUST the initial(s) of the surname
* `initials[primary[sur]]` would give JUST the initial(s) of the primary surname without prefix or connector.


Thus for my father, (Robert "Bob" Daniel McCullough, Sr.) I could almost replicate his how he inscribed his belongings at work with the lowercase monogram 'rdm' with:
initials initials[primary] 
(which generates "r. d. m.")

Or where he had to fill-out a signin sheet: "B. McCullough"  would be:
Initials[nickname] Primary[sur]


Of course by this logic, the rawsurnames "raw" should have been used in the [sur] modifier since it does the same thing. Or may in the description as "rawSURnames" instead of "surname only"

Another interesting edgecase is a surname like "van der Waals" which is abbreviated in this way for the unit of measure: "van der Waals (vdW) forces and interactions" postulate by Dutch physicist [Johannes Diderik van der Waals](https://www.britannica.com/biography/Johannes-Diederik-van-der-Waals)

PR #1597, opened 2023-09-19
Use imagesize rather than magic

Nick-Hall:
Using python-imagesize rather than python-magic would seem to be a better choice when finding the size of an image.

### DIFF
--- a/gramps/gen/utils/image.py
+++ b/gramps/gen/utils/image.py
@@ -182,31 +182,12 @@ def image_size(source):
     from gi.repository import GLib
 
     try:
-        import re
-        import magic
+        # For performance reasons, we'll try to get image size from imagesize.
+        import imagesize
 
-        # For performance reasons, we'll try to get image size from magic.
-        # This avoid to load the image in memory. This is a real improvement
-        # when we have many big images.
-        # Used in odfdoc, rtfdoc and webreport and tested with png, gif, jpeg,
-        # bmp, tiff
-        #
-        #            file size     with magic  without (Gdk)   ratio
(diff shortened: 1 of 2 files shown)


gramps/gen/utils/image.py
```diff
@@ -182,31 +182,12 @@ def image_size(source):
...
-            return (int(width[0]), int(height[0]))
-        found = img.find("precision")
-        if found > 0:
-            img = img[found:]
-        size = re.search("(\d+)\s*x\s*(\d+)", img).groups()
-        return (int(size[0]), int(size[1]))
+        # For performance reasons, we'll try to get image size from imagesize.
+        import imagesize
+
+        return iamgesize.get(source)
```
QuLogic, line 188:
```suggestion
        return imagesize.get(source)
```

Nick-Hall: Thank you.

dsblank (commented):
Should "imagesize" be added to requirements (it is a pure-python library with no dependencies). It can then be assumed to be installed.

Otherwise, there is no indication that this module can be used.

Nick-Hall:
I've moved the _python-imagesize_ package from the optional to the strongly recommended section for now.

We can make it a required package if it doesn't cause a problem for our distribution maintainers.

emyoulation:
We need a few broad-appeal items that exercise the new 5.2 pre-requisites management features of the Addon Manager.

This seems like most users will want  this for the performance boost. So it should generate a spectrum of feedback if it doesn't work or is clumsy.

PR #1598, opened 2023-09-19
Fix sizing of shared information in the reference editors

Nick-Hall:
Sets the packing of the expander widget to depend on whether or not it is expanded.

Fixes [#13030](https://gramps-project.org/bugs/view.php?id=13030).


emyoulation:
Patching the 4 files resolves the resizing issue on my Fedora box. Thanks!

PR #1600, opened 2023-09-20
Fix GEDCOM import for SUBN record containing note

prculley:
Fixes [#13024](https://gramps-project.org/bugs/view.php?id=13024), [#12152](https://gramps-project.org/bugs/view.php?id=12152)

A note attached to a SUBN (submission) record was causing a crash if the use default source preference was not set.

While testing I also discovered that when "use default source" preference WAS set, the note never was attached to the source, but ended up as an unconnected object.

### DIFF
--- a/data/tests/imp_sample.gramps
+++ b/data/tests/imp_sample.gramps
@@ -3,647 +3,647 @@
 "http://gramps-project.org/xml/1.7.1/grampsxml.dtd">
 <database xmlns="http://gramps-project.org/xml/1.7.1/">
   <header>
-    <created date="1999-12-25" version="5.1.3"/>
+    <created date="1999-12-25" version="5.2.0-rc1"/>
     <researcher>
       <resname>Alex Roitman,,,</resname>
       <resaddr>Not Provided</resaddr>
     </researcher>
   </header>
   <events>
-    <event handle="_0000000600000006" change="1" id="E0000">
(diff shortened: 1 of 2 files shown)


emyoulation:
The GEDCOM Assessment testing results grid by John Cardinal have opportunities for improvement due to new features in Gramps 5.2 

New in 5.2 is support for open-ended date spans.  10 samples were identified as "Imported incorrectly" in the [Date Valid](https://www.gedcomassessment.com/en/assessment-gramps-5.htm#area-date-valid.htm) section. Can they be added? It would drop the count of "Imported incorrectly" from 18 to 8.
https://github.com/gramps-project/gramps/pull/1124

The [Event Shared](https://www.gedcomassessment.com/en/assessment-gramps-5.htm#area-event-shared.htm) section indicates that the **non-standard** (but common) Role and Association data is not supported by the Gramps GEDCOM import.  Since Roles and Associations have existed for a while in Gramps, can they be handled by the import and export?


References:
Cardinal, John. "GEDCOM Assessment v1.05" [evaluating GEDCOM v5.5.1 import using Gramps AIO64-5.1.2-1](https://www.gedcomassessment.com/en/assessment-gramps-5.htm) with the " [assess.ged](https://www.gedcomassessment.com/en/index.htm) v1.05" test file, 7 Mar 2020 (GEDCOM Assessment [change log](https://www.gedcomassessment.com/en/changelog.htm)

PR #1601, opened 2023-09-21
Change Citation view Filter gramplet defaults to find all

emyoulation:
The original default for Min.Conf. was "Normal" 
This filters out "Low" and "Very Low" confidence Citations

Change defaults to CONF_VERY_LOW


emyoulation:
PR picked up 148 changed files instead of 1 changed file

PR #1604, opened 2023-09-25
Add event reference citations to the complete individual report

Nick-Hall:
Fixes [#13035](https://gramps-project.org/bugs/view.php?id=13035).


DaveSch-gramps:
This works for the CIR.

Will other reports have the same issue?

PR #1606, opened 2023-10-01
Add URL to citation data form Ancestry ged

hogenf:
“Other sources” with url attached in Ancestry.com generate a invalid ged. Solution to handle "Tag WWW is not allowed under DATA" and save it as NoteType.SOURCE_TEXT.


hogenf:
I have no idea way Lint errors.

Nick-Hall:
We use [black](https://black.readthedocs.io/en/stable/index.html) as a code formatter.

Run:

`black gramps/plugins/lib/libgedcom.py`

PR #1608, opened 2023-10-02
New Citations not included on the Narrative Web - Re: Add citations to event references

SNoiraud:
Fixes [#13046](https://gramps-project.org/bugs/view.php?id=13046).

This PR contains a fix for an old CacheProxyDb bug
The event ref attributes have no citations. I added them.


DaveSch-gramps:
Worked great

PR #1609, opened 2023-10-08
Allow user format option

DaveSch-gramps:
Allows all place names to reflect the format created by the user. The fmt=0 was forcing all place names to utilize the default "Full" option.

Bug Report [#13058](https://gramps-project.org/bugs/view.php?id=13058).

### DIFF
--- a/gramps/plugins/webreport/basepage.py
+++ b/gramps/plugins/webreport/basepage.py
@@ -1085,7 +1085,7 @@ class BasePage:
 
         place_hyper = None
         if place:
-            place_name = _pd.display(self.r_db, place, evt.get_date_object(), fmt=0)
+            place_name = _pd.display(self.r_db, place, evt.get_date_object())
             place_hyper = self.place_link(place_handle, place_name, uplink=uplink)
 
         evt_desc = evt.get_description()
@@ -2307,7 +2307,7 @@ class BasePage:
                 _linkurl = self.report.build_url_fname_html(_obj.handle, "evt", True)
             elif classname == "Place":
                 _obj = self.r_db.get_place_from_handle(newhandle)
-                _name = _pd.display(self.r_db, _obj, fmt=0)
+                _name = _pd.display(self.r_db, _obj)
                 if not _name:
                     _name = self._("Unknown")
                 _linkurl = self.report.build_url_fname_html(newhandle, "plc", True)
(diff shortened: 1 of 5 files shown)


SNoiraud:
This was done to resolve the 12821 bug request.
So your proposal broke that solution.
I think we need to add an option for what you want to do

Nick-Hall:
See commit [35f20a8](https://github.com/gramps-project/gramps/commit/35f20a8893599710f0cfb49685249759cdada6db).

emyoulation:
It would be very nice to have the Researcher and Owner enhanced to use the Place Tree and Person Address.  Then have Places relative to the "Home Place" for one of those.  

e.g., if the Home Place of the Researcher was something broad like "USA, North America, Terra, Sol system, Milky Way galaxy" then "Paris, Île-de-France, France, Europe, Terra, Sol system, Milky Way galaxy" would be cropped to "Paris, Île-de-France, France, Europe". And "Houston, Harris county, Texas, USA, North America, Terra, Sol system, Milky Way galaxy" would show "Houston, Harris county, Texas"

DaveSch-gramps:
In reading [0012821](https://gramps-project.org/bugs/view.php?id=12821) the issue I took away from it was how a place record is displayed in the narrative web's Place listing. Not how it is displayed within a person/family's event listing.

And that is a function of building the place hierarchy not the fine tuning of how the end results are displayed using the Place Formater.

In the database, places attached to a date show the name as it existed at that time. But in the Place view, the place displays at it does today. Do the same when building the Narrative Web. Have the Place list and the individual Place record display their main place name as it exists on the day the narrweb was built. I can see these place list/records reverting to the default "Full" formatting.

How a user chooses to fine tune the format in the final displayed place name will only become more important if/when GEP045 is implemented.

Nick-Hall:
> It would be very nice to have the Researcher and Owner enhanced to use the Place Tree and Person Address.

In Gedcom 7.0 there is now a top-level submitter record.  This is used for researcher/owner information as well as for 
recording who submitted particular records.

DaveSch-gramps:
The new commit solves the place name attached to events.

> Have the Place list and the individual Place record display their main place name as it exists on the day the narrweb was built. 

I really think this is the ultimate way to solve which place name is displayed on the Place list and in the individual Place records. Unfortunately it is beyond my hacking skill set. It was confusing enough determining which of the place lines was doing what. Some I could not determine what they did.

Nick-Hall:
I'll close this because it breaks a previous fix.

PR #1613, opened 2023-10-29
Fix context menu in all family events quick view screen

uwagura:
Fixes [#12759](https://gramps-project.org/bugs/view.php?id=12759).

Replaces the string "family" with the actually family object, allowing Quick view menu to display functional "see the family details" option instead of previous non-functional "see see details details" option.  This also displays the name of the couple instead of just the word "family", while also allowing drag and drop to the clipboard.

### DIFF
--- a/gramps/plugins/quickview/all_events.py
+++ b/gramps/plugins/quickview/all_events.py
@@ -81,7 +81,7 @@ def run_fam(database, document, family):
     stab = QuickTable(sdb)
 
     # get the family events
-    event_list = [(_("Family"), x) for x in sdb.events(family)]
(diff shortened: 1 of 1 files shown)


emyoulation:
Nice clean fix. Thanks!

But it should point at:
[0012758](https://gramps-project.org/bugs/view.php?id=12758): [Descendant Report] birth date display needs to be select optional

(0012497: Development of a distinctive Gramps icon theme stencil)

uwagura:
My mistake, I've now pointed this pull request to the correct bug report. I think this pull request only resolves the quick view issue, and not the Descendant report one, so it should be alright to point this PR to 12759 alone and not 12758

PR #1614, opened 2023-11-02
more detection of resource path

jmichault:
automatically detects the resource path in two additional cases:
* distribution of sources without git
* egg distribution

### DIFF
--- a/gramps/gen/utils/resourcepath.py
+++ b/gramps/gen/utils/resourcepath.py
@@ -74,7 +74,7 @@ class ResourcePath:
         package_path = os.path.abspath(
             os.path.join(os.path.dirname(__file__), "..", "..", "..")
         )
-        installed = not os.path.exists(os.path.join(package_path, ".git"))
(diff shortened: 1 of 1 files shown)


Nick-Hall:
We distribute wheels rather than eggs.  Haven't eggs been deprecated?

QuLogic:
They are very much deprecated; [`pip` has plans to remove their installation](https://github.com/pypa/pip/issues/12330), and [PyPI doesn't allow their upload](https://blog.pypi.org/posts/2023-06-26-deprecate-egg-uploads/).

jralls:
> We distribute wheels rather than eggs.

But if you're old-fashioned and say `python3 setup.py install` you get an egg.

QuLogic:
That is [also deprecated](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/), as you should be using a build frontend (`build` or `pip`).

jralls:
There's a difference between deprecated and forbidden. It's easy to make eggs so Gramps should run from them.

PR #1615, opened 2023-11-02
Allow user to round or not the year when display age precision is "Years"

SNoiraud:
Fixes [#12422](https://gramps-project.org/bugs/view.php?id=12422).


DaveSch-gramps:
This works and a good compromise.

Closed my PR https://github.com/gramps-project/gramps/pull/1301

PR #1616, opened 2023-11-02
Changes to allow building the Debian package.

jmichault:
This pull request allow to build the Debian package for 5.2.0-rc1.
The package can be built with these commands:
```
python3 setup.py sdist
cd dist
tar zxf gramps-5.2.0rc1.tar.gz
mv gramps-5.2.0rc1.tar.gz gramps_5.2.0.orig.tar.gz
cd gramps-5.2.0rc1
export LANG=en_US.utf8
debuild
```

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+gramps (5.2.0-rc1) unstable; urgency=medium
+
+  * New release
+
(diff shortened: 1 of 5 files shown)


Nick-Hall:
We need to get our Debian maintainer, @RossGammon, to comment on this.

PR #1617, opened 2023-11-11
Update openstreetmap.py

DaveSch-gramps:
OpenStreetMap's Nomiatim does not like abbreviations in place like "U.S.A." or "p.E.I."; Prince Edward Island.

The added line strips periods "." from the place name before sending it out to the URL

Bug report https://gramps-project.org/bugs/view.php?id=13085

### DIFF
--- a/gramps/plugins/mapservices/openstreetmap.py
+++ b/gramps/plugins/mapservices/openstreetmap.py
@@ -62,6 +62,7 @@ class OpensStreetMapService(MapService):
             return
 
         titledescr = place_displayer.display(self.database, place)
+        titledescr = titledescr.replace(".", "")
         self.url = "http://nominatim.openstreetmap.org/" "search?q=%s" % "+".join(
             titledescr.split()
         )


DaveSch-gramps:
@hgohel 

I just tried a search "Segovia, Castile and León, España, E.U." and no results were returned. and again with "EU" with no results. Removing any reference to the E.U. returned results. And yes, wherever possible, I use the endonymn for place names even though I am in the U.S..

https://nominatim.openstreetmap.org/ui/search.html?q=Segovia%2C+Castile+and+Le%C3%B3n%2C+Espa%C3%B1a%2C+E.U.

Could you provide specific searches and their results.

It may turn out that a universal fix is not be possible from within Gramps. I'll raise an issue report in the Nominatim github.

hgohel:
Hi @DaveSch-gramps In my response I didn't make it clear that "E.U." was an abbreviation for Estados Unidos, United States in Spanish, rather than European Union. So any US address with a country name "E.U." is found but not found with "EU". As an example, "empire state building, new york, ny, e.u."
I tend to agree with you that Gramps may not be the place to try and fix this without understanding of the impact; it may be better to learn more about Nominatim search. Thanks.

DaveSch-gramps:
@hgohel My fault. Rereading (more carefully) your initial post does say E.U. for the United States.

Looking at the Nominatim page for the United States, there is a long list of all the variations of the name by language including official_name, short_name, old_name as well as both 2 and 3 letter language codes. According to the list, none of them are E.U.

https://nominatim.openstreetmap.org/ui/details.html?osmtype=R&osmid=148838&class=boundary

EE.UU. (short_name:es)
É.-U. (short_name:fr) 

Maybe it thinks your country entry is French.

As a test, I just changed my U.S.A. to "Amerikan yhdysvallat", Finnish for the United States. "Boston, Suffolk, Massachusetts, Amerikan yhdysvallat" is found! where "Boston, Suffolk, Massachusetts, U.S.A." is not! The key appears to be this list.

hgohel:
Hmm. Right, no "E.U." in the list, but "E.-U." with the dash/hyphen is odd. Anyway, the list you shared shows that (1) periods in names are valid, and removing isn't always the right thing to do, and (2) U.S.A. (with periods) isn't in the list so not receiving results is "expected." Should it be up to Gramps to always remove periods? My vote would be no. Maybe Gramps online help can point users to a wiki to to get better results (if there is one)?

DaveSch-gramps:
I can change the PR to only change U.S.A. to USA without affecting any other periods users may have  But does make the PR specific for my use.

The alternative is to sign up for an account to OpenStreetMaps and find out how to add U.S.A. as a name possibility.

Nick-Hall:
The punctuation is stripped out of the search anyway.  The problem is that "U.S.A." becomes "U S A" in the query which is not listed.

In general, I think that stripping out the periods would probably be beneficial.

DaveSch-gramps:
I stripped out my patch and sent a query and the punctuation was not stripped out. My query: Basin, Big Horn Co, Wyoming, U.S.A. 

Is OSM doing something different with queries from the U.K.??

@hgohel pointed out that searches from Spain would NOT benefit having the periods stripped out.

The patch strips the period (.) out without adding spaces. I expected this patch would not be merged but left as a fix for those that would need it. I myself could just change my Gramps record for the "United States" to make it work.

I signed up for an OSM account but cannot see where I can have "USA" added to the list of options.

Nick-Hall:
Look at the debug output.  The transliterated query is:
```
<basin,big horn co,wyoming,u s a>
```
If we strip out the periods the query becomes:
```
<basin,big horn co,wyoming,usa>
```
@hgohel found the opposite case. Since "E.-U." is in the list, it will match "E U", and any query such as "E*U" will also match.  "EU" is not in the list, so in this case stripping out the "." is bad.

The UK works quite well.  It will match "UK", "U.K." and "GB", but not "G.B.".

PR #1620, opened 2023-11-18
Convert images in tree reports to thumbnails for embedding

azrdev:
Images are included by the latex code in full into the rendering, instead of only the rectangle selected in gramps' media DB. Having lots of big photos with small faces in my setup, this leads to unreasonably huge PDF files with unrecognizably small faces in them.

Also, some characters (e.g. comma, spaces) in filepaths might trouble the latex parser, see 75921ceaf and 2da93aad1 -- using thumbnails this is circumvented.

However, note this makes the resulting .tex file not standalone anymore.

### DIFF
--- a/gramps/gen/plug/docgen/treedoc.py
+++ b/gramps/gen/plug/docgen/treedoc.py
@@ -513,6 +513,8 @@ class TreeDocBase(BaseDoc, TreeDoc):
         self.write(level, "}\n")
 
     def write_node(self, db, level, node_type, person, marriage_flag, option_list=None):
+        from ...utils.thumbnails import get_thumbnail_path
+
         options = ["id=%s" % person.gramps_id]
         if option_list:
             options.extend(option_list)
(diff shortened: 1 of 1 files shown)


azrdev:
Already tried to submit the patch to the mailinglist for discussion, but the attachment was stripped.

Nick-Hall:
Rebased.

Nick-Hall:
This PR is causing unit tests to fail.

gramps/gen/plug/docgen/treedoc.py
```diff
@@ -47,6 +47,7 @@
 from ...config import config
 from ...errors import ReportError
 from ...const import GRAMPS_LOCALE as glocale
+from ...utils.thumbnails import get_thumbnail_path
```
dsblank, line 50: Move import to `def write_node()`

```suggestion
```

dsblank (changes requested):
Move import

gramps/gen/plug/docgen/treedoc.py
Nick-Hall: I've moved the import.

hgohel:
@azrdev Is there an enhancement issue associated with this change? This has caused regression [bug 13888](https://gramps-project.org/bugs/view.php?id=13888).

giotodibondone:
Yes restore the previous behavior and make this pr as an option or the other way around please!

azrdev:
> @azrdev Is there an enhancement issue associated with this change? This sounds like a helpful enhancement, but it has caused regression [bug 13888](https://gramps-project.org/bugs/view.php?id=13888).

I think if there was an issue, I'd have linked it here and in the commit message

PR #1622, opened 2023-11-27
importgeneweb.py broken with Python 3.11

olivierberten:


hgohel:
@olivierberten Thanks for the PR. What's the behavior without this fix? Do we need to get this into Gramps 5.2? If so we should get a bug and target it for 5.2.

emyoulation:
Rather than impede progress toward a 5.2 release, perhaps the question is whether this could be done as an add-on? Whether it could  supersede the built-in (or the selector for which plug-in is associated with a file extension)?

This question has direct implication for superseding import/export modules that recognize GEDCOM7  (even if all the GEDCOM features are not supported) or vCard that recognized version 2 AND 3 (not just 3).

olivierberten:
> @olivierberten Thanks for the PR. What's the behavior without this fix? Do we need to get this into Gramps 5.2? If so we should get a bug and target it for 5.2.

Importing a gw file will fail with
```
45320: ERROR: dbloader.py: line 569: Failed to import database.
  File "/usr/lib/python3/dist-packages/gramps/gui/dbloader.py", line 555, in do_import
    self.import_info = importer(self.dbstate.db, filename,
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/gramps/plugins/importer/importgeneweb.py", line 151, in importData
    g = GeneWebParser(database,filename)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/gramps/plugins/importer/importgeneweb.py", line 173, in __init__
    self.f = open(file, "rUb")
             ^^^^^^^^^^^^^^^^^
ValueError: invalid mode: 'rUb'
```

hgohel:
@emyoulation this will be a regression from 5.1.6 if the code ships as-is for 5.2.0 and would need a patch right after release as users run into this. Could we do both, i.e. approve this one character bug fix, and discuss your longer-term thinking in the Development forum?

BTW, there isn't a bug report in the tracker. @olivierberten would you mind creating one? Thanks.

PR #1623, opened 2023-12-06
Fixed Improper Method Call: `__enter__` and `__exit__`

fazledyn-or:
## Details
While triaging your project, our bug fixing tool generated the following message(s)-

> In file: [undoredo.py](https://github.com/gramps-project/gramps/blob/master/gramps/gen/db/undoredo.py#L55), class: DbUndo, there is a special method `__enter__` that contains an unexpected number of parameters. Each call to this method will result in a TypeError. iCR suggested that special methods should have an expected number of parameters. More infomation can be found in the Python [documentation on the various special method](https://docs.python.org/3/reference/datamodel.html#special-method-names).

Based upon close inspection, I found out that the abstract class `DbUndo` was implemented by the [`DbGenericUndo`](https://github.com/gramps-project/gramps/blob/master/gramps/gen/db/generic.py#L114) class. And that class didn't have any `__enter__` or `__exit__` method defined. Since the implemented class didn't implement any of those method, it made sense to remove the methods from the abstract class itself than decorating those methods with `@abstractmethod`.

However, if you feel that those methods need to be present in the source code, then let's discuss how we can make it happen.

## CLA Requirements
*This section is only relevant if your project requires contributors to sign a Contributor License Agreement (CLA) for external contributions.*

All contributed commits are already automatically signed off.

> The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit this work under the same license and agrees to a Developer Certificate of Origin (see [https://developercertificate.org/](https://developercertificate.org/) for more information).
\- [Git Commit SignOff documentation](https://developercertificate.org/)

## Sponsorship and Support
This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/gramps/gen/db/undoredo.py
+++ b/gramps/gen/db/undoredo.py
@@ -52,21 +52,6 @@ class DbUndo(metaclass=ABCMeta):
         self.redoq.clear()
         self.undo_history_timestamp = time.time()
 
-    def __enter__(self, value):
-        """
-        Context manager method to establish the context
-        """
-        self.open(value)
-        return self
-
-    def __exit__(self, exc_type, exc_val, exc_tb):
-        """
-        Context manager method to finish the context
-        """
-        if exc_type is None:
-            self.close()
-        return exc_type is None
-
     @abstractmethod
     def open(self, value):
         """


Nick-Hall:
The DbUndo class is part of a public API.  It is possible, although unlikely, that this code is used by a third-party addon.  We should deprecate the functionality before removing it.

PR #1625, opened 2023-12-21
Report simpler Python version string in About and Error dialogs

hgohel:
A gramps user reported on the discourse forum ([thread](https://gramps.discourse.group/t/help-about-showing-system-details/4534)) that information in the about box was truncated, and he could not expand the box to view the rest of the text.

A look at the text in question suggested a simple change: the truncated string is intended to be the Python version which is typically major.minor.patch version, but the text being displayed is the output of sys.version() which includes additional information including build number and compiler used. For the About box this isn't necessarily useful, so this PR replaces that call with platform.python_version() which returns just the version and resolves the truncated string issue. The same change is also made in the error report dialog for consistency.

An unrelated change is also included in this PR: after building the AIO bundle locally several artifacts are generated (or downloaded) to the source folder, which have been added to .gitignore to avoid tracking with git.

Here's what the Python strings look like with the change:

gramps52-about-box-python-ver-short
gramps52-errorreport-python-ver-short

FYI @emyoulation

### DIFF
--- a/gramps/gui/aboutdialog.py
+++ b/gramps/gui/aboutdialog.py
@@ -179,7 +179,7 @@ class GrampsAboutDialog(Gtk.AboutDialog):
             + distro
         ) % (
             ellipses(str(VERSION)),
-            ellipses(str(sys.version).replace("\n", "")),
+            ellipses(platform.python_version()),
             BSDDB_STR,
             ellipses(get_env_var("LANG", "")),
             ellipses(platform.system()),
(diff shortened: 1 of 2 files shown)


Nick-Hall:
> An unrelated change is also included in this PR: after building the AIO bundle locally several artifacts are generated (or downloaded) to the source folder, which have been added to .gitignore to avoid tracking with git.

Please open a separate PR for the unrelated change.

PR #1628, opened 2023-12-27
Fix gramps example media source

dsblank:
Apparently, the media path has changed relative to {GRAMPS_RESOURCES}.

### DIFF
--- a/example/gramps/example.gramps
+++ b/example/gramps/example.gramps
@@ -7,7 +7,7 @@
     <researcher>
       <resname>Alex Roitman,,,</resname>
     </researcher>
-    <mediapath>{GRAMPS_RESOURCES}/doc/gramps/example/gramps</mediapath>
+    <mediapath>{GRAMPS_RESOURCES}/example/gramps</mediapath>
   </header>
   <name-formats>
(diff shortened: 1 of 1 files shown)


emyoulation:
For 5.2, some directories have been relocated... apparently without planning or discussion beforehand.

In Windows,  the User Directory relocated from the %appData% to %LocalAppDatta% path. This might be arguably appropriate. But it is bad to relocate resources unless it is by design.

There ought to be a discussion of structured default locations of various files: the examples, css (icons and themes), .ini files for addons, .md markdown files for addons, and others. Many should be relative to the User Directory... regardless of OS. 

And the Gtk File Chooser could be made more context aware for its application assigned Shortcut button.

Nick-Hall:
Yes.  The examples files are now installed to the `doc` directory instead of the Gramps resources directory.

So for a standard Linux installation we would find the examples in `/usr/share/doc/gramps` instead of `/usr/share/gramps`.

Is the `GRAMPS_RESOURCES` environment variable set to `/usr/share` or `/usr/share/gramps`? I forget without checking.

Nick-Hall:
> For 5.2, some directories have been relocated... apparently without planning or discussion beforehand.

There was some discussion in PR #1368 when we added support for the XDG base directory specification.

Nick-Hall (approved):

PR #1629, opened 2023-12-28
Update and fix CSV import

dsblank:
This might be too late for gramps52, but this does fix the bug (overwrites firstname) and adds tags to all import types (in CSV), both mentioned in:

https://gramps.discourse.group/t/how-can-people-be-deleted-according-to-a-list/4511/13?u=dsblank

It is ok to apply to master.

### DIFF
--- a/gramps/plugins/importer/importcsv.py
+++ b/gramps/plugins/importer/importcsv.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 #
 # Gramps - a GTK+/GNOME based genealogy program
 #
@@ -24,15 +25,8 @@
 
 "Import from CSV Spreadsheet"
 
-# -------------------------------------------------------------------------
-#
-# Standard Python Modules
-#
-# -------------------------------------------------------------------------
-import time
-import csv
 import codecs
-from io import TextIOWrapper
+import csv
 
 # ------------------------------------------------------------------------
 #
@@ -41,7 +35,15 @@ from io import TextIOWrapper
 # ------------------------------------------------------------------------
 import logging
 
-LOG = logging.getLogger(".ImportCSV")
+# -------------------------------------------------------------------------
+#
+# Standard Python Modules
+#
+# -------------------------------------------------------------------------
+import time
+from io import TextIOWrapper
+
+from gramps.gen.config import config
 
 # -------------------------------------------------------------------------
 #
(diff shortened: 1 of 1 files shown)


dsblank:
Looks like I broke a test. Fixing...

emyoulation:
May I ask why the CSV parser allows adding a Source, but not a Citation? (Or maybe it does and the Wiki overlooks documenting it?)

emyoulation:
Is it possible that the import was originally intended to parse a Full Name ("Surname, Given name")? If so, that would be a useful functionality to keep.

The date is passed through the Date string validator. Perhaps the Place was expected to be a Place Title? From before the Place hierarchy was implemented.

dsblank:
> Is it possible that the import was originally intended to parse a Full Name ("Surname, Given name")? 

No, I don't think I every allowed that as it would be too complicated considering this is CSV.

> May I ask why the CSV parser allows adding a Source, but not a Citation? 

I can add that in the future if it isn't urgent for now.

> The date is passed through the Date string validator. Perhaps the Place was expected to be a Place Title? From before the Place hierarchy was implemented.

Is there a bug related to this?

emyoulation:
No bug. It works fine. Just trying to discover the design approach. So I can determine whether it makes more sense to leverage it... or to embed functions if it is likely to evolve in an opposite direction.

dsblank:
| May I ask why the CSV parser allows adding a Source, but not a Citation?

Looking at the code, I guess for simplicity (or was just based on my workflow). It does create a default citation, but doesn't allow you to use an existing one, or to provide values.

dsblank:
All tests passing now.

emyoulation:
> Looking at the code, I guess for simplicity (or was just based on my workflow). It does create a default citation, but doesn't allow you to use an existing one, or to provide values.

My common case when flowing data from a source via CSV is that they share the same citation. i.e., all the data was transcribed from pages 7-12 of that source. 

So maybe there could be an option to create a single shared citation for the source as an alternative to serialized citations?

dsblank:
| So maybe there could be an option to create a single shared citation for the source as an alternative to serialized citations?

Good idea! I'll do this for an updated 5.2 once it is out.

emyoulation:
please add 'tags' into the recognized headers.  This is what is used for export.

It outputs a CSV string. e.g. "ToDo, Complete, Imported 2024/01/11 10:18:02, Another tag"

Import by relabeling the 'Tags' to "tag" creates Tag named for the various combination. Such as named "ToDo, Complete, Imported 2024/01/11 10:18:02, Another tag" and "ToDo, Another tag"

Nick-Hall (approved):

PR #1630, opened 2024-01-16
Update en.rst by removing reference to gramps-xml (Issue ID 0009723)

kenhara:
I saw this old ticket and thought I might be able to clean it up: https://gramps-project.org/bugs/view.php?id=9723
 This is my first contribution, sorry if I'm doing this incorrectly!

### DIFF
--- a/data/man/en.rst
+++ b/data/man/en.rst
@@ -62,15 +62,15 @@ OPTIONS
     If the **-f** option is not given for any *FILE*, the format of that file
     is guessed according to its extension or MIME type.
 
-    Formats available for export are **gramps-xml** (guessed if *FILE* ends
+    Formats available for export are **gramps** (guessed if *FILE* ends
     with `.gramps`), **gedcom** (guessed if *FILE* ends with `.ged`), or any
     file export available through the Gramps plugin system.
 
-    Formats available for import are **gramps-xml**, **gedcom**, **gramps-pkg**
+    Formats available for import are **gramps**, **gedcom**, **gramps-pkg**
     (guessed if *FILE* ends with `.gpkg`), and **geneweb** (guessed if *FILE*
     ends with `.gw`).
 
-    Formats available for export are **gramps-xml**, **gedcom**,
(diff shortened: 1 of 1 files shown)


data/man/en.rst
```diff
@@ -221,7 +221,7 @@ To explicitly specify the formats in the above example, append filenames with
 appropriate **-f** options::
 
     gramps -i file1.ged -f gedcom -i file2.tgz -f gramps-pkg \
-    -i ~/db3.gramps -f gramps-xml -i file4.wft -f wft -a check
+    -i ~/db3.gramps -f gramps -i file4.wft -f wft -a check
```
hgohel, line 224: As an argument, changing to `-f gramps` is correct. All the other instances of `gramps-xml` on this page refer to the file format which is ["Gramps XML" format](https://www.gramps-project.org/wiki/index.php/Gramps_XML) so I believe it is OK to leave them as-is. However I would defer to someone who has some historical knowledge of the change. @Nick-Hall or @emyoulation or...?

hgohel (changes requested):
@kenhara Thanks for your first PR to Gramps!

A quick search of the code for `gramps-xml` shows that two similar command lines are part of the usage string in [gramps\gramps\cli\argparser.py](https://github.com/gramps-project/gramps/blob/892fc270592095192947097d22a72834d5c70447/gramps/cli/argparser.py#L116) and should be updated with this PR.

hgohel:
@kenhara BTW, I noticed this PR is against your master branch. Typically you'd want to create a branch in your fork for each PR you submit to Gramps. It can take a while for PRs to get merged and you don't want to have multiple fixes pending in your master branch - it will get difficult to manage.

kenhara:
@hgohel Thanks so much for your help, apologies for the newbie error. I've submitted a fresh PR that incorporates this feedback. It is from a branch (not master), it just focuses on the command line changes and adds the additional one you found. Feel free to merge the two or simply close this one, I hope creating another was alright:

([https://github.com/gramps-project/gramps/pull/1631])

PR #1631, opened 2024-01-18
Changing -f gramps-xml to -f gramps (Issue ID 0009723)

kenhara:
Found instances of command line -f prompts for gramps-xml when it should be gramps

This is a do-over for my first pull request (https://github.com/gramps-project/gramps/pull/1630) which I goofed up by using master. It also incorporates changes from @hgohel to update another file and retain the references to Gramps XML. This only applies to the command line prompts now. 

Fixes bug [#9723](https://gramps-project.org/bugs/view.php?id=9723).


hgohel (approved):

hgohel:
Looks good! Thanks again for your contribution to Gramps.

PR #1632, opened 2024-01-21
Fix command line progress output

V-Smeets:
During CLI invocation (e.g.: gramps --action=report ...) a progress is shown on stdout. This progress consists of a procent value and some text. The text is however partly overwriten whenever the percent value is updated.

This PR fixes it in a way that the percent value and text are correctly alligned.


Nick-Hall:
Added unit tests and rebased onto _gramps52_.

PR #1633, opened 2024-01-21
Show other roles for an event in the narrative web report

V-Smeets:
Include an option to the Narrative Web Page generator to also include other roles for an event.

A person can have multiple events. The events where the person is the primary person, are normally shown in the report. With the additional option "Include other roles" enabled (default disabled), the report will also show the persons that also have some role in that event.

The result looks like this:
image

### DIFF
--- a/gramps/gen/utils/db.py
+++ b/gramps/gen/utils/db.py
@@ -664,6 +664,32 @@ def get_source_and_citation_referents(source_handle, db):
     return the_lists
 
 
+def get_event_person_referents(event_handle, db):
+    """
+    Find persons that refer to the event object.
+
+    This function finds all primary objects that refer
+    to the given event handle in a given database.
+    """
+    _primaries = ("Person",)
+
+    lists = get_referents(event_handle, db, _primaries)
+    return [x for xs in lists for x in xs]
+
+
+def get_event_family_referents(event_handle, db):
+    """
+    Find families that refer to the event object.
+
+    This function finds all primary objects that refer
+    to the given event handle in a given database.
+    """
+    _primaries = ("Family",)
+
+    lists = get_referents(event_handle, db, _primaries)
+    return [x for xs in lists for x in xs]
+
+
 def get_media_referents(media_handle, db):
     """
     Find objects that refer the media object.
--- a/gramps/plugins/webreport/basepage.py
+++ b/gramps/plugins/webreport/basepage.py
@@ -92,7 +92,12 @@ from gramps.gen.utils.thumbnails import get_thumbnail_path
 from gramps.gen.display.name import displayer as _nd
 from gramps.gen.display.place import displayer as _pd
 from gramps.plugins.lib.libhtmlconst import _CC
-from gramps.gen.utils.db import get_birth_or_fallback, get_death_or_fallback
+from gramps.gen.utils.db import (
+    get_birth_or_fallback,
+    get_death_or_fallback,
+    get_event_person_referents,
+    get_event_family_referents,
+)
 from gramps.gen.datehandler import parser as _dp
 from gramps.plugins.lib.libhtml import Html, xml_lang
 from gramps.plugins.lib.libhtmlbackend import HtmlBackend, process_spaces
@@ -210,6 +215,7 @@ class BasePage:
         self.create_thumbs_index = report.options["create_thumbs_index"]
         self.inc_families = report.options["inc_families"]
         self.inc_events = report.options["inc_events"]
+        self.inc_other_roles = report.options["inc_other_roles"]
         self.usecms = report.options["usecms"]
         self.prevnext = report.options["prevnext"]
         self.target_uri = report.options["cmsuri"]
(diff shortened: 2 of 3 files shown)


emyoulation:
How timely!  A user asked for a comparable enhancement to show all Shared Event participants in the Complete Individual Report on Discourse

https://gramps.discourse.group/t/shared-events-and-participants-reporting/4712

drjimbouk:
Hi Vincent.
I'm that user!
How do I get copies of the updated files? I don't know my way around GitHub.
I would like to see whether it would work for the forms construct for data entry of census data, for example.
TIA,
James

emyoulation:
There is a discussion about how to introduce non-programmers to Pull Requests in GitHub. There is a 1st draft describing how the patches can be applied to the Gramps on their local machine :

https://gramps.discourse.group/t/create-guidance-document-for-users-about-patching/4699

SNoiraud:
It will not work on a gramps 5.1
You need to have gramps 5.2.

SNoiraud:
It does not work for me.

  File "/home/gramps/pull-gramps-1/gramps/plugins/webreport/basepage.py", line 929, in display_event_row
    witnessed_person_handles = find_witnessed_people(self.r_db, self.person)
AttributeError: 'FamilyPages' object has no attribute 'person'

SNoiraud:
You added a new page name "other roles".
I have two things to say:
 1 - either your patch is incomplete and we have a missing module that is not called.
 2 - either the new page should be removed in the gramps navigation sidebar
If you add no module, you should remove the lines 1897 and 2078

Another thing:
If you want your modification works in family pages, you should replace the line 927:

        if self.inc_other_roles:
by
        if self.inc_other_roles and not omit:

V-Smeets:
I have removed the lines in 1897 and 2078. That was a copy/paste error on my side.

The problem on line 927 is different. I only considered events on persons and not on families. Your fix "and not omit" is not correct. I'm working on a better fix that will work.

SNoiraud:
> Your fix "and not omit" is not correct

Yes, it is the only solution for that problem. The omit variable is set only in the family page. I tested on my database with 15000 persons, 23000 events and 5200 families. No crash and I have the other roles for the persons.

fooltech:
Thanks Serge your fix worked for me by tree is about the same size as yours
to V-Smeets good patch
Phil

SNoiraud:
> I'm working on a better fix that will work.

Do you want to add "other roles" for familes events ?

V-Smeets:
> > I'm working on a better fix that will work.
> 
> Do you want to add "other roles" for familes events ?

Yes, that's what I'm working on.

fooltech:
OK so where I am now the Report is working, I now have a Tab for other roles ( which generates the error "file not found" which is to be expected) and I can see other roles when I go into either of Individuals or Surnames. Could I ask what is the logic by behind extending to Families when by definition "another role" is an individual who is present at the Event of a second individual. I am not saying it would not be valid I am just trying to get my head round the concept.
Phil

V-Smeets:
I just pushed some commits. The tab "Other roles" is now removed.

A Person can have events (with a role) and a Family can also have events (with a role). That means that an event can be referenced by a Person or a Family.
My current commit can only show multiple roles for a person on a PersonPage. On that PersonPage, it will not show any families that also have a role in that same event.

My goal is to extend everywhere where an event is shown (PersonPage and FamilyPage), to also show the other objects (Person or Family) that also have a role in that event.

fooltech:
Hi Vincent

Still cannot conceive any actual event where a Family as group will have 
a role in an individuals life. Or do you just consider Mother & Father 
as the Family Group.
Maybe as group of individuals attending a significant birthday or similar.
Will sort "other roles" Tab
Thanks
Phil

emyoulation:
In general, Family secondary objects are things that apply to both spouses (not the Family's children or one spouse) in equal measure as "Primaries"

So they are good for a report/article that might list "Mr. and Mrs. John and Mary Smith" ... or "Dr. and Mrs. Jones". Both are considered as 'Primary' without the hassles of endless sharing.

fooltech:
Hi

OK can take that but it occurs so infrequently in my world that it is 
not a priority not even on the to do list.
Phil

emyoulation:
I'm surprised at that. Some people only use Family for Events that change the relationship: engagement, marriage, divorce, et cetera.

But when you consider Gallery, Notes, Sources; it is a lot broader. And then there are Residences, Property and such where to don't want doubling of pins on the Maps.

fooltech:
Hi

I have no problem with that approach, however engagement and divorce are 
relatively modern events and certainly I have never come across any 
record of engagements in my family other than those that occurred in my 
living memory, and divorce is rare or again at least in my family so not 
my preferred route, I do not geolocate residencies I could spend the 
rest of my life doing the existing properties I have listed and nothing 
else so I only locate churches and places of Burial, Baptism and 
Marriage, Birth if occuring in hospital.
Phil

SNoiraud:
> I'm surprised at that. Some people only use Family for Events that change the relationship: engagement,
> marriage, divorce, et cetera.

This is how you should do. These are not person events.

fooltech:
Hi Serge

I think on this one we will have to agree to differ and leave it at that

Phil

SNoiraud:
For me you are wrong and this is how gramps is working.
A family start with a marriage event and end with either a divorce either the death of one of the two partners.
If you have a divorce then the two same partners remarry, you must create a second marriage with the new marriage event.
etc...

Nick-Hall:
> Some people only use Family for Events that change the relationship: engagement,
marriage, divorce, et cetera.

Yes.  We inherited this functionality from Gedcom.

When a Family is a participant in an event, the participants are the two people in the relationship.  These events will typically be relationship events and the Family will have the role of "Family".  This indicates that the role of the two people in the family is primary. 

Although it is possible, I can't see a use for adding more than one family to an event, or setting the role for a Family to anything other than "Family".  Does anyone do this?

It is also possible to add two people to a marriage event with the roles "Husband" and "Wife".   Using this approach is likely to cause problems and is not recommended.

V-Smeets:
I have rewritten the code. It now works correctly on the person page and the family page. It shows for every event also the other percipitence (person or family) dat have some role in the event.

The lint check failed on my last commit. It complains about files that I haven't touched. The files I touch are correctly formatted.

To me, this PR is ready to be merged.

Nick-Hall:
> The lint check failed on my last commit. It complains about files that I haven't touched. The files I touch are correctly formatted.

Don't worry about this.  I'll fix it.

Nick-Hall:
Rebased.

drjimbouk:
I've patched my system and run the web report against my data entered via forms. No hiccups and the changes address some of my wishes. A couple of observations: What is the purpose of the list of roles in each individual events section? Where a hyperlink exists it take the user to an individual's page rather that the event where the individual's role is detailed. Whilst  I appreciate that this may be beyond the scope of the submitter's use case, it would be invaluable if other actors in the event where an individual had a role were also reported, together with associated attributes as is the case for marriage events.

V-Smeets:
Hi @drjimbouk,

To your questions:

- What is the purpose of the list of roles in each individual events section?
There can be multiple people that have some role in an event. For a baptising, the primary person is the one that is baptized. There can be some other people that were a witniss to the event. They can all be linked to the same event.
When a family is moving to an other place, then you can assign all the persons of the family as having a primary role in that event. They all move to the new place.
- Where a hyperlink exists it take the user to an individual's page rather that the event where the individual's role is detailed.
There is no page that shows only a single event. So it is not possible to link to such a page. I had no intend to create such an event page.
The show the other people that have a role in that same event. The name of the other person contains a link to the page of that person. There you will see the same event, but then viewed from that other person.
- ... together with associated attributes
What additional assiciated attribute are you expecting? It is not wise to add too much information (birthday, ...) to the names of the other persons. It would cluther the table with unrelated (to the primary person) information.

drjimbouk:
Hi Vincent,

Thanks for the clarification re: list of roles and hyperlink capability. It will still be of use to me as envisaged.

Wrt other actors + attributes. I was thinking along the lines of the marriage event person attributes as currently reported.

Another example might be associated actors in a census event + the same attributes as reported for the principal actor.

Hopefully, someone can pick up the baton and extend what you have achieved.

JS

Nick-Hall:
@SNoiraud Is this PR ready to merge?

SNoiraud:
> @SNoiraud Is this PR ready to merge?

I did not test it because I don't use this kind of feature.

PR #1635, opened 2024-01-26
Reformat code using the latest stable version of black

Nick-Hall:
Version 24.1.0 was released earlier today.

### DIFF
--- a/aio/setup.py
+++ b/aio/setup.py
@@ -2,6 +2,7 @@
  Use with:
  python3 setup.py build_exe --no-compress -O1
 """
+
 import sys
 import os
 import site
--- a/gramps/cli/user.py
+++ b/gramps/cli/user.py
@@ -78,9 +78,9 @@ class User(user.UserBase):
         if auto_accept:
             self.prompt = yes
         if quiet:
-            self.begin_progress = (
-                self.end_progress
-            ) = self.step_progress = self._default_callback = yes
+            self.begin_progress = self.end_progress = self.step_progress = (
+                self._default_callback
+            ) = yes
 
     def begin_progress(self, title, message, steps):
         """
--- a/gramps/gen/datehandler/_date_ja.py
+++ b/gramps/gen/datehandler/_date_ja.py
@@ -249,7 +249,11 @@ class DateDisplayJA(DateDisplay):
                 else:
                     value = "%s年%s" % (year, self.short_months[date_val[1]])
             else:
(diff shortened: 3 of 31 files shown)


jralls:
The "code quality" claim is dubious at best considering the moving around of gratuitous parentheses in long chains of `foo = bar = … = quux = None`.

Nick-Hall:
Yes.  The main reason for the PR (and the quick merge) was to fix the failing lint check.

An alternative would be to force the action to use an older version of black as described [here](https://black.readthedocs.io/en/stable/integrations/github_actions.html) in the documentation.

It seems that they do a major upgrade like this once a year that may change some formatting.   How should we handle this now and in the future?

jralls:
> An alternative would be to force the action to use an older version of black as described [here](https://black.readthedocs.io/en/stable/integrations/github_actions.html) in the documentation.

That's illusory: It says the version has to be on PyPi and the [PyPi](https://pypi.org/project/black/#files) has only the latest version.

> How should we handle this now and in the future?

Split up the lines so that Black doesn't want to add parentheses to make a line continuation. e.g:
```
self.begin_progress = self.end_progress = yes
self.step_progress = self._default_callback = yes
```
is clearer than either
```
    self.begin_progress = (
           self.end_progress
      ) = self.step_progress = self._default_callback = yes
```
or
```                       
self.begin_progress = self.end_progress = self.step_progress = (
       self._default_callback
) = yes
```
and will be immune to getting reformatted every time the Black team's whim changes about where the line should break.

QuLogic:
> That's illusory: It says the version has to be on PyPi and the [PyPi](https://pypi.org/project/black/#files) has only the latest version.

I don't love black either, but there are plenty of [older releases on PyPI](https://pypi.org/project/black/#history).

kulath:
I don’t love black either. I suppose the alternative would be (liberal) use of Pragma (pragmata?) to tell black (and if necessary lint) to ignore certain lines. TimSent from my iPhoneOn 27 Jan 2024, at 04:52, Elliott Sales de Andrade ***@***.***> wrote:﻿

That's illusory: It says the version has to be on PyPi and the PyPi has only the latest version.

I don't love black either, but there are plenty of older releases on PyPI.

jralls:
> I don't love black either, but there are plenty of older releases on PyPI.

Ah, I stand corrected. I expected history to be a simple list, not installable links.

PR #1636, opened 2024-01-26
Add a name parameter to the get_person_mark utility

Nick-Hall:
This allows an alphabetical index entry for a person to use the selected name format.  Previously a "Surname, Given Suffix" format was always used.

Fixes [#13155](https://gramps-project.org/bugs/view.php?id=13155).

A quick _grep_ reveals that about a dozen reports are affected.

Another possible solution would be to pass a name displayer to the `get_person_mark` utility rather than a name.  This needs further investigation.  I don't have time right now.

### DIFF
--- a/gramps/gen/plug/report/utils.py
+++ b/gramps/gen/plug/report/utils.py
@@ -218,7 +218,7 @@ def find_marriage(database, family):
 # Indexing function
 #
 # -------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


emyoulation:
Is the "needed for Arabic" annotation used everywhere in the code? Or is it a mix of notations using a generic "RTL" and others with specific languages?

Nick-Hall:

PR #1638, opened 2024-01-28
Make subprocess inherit environment variables

TomaSajt:
I was trying to build the latest beta versions of the package, but I got an error coming from the `merge(...)` function in `setup.py`

> error: [Errno 2] No such file or directory: 'msgfmt'

It turns out, a call to `msgfmt` was setting an environment variable without inheriting the current environment variables. Because of this, the `PATH` environment variable got unset for that subprocess, thus not being able to find `msgfmt`.

Luckily, I found a quick fix for it.


Nick-Hall:
Rebased onto _gramps52_.

PR #1641, opened 2024-02-08
Ignore AIO artifacts in gitignore

hgohel:
Several artifacts are generated (or downloaded) when building the AIO bundle, which have been added to .gitignore since they don't need to be tracked with git.


hgohel:
@Nick-Hall This was split from the Python version PR as you suggested.

PR #1656, opened 2024-02-23
Remove a reverted change from Changelog

tosky:
It was wrong translation and it was reverted.

Unfortunately there is no a proper way to mark "do not translated this string", maybe some strings should be moved to a different file.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,11 +7,6 @@
 	* po/zh_CN.po: Translated using Weblate (Chinese (Simplified)) Currently translated at 90.9% (6474 of 7115 strings) Translation: Gramps/Program Translate-URL:
 	https://hosted.weblate.org/projects/gramps-project/gramps/zh_Hans/
 
-2024-02-23  Marco Ciampa <ciampix@posteo.net>
-
-	* po/it.po: Translated using Weblate (Italian) Currently translated at 98.3% (7000 of 7115 strings) Translation: Gramps/Program Translate-URL:
-	https://hosted.weblate.org/projects/gramps-project/gramps/it/
-
 2024-02-22  Nick Hall <nick-h@gramps-project.org>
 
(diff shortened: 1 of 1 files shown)


jralls:
> Unfortunately there is no a proper way to mark "do not translated this string", maybe some strings should be moved to a different file.

There is: Remove the gettext alias call (`_(…)`) around the string in the source. 

But that's not the right thing to do for most of these strings. They hava a comment [`Translators: If "from <Month>" needs a special inflection in your language, translate to "{long_month.forms[X]} {year}" (where X is one of the month-name inflections you defined) else leave it untranslated`](https://github.com/gramps-project/gramps/blob/f41cddcd2f3f2164de43596d39ca1aa835fff38f/gramps/gen/datehandler/_datedisplay.py#L185C13-L188C41) explaining when it should be translated. That's displayed in white-on-black above the string on the [weblate page](https://hosted.weblate.org/translate/gramps-project/gramps/it/?q=%7Blong_month%7D&sort_by=-priority%2Cposition&offset=2).

Nick-Hall:
I don't see much point to this commit.  The entry will be in both `ChangeLog` and `po/ChangeLog` which will both be overwritten in the next release anyway.

tosky:
Are those ChangeLog files overwritten even before minor releases, i.e. they will be regenerated also for 5.2.1?

For the record, the commit was useful before I realized the release was created.

Nick-Hall:
Yes, we create new files for every release.  There is a step in the [What to do for a release](https://gramps-project.org/wiki/index.php/What_to_do_for_a_release#Changelog_and_NEWS_file) wiki page.

The extra commit would just create an extra entry in the ChangeLog.

Your revert commit isn't in the ChangeLogs because I had already generated them when I saw your commit.  I suppose strictly speaking I should have gone back and generated them again, but I decided that it wasn't that important.

PR #1657, opened 2024-02-24
Update AppStream MetaInfo file to conform to latest spec

Alexander-Wilms:
See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html for reference


Nick-Hall:
Thanks for bringing this to our attention.

This PR looks good, except that the `gramps.pot` and `*.po` files should not be updated. The `gramps.pot` is generated automatically and is merged into the `po` files by Weblate.

Alexander-Wilms:
Done

Nick-Hall:

File rename already done in PR #1674 instead.

Changed "Pedigree Chart" to "Pedigree" because it was already translated.  Changed "org.gramps-project" to "org.gramps_project" for consistency.

Need to include metainfo rather than appdata its files.

PR #1660, opened 2024-02-26
Use regular buttons to copy owner information from/to preferences

QuLogic:
It's strange to me to use a context menu *anywhere in the window* for this, and in the space used for the label suggesting how to use it, we can fit the two buttons instead.

Also, GtkMenu is gone in GTK4, so removing one of its usages will make such a port easier.

Also, modifying the Glade file with newer Glade changed all underscores to dashes, making the diff a bit large; I could revert that or perhaps it should be a separate update PR for _all_ Glade files.

Before:
image
After:
image

### DIFF
--- a/gramps/plugins/tool/ownereditor.glade
+++ b/gramps/plugins/tool/ownereditor.glade
@@ -2,431 +2,397 @@
 <!-- Generated with glade 3.40.0 -->
 <interface>
   <requires lib="gtk+" version="3.24"/>
-  <object class="GtkAccelGroup" id="accelgroup1"/>
-  <object class="GtkMenu" id="popup_menu">
-    <property name="can-focus">False</property>
-    <child>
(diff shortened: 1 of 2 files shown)


emyoulation:
There are several enhancement requests related to integrating Owner and Researcher editors in a cleaner way. And using XML structures for this data that exist in the Person (and secondary) objects instead having another structure.

Just including this note so there is something to connect this change to a possible 5.3 change.

https://gramps-project.org/bugs/view.php?id=12430

Nick-Hall:
Rebased.

giotodibondone:
It would have be helpful to keep the explanation message **"Copy from/to Researcher Preferences"**  for usability as either an inline message like it was (preferred for screenshots) or a tooltip on each button.

https://gramps-project.org/wiki/index.php/Gramps_6.0_Wiki_Manual_-_Tools#Edit_Database_Owner_Information

PR #1661, opened 2024-02-26
Replace GtkArrow with GtkImage and symbolic icons

QuLogic:
The former was deprecated in GTK 3.14, and removed in GTK4.


QuLogic:
I've reverted the changes to the Gramplet bar and Navigator, because I've opened #1663 and #1664 which rewrite those in better ways.

Nick-Hall:
Rebased.

PR #1663, opened 2024-03-02
Replace GtkMenu with GioMenu in Gramplets bar

QuLogic:
The former is removed in GTK4. And this change also removes the deprecated `GtkArrow`.


Nick-Hall:
Rebased.

PR #1664, opened 2024-03-02
Update Navigator to use modern widgets

QuLogic:
A combo box for the page selection tracks the popup menu for you, which simplifies that bit of the code, and also removes the deprecated `GtkMenu` and `GtkArrow`.

A stack instead of a notebook for the pages can be set to non-homogeneous, so there's no need to mess with hiding/showing widgets.

With the available properties (`GtkComboBox:active-id` and `GtkStack:visible-child-name`), we can bind the stack/combo together and not have to implement this ourselves.

Finally, fix a typo in the `ViewsInCategory` action name.

I broke each of these changes into separate commits for easier review.

### DIFF
--- a/gramps/gui/grampsgui.py
+++ b/gramps/gui/grampsgui.py
@@ -215,7 +215,7 @@ UIDEFAULT = (
           <attribute name="label" translatable="yes">F_ull Screen</attribute>
         </item>
       </section>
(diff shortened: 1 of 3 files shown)


Nick-Hall:
Rebased.

Breaking this down into separate commits was helpful.  Thank you.

PR #1666, opened 2024-03-02
Remove unused Subversion keyword substitutions

QuLogic:
These are not used with Git and are clearly outdated, as these files have all been touched since the indicated dates.


hgohel:
This cleanup makes sense. Same rationale applies the addons-source project as well.

Nick-Hall:
Yes.  We could even reformat _addons-source_ using black.

PR #1669, opened 2024-03-03
Addon Manager Requirements mark symbols on Mac

GaryGriffin:
Change the cross mark and check mark symbols used in the Addon Manager Requirements screen to be valid characters on Mac.

Addresses bug [#13194](https://gramps-project.org/bugs/view.php?id=13194).

### DIFF
--- a/gramps/gen/utils/requirements.py
+++ b/gramps/gen/utils/requirements.py
@@ -182,4 +182,4 @@ def tick_cross(value):
     """
     Return a tick for True or a cross for False
     """
-    return "\u2714" if value else "\u2718"
+    return "\N{CHECK MARK}" if value else "\N{CROSS MARK}"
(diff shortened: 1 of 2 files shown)


jralls:
@GaryGriffin your commit messages about code points not being valid on mac are wrong. The code points are part of the Unicode standard and are valid everywhere. The problem on macOS is that the current system UI font, San Francisco, doesn't provide glyphs for them. Other fonts do, something you can prove to yourself by searching the codepoints in the Emoji and Symbol chooser.

I have no objection to the actual changes, though @Nick-Hall and @SNoiraud  might like them guarded with `if mac():` so that the glyphs they like are kept on Linux and Windows where font fallback isn't broken. I do prefer using the `\N{…}` code-point name escape over the code point number, it's more readable.

GaryGriffin:
Added a second fix (tied to same bug report) for the fallback symbol for Death. In previous gramps versions (5.1.6), it was '+'. It was just changed (in 5.2) to LATIN CROSS (unicode 0271d), which does not have a glyph in the default Mac font. This change reverts back to using a fallback symbol of +. 

The current 5.2 code has LATIN CROSS as both the default and the fallback. This displays a bad glyph (text of 271d) in the Pedigree chart death date line, when using the pure standard install dmg.

jralls:
Note that the death symbol is just a default and can be overridden in Preferences>Genealogy Symbols.

PR #1670, opened 2024-03-05
Fix fan charts on HiDPI screens

QuLogic:
Most DrawingArea-based widgets are fine, but the fan charts use a backing image surface that is blitted when needed. On HiDPI displays, this image surface is normal size, and thus everything ends up blurry when scaled up.

Instead the backing surface should be scaled to match the display, and also re-created if the display scale changes (e.g., if display settings change, or the window is moved between displays with different settings.)

I suppose this could possibly be backported to maintenance branches as well.

### DIFF
--- a/gramps/gui/widgets/fanchart.py
+++ b/gramps/gui/widgets/fanchart.py
@@ -165,6 +165,7 @@ class FanChartBaseWidget(Gtk.DrawingArea):
         self.set_can_focus(True)
         self.connect("key-press-event", self.on_key_press)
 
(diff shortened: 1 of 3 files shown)


Nick-Hall:
Is there a bug report for this?

QuLogic:
I could not find one when I looked.

Nick-Hall:
@QuLogic Did you intend this bug fix for the _gramps52_ branch?

QuLogic:
Up to you; I can re-target or backport.

Nick-Hall:

PR #1671, opened 2024-03-05
Simplify some Cairo drawing

QuLogic:
Instead of generating paths twice, once for filling and once for stroking, use the path-preserving `stroke_preserve` operator instead. As well, due the filling before stroking for consistency, or else the stroke might be accidentally covered, as occurred with the child rings.

Also, simplify the map marker which seems to have drawn the elliptical portion twice.


hgohel:
@QuLogic Love all your enhancements. How do we validate the changes?

QuLogic:
For the fan charts, you can just open them like normal, and they should not have changed (except the strokes on the child rings).

For the map markers, I'm actually not sure how to get the maps to use the coloured markers. I just edited the view files and forced it to use a hard-coded colour instead of the stock image while testing.

Nick-Hall:
Rebased.

PR #1672, opened 2024-03-09
Fix unit tests when run with Python 3.12

Nick-Hall:
When running in command line mode, the Gedcom and ProGen importers were being run before the `gi.require_version` method, causing Gtk 4.0 to be incorrectly loaded. This caused some unit tests to fail.

Fixes [#13212](https://gramps-project.org/bugs/view.php?id=13212).


RossGammon:
I just added a note on the bug. Unfortunately this does not fix the build failure on Ubuntu. Would it be worth temporarily using python 3.12 on this bug fix branch in Github-CI?
https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#using-a-specific-python-version

Nick-Hall:
I don't think that changing the python version will work because of the build dependencies.

For my testing, I installed a nightly build of Ubuntu Noble in a virtual machine.  Then I just run the unit tests from the git clone.

What is the easiest way for me to replicate your build environment?

RossGammon:
Install needed packages for the build (I hope this is everything):
sudo apt-get install pbuilder debootstrap devscripts
sudo apt-get install build-essential quilt

Get the debian packaging and source:
dget https://deb.debian.org/debian/pool/main/g/gramps/gramps_5.1.6+dfsg-1.dsc

cd gramps-5.1.6+dfsg/
Create a new version in the Debian changelog (e.g. append ubuntu0 to the version):
dch -r
Build a new Debian source package
debuild -S --no-sign
cd ..

Create the pbuilder chroot:
sudo pbuilder create --distribution noble  --debootstrapopts --variant=buildd  --othermirror "deb http://archive.ubuntu.com/ubuntu noble universe"

Build the package:
sudo pbuilder build gramps_5.1.6+dfsg-1ubuntu0.dsc

RossGammon:
With the extra commits as a patch, the package built fine in a ppa. So I have uploaded it to Ubuntu. Unfortunately, it goes into the NEW queue which needs someone to manually approve it. As it is a bug fix upload, I don't expect any problems with that. I will keep an eye on it, and contact the archive managers if it gets ignored for too long.
Thanks a lot for your help Nick!

PR #1673, opened 2024-03-10
Add release information to appdata

a-andre:


a-andre:
@Nick-Hall Before tagging a (stable) release an entry should be added to the appdata file, so that the [flatpak repo](https://github.com/gramps-project/flatpak) doesn't have to maintain a version of the appdata file.

Nick-Hall:
Do we also need to rename this file for the current flatpak?

Nick-Hall:
Merged PR #1674 into _gramps52_ instead.

PR #1674, opened 2024-03-10
Add release information + rename metadata file to follow spec

a-andre:
See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location


Nick-Hall:
I'll add a note in the release documentation to update this file.

Nick-Hall:
Rebased onto _gramps52_ branch.

PR #1676, opened 2024-03-11
Add forwards and backwards navigation using mouse buttons.

carlesalbasboix:
I have added the functionality to go backwards and forwards in NavigationView using the side mouse buttons, just like in a browser.

### DIFF
--- a/gramps/gui/views/navigationview.py
+++ b/gramps/gui/views/navigationview.py
@@ -486,6 +486,19 @@ class NavigationView(PageView):
                     return True
         return super(NavigationView, self).key_press_handler(widget, event)
 
+    def button_press_handler(self, widget, event):
+        """
+        Handle forward and backward buttons, or pass it on.
+        """
+        if self.active:
+            if event.button == 8:
+                self.back_clicked()
+                return True
+            elif event.button == 9:
+                self.fwd_clicked()
+                return True
+        return super(NavigationView, self).button_press_handler(widget, event)
+
     def call_copy(self):
         """
         Navigation specific copy (control+c) hander. If the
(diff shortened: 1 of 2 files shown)


Nick-Hall:
Good idea.

Nick-Hall:
Rebased.

Nick-Hall:
In testing this, I am getting a signal recursion:

`Warning: History: Signal recursion blocked. Signals was : active-changed`

This occurs at lines 141, 156 and 173 in the file [gramps/gui/displaystate.py](https://github.com/gramps-project/gramps/blob/master/gramps/gui/displaystate.py#L85).

carlesalbasboix:
I cannot see this warning. Are you enabling extra warnings somehow?

Nick-Hall:
The warnings appear in the terminal window.  I have no extra warnings or debug options enabled.

I'm testing using the example tree in the person view.  I cursor down to add about 5 people to the history and then use the mouse buttons to backwards and forwards.  The navigation works but the warnings are concerning.

carlesalbasboix:
I was testing it only in the Charts view. Now I tried the person view and I get the warning, but only when I click the forwards and backwards buttons on the mouse very fast. If I hold down Alt + Left I get the same warning. I think the problem is emitting a signal to change the active person before the previous one has finished processing.

carlesalbasboix:
I think I have fixed the problem by preventing forwards/backwards navigation while history is locked.

emyoulation:
An observation about something that may give a false impression of 'flakeyness' the improved mouse forward/back navigation...


When committing a new object to Gramps via the Object Editors, that new object is not added to the History. 


I'm NOT suggesting that the active object focus should shift to the new object after an Add. Effectively, the active object briefly shifts to the new object as the commit occurs, then back to where it was. So maybe the History should push an item into the list?

Nick-Hall:
@carlesalbasboix Thank you. It looks like you fixed the problem.  I'll merge this now.

PR #1678, opened 2024-03-14
Allow 0 ancestors/descendants in hourglass graph

carlesalbasboix:
For some reason, the hourglass graph doesn't allow 0 ascendants or descendants. The graph works well with 0, so I suggest to change it.

### DIFF
--- a/gramps/plugins/graph/gvhourglass.py
+++ b/gramps/plugins/graph/gvhourglass.py
@@ -443,13 +443,13 @@ class HourGlassOptions(MenuReportOptions):
         pid.set_help(_("The Center person for the graph"))
         menu.add_option(category_name, "pid", pid)
 
-        max_gen_d = NumberOption(_("Max Descendant Generations"), 10, 1, 15)
+        max_gen_d = NumberOption(_("Max Descendant Generations"), 10, 0, 15)
         max_gen_d.set_help(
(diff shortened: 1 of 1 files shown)


emyoulation:
Without at least 1 Ancestor and 1 Descendant generations, it is no longer an "Hourglass" shape.

Is this an appropriate enhancement if it pushes outside the graph type?

carlesalbasboix:
In math we'd call it a degenerated hourglass :)

Regardless of naming, I don't see any reason for artificial limitations. The exact "upper half" or "lower half" of the hourglass graph cannot be obtained in any other way as far as I know.

Nick-Hall:
I don't see a problem with this.

emyoulation:
Ok. I can see the value

In fact, there have also been calls for the graph without the "narrow neck" node.

e.g., an hourglass graph for a family rather than a person... 
1) of a couple without children, or 
2) of siblings with unknown parents.

emyoulation:
Ok. I can see the value

In fact, there have also been calls for the graph without the "narrow neck" node.

e.g., an hourglass graph centered on a Family node rather than a Person node... 
1) of a couple without children, or 
2) of siblings with unknown parents.

Nick-Hall:
Rebased.

PR #1679, opened 2024-03-15
Update AIO minimum Windows version requirement to 8.1 64-bit

Nick-Hall:
The MSYS2 build environment [dropped support](https://www.msys2.org/news/#2023-01-15-dropping-support-for-windows-7-and-80
) for Windows 7 and 8.0 on 15th Jan 2023.

See the MSYS2 [supported versions](https://www.msys2.org/docs/windows_support/).

### DIFF
--- a/aio/grampsaio64.nsi.template
+++ b/aio/grampsaio64.nsi.template
@@ -718,8 +718,8 @@ Function .onInit
     Quit
   ${EndIf}
 
-  ${If} ${AtMostWinXP}
-    MessageBox MB_OK "Windows Vista and above required"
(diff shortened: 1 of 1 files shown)


jralls:
> The MSYS2 build environment [dropped support](https://www.msys2.org/news/#2023-01-15-dropping-support-for-windows-7-and-80) for Windows 7 and 8.0 on 15th Jan 2013.

Err, 15 Jan 2023. Commit message needs fixing too.

Nick-Hall:
@jralls Thanks.  I'll fix it.

PR #1688, opened 2024-03-23
Allow Gramps to run from an extracted source distribution

jmichault:
When trying to run gramps directly from sources extracted from tar.gz file, gramps don't run : 

```
$ python3 Gramps.py
ResourcePath.ERROR: Unable to determine resource path
```
This is because gramps is testing for the existence of the ".git" folder, which does not exist in the sources extracted from the tar.gz file.

I suggest testing for the existence of the "data" folder, which exists both in the sources obtained by "git clone" and those extracted from tar.gz.

### DIFF
--- a/gramps/gen/utils/resourcepath.py
+++ b/gramps/gen/utils/resourcepath.py
@@ -74,7 +74,7 @@ class ResourcePath:
         package_path = os.path.abspath(
             os.path.join(os.path.dirname(__file__), "..", "..", "..")
         )
-        installed = not os.path.exists(os.path.join(package_path, ".git"))
+        installed = not os.path.exists(os.path.join(package_path, "MANIFEST.in"))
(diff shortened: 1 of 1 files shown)


Nick-Hall:
The problem is that it is possible for a package called "data" to exist in the `site-packages` directory.  A check for the `MANIFEST.in` file would be safer.

Please rebase this on the _master_ branch.  It is not a bug.

Nick-Hall:
As a workaround, run the command `touch .git` to create an empty file.

jmichault:
All right.
Pull request rebased, and test changed for MANIFEST.in

Nick-Hall:
That should be safer.  There is a project called [data](https://pypi.org/project/data/), but calling a project "MANIFEST.in" is unlikely.

PR #1692, opened 2024-03-26
Don't show Navigation when we print a page.

SNoiraud:
Fixes #13160


Nick-Hall:
Merged in commit c30de3c6d8acbf3432a847a867b585b892ccb2ac.

PR #1697, opened 2024-04-07
add workflow for .deb build.

jmichault:
little changes in debian/rules to allow build.

And workflow to build .deb package for ubuntu 20.04 and ubuntu latest (actually 22.04 for github).

### DIFF
--- a/.github/workflows/build_deb.yml
+++ b/.github/workflows/build_deb.yml
@@ -1,0 +1,97 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2021       Nick Hall
+# Copyright (C) 2024       Jean Michault
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+name: Build Debian packages
+
+on:
+  workflow_dispatch:
+    inputs:
+      build-number:
+        description: 'Build number'
+        type: string
+        default: '1'
+
+jobs:
+  build:
+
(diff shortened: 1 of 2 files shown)


Nick-Hall:
Excellent!  I'll test this tomorrow.

Nick-Hall:
This looks good.  I only made a few minor changes:

* For a non-development version, use a build number instead of a short commit sha.
* Rename the deb file to include the Ubuntu version.
* Specify ubuntu-22.04 instead of ubuntu-latest.

emyoulation:
The releases for Debian 5.2.2 were just posted to the Release Assets.

Where is there are description of the file naming convention?  We had to explain the AIO for the Windows assets.  What does the "all." mean in the Debian assets.


gramps_5.2.2-1_all.ubuntu-20.04.deb

Nick-Hall:
This workflow gives us two builds using different [runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).  At the moment they are Ubuntu 20.04 and Ubuntu 22.04.

The usual naming convention would be:  gramps_5.2.2-1_all.deb

The "all" just means all architectures.  I just added the runner to differentiate the files.  Jean used a similar approach in his repository.

emyoulation:
the "All" in the All-in-one for windows refers to "all dependencies for the Windows platform".   (Further explained as "all the related programs and library files which Gramps needs on Windows .")

The "All Architectures" means what specifically?   

According to the search results, Debian officially supports 10 major architectures:
-     amd64 (64-bit x86 processors)
-     arm64 (64-bit ARM processors)
-     armel (32-bit little-endian ARM processors)
-     armhf (32-bit ARM processors with hardware floating-point support)
-     i386 (32-bit x86 processors)
-     mips (32-bit big-endian MIPS processors)
-     mips64el (64-bit little-endian MIPS processors)
-     mipsel (32-bit little-endian MIPS processors)
-     ppc64el (64-bit little-endian PowerPC processors)
-     s390x (64-bit IBM System z mainframe processors)

emyoulation:
> ...different [runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories). At...

Interesting read.  Thanks

Nick-Hall:
> The "All Architectures" means what specifically?

It just means that the package is not architecture specific.  In our case Gramps, doesn't have to be compiled separately for each architecture.

Here are a couple of links to read:

https://askubuntu.com/questions/330018/what-is-the-standard-for-naming-deb-file-name
https://kerneltalks.com/tools/understanding-package-naming-convention-rpm-deb

PR #1698, opened 2024-04-08
Narrative web: Handle not found when creating the updates page

SNoiraud:
Fixes #13216


gramps/plugins/webreport/updates.py
```diff
@@ -309,13 +309,13 @@ def sort_on_change(handle):
             fct = self.report.database.get_place_from_handle
             fct_link = self.place_link
         elif object_type == Source:
-            fct = self.report.database.get_place_from_handle
+            fct = self.report.database.get_source_from_handle
```
hgohel, line 312: `get_*_from_handle` methods raise errors which if trapped would give better call stacks than the one in the bug report associated with this defect. Suggest adding exception handling.

SNoiraud: Ok.

hgohel: I see the exception handling added below, however, `fct` gets invoked in the call to `sorted()` so exception handling prior to that would be more effective.

SNoiraud: I can't . I already make the exception handling in the sort:

        def sort_on_change(handle):
            """sort records based on the last change time"""
            try:
                obj = fct(handle)
                timestamp = obj.get_change_time()
            except exception:
                print("In sort :", handle, " not found for", object_type)
                timestamp = 0
            return timestamp

...

        obj_list = sorted(
            self.report.obj_dict[object_type], key=sort_on_change, reverse=True
        )

hgohel (approved):

PR #1700, opened 2024-04-10
Update citation.py

emyoulation:
change default in Citation filter for Confidence from "Normal" to "Very Low" so that nothing is excluded by default.

https://gramps-project.org/bugs/view.php?id=11797

### DIFF
--- a/gramps/gen/lib/citation.py
+++ b/gramps/gen/lib/citation.py
@@ -81,7 +81,7 @@ class Citation(
         DateBase.__init__(self)  #  2
         self.source_handle = None  #  5
         self.page = ""  #  3
-        self.confidence = Citation.CONF_NORMAL  #  4
+        self.confidence = Citation.CONF_VERY_LOW  #  4
         SrcAttributeBase.__init__(self)  #  8
 
     @classmethod


Nick-Hall:
We should probably the keep the default confidence for a citation as "Normal".  This equates to no Gedcom confidence being specified.

emyoulation:
I strongly disagree.  The default should be unfiltered.  This feature and the Addon Manager are the 2 features that filter by default.  Both cause issues.  

The Addon Manager non-intuitive complaints could be addressed by conspicuous feedback (by color) that a Filter is occurring. 

That luxury does not exist for the Confidence.

And I do not think that 'Normal' equates to no confidence specified.  Wouldn't that a NULL ?

Nick-Hall:
Have a look at the [QUALITY_MAP](https://github.com/gramps-project/gramps/blob/master/gramps/plugins/export/exportgedcom.py#L148) in the Gedcom export.  We do not map CONF_NORMAL to any Gedcom value.

You code would set the default confidence for any new citation created to CONF_VERY_LOW.

I see nothing in this PR relating to filters.

emyoulation:
OK. Thanks for catching my mistake.

I was thinking this was where the "Reset" for the Citation Filter gramplet got its values. Instead, it is the default for creation of a New Citation. And yes, new citations should be "Normal" by default.

SNoiraud:
On my gramps version, I have a preference option to choose the default confidence level.
I can make a PR if you are intetested

Nick-Hall:
Closed in favour of #1721.

emyoulation:
> On my gramps version, I have a preference option to choose the default confidence level.
> I can make a PR if you are interested

That sounds interesting. But if a default citing option was added, it would be nice if it was a Citation WITH its Confidence as a default.

That's because researchers will normally be transcribing a full page from a source during a session. Then move on to the next page or another source. So the Citation page and confidence should be the same on each 'Ok' click.

Or else they will proving a single fact by adding multiple sources. In that case, the page and confidence will differ for each added source.

PR #1701, opened 2024-04-10
Change default confidence level in citation filter

emyoulation:
change default in Citation filter for Confidence from "Normal" to "Very Low" so that nothing is excluded by default.

https://gramps-project.org/bugs/view.php?id=11797
missed line for  PR Update citation.py #1700

### DIFF
--- a/gramps/gui/filters/sidebar/_citationsidebarfilter.py
+++ b/gramps/gui/filters/sidebar/_citationsidebarfilter.py
@@ -142,7 +142,7 @@ class CitationSidebarFilter(SidebarFilter):
         self.filter_id.set_text("")
         self.filter_page.set_text("")
         self.filter_date.set_text("")
-        self.filter_conf.set_active(Citation.CONF_NORMAL)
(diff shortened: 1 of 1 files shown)


Nick-Hall:
We should also consider renaming the level to correspond to the Gedcom definitions:

* Very high -> Primary evidence
* High -> Secondary evidence
* Low -> Questionable evidence
*  Very low -> Unreliable/Estimated evidence

emyoulation:
The bad thing about using the GEDCOM strings/phrases is the extra Column width it would consume in the Views. (They are bad enough in English. But could be prohibitive in locales that are often double the length of the English term... like German) Could the GEDCOM strings be tooltips instead of the visible values?

    Very high -> Primary evidence
    High -> Secondary evidence
    Low -> Questionable evidence
    Very low -> Unreliable/Estimated evidence

Discussion and some stats on word length expansion during translation:
https://www.inter-contact.de/en/blog/text-length-languages

Nick-Hall:
Closed in favour of #1721.

PR #1708, opened 2024-04-18
Installing 5.1.x after 5.2.x will no longer override config/data locations

hgohel:
Fix [#13261](https://gramps-project.org/bugs/view.php?id=13261)

In an existing 5.2.x installation with XDG directory layout for data/config/cache, installing 5.1.x later would create a second config folder in the legacy layout, overriding the existing config.

This change makes it so that an existing configuration for 5.2.x will be given preference over the 5.1.x config. If there isn't one, previous logic will continue to be used.

Testing passed on Windows
1. Gramps 5.2.2 only
2. Gramps 5.2.2 followed by Gramps 5.1.6
3. Gramps 5.1.6 followed by Gramps 5.2.2
* Create family trees in each version of Gramps and ensure that they are available when that version is launched.
* Ensure that if 5.1.x is installed before 5.2.x, config is imported.

- [X] Testing on Windows
- [x] Testing on Linux
- [X] Testing on MacOS 
- [ ] Testing with GRAMPSHOME and SAFEMODE (cli?).

Thanks!

### DIFF
--- a/gramps/gen/const.py
+++ b/gramps/gen/const.py
@@ -114,15 +114,16 @@ if "SAFEMODE" in os.environ:
         USER_HOME = get_env_var("HOME")
     HOME_DIR = get_env_var("SAFEMODE")
 
-
-if os.path.exists(HOME_DIR) or "GRAMPSHOME" in os.environ or "SAFEMODE" in os.environ:
+USER_DATA = os.path.join(GLib.get_user_data_dir(), "gramps")
+USER_CONFIG = os.path.join(GLib.get_user_config_dir(), "gramps")
+USER_CACHE = os.path.join(GLib.get_user_cache_dir(), "gramps")
+if os.path.exists(USER_DATA) and os.path.exists(USER_CONFIG):
+    # Use existing XDG-based directory layout
+    pass
+elif os.path.exists(HOME_DIR) or "GRAMPSHOME" in os.environ or "SAFEMODE" in os.environ:
     USER_DATA = HOME_DIR
     USER_CONFIG = HOME_DIR
     USER_CACHE = HOME_DIR
-else:
-    USER_DATA = os.path.join(GLib.get_user_data_dir(), "gramps")
-    USER_CONFIG = os.path.join(GLib.get_user_config_dir(), "gramps")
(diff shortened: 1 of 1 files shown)


gramps/gen/const.py
```diff
@@ -114,15 +114,16 @@
         USER_HOME = get_env_var("HOME")
     HOME_DIR = get_env_var("SAFEMODE")
 
-
-if os.path.exists(HOME_DIR) or "GRAMPSHOME" in os.environ or "SAFEMODE" in os.environ:
+USER_DATA = os.path.join(GLib.get_user_data_dir(), "gramps")
+USER_CONFIG = os.path.join(GLib.get_user_config_dir(), "gramps")
+USER_CACHE = os.path.join(GLib.get_user_cache_dir(), "gramps")
+if os.path.exists(USER_DATA) and os.path.exists(USER_CONFIG):
```
hgohel, line 120: This condition could be merged into elif later on but readability/understandability is lost, so I left it this way with a pass statement.

jralls:
macOS, at least in the AIO bundle, isn't affected by the bug and this change is harmless.

hgohel:
> macOS, at least in the AIO bundle, isn't affected by the bug and this change is harmless.

Thanks for reviewing the change.

Nick-Hall (changes requested):
This works on Linux, but the `GRAMPSHOME` and `SAFEMODE` options are ignored if the new directory structure exists.

The `GRAMPSHOME` environment variable is used to specify a base directory for all data and configuration file storage.  The safe mode works by setting this to an empty directory.

hgohel:
@Nick-Hall Returning to this defect after a while. Discovered a related bug while trying out safe mode, which affects the testing of this one so I think it needs to be fixed first: [13300](https://gramps-project.org/bugs/view.php?id=13300). It would be very convenient if the root cause of that turns out to be the same as this one.

hgohel:
@Nick-Hall Turns out that this defect is a side-effect of [13300](https://gramps-project.org/bugs/view.php?id=13300), and the fix for that is in PR #1726 (gramps52) and PR #1725 (master). I'm closing this PR in favor of those.

PR #1709, opened 2024-04-21
Add Gramps ID option to Kinship Report

emyoulation:
prefix Gramps ID on each row  in Maintenance52 branch. But would be compatible with 5.1, 5.3 and master branches. gramps/plugins/textreport/kinshipreport.py

### DIFF
--- a/gramps/plugins/textreport/kinshipreport.py
+++ b/gramps/plugins/textreport/kinshipreport.py
@@ -6,6 +6,7 @@
 # Contribution  2009 by   Reinhard Mueller <reinhard.mueller@bytewise.at>
 # Copyright (C) 2010      Jakim Friant
 # Copyright (C) 2013-2014 Paul Franklin
+# Copyright (C) 2024      Brian McCullough
 #
 # This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


emyoulation:
Line 109 may be unnecessary once the configuration file is updated.
It sets a default inc_ids value for the new Options switch.   But it was doing weird complaints without that line.

emyoulation:
Created in response the Discourse thread:
https://gramps.discourse.group/t/adding-data-in-kinship-report/5287

emyoulation (commented):
Still needs Lint fixes

hgohel:
> Still needs Lint fixes

Done with commit a68f57c297753a85c2da10b64114d30774982e9c

emyoulation:
Could this be merged into the master now?

While doing support for the "[merging family Trees](https://gramps.discourse.group/t/merging-family-trees/6574/9)" Discourse thread, I discovered that none of the reports were usable for a sanity check.  This Kinship report was only useful for that if the IDs were included.
image

Nick-Hall:
Reased onto the _master_ branch.

PR #1710, opened 2024-04-21
Prevent hyperlink tagging being added as an undo action

Nick-Hall:
When the mouse hovers over a link in the note editor it is highlighted. This was stacked as an undoable action and as a result clicking the undo button sometimes had no noticeable effect.

Fixes [#13267](https://gramps-project.org/bugs/view.php?id=13267).


emyoulation:
Another bug related to dynamic styling of Note text (spellcheck underline) was identified in Discourse:

https://gramps.discourse.group/t/error-when-changing-font-size-in-notes/4724

Could it be related?

emyoulation:
The dynamic spellcheck styling also seems to make other edits Undoable.

emyoulation:
Undo in the Note Editor is still rendered unusable if Spellcheck is enabled.

Will file a bug report.

PR #1712, opened 2024-04-24
URL-Encode paths when creating URLs to pass to Gio.AppInfo.launch_default_for_uri.

jralls:
Paths submitted to `Gio.AppInfo.launch_default_for_uri` must be url-encoded to ensure that paths with spaces make correct URIs.


Nick-Hall:
Fixes bug [#13218](https://gramps-project.org/bugs/view.php?id=13218).

PR #1714, opened 2024-04-26
Remove the spaces padding the connector in a hyphenated surname

Nick-Hall:
This fixes the previous code which replaced " - " with "-" within the entire name.

Reverts 3c7d6ebce5d3164c99181254818fc2511b9d775e

Fixes [#13274](https://gramps-project.org/bugs/view.php?id=13274).


DaveSch-gramps:
Best of both worlds.

When the hyphen is the connector, no spaces. 

When the hyphen is part of the Given or a Surname, leave it as entered by the user.

DaveSch-gramps:
Worked great and I double checked all other name fields. Working as now expected.

PR #1717, opened 2024-05-04
Add mypy static type checker to CI and necessary fixes to the source code

DavidMStraub:
This adds mypy static type checks to CI. The PR contains the minimal set of changes to the code base needed to make mypy pass. Type hints are only added where needed by mypy, in most cases class/instance attributes initialized to empty lists or dicts. Some smaller bugs or inconsistencies discovered by mypy are also fixed.


gramps/cli/clidbman.py
```diff
@@ -113,7 +115,7 @@ class CLIDbManager:
     ICON_LOCK = 2
     ICON_OPEN = 3
 
-    ICON_MAP = {
+    ICON_MAP: Dict[int, Optional[str]] = {
```
QuLogic, line 118: I think you can use `from __future__ import annotations` and then use the simpler:
```suggestion
    ICON_MAP: dict[int, str | None] = {
```
so long as `mypy` is new enough. And then apply as necessary to most of the `typing` imports in other files.

DavidMStraub: That's correct but eventually it's a style decision as long as we have to support Python 3.8 or earlier. When we drop 3.8 support, we can switch to the new syntax without needed the future import.

QuLogic: True, it's a matter of style, but it's a lot easier to have it go in the new way now, and review that once, plus a simple removal of future imports, rather than review this now plus have to do some cleanup that touches everything all over again at some point in the future.

DavidMStraub: I don't mind changing it. Perhaps we will anyway require Python 3.9 for Gramps 5.3? Python 3.8 EOL is in 5 months from now.

DavidMStraub: Changed everywhere now!

gramps/plugins/gramplet/filter.py
```diff
@@ -52,7 +59,7 @@ class Filter(Gramplet):
     The base class for all filter gramplets.
     """
 
-    FILTER_CLASS = None
+    FILTER_CLASS: Any = None
```
QuLogic, line 62: I think this could be
```suggestion
    FILTER_CLASS: type[SidebarFilter] | None = None
```
It may even be a good idea to wrap in `ClassVar`.

gramps/gen/utils/configmanager.py
```diff
@@ -74,7 +75,7 @@ class ConfigManager:
     settings are stored.
     """
 
-    PLUGINS = {}
+    PLUGINS: dict[str, "ConfigManager"] = {}
```
QuLogic, line 78: I don't think the quotes are necessary with the future import.

gramps/plugins/lib/libhtml.py
```diff
@@ -353,7 +353,7 @@ def __init__(self, tag="html", *args, **keywargs):
                 self[len(self) :] = ["</%s>" % tag]  # add it on the end
 
     #
-    def __add(self, value):
+    def __add(self, value) -> "Html":
```
QuLogic, line 356: If you add the future import, the quotes shouldn't be necessary.

test/GrampsLogger/GtkHandler_Test.py
```diff
@@ -23,15 +23,16 @@
 import unittest
 import logging
 import sys
+
 from gi.repository import Gtk
 import os
```
QuLogic, line 28: This should probably move up, if you're splitting stdlib/third-party.

gramps/gen/datehandler/_datestrings.py
```diff
@@ -335,13 +335,13 @@ def __init__(self, locale):
     except AttributeError:
         localized_months = old_long
 
-    def print_po_snippet(en_loc_old_lists, context):
+    def print_po_snippet(en_loc_old_lists, context, line: int = 10000):
```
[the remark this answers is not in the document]
DavidMStraub: Yes, it does work the same way. I am pretty sure my slight refactor (needed for mypy checks to pass) does not affect the way it was implemented before. That being said, the function as such is quite a terrible way to do things, but my goal in this PR is to do minimal changes without affecting functionality.

gramps/plugins/gramplet/filter.py
DavidMStraub: I think this didn't work; anyway the goal here is not to *add* type hints consistenly yet, but just to make mypy pass to *enable* the addition of type hints from now on, so I'll leave this here for now.

DavidMStraub:
Thanks for your review @QuLogic. This is now ready for merge. Then I can work on adding type hints.

gramps/gen/datehandler/_datestrings.py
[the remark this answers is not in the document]
QuLogic: No, it does not; `print_po_snippet.line` is effectively global state, so at the end of the function, it will be increased by however number of times the `+= 1` is called, and remain that way for the next call. In the new code, the default value for parameter is _not_ saved anywhere, and begins at 10000 for every call.

gramps/gen/db/dummydb.py
```diff
@@ -149,23 +153,17 @@ class M_A_M_B(ABCMeta, MetaClass):
...
-        "NewBaseClass",
-        (
-            DbReadBase,
-            Callback,
-            object,
-        ),
-        {},
-    )
-):
+class DummyDbMeta(M_A_M_B):
```
QuLogic, line 156: Do you need the extra class? Can you not pass `M_A_M_B` as the metaclass directly?

DavidMStraub:
Thanks @QuLogic, I address the latest 2 comments and rebased on `master`. Would love to see this merged so I can start adding type hints.

Nick-Hall:
I'm planning a maintenance release this weekend.  There are a few bugs fixed and two new translations added.

After that I'll merge some of the PR s into _master_.

QuLogic:
This [comment](https://github.com/gramps-project/gramps/pull/1717#discussion_r1633977287) about `print_po_snippet` is not resolved, though it's marked as such in the GitHub interface.

DavidMStraub:
Right, now it should be resolved.

dsblank (approved):
Thank you! A lot of work here, and a good idea. Brings code base up to modern standards. Looks good to me.

gramps/gen/lib/test/grampstype_test.py
```diff
@@ -42,7 +43,7 @@ class GT0(GrampsType):
 # A migration utility might instantiate several of these with
 #   varying blacklist-specs
 class GT1(GT0):
-    _BLACKLIST = BLIST
+    _BLACKLIST: list[int] | None = BLIST
```
bartfeenstra, line 46: Instead of typing on `list` (and `dict`, both in general and not just on this particular line), can we instead type on `collections.Sequence`, `collections.MutableSequence`, `collections.Mapping`, and `collections.MutableMapping`? In my experience it makes things slightly less dependent on internals (one historical example was `dict` being ordered as an internal implementation detail until... 3.7?). Concrete benefits are also that return values can be marked immutable, reducing the risk of other code attempting to modify data that should be read-only.

gramps/gui/plug/report/_graphreportdialog.py
```diff
@@ -53,6 +56,7 @@
 from gramps.gen.config import config
 from gramps.gen.plug.report import CATEGORY_GRAPHVIZ
 from ._reportdialog import ReportDialog
+
 from ._papermenu import PaperFrame
```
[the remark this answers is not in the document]
bartfeenstra: Possibly added by an automatic formatter? There are other changes that solely alter the code's formatting.

gramps/plugins/gramplet/filter.py
bartfeenstra: @DavidMStraub Understood, but please be aware that changing `Any` to **literally anything else** is a BC break. Based on my experience, `Any` should be used sparingly (even when you think it's what you need, most of the time you actually need `type` or `object` instead).

gramps/plugins/webreport/common.py
```diff
@@ -646,7 +647,7 @@ def __init__(self, rlocale):
                     super().addLabels(loc)
 
 else:
-    AlphabeticIndex = localAlphabeticIndex
+    AlphabeticIndex = localAlphabeticIndex  # type: ignore
```
bartfeenstra, line 650: mypy allows a `# type: ignore[$reason-code]` syntax as well where `$reason-code` is the code (readable machine name) of the specific violation you want to ignore. Doing that, in combination with enabling `warn_unused_ignores` in your favorite mypy configuration file means that you won't end up accidentally ignoring the wrong things as the codebase changes over time.

DavidMStraub:
@bartfeenstra thanks for yet another review, but to reiterate, the sole purpose of this PR is to add type checks to the CI and add the minimal set of type hints that silence the type checker.

It is NOT a PR that adds consistent, complete, or meaningful type hints to the Gramps code base. So I'm not willing to spend more work on this. I will wait for it to be merged and then start adding type hints in follow-up PRs.

emyoulation:
When the Type Hints start being added, please add a line to that Pull Request/Discussion to maintain a path for people who have a vested interest in the implementation

DavidMStraub:
Now might be a good time to merge this :hugs:

Nick-Hall:
Rebased and fixed conflict.

gramps/gen/utils/alive.py
```diff
@@ -461,7 +462,7 @@ def ancestors_too_old(person, year):
                 return (None, None, "", None)
             self.pset.add(person.handle)
             LOG.debug(
-                "ancestors_too_old('%s', %s)".format(
+                "ancestors_too_old('{}', {})".format(
                     name_displayer.display(person), year
                 )
             )
```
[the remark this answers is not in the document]
Nick-Hall: After the merge of PR #1790 the debug messages now contain `%s` rather than `{}`.

gramps/gen/utils/alive.py
```diff
@@ -672,7 +673,7 @@ def probably_alive(
     if current_date is None:
         current_date = Today()
     LOG.debug(
```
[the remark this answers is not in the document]
Nick-Hall: Also resolved after #1790 was merged.

gramps/gen/datehandler/_datestrings.py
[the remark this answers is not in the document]
Nick-Hall: I'll check this code before merging, but I suspect that it isn't used anymore.

Nick-Hall:
Fixed annotation in GrampsType.

DavidMStraub:
Passing again

gramps/gen/lib/grampstype.py
```diff
@@ -24,12 +24,20 @@
...
+#
+# ---------------------------------------------------------------
+from __future__ import annotations
+from typing import Any
+
 # -------------------------------------------------------------------------
 #
 # Python modules
 #
 # -------------------------------------------------------------------------
```
stevenyoungs, line 40:
```suggestion

```
Remove duplicate comment block

gramps/gen/db/dummydb.py
```diff
@@ -115,7 +119,7 @@ def wrapped(*args, **keywargs):
 
 class MetaClass(type):
     """
-    transform class by wrapping it with a diagnostic wrapper (if __debig__ is
+    transform class by wrapping it with a diagnostic wrapper (if __debug__ is
     not set
```
stevenyoungs, line 123:
```suggestion
    not set)
```
Close parenthesis in comment

Nick-Hall: OK.  I added this by mistake when I fixed the conflict.

gramps/gen/lib/grampstype.py
Nick-Hall: OK. I added this by mistake when I fixed the conflict.

stevenyoungs:
As David mentions at the top, there are some bug fixes mixed into this PR. It will be hard to recall these when writing the ChangeLog. It's more work, but is it worth separating these bug fixes out?
The whitespace changes, already mentioned by another reviewer, also make it harder to see the substantive changes. Probably not worth splitting out at this stage though.

Nick-Hall:
> As David mentions at the top, there are some bug fixes mixed into this PR. It will be hard to recall these when writing the ChangeLog. It's more work, but is it worth separating these bug fixes out?

Yes, they should probably be separated out.

DavidMStraub:
Honestly, I find it very hard to accept to ask separating something 9 months after a PR has been raised, and around like half a year after people have reviewed it. If this is a requirement, why not raise it immediately?

Also, I don't think the request makes much sense as this PR is about solving small inconsistencies to make the type checker happy, so what is considered a bug and what is just a style issue is not always clear. So there is no clear separation line.

dsblank:
In my opinion, it only makes sense to separate bug fixes from the rest if there is a chance that we'd need to unmerge one or the other. But there is no reason to do that. We are a small team, and being too pedantic is not useful if we drive away developers.

gramps/gui/plug/report/_graphreportdialog.py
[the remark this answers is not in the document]
Nick-Hall: OK. Removed.

gramps/gen/lib/grampstype.py
Nick-Hall: Done.

gramps/gen/db/dummydb.py
Nick-Hall: Added.

gramps/gen/lib/test/grampstype_test.py
Nick-Hall: Thanks.  We can consider this later.

gramps/gen/datehandler/_datestrings.py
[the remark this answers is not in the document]
Nick-Hall: This code is no longer used, but we can remove it later.

gramps/plugins/webreport/common.py
Nick-Hall: Added reason codes.

dsblank:
Congrats Gramps! This is the first step in a new level of code quality. Thanks @DavidMStraub!

PR #1718, opened 2024-05-04
Fix error when changing the style of misspelt words

Nick-Hall:
In the note editor, changing the style of text failing the spell check caused an error.  The style tag used by the spell checker to indicate spelling errors has no tag name.

Fixes [#13282](https://gramps-project.org/bugs/view.php?id=13282).

### DIFF
--- a/gramps/gui/widgets/styledtextbuffer.py
+++ b/gramps/gui/widgets/styledtextbuffer.py
@@ -492,7 +492,7 @@ class StyledTextBuffer(UndoableBuffer):
         start, end = self._get_selection()
         tags = self._get_tag_from_range(start.get_offset(), end.get_offset())
         for tag_name, tag_data in tags.items():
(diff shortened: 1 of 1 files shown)


emyoulation:
Do I need to do anything special to test this patch with the Undo path done for the rollover of URLs in Notes?

Nick-Hall:
This is a simple one line fix and is unrelated to any other fix.

emyoulation:
OK.  But I'll apply the patch in https://github.com/gramps-project/gramps/pull/1710 and see if the Note Editor Undo button starts working with the Spell Checker.

Have to set up the Windows box again. This Fedora argues any time I try to patch the application.

emyoulation:
Thanks for the quick fix to styling with Spellcheck! 

The Undo issue is still a (less critical) problem. Will file on PR1710 since it is more closely related.

PR #1720, opened 2024-05-07
Check that pdf output file exists in genealogy tree reports

Nick-Hall:
Fixes [#10696](https://gramps-project.org/bugs/view.php?id=10696).

### DIFF
--- a/gramps/gen/plug/docgen/treedoc.py
+++ b/gramps/gen/plug/docgen/treedoc.py
@@ -131,7 +131,6 @@ _NOTESIZE = [
 
 if win():
     _LATEX_FOUND = search_for("lualatex.exe")
-    DETACHED_PROCESS = 8
 else:
     _LATEX_FOUND = search_for("lualatex")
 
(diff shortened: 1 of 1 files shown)


ennoborg:
We may get more information by processing the return value of the communicate call, like this:

outs, errs = proc.communicate()

The variables shown here store the stdout and stderr provided by the process, and may provide more information when the process runs out of memory, or when an incomplete install leads to missing styles.

Nick-Hall:
The process outputs a huge amount of information.

For debugging, it is best to create a `tex` file and then run _lualatex_ manually, then you can examine the log files.

gramps/gen/plug/docgen/treedoc.py
```diff
@@ -815,7 +816,14 @@ def close(self):
             else:
                 proc = Popen(args, stdin=PIPE, stdout=PIPE, stderr=PIPE)
```
hgohel, line 817: The current change isn't enough to fully fix [10696](https://gramps-project.org/bugs/view.php?id=10696). By changing the call to Popen() in Windows to match the other platforms, report generation works on Windows.

Nick-Hall: Thanks.  I have changed this, but cannot test it because I don't have a copy of Windows.

hgohel: @Nick-Hall I'll be able to test the change on Windows 10 and 11 in a couple of days and report results.

hgohel: @Nick-Hall Tested on both Win 11 and 10 and unfortunately report generation continues to fail. When this issue was being discussed on the forums, the local change I had did seem to resolve it. Will need experiment some more.

It would be helpful if anyone else could test on Windows to confirm.

PR #1721, opened 2024-05-11
Enhanced version of the Filter gramplet

kkujansuu:
This is an enhanced version of the Filter gramplet. The enhancements are

1. The elapsed time is displayed.
2. A new 'Define filter' button that opens a filter editor dialog and automatically populates it with the rules that correspond to values given in the gramplet.
3. The person filter has two new fields: Birth place and Death place
4. The default confidence value in the Citation filter is 'Very low'.
5. Place filter: An error message is displayed if the user fills in the 'Within' value but no place is active

### DIFF
--- a/gramps/gen/filters/rules/event/__init__.py
+++ b/gramps/gen/filters/rules/event/__init__.py
@@ -24,15 +24,13 @@
 Package providing filter rules for Gramps.
 """
 
-from .._haseventbase import HasEventBase as HasEvent
-from .._haseventbase import HasEventBase
-
 from ._hastype import HasType
 from ._allevents import AllEvents
 from ._hasgallery import HasGallery
 from ._hasidof import HasIdOf
 from ._regexpidof import RegExpIdOf
 from ._hascitation import HasCitation
+from ._hasevent import HasEvent
 from ._hasnote import HasNote
 from ._hasnoteregexp import HasNoteRegexp
 from ._hasnotematchingsubstringof import HasNoteMatchingSubstringOf
--- a/gramps/gen/filters/rules/event/__init__.py
+++ b/gramps/gen/filters/rules/event/__init__.py
@@ -25,6 +25,7 @@ Package providing filter rules for Gramps.
 """
 
 from .._haseventbase import HasEventBase as HasEvent
+from .._haseventbase import HasEventBase
 
 from ._hastype import HasType
 from ._allevents import AllEvents
--- a/gramps/gui/filters/sidebar/_citationsidebarfilter.py
+++ b/gramps/gui/filters/sidebar/_citationsidebarfilter.py
@@ -78,7 +78,7 @@ class CitationSidebarFilter(SidebarFilter):
         for conf_value in sorted(conf_strings.keys()):
             model.append((_(conf_strings[conf_value]),))
         self.filter_conf.set_model(model)
-        self.filter_conf.set_active(Citation.CONF_NORMAL)
+        self.filter_conf.set_active(Citation.CONF_VERY_LOW)
 
         self.filter_note = Gtk.Entry()
 
--- a/gramps/gui/filters/sidebar/_sidebarfilter.py
+++ b/gramps/gui/filters/sidebar/_sidebarfilter.py
@@ -167,6 +167,7 @@ class SidebarFilter(DbGUIElement):
         t2 = time.perf_counter()
         msg = _("Elapsed time: %.2fs") % (t2 - t1)
         self.msg_label.set_text(msg)
+        self.msg_label.get_style_context().remove_class("error")
         self.uistate.set_busy_cursor(False)
 
     def clicked_func(self):
@@ -373,7 +374,8 @@ class SidebarFilter(DbGUIElement):
             if isinstance(rule, gramps.gen.filters.rules.place.WithinArea):
                 if rule.list[0] is None:  # No active place
                     msg = _("You should select a place when using the 'Within' rule")
-                    self.msg_label.set_markup("<span color='red'>" + msg + "</span>")
+                    self.msg_label.set_text(msg)
+                    self.msg_label.get_style_context().add_class("error")
                     return False
         return True
 
@@ -381,9 +383,8 @@ class SidebarFilter(DbGUIElement):
         self.filterdb = gramps.gen.filters.CustomFilters
         the_filter = self.get_filter()
         if the_filter is None:
-            self.msg_label.set_markup(
-                "<span color='red'>" + _("Supply at least one value") + "</span>"
-            )
+            self.msg_label.set_text(_("Supply at least one value"))
+            self.msg_label.get_style_context().add_class("error")
             return
 
         if not self.filter_is_ok():
--- a/gramps/gui/filters/sidebar/_sidebarfilter.py
+++ b/gramps/gui/filters/sidebar/_sidebarfilter.py
@@ -398,8 +398,8 @@ class SidebarFilter(DbGUIElement):
                 rule.list[2] = str(rule.list[2])
             new_rules.append(rule)
         the_filter.flist = new_rules
-        comment = "Created by Filter gramplet on " + time.strftime(
-            "%Y-%m-%d", time.localtime(time.time())
+        comment = _("Created by Filter gramplet on {today}").format(
+            today=time.strftime("%Y-%m-%d", time.localtime())
         )
         the_filter.set_comment(comment)
 
--- a/gramps/gui/filters/sidebar/_sidebarfilter.py
+++ b/gramps/gui/filters/sidebar/_sidebarfilter.py
@@ -19,7 +19,21 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 
+# -------------------------------------------------------------------------
+#
+# Python modules
+#
+# -------------------------------------------------------------------------
+import time
+
+# -------------------------------------------------------------------------
+#
+# Gramps modules
+#
+# -------------------------------------------------------------------------
+
 from gramps.gen.const import GRAMPS_LOCALE as glocale
+from gramps.gen.filters import reload_custom_filters
 
 _ = glocale.translation.gettext
 from bisect import insort_left
@@ -31,6 +45,11 @@ from ... import widgets
 from ...dbguielement import DbGUIElement
 from gramps.gen.config import config
 from ...utils import no_match_primary_mask
+
+from ...editors import EditFilter
+import gramps.gen.filters
+
+# import gramps.gen.filters.rules.place
 
 _RETURN = Gdk.keyval_from_name("Return")
 _KP_ENTER = Gdk.keyval_from_name("KP_Enter")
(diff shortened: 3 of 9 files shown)


emyoulation:
Very nice!

You have one of the few functional aborts in Gramps for your Filter Params test feature.

I cannot count the number of times that the Filter progress dialog has appeared and I have wished for an abort. (Usually because I had set a parameter incorrectly.)


Could that test abort routine be made compatible with the Filter Progress dialog?

It would probably have to abort to "unfiltered"... would that intuitive enough?  I

kkujansuu:
> Could that test abort routine be made compatible with the Filter Progress
dialog?

Not easily.

la 11. toukok. 2024 klo 21.07 Emyoulation or BAMaustin (
***@***.***) kirjoitti:

emyoulation:
> Not easily. 

Understood and thanks for looking... and for not being aggravated by the extra ask.

Looks like the PR test fails are formatting compliance. I ran into similar fails for the last PR that I tried. And they would have had the same fails for the module _before_ my changes.

Maybe there's a way to start re-submit the original modules for a compliance check with the new requirements. Although the new proposal (for Type Hints and static type checking) might make that necessary again in the near future.

https://sourceforge.net/p/gramps/mailman/gramps-devel/thread/46818942-927a-463d-ba3c-4d04de3aaf41%40gramps-project.org/#msg58766730

gramps/gui/filters/sidebar/_sidebarfilter.py
```diff
@@ -316,3 +363,47 @@ def set_filters_to_name(self, filter_name):
...
+
+    def filter_is_ok(self):
+        the_filter = self.get_filter()
+        if the_filter is None:
+            return True
+        for rule in the_filter.get_rules():
+            if isinstance(rule, gramps.gen.filters.rules.place.WithinArea):
+                if rule.list[0] is None: # No active place
+                    msg = _("You should select a place when using the 'Within' rule")
+                    self.msg_label.set_markup("<span color='red'>" + msg + "</span>")
```
Nick-Hall, line 375: It is better to add the css error class to the widget style context rather than hard-coding the colour "red".   For example you could use:
```
self.msg_label.set_text(msg)
self.msg_label.get_style_context().add_class("error")
```
This will use the error colour defined in the theme.  This is useful for visually impaired users and non-European cultures.

gramps/gui/filters/sidebar/_sidebarfilter.py
```diff
@@ -316,3 +363,47 @@ def set_filters_to_name(self, filter_name):
...
+                    return False
+        return True
+    
+    def define_filter(self, _obj):
+        self.filterdb = gramps.gen.filters.CustomFilters
+        the_filter = self.get_filter()
+        if the_filter is None:
+            self.msg_label.set_markup("<span color='red'>" + _("Supply at least one value") + "</span>")
```
Nick-Hall, line 383: As above.

gramps/gen/filters/rules/event/__init__.py
```diff
@@ -71,4 +72,5 @@
     HasTag,
     MatchesPlaceFilter,
     HasDayOfWeek,
+    HasEventBase,
```
Nick-Hall, line 75: How is `HasEventBase` and `HasSourceBase` related to this PR?  I can't see them referenced anywhere.

Nick-Hall (changes requested):
This looks good.  I just made a few suggestions.

To get rid of the formatting errors, run [black](https://black.readthedocs.io/en/stable/index.html) on the files you changed.

gramps/gui/filters/sidebar/_citationsidebarfilter.py
```diff
@@ -78,7 +78,7 @@ def __init__(self, dbstate, uistate, clicked):
         for conf_value in sorted(conf_strings.keys()):
             model.append((_(conf_strings[conf_value]),))
         self.filter_conf.set_model(model)
-        self.filter_conf.set_active(Citation.CONF_NORMAL)
+        self.filter_conf.set_active(Citation.CONF_VERY_LOW)
```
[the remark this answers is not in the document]
emyoulation: With Kari's PR, we can cancel/close both PRs 1700 and 1701.

Also, I do not believe that relabeling the Confidences to harmonize with 5.5.1 GEDCOM confidence labels would be a good choice at this point. GEDCOM7 is discussing the addition of negative Confidence options (disproven & similar levels). So it would be better to do a single change that encompasses both labeling and scope changes.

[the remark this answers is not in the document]
Nick-Hall: OK.  I'll close #1700 and #1701.

We can leave the confidence level labels for now.  The Gedcom text is already in the tooltips.

gramps/gui/filters/sidebar/_sidebarfilter.py
kkujansuu: I don't quite understand how css and themes work in Gramps. If I add the "error" class then the label still doesn't change color.

gramps/gen/filters/rules/event/__init__.py
kkujansuu: The 'Define filter' feature does not work without it.

Event filtering uses the rule class HasEventBase and 'Define filter' will store that class name in custom_filters.xml. For example:

  <object type="Event">
    <filter name="test" function="and" comment="Created by Filter gramplet on 2024-05-12">
      <rule class="HasEventBase" use_regex="False" use_case="False">
        ...
      </rule>
  </object>

However, when custom_filters.xml is read and interpreted then the corresponding rule is silently ignored because the class is not imported in gramps.gen.filters.rules.event.__init__.py. There is only this message on the console:

    ERROR: Filter rule 'HasEventBase' in filter 'test' not found!

My solution is to import the class in the __init__.py file. At first i thought it also needed to be added in 'editor_rule_list' but that does not seem to be necessary.

emyoulation:
Kari,  

Could you add your 2024 copyright to the list of contributors for these 5 files?  It isn't just about taking credit. It also agrees to the terms of the licensing for your contribution. And it also builds confidence by showing that the modules are continuing to evolve.   Most of these show 2010 as the last touch.

Cross-link to Discourse forum discussion:
https://gramps.discourse.group/t/can-add-on-gramplets-use-the-lib-plugin-type-to-supersede-built-ins/5422/10

MantisBT request [0013289](https://gramps-project.org/bugs/view.php?id=13289): Roll Filter+ gramplet plugins into the core (built-in) plugins for 5.3

gramps/gui/filters/sidebar/_sidebarfilter.py
Nick-Hall: Interesting.  I've pushed a commit to your branch which works for me.  The errors are red, but the elapsed time is black.

gramps/gui/filters/sidebar/_sidebarfilter.py
```diff
@@ -316,3 +363,47 @@ def set_filters_to_name(self, filter_name):
...
+        # fix some rules:
+        new_rules = []
+        for rule in the_filter.get_rules():
+            # The Place rule WithinArea might have numeric values while custom_filters.xml only accepts strings.
+            if isinstance(rule, gramps.gen.filters.rules.place.WithinArea):
+                rule.list[1] = str(rule.list[1])
+                rule.list[2] = str(rule.list[2])
+            new_rules.append(rule)
+        the_filter.flist = new_rules
+        comment = "Created by Filter gramplet on " + time.strftime("%Y-%m-%d", time.localtime(time.time()))
```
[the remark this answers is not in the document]
Nick-Hall: I've pushed a commit to fix this.  I think we can leave the date in ISO format.

gramps/gui/filters/sidebar/_sidebarfilter.py
kkujansuu: I can make the css style work if I manually make changes in the file ~/.config/gtk-3.0/gtk.css. But that is probably not the correct the way to do this, is it?

kkujansuu: Anyway, it seems that we are doing changes simultaneously. Did I mess things up?

Nick-Hall: What operating system and theme are you using?  No changes to the css should be necessary.

When you run Gramps in the _master_ branch, what colour is the version number in the status bar?  That is set using the "destructive-action" class.

You can either use my changes or update the code yourself.

gramps/gen/filters/rules/event/__init__.py
Nick-Hall: I don't see any warnings or errors if I comment these lines out.

The `HasEventBase` is used as the "Events with parameters" rule.  It works if I click the 'Define filter' button in the sidebar, or if I invoke the filter editor from the menu.

kkujansuu:
I am using Linux Mint and I don't know what theme I am using! How to check
that?



su 12. toukok. 2024 klo 20.56 Nick Hall ***@***.***)
kirjoitti:

kkujansuu:
Yes,  it is sufficient to have HasEventBase imported in the module. I
noticed that also.

su 12. toukok. 2024 klo 20.59 Nick Hall ***@***.***)
kirjoitti:

Nick-Hall:
> I am using Linux Mint and I don't know what theme I am using! How to check that?

From the command line, start Gramps with:

```
GTK_DEBUG=interactive python Gramps.py
```

Then, in the debugger, look in the "Visual" tab.

kkujansuu:
GTK Theme = Mint-Y

su 12. toukok. 2024 klo 23.14 Nick Hall ***@***.***)
kirjoitti:

kkujansuu:
The version number is red.

If I add the "destructive-action" class to the Reset button then it turns
red. But using it on the label does not work. The "error" class does not
work for the button either.

su 12. toukok. 2024 klo 20.56 Nick Hall ***@***.***)
kirjoitti:

Nick-Hall:
> GTK Theme = Mint-Y

The Mint-Y theme doesn't seem to define the "error" class for labels, although the Mint-X theme does. The "destructive-action" class is only used by buttons.

I suggest that we add the following line to our `gramps.css` file:
```css
label.error {color: @error_color}
```
This will let us use classes as recommended.  I expect that the @error_color will be used by themes anyway, so overriding it in this way shouldn't cause a problem.

emyoulation:
Have you tried the GTK+ Inspector tool?  It was very enlightening when exploring what CSS was applied to each element of the Gramps GUI.

For Linux only. Wish there were Windows and macOS versions.

Nick-Hall:
> Have you tried the GTK+ Inspector tool?

What happens if you try to start it in Windows?

emyoulation:
>> Have you tried the GTK+ Inspector tool?
> What happens if you try to start it in Windows?

I only found GTK+ Inspector in the Gnome project. Haven't found info about installation for Windows or macOS

jralls:
The inspector is part of Gtk. Just start Gramps from the command line with e.g. `GTK_DEBUG=interactive gramps` and it pops up behind the gramps window. On Windows you probably have to do it in two lines, e.g.
```
set GTK_DEBUG interactive
gramps
```
There are also two keyboard shortcuts, `<ctrl><shift>I` and `<ctrl><shift>D`, to open it when `$GTK_DEBUG` isn't set. They don't work on macOS but might on Windows.

kkujansuu:
I am travelling this week and can't work on this. Please proceed with the
PR if you wish. Anyway, I don't know what I should do next.

la 11. toukok. 2024 klo 23.11 Nick Hall ***@***.***)
kirjoitti:

Nick-Hall:
@kkujansuu I'll finish this for you.

emyoulation:
> There are also two keyboard shortcuts, `<ctrl><shift>I` and `<ctrl><shift>D`, to open it when `$GTK_DEBUG` isn't set. They don't work on macOS but might on Windows.

They keybindings don't work on Fedora.  (`<ctrl><shift>D` is the keybinding for Organize Bookmarks in Gramps)

Nick-Hall (approved):

giotodibondone:
Updating documentation 
https://gramps-project.org/wiki/index.php/Gramps_6.0_Wiki_Manual_-_Gramplets#Filter

PR #1723, opened 2024-05-17
Fix some Glade deprecations

QuLogic:
`margin-{left,right}` have been replaced by `margin-{start,end}`, and `xalign` with `halign`. In some cases, both were specified, so I deleted the deprecated one.


QuLogic:
I rebased this, and also added removal of stock items from the Glade files.

PR #1724, opened 2024-05-28
Some short french month names can not be entered.

SNoiraud:
Fixes [#13307](https://gramps-project.org/bugs/view.php?id=13307).

Short months like 7bre (september), 8bre (october), 9bre (november) and xbre (december) cannot be entered as valid date with french language.


emyoulation:
Shouldn't this be a 5.3 milestone?

PR #1726, opened 2024-05-29
Safe mode working on Windows

hgohel:
Safe mode which was broken with Gramps 5.2.0-5.2.2 on Windows, has been fixed with this commit. Fixes [#13300](https://gramps-project.org/bugs/view.php?id=13300)

Existing code was looking for user configuration in APPDATA, but with the move to XDG-based directory layout (bug 8025), this is no longer correct (see [doc](https://docs.gtk.org/glib/func.get_user_data_dir.html)). With the fix code looks for the new layout first, and then APPDATA.

The modified code is probably executed only on Windows, but it would be good to get confirmation for Linux and MacOS.

- [X] Testing on Windows with Gramps 5.2.2 only, and side-by-side with Gramps 5.1.6.
- [ ] Testing on Linux, if affected
- [ ] Testing on MacOS, if affected

Test to verify that [safe mode](https://gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Command_Line#Safe_mode) functions as described, i.e. starting from a user's customized environment, verify that existing family trees are available in safe mode but settings and configuration are not.

NOTE: This also fixes [#13261](https://gramps-project.org/bugs/view.php?id=13261) which documents another way in which this bug manifests. However, this is bug is closer to the root cause and easier to test.

### DIFF
--- a/gramps/gen/const.py
+++ b/gramps/gen/const.py
@@ -101,10 +101,11 @@ elif "USERPROFILE" in os.environ:
     USER_HOME = get_env_var("USERPROFILE")
     HOME_DIR = os.path.join(GLib.get_user_config_dir(), "gramps")
     if not os.path.exists(HOME_DIR):
-        if "APPDATA" in os.environ:
-            HOME_DIR = os.path.join(get_env_var("APPDATA"), "gramps")
-        else:
-            HOME_DIR = os.path.join(USER_HOME, "gramps")
+        homedir = get_env_var("APPDATA") if "APPDATA" in os.environ else USER_HOME
+        homedir = os.path.join(homedir, "gramps")
+        dbpath = os.path.join(homedir, "grampsdb")
+        if os.path.isdir(dbpath) and os.listdir(dbpath):
+            HOME_DIR = homedir
 else:
     USER_HOME = get_env_var("HOME")
     HOME_DIR = os.path.join(USER_HOME, ".gramps")
--- a/gramps/gen/const.py
+++ b/gramps/gen/const.py
@@ -99,10 +99,12 @@ if "GRAMPSHOME" in os.environ:
     HOME_DIR = os.path.join(USER_HOME, "gramps")
 elif "USERPROFILE" in os.environ:
     USER_HOME = get_env_var("USERPROFILE")
-    if "APPDATA" in os.environ:
-        HOME_DIR = os.path.join(get_env_var("APPDATA"), "gramps")
-    else:
-        HOME_DIR = os.path.join(USER_HOME, "gramps")
+    HOME_DIR = os.path.join(GLib.get_user_config_dir(), "gramps")
+    if not os.path.exists(HOME_DIR):
+        if "APPDATA" in os.environ:
+            HOME_DIR = os.path.join(get_env_var("APPDATA"), "gramps")
+        else:
+            HOME_DIR = os.path.join(USER_HOME, "gramps")
 else:
     USER_HOME = get_env_var("HOME")
     HOME_DIR = os.path.join(USER_HOME, ".gramps")


hgohel:
This PR is for Gramps52 branch.

hgohel:
@Nick-Hall Any comments on the approach taken by this fix?

emyoulation:
Just a question, is it possible that the breaking of Safe Mode (due to the XDG support change) is also related to Debug/optimise mode always being enabled?

https://gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Tools#Debug

hgohel:
@Nick-Hall 
> We need to use the old directory structure if the `GRAMPSHOME` environment variables is set or if we are starting in safe mode. Safe mode just sets the `GRAMPSHOME` to a temporary directory.
> 
> If the XDG base directory structure is present we should use it.

This PR doesn't change behavior when GRAMPSHOME is set, or when SAFEMODE is invoked. I think this should satisfy requirement in para 1.

If it's a normal startup and the XDG base directory is present, it will be used. This should satisfy para 2.

> We should only use the old directory structure if a database in the old location exists and the new structure in not present. This differs from the existing code where the old structure is always used if present.

Since this was a regression bug fix for safe mode behavior, the last requirement is not met by this PR yet. Lmk if we want to include checking for existence of entries under grampsdb directories in this PR. I'll add PR comments where I think those need to be made.

Testing this issue is tedious and coming back to it after months even more so, so I want to wrap it soon. Thanks.

gramps/gen/const.py
```diff
@@ -99,10 +99,12 @@
     HOME_DIR = os.path.join(USER_HOME, "gramps")
 elif "USERPROFILE" in os.environ:
     USER_HOME = get_env_var("USERPROFILE")
-    if "APPDATA" in os.environ:
-        HOME_DIR = os.path.join(get_env_var("APPDATA"), "gramps")
-    else:
-        HOME_DIR = os.path.join(USER_HOME, "gramps")
+    HOME_DIR = os.path.join(GLib.get_user_config_dir(), "gramps")
+    if not os.path.exists(HOME_DIR):
+        if "APPDATA" in os.environ:
```
hgohel, line 104: @Nick-Hall To use the existing configuration only if there are any databases present, I would add a check for contents under grampsdb in both cases of this `if`, and modify `HOME_DIR` only if there are any contents present (minimal check for any entries in that directory, no validation that they are actual, valid databases). If there are no databases, `HOME_DIR` would be unchanged from the setting in L102 above.

This should achieve the intent you mentioned.

hgohel: Implemented in the commit 8648ecf279d0e21f750d0ceeea46abb47a21788e

gramps/gen/const.py
```diff
@@ -99,10 +99,13 @@
     HOME_DIR = os.path.join(USER_HOME, "gramps")
 elif "USERPROFILE" in os.environ:
```
stevenyoungs, line 100: The code relies on the existence of an environment variable to infer the platform and run platform specific code.
Is it better to directly query the platform using `sys.platform == "win32"` rather than inferring it? This would cover the scenario of a Linux or mac user not having `GRAMPSHOME` set and having `USERPROFILE` set. Maybe unlikely, but it won't be easy to debug if it does occur.

This might be better handled in a separate PR as it is outside the scope of this PR

hgohel: @stevenyoungs Thanks for the review. It does appear that the USERPROFILE block is intended only for Windows, so your suggestion would make it explicit. OTOH, I don't know if Linux/Mac users setting USERPROFILE is supposed to be a supported case. Agree with creating another PR if its decided to change this.

Nick-Hall: > I don't know if Linux/Mac users setting USERPROFILE is supposed to be a supported case.

No. The `USERPROFILE` environment variable is Windows only.

hgohel:
New PR created for master branch #1881

PR #1727, opened 2024-06-11
Avoid empty media section in narrative web report

SvenWe:
When creating a WebReport ("Reports" → "Web Pages" → "Narrated Web Site") all pages for individuals with any photos or any other media entry include a "Other media: videos, pdfs..." even if there are only images, which are not in this section.
Example:
```
<div id="toggle_media" style="display:block">
  <div class="MediaRow">…</div>
  <div class="ModalClass" id="MediaModal">…</div>
  <h3>Other media: videos, pdfs...</h3>
    <div id="medias">
    </div>
</div>
```

Changes:
- Don't show "Other media: videos, pdfs..." section if there are no non-image media entries.

### DIFF
--- a/gramps/plugins/webreport/basepage.py
+++ b/gramps/plugins/webreport/basepage.py
@@ -2582,11 +2582,10 @@ class BasePage:
                 if lightboxes_1:
                     toggle += lightboxes_1
                     toggle += lightboxes_2
-                    if lightbox < len(photolist):
-                        toggle += Html(
-                            "h3", self._("Other media: videos, pdfs..."), inline=True
-                        )
-                if medias:
+                if medias.inside:
+                    toggle += Html(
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Is there a related bug report for this PR?

@SNoiraud Are you happy with this change?

emyoulation:
Is it related to
https://gramps-project.org/bugs/view.php?id=13272

0013272: "Narrated Web Site" only shows the media from the "main gallery"

Nick-Hall:
@SNoiraud Should I merge this PR?

Nick-Hall:
@SNoiraud Is this fix acceptable?  We should either merge this or reject the change and close the PR.

SNoiraud:
I don't see the problem.
When we have only images, it works correctly. If we have pdfs, it works correctly too.
The section "Other media: videos, pdfs..." is already only visible when we have other media.

PR #1738, opened 2024-07-12
Add missing tooltips in the Addon Manager

emyoulation:
Add tooltips to 7 items in the Addon Manager  

0013358: [Addon Manager] tooltips needed for unlabeled Projects management buttons  
https://gramps-project.org/bugs/view.php?id=13358

### DIFF
--- a/gramps/gui/plug/_windows.py
+++ b/gramps/gui/plug/_windows.py
@@ -460,6 +460,7 @@ class AddonManager(ManagedWindow):
         hbox.pack_start(self.status_combo, False, False, 0)
 
         clear = Gtk.Button.new_from_icon_name("edit-clear", Gtk.IconSize.BUTTON)
+        clear.set_tooltip_text(_("Reset addon filters to defaults"))
         clear.connect("clicked", self.__clear_filters)
         hbox.pack_start(clear, False, False, 0)
 
@@ -699,10 +700,15 @@ class AddonManager(ManagedWindow):
 
         hbox = Gtk.Box()
         add_btn = SimpleButton("list-add", self.__add_project)
+        add_btn.set_tooltip_text(_("Insert a project"))
         del_btn = SimpleButton("list-remove", self.__remove_project)
+        del_btn.set_tooltip_text(_("Remove the selected project"))
         up_btn = SimpleButton("go-up", self.__move_up)
+        up_btn.set_tooltip_text(_("Move project upwards in the list"))
         down_btn = SimpleButton("go-down", self.__move_down)
+        down_btn.set_tooltip_text(_("Move project downwards in the list"))
         restore_btn = SimpleButton("document-revert", self.__restore_defaults)
+        restore_btn.set_tooltip_text(_("Revert to the project defaults"))
         self.buttons = [add_btn, del_btn, up_btn, down_btn, restore_btn]
         for button in self.buttons:
             hbox.pack_start(button, False, False, 0)
@@ -740,6 +746,9 @@ class AddonManager(ManagedWindow):
         row = 1
         install = Gtk.CheckButton()
         install.set_label(_("Allow Gramps to install required Python modules"))
+        install.set_tooltip_text(
+            _("Where possible, try to install any missing prerequisite Python modules")
+        )
         active = config.get("behavior.addons-allow-install")
         install.set_active(active)
         install.connect("toggled", self.install_changed)


emyoulation (commented):
Black found trailing whitespace on added lines

gramps/gui/plug/_windows.py
```diff
@@ -460,6 +460,7 @@ def __init__(self, dbstate, uistate, track):
         hbox.pack_start(self.status_combo, False, False, 0)
 
         clear = Gtk.Button.new_from_icon_name("edit-clear", Gtk.IconSize.BUTTON)
+        clear.set_tooltip_text(_("Reset add-on filters to defaults"))
```
wroldwiedbwe, line 463: You might want to use the dominant spelling of "add-on" used by the Gramps project eg:  "addon" without the hyphen this will help the translators!  Even your PR title shows the correct spelling eg "Update _windows.py with 7 tooltips additions to **Addon** Manager"

emyoulation: I am resigned to the idea that the following is a losing argument.  It's  just a symptom of my disagreement about not using the hyphenated variant in sentences. The terms "add-on" and "plug-in" are better dissected / parsed by automated translation.    For English-as-a-second-language readers, it would be too easy to sound it out as ad-dons instead of add-ons 

As to the Addon Manager... that is the spelling in the GUI.  So there's no arguing a branding.

Sigh.... I'll change it.

wroldwiedbwe: Thanks 👍

Unfortunately that's how it goes with projects that have been around a long time :)  

Checked the Gramps documentation website and the alternate spelling seems to all be from somebody called Bamaustin ?

emyoulation: Just want to leave a note about this.  

 "addon", "plugin", and "gramplet" are Gramps branding bastardizations of the industry standard terms "add-on", "plug-in" and "applet".

So anchored cross-references from standard terms to "introductions" of the branded bastardizations need to exist in the wiki and Gramps Glossary.

Nick-Hall: The terms "addon", "add-on", "addin", "add-in", "plugin", "plug-in" and "extension" are largely synonyms.  Their exact meaning depends on the type of application in which they are used.

The term "gramplet" is purely a Gramps invention used to describe a certain category of plugin.

The use of the hyphenated forms is on the decline, but is still widely used in "add-on", but "plug-in" is no longer common.  We see a similar trend with "web-page" becoming "webpage".

emyoulation: The "plugin" _**does**_ seem to have overtaken "plug-in" in the common vernacular although Oxford English Dictionary (OED) still lists "plug-in" as the correct form. But I think that is because they are trying to distance the software term from metaphor for which it was named... since real world things are regularly "plugged into" computers. (Similar to the way I often use Windoze for the more rarely referenced MS Windows. Because otherwise, I'd have to always indicate OS/MS or GUI in every use of the word "window".) Still, there is arguable validity in this usage.

However, addon is still deeply in the minority. And "buitin" is rare. (Although it IS frequent in the GNU.org docs.)

My primary objection is due to automated translation. Particularly in the technologies outputting speech. Hyphenated is more likely to translate the words separately rather than spew English solid compounds. This loses the metaphor's leveraging of existing knowledge and experiences to facilitate understanding...  and instead makes it a memorization exercise. And documentation is primarily for the uninitiated audience.  

Counter to my stance those hyphenated examples, I will continue to argue that "GrampsWeb" compounding is more appropriate than "Gramps Web" because that CANNOT be a metaphor. "Gramps" is already a branding and so requires a memorization, albeit a simple one. **MOREOVER** both "Gramps" and "web" are _SO_ pervasive in our documentation that the 2 word form becomes unsearchable. 

Further, while you've branded "Edit -> Addon Manager" in an apparent attempt to associate it with the "Help -> Plugin Manager" feature, that "Addon" branding is a false economy. Similarly to the "Edit -> Preferences" versus "View -> Configure" subtleties, the separation of closely such related features amplifies Gramps' reputation for being unnecessarily complex.

Nick-Hall: This has been discussed previously and we agreed on the term "Addon".  Feel free to raise this topic again.  As long as we are consistent I don't really mind.

emyoulation:
A GitHub workflow question...
Do I delete this branch now that the commit has been force merged?

Nick-Hall:
Yes.  You should delete the branch.

PR #1741, opened 2024-07-12
Update aboutdialog.py fix linkrot

emyoulation:
fix https://gramps-project.org/bugs/view.php?id=13152

Original:
http://ch.tudelft.nl/~arthur/svn2cl/

Wayback Machine:
https://web.archive.org/web/20070323201820/http://ch.tudelft.nl/~arthur/svn2cl/

Also affect the Sphinx in the Source linked from:
https://gramps-project.org/api_5_1_x/_modules/gramps/gui/aboutdialog.html#AuthorParser

### DIFF
--- a/gramps/gui/aboutdialog.py
+++ b/gramps/gui/aboutdialog.py
@@ -194,9 +194,9 @@ class GrampsAboutDialog(Gtk.AboutDialog):
 class AuthorParser(handler.ContentHandler):
     """Parse the ``authors.xml`` file to show in the About dialog.
 
-    The ``authors.xml`` file has the same format as the one in the `svn2cl
-    <http://ch.tudelft.nl/~arthur/svn2cl/>`_ package, with an additional
-    ``title`` tag in the ``author`` element. For example::
+    The ``authors.xml`` file has the same format as the one in the `svn2cl`
+    package, with an additional ``title`` tag in the ``author`` element.
+    For example::
 
       <author uid="dallingham" title="author">
         Don Allingham &lt;<html:a href="mailto:don@gramps-project.org">don@gramps-project.org</html:a>&gt;


Nick-Hall:
We now use `git2cl` rather than `svn2cl`.

emyoulation:
>  We now use git2cl rather than svn2cl.

OK.  Then does the comment needs to be completely replaced? 
Or just updated with a definitive link for `git2cl`

"""Parse the ``authors.xml`` file to show in the About dialog.
    The ``authors.xml`` file has the same format as the one in the `svn2cl
    <http://ch.tudelft.nl/~arthur/svn2cl/>`_ package, with an additional
    ``title`` tag in the ``author`` element. For example::
    <https://web.archive.org/web/20070323201820/http://ch.tudelft.nl/~arthur/svn2cl/>`_ package,
    with an additional ``title`` tag in the ``author`` element. For example::
      <author uid="dallingham" title="author">
        Don Allingham &lt;<html:a href="mailto:don@gramps-project.org">don@gramps-project.org</html:a>&gt;
      </author>}
    """

emyoulation:
Looked for a good reference to insert. It looks like the canonical home page is gone.

The last good Wayback Machine archive for the josefsson.org page is from 2009.  
https://web.archive.org/web/20090411073845/http://josefsson.org/git2cl/

from: https://github.com/zwizwa/git2cl
The canonical home page for git2cl is:
link: http://josefsson.org/git2cl/[] and its repository can be found at
link: http://repo.or.cz/w/git2cl.git[].

Found this definition:
Git2cl is a utility tool designed to convert Git logs into GNU ChangeLog format. It serves as a quick and simple way to transform the output of Git's version control history into a more standardized changelog format commonly used in GNU projects.

Nick-Hall:
I've just removed the link.  It wasn't very helpful.

PR #1749, opened 2024-07-26
Prevent debug plugin loading for Gramps AIO

ennoborg:
The new AIO build, using cx_freeze, leads to a release version that loads debug plugins. This can be prevented by checking for 'frozen' code. This fix also changes appbuild to a simple number, to stay consistent with normal release policies.

The changed code still allows loading of debug plugins on Linux, where it can be suppressed in the usual way, by specifying the Python optimize parameter (-O)

Note: The \__debug__ constant is checked in more places than the changed _pluginreg.py. Those other files were not changed.

### DIFF
--- a/gramps/gen/plug/_pluginreg.py
+++ b/gramps/gen/plug/_pluginreg.py
@@ -49,6 +49,9 @@ import logging
 
 LOG = logging.getLogger("._manager")
 
+# Fix DEBUG for AIO built with cx_Freeze
+DEBUG = __debug__ and not hasattr(sys, "frozen")
+
 # -------------------------------------------------------------------------
 #
 # PluginData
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Is there a bug report for this?

ennoborg:
No. Should I make one?

Nick-Hall:
@ennoborg We usually like bug reports for fixes in the maintenance branch, but in this case I don't think that it is necessary.

I'll split the change to the AIO build number into a separate commit.

emyoulation:
Perhaps it would be appropriate to document hacking Gramps to make "debug mode" the default?  

Until SuperTool (addon, expert) , Import Text (dev), and/or the Python Evaluation (dev) are part of the built-in plugins as "Expert" mode, then the mass editing options are too doggone limited. 

Mass edits are likely during the early experiences with Gramps following the initial GEDCOM import. Each dialect (including the Gramps GEDCOM Extensions) tends to needs mass tweaks.  Even imports of old Gramps XML might have tweaks needed to re-align old workarounds to new features. (e.g.: converting workarounds for before the Other gender or Grandparent role existed in the XML schema)

PR #1758, opened 2024-08-25
Fix unicode conversion bug when upgrading from schema 16 to 17

patrick-nicholson:
Fixing a string behavior error (py2->py3 change) that prevented me from upgrading a very old database file

### DIFF
--- a/gramps/gen/db/upgrade.py
+++ b/gramps/gen/db/upgrade.py
@@ -317,7 +317,9 @@ def gramps_upgrade_17(self):
             n -= 1
 
         if parent_handle is not None:
-            placeref_list = [(parent_handle.decode("utf-8"), None)]
+            if isinstance(parent_handle, bytes):
+                parent_handle = parent_handle.decode("utf-8")
+            placeref_list = [(parent_handle, None)]
         else:
             placeref_list = []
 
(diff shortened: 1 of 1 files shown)


gramps/gen/db/upgrade.py
```diff
@@ -317,6 +317,8 @@ def gramps_upgrade_17(self):
             n -= 1
 
         if parent_handle is not None:
+            if isinstance(parent_handle, str):
+                parent_handle = parent_handle.encode()
             placeref_list = [(parent_handle.decode("utf-8"), None)]
```
QuLogic, line 322: This seems only used here; it would be simpler to set `placeref_list` directly instead of encoding and decoding again.
```suggestion
            if isinstance(parent_handle, str):
                placeref_list = [(parent_handle, None)]
            else:
                placeref_list = [(parent_handle.decode("utf-8"), None)]
```

gramps/gen/db/upgrade.py
```diff
@@ -487,6 +489,8 @@ def add_place(self, name, level, parent, title):
     gid = self.place_prefix % self.max_id
     placetype = (7 - level, "")
     if parent is not None:
+        if isinstance(parent, str):
+            parent = parent.encode()
         placeref_list = [(parent.decode("utf-8"), None)]
```
QuLogic, line 494: Same here.
```suggestion
        if isinstance(parent, str):
            placeref_list = [(parent, None)]
        else:
            placeref_list = [(parent.decode("utf-8"), None)]
```

PR #1761, opened 2024-08-26
Clean up `exec` and `globals`

QuLogic:
Using `exec` and `globals` can sometimes obscure what code is doing.

The last commit also fixes bug [#13347](https://gramps-project.org/bugs/view.php?id=13347).

### DIFF
--- a/gramps/gen/display/name.py
+++ b/gramps/gen/display/name.py
@@ -749,7 +749,7 @@ class NameDisplay:
 
         The new function is of the form::
 
-        def fn(first, raw_surname_list, suffix, title, call,):
+        def fn(first, raw_surname_list, suffix, title, call, nick, famnick):
             return "%s %s" % (first,suffix)
 
         Specific symbols for parts of a name are defined (keywords given):
@@ -1172,8 +1172,7 @@ class NameDisplay:
 
     def _make_fn(self, format_str, d, args):
         """
-        Create the name display function and handles dependent
-        punctuation.
+        Create the name display function and handles dependent punctuation.
         """
         # d is a dict: dict[code] = (expr, word, translated word)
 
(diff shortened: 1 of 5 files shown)


gramps/gen/display/name.py
```diff
@@ -1288,18 +1287,17 @@ def ifNotEmpty(str,p,s):
             ",".join(param),
         )
         try:
-            exec(s) in globals(), locals()
-            return locals()["fn"]
-        except:
+            result = {}
+            exec(s, globals(), result)
+            return result["fn"]
+        except SyntaxError:
```
QuLogic, line 1293: Note, since the string above only defines a function, I can't see how this could fail in a way other than a syntax error.

gramps/plugins/tool/verify.py
```diff
@@ -547,7 +546,9 @@ def process_family(
         yngmar,
         oldmar,
         oldmom,
+        olddad,
         yngmom,
+        yngdad,
```
QuLogic, line 551: Two examples of undefined variables that happened because `globals()` modifications set them, instead of passing them as parameters.

gramps/gen/display/name.py
hroncok, line 1293: Wrong format strings produce TypeErrors:

```pycon
>>> '' % 'a'
  File "<python-input-0>", line 1, in <module>
    '' % 'a'
    ~~~^~~~~
TypeError: not all arguments converted during string formatting
>>> '%x' % 'a'
  File "<python-input-1>", line 1, in <module>
    '%x' % 'a'
    ~~~~~^~~~~
TypeError: %x format: an integer is required, not str
>>> '%s %s' % 'a'
  File "<python-input-2>", line 1, in <module>
    '%s %s' % 'a'
    ~~~~~~~~^~~~~
TypeError: not enough arguments for format string
```

QuLogic: Format strings are only evaluated on the line defining `s` earlier, or within `fn` itself, but not at the outer level at which the `exec` is run.

All the examples you've given pass when wrapped in `def fn():`.

Note, `fn` is not called here, but cached for use in the callers of this function.

hroncok: Alright then, sorry for the noise.

QuLogic: PS, this file could be written in a way that doesn't use `exec` with some `lambda`s, but as it seemed that this was done for speed, I'm not sure if I should change that without having any benchmarks available. So I wonder if there are any benchmarks for this bit of code?

usdanskys:
Applied all changes to the individual files in my Fedora 41 install. Seems to work!

justbyitself:
Please, merge this pull request. When you update to Python 3.13, you'll get ugly errors everywhere. This is a critical patch, not just an code quality enhancement.
Tested on Arch Linux and it works. Thanks @QuLogic.

dsblank (approved):
Changes look good. Nice cleanup!

dsblank:
@Nick-Hall, this is a bug fix for Python 3.13, and some nice cleanup.

Nick-Hall:

PR #1763, opened 2024-09-04
Fix Python 3.12 deprecation warnings

QuLogic:
[`sqlite3.version`](https://docs.python.org/3/library/sqlite3.html#sqlite3.version) and [truth testing on an XML `Element`](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.remove) are now deprecated.


dsblank (approved):
LGTM. Thanks!

PR #1764, opened 2024-09-05
Replace deprecated Gdk.Color by Gdk.RGBA

QuLogic:
It's been deprecated since GDK 3.14.

### DIFF
--- a/gramps/gui/plug/report/_styleeditor.py
+++ b/gramps/gui/plug/report/_styleeditor.py
@@ -309,7 +309,7 @@ class StyleEditor(ManagedWindow):
         for widget_name in ("color", "bgcolor", "line_color", "fill_color"):
             color = self.top.get_object(widget_name)
             label = self.top.get_object(widget_name + "_code")
-            color.connect("notify::color", self.color_changed, label)
+            color.connect("notify::rgba", self.rgba_changed, label)
 
         self.top.get_object("style_name").set_text(name)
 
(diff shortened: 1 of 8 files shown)


gramps/gui/plug/report/_styleeditor.py
```diff
@@ -395,10 +395,10 @@ def draw_graphics(self):
         self.top.get_object("line_style").set_active(g.get_line_style())
         self.top.get_object("line_width").set_value(g.get_line_width())
 
-        self.line_color = rgb2color(g.get_color())
-        self.top.get_object("line_color").set_color(self.line_color)
-        self.fill_color = rgb2color(g.get_fill_color())
-        self.top.get_object("fill_color").set_color(self.fill_color)
```
QuLogic, line 401: Note that I couldn't find any use of `self.line_color`/`self.fill_color` in this file (e.g., the corollary to this function, `save_graphics`, does not use them), so I think that assignment was a mistake.

gramps/plugins/view/geomoves.py
```diff
@@ -659,28 +660,26 @@ def animate_moves(self, index, person, color):
                     death = high_date
             new_list.append([level, plxp, birth, death])
         pidx = 0
-        if isinstance(color, str):
-            color = Gdk.color_parse(color)
```
QuLogic, line 663: This function (`animate_moves`) is only a callback from a timeout added in `_createmap`, or a timeout added by itself, so we know the input is a `Gdk.RGBA`, and this check can be removed.

PR #1765, opened 2024-09-05
Convert 'Tip of the Day' window to Gtk.Template

QuLogic:
I don't really understand the reason for the `Glade` class, but here's an example of using PyGObject's `Gtk.Template` integration instead on the simplest window.

### DIFF
--- a/gramps/gui/glade/tipofday.glade
+++ b/gramps/gui/glade/tipofday.glade
@@ -2,7 +2,7 @@
 <!-- Generated with glade 3.40.0 -->
 <interface>
   <requires lib="gtk+" version="3.24"/>
-  <object class="GtkWindow" id="tipofday">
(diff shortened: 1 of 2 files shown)


Nick-Hall:
I have posted a [Proposal to replace the Glade class with Gtk Template](https://sourceforge.net/p/gramps/mailman/message/59114120/) to the _gramps-devel_ mailing list asking for opinions.

This PR looks good to me though.

jralls (approved):
Looks OK to me.

emyoulation:
Filed a [bug report 13794](https://gramps-project.org/bugs/view.php?id=13794).  The horizontal centering is calculating incorrectly.  The dialog appears partially off-screen to the left 

Could be related to having no gramps.ini on a fresh install.  But the re-positioning of the dialog is not saved

PR #1766, opened 2024-09-06
Remove Mouse scroll direction option from Pedigree view

QuLogic:
This currently does nothing (probably due to a change in GTK event handling), and while it could probably be fixed, the scrolled window responds to the standard Shift+mouse wheel for horizontal scrolling, so it seems better to just work with standard paradigms.


Nick-Hall:
I have created PR #1838 which should fix the bug.

Nick-Hall:
Closed in favour of PR #1838.

PR #1768, opened 2024-09-06
Retarget Python Eval audience, status, help_url

emyoulation:
Everyone > Expert
Unstable > Stable
generic debug to specific section (5.2 specific version URL. How can this adapt to 5.3?)

fulfill 2nd half of  	0013422: Change "Audience" registrations of plug-ins to 5.2 "Expert"
https://gramps-project.org/bugs/view.php?id=13422

### DIFF
--- a/gramps/plugins/gramplet/gramplet.gpr.py
+++ b/gramps/plugins/gramplet/gramplet.gpr.py
@@ -1452,12 +1452,13 @@ register(
     description="Gramplet allowing the evaluation of python code",
     version="1.0.0",
     gramps_target_version=MODULE_VERSION,
-    status=UNSTABLE,
+    audience=EXPERT,
+    status=STABLE,
     fname="eval.py",
     height=200,
     gramplet="PythonEvaluation",
     gramplet_title="Python Evaluation",
-    help_url=DEBUG_HELP,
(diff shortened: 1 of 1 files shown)


Nick-Hall:
There are three problems I see  with this PR:
* The audience is used by the Addon Manager to select relevant addons for the user.  Is it useful for built-in plugins?
* Changing the status from "Unstable" to "Stable" will make this visible to all users, not just developers.
* The help URL has the version number in it.  This means that we will have to remember to update it for each feature release.

emyoulation:
I am disappointed but understand your objections.

This gramplet would allow diagnostic scripts ( shared via the Discourse forum ) that end‐users could run via a simple copy'n'paste. That really isn't an option with it having to be run in debug mode. It requires too much understanding about setting environmental variables and running Gramps from a console. End-users will exceed their capacity for patience before getting to productive answers. 

And importing scripts takes too many steps in SuperTool. (I don't know if the Python Shell allows a script or only single statements. I've never been able to get the shell gramplet to be productive)

daleathan:
@Nick-Hall The other option for the built-in [Python Evaluation grampet](https://gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Gramplets#Python_Evaluation)  is move it to addons as I don't believe anything like gramps python test rely on it?

PR #1769, opened 2024-09-08
Remove InteractiveSearchBox entirely

QuLogic:
This widget was added in d426f6232e2069de9f051ef5b0f45513bfa55b9c as a copy of the original `GtkTreeView` code, with a modified timeout to the start of searching. Since then, there has been 1 additional change to functionality in bb26e53fc0cb3989c05ee302052d635a61ec6b3d to add support for a slightly different version of `PersistentTreeView` on `ListView`. All other changes have been minimal maintenance: fixing small bugs, replacing deprecated GTK calls, or code formatting.

I contend that this is too much code to maintain for such a small change in functionality from upstream GTK, and any migration to new widgets in GTK4 will have to implement this differently anyway, so I'm proposing to remove it all. AFAICT, search still works without this code, just with GTK's timing instead.


Nick-Hall:
We should ask @bastienjacquet for his opinion on removing the `InteractiveSearchBox` class.  I seem to remember that it was added for performance reasons.

It is difficult for me to assess the impact of this PR since I don't use the interactive search myself.

bastienjacquet:
Wow 😳 that code is 10 year old.
But, yes, the whole point was to deal with large Trees having 50000 people, where standard treeview or Listview search just doesn't work as it is implemented as a line by line search.
As far as I remember, I did 2 things:
- implemented switching to binary search for sorted Listview , especially with views like "Lastname, First name", 
- added the delayed search that was not in the gtk version used at the time.

So, if we have any other (newer?) way to effectively have CTRL+F search on Listviews with 50k item, I'm happy to remove this interactiveSearchBox entirely.

Nick-Hall:
@bastienjacquet Thanks for the quick reply. It looks like we should keep this code until a suitable replacement is suggested.

PR #1770, opened 2024-09-09
Fix printing of Books

QuLogic:
As noted in the original issue, PR#372 changed `GtkPrint` to try to parent its window via `self.uistate`, set in its constructor.

In `DocReportDialog.make_document`, the `uistate` argument is passed to the format class, but `BookDialog` overrides that method and is missing that argument, causing the `GtkPrint.uistate` to remain `None`. This allowed the 'Print...' option to work in plain Reports, but not Books.

Fixes [#12804](https://gramps-project.org/bugs/view.php?id=12804).


hgohel (approved):
Thanks for fixing this bug.

PR #1771, opened 2024-09-09
Use standard decorator form of defining properties

QuLogic:
... for properties that are defined with private methods.


dsblank:
Good work! Long overdue. Thank you!

Nick-Hall (approved):

PR #1775, opened 2024-09-13
Fixes Bug#0013433 -  ancestor_too_old() tests were being skipped uncondirionally

CameronD73:
When checking for probably_alive, previous code did not clear the self.pset list between running descendant test and ancestor test, so the ancestor test code assumed all persons had already been checked and was never executed.

This bug and fix is easier to evaluate if the patch for bug #13431 has been applied - simply to improve debug output.


CameronD73:
oops, wrong branch

PR #1780, opened 2024-09-30
Fix list size option in the top surnames gramplet

DaveSch-gramps:
Allow users to specify how many surnames appear in the list from 10 to 1000

from discourse

https://gramps.discourse.group/t/diversely-transliterated-sur-names-and-statistics/5768/8

### DIFF
--- a/gramps/plugins/gramplet/topsurnamesgramplet.py
+++ b/gramps/plugins/gramplet/topsurnamesgramplet.py
@@ -31,6 +31,7 @@ from collections import defaultdict
 from gramps.gen.plug import Gramplet
 from gramps.gen.config import config
 from gramps.gen.const import GRAMPS_LOCALE as glocale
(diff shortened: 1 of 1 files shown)


DaveSch-gramps:
Bug #0013448

hgohel:
@DaveSch-gramps Do you have Black installed locally? Running that on the source file will eliminate the lint check failure.

DaveSch-gramps:
No to Black. I am not a real coder. I modified code I found in LastModified to achieve the menu options.

I just use Github Desktop for Windows and Atom. I do understand Atom is no longer supported.

hgohel:
> No to Black.

No worries. Looks like Black would like to see two empty lines after `NUM_SURNAMES =" and there's only one right now. Simple fix.

DaveSch-gramps:
Line added and both checks are green!

PR #1781, opened 2024-10-06
Fix broken encoding

smurfix:
UTF-8 vs. Latin1 raises its ugly head again.

Possible TODO: fix the code that adds these entries.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -139,7 +139,7 @@
 	zh_TW Add span characters so that the new display format is accepted as a
 	valid input format.  This is necessary to make the unit tests pass.
 
-2024-07-07  Lajos Nemeséri <nemeseril@gmail.com>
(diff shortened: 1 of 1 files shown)


Nick-Hall:
The `ChangeLog` files are generated as part of the [release process](https://gramps-project.org/wiki/index.php/What_to_do_for_a_release#Changelog_and_NEWS_file), so there isn't much point in fixing them.

We use the [git2cl](https://github.com/gitGNU/gnu_git2cl/blob/master/README) script.  You could try to fix that script or write a python version for us.

I suggest that for the next release I try to fix the issue using my editor.

smurfix:
Thanks for the pointer to the culprit. Fixed.

https://github.com/gitGNU/gnu_git2cl/pull/1

Nick-Hall:
I have downloaded the updated version of `git2cl` and will use it in the next release.

PR #1782, opened 2024-10-07
Check events of type CENSUS as well as OCCUPATION when looking for occupations

davejbur:
At present, when an "Include occupation" option is selected, only events of type OCCUPATION are searched. This mod includes events of type CENSUS as well, which usually include occupation information too.

An alternative, far more complex, approach would be to create another option set for "Include census", independent of the "Include occupation" option. However, for most people, this 2 line mod would probably suffice. (Actually, it would have been just a one line mod, but I've just spent 15 minutes going round in circles working out what's wrong with my code formatting style, only to find my one line needs to be split across two with no trailing spaces:-) )

### DIFF
--- a/gramps/plugins/graph/gvrelgraph.py
+++ b/gramps/plugins/graph/gvrelgraph.py
@@ -749,7 +749,8 @@ class RelGraphReport(Report):
                 for event in [
                     self._db.get_event_from_handle(ref.ref) for ref in event_refs
                 ]
-                if event.get_type() == EventType(EventType.OCCUPATION)
+                if (event.get_type() == EventType(EventType.OCCUPATION))
+                or (event.get_type() == EventType(EventType.CENSUS))
             ]
             if len(events) > 0:
                 events.sort(key=lambda x: x.get_date_object())


davejbur:
Not sure if this is being looked at? or ignored? or just been overlooked?

Nick-Hall:
Sorry for the delay.  I did have a quick look and it just looks wrong to gather occupation details from the description of a census event.

Nick-Hall:
In Gramps, the census event description is shared with all participants of the event.  With a shared event using the description for the occupation of one of the participants is not ideal.   I have seen some people use this field for address information.

> I don't think I would record people in a census event, then add separate occupation & residence events for each census entry, duplicating the information - or is that what others do?

Some people like to do this, but not everyone.

> Perhaps there's a case for adding an option to let people choose whether to include census events?

Yes.  I think that it would have to be an option.

Nick-Hall:
I've create a [topic](https://gramps.discourse.group/t/what-do-people-put-in-the-event-description-field-for-a-census/6765) on our Discourse forum to get some feedback.

Nick-Hall:
The feedback from the Discourse poll suggests that most people leave the event description field empty.  However, those who use it don't put an occupation in it.

I'm closing this because I think that it will cause more problems than it will solve.

PR #1783, opened 2024-10-08
Crash fix

smurfix:
Fixes https://gramps-project.org/bugs/view.php?id=13461


Nick-Hall:
I'm closing this PR because this is not a good fix.  It masks a potential database upgrade bug leaving an old schema in a new database.

PR #1784, opened 2024-10-10
Downgrade informational messages from warning level

hgohel:
Fixes [13464](https://gramps-project.org/bugs/view.php?id=13464)

In upgrade scenarios informational messages that indicate old config files are being imported into the new config are now being treated as informational rather than warnings which may alarm the reader.


hgohel:
Thanks for reviewing the PR. This change does not remove the message from being shown in the console so whatever the reader did previously, they can continue to do that after this change, without being alarmed :) 

If there is a desire to make the messages more conspicuous, a new feature request would be best to handle that.

emyoulation:
Agreed about a feature request. Just wanted to see if there is a consensus first.

PR #1786, opened 2024-10-10
Switch from pickled blobs to JSON data

dsblank:
This PR converts the database interface to use JSON data rather than the pickled blobs used since the early days.

1. Uses a new abstraction in the database: `db.serializer`
  a. abstracts data column name
  b. contains serialize/unserialize functions
2. Updates database format to 21
3. The conversion from 20 to 21 reads pickled blobs, and writes JSON data.
  a. It does this by switching between serializers
4. New databases do not contain pickled blobs
5. Converted databases contain both fields


dsblank:
One thing I am not sure about. If you need to update a table from version 20 or older, you need to open it with the old schema. How does that work?

Nick-Hall:
If we are moving from BLOBs to JSON then we should really use the new format.   See PR #800.

The new format uses the `to_json` and `from_json` methods in the [serialize](https://github.com/gramps-project/gramps/blob/master/gramps/gen/lib/serialize.py) module to build the json from the underlying classes.  It comes with `get_schema` class methods which provide a [JSON Schema](https://json-schema.org) that allow the validation that we already use in our unit tests.

The main benefit of the new format is that it is easier maintain and debug.  Instead of lists we use dictionaries.  So, for example, we refer to the field "parent_family_list" instead of field number 9.

Upgrades are no problem.  We just read and write the raw data.

When I have more time I'll update you on discussion whilst you have been away.

dsblank:
Oh, that sounds like a great idea! I'll take a look at the JSON format and switch to that. Should work even better with the SQL JSON_EXTRACT().

Nick-Hall:
There are a few places where the new format is used, so we will get some bonus performance improvements.

Feel free to make changes to my existing code if you see a benefit.

You may also want to have a quick look at how we serialize `GrampsType`.  Enough information is stored so that we can recreate the object, but I don't think that I chose to store all fields.

dsblank:
Making some progress. Turns out, the serialized format had leaked into many other places, probably for speed. Probably good candidates for business logic.

dsblank:
I added a `to_dict()` and `from_dict()` based on the `to_json()` and `from_json()`. I didn't know about the object hooks. Brilliant! That saves so much code.

dsblank:
@Nick-Hall , I will probably need your assistance regarding the complete save/load of the to_json and from_json functions. I looked at your PR but as it touches 590 files, there is a lot there.

In this PR, I can now upgrade a database, and load the people views (except for name functions which I have to figure out).

image

Nick-Hall:
@dsblank I have rebased PR #800 on the _gramps51_ branch.  Only 25 files were actually changed.

You can also see the changes suggested by @prculley resulting from his testing and performance benchmarks.

dsblank:
Thanks @Nick-Hall, that was very useful. I think that I will cherry pick some of the changes (like attribute name changes, elimination of private attributes).

You'll see that I did many of the same changes you made. But, one thing I found is that if we want to allow upgrades from previous versions, then we need to be able to read in blob_data, and write out json_data. I think my version has that covered.

I'll continue to make progress.

Nick-Hall:
@dsblank Why are you removing the properties?  The validation in the setters will no longer be called.

dsblank:
@Nick-Hall , I thought that was what @prculley did for optimization, and I thought was needed. I can put those back :)

Nick-Hall:
Perhaps we could consider a solution similar to that provided by the pickle `__getstate__` and `__setstate__` methods.

A `get_state` method in a base class could return a dictionary of public attributes by default.  This could be overridden to add properties if required.

A`set_state` method could write the values back.  In the case of properties we could just set the corresponding private variable rather than calling the setter.  The list to tuple conversion could also be done in this method.

I expect that only a handful of classes would need to override the default methods.

dsblank:
Thanks @Nick-Hall, I think you are right. That will allow us to have a 100% JSON encoding and decoding. I got stuck with the current GrampsType, but a get_state/set_state would be completely customizable for each variation.

Nick-Hall:
@dsblank See PR #1787. The code needs tidying up but it essentially works.

dsblank:
@Nick-Hall, yes #1787 looks clean and consistent. I'll wait until that is complete, and rebase this PR on master after that.

dsblank:
Major milestone reached: database upgraded, and all of the views can now be rendered using the JSON raw data. Lots of work still to be done, but wanted to take a minute to update on the status.

image

emyoulation:
Is there something that could be posted to the Discourse forum to stir up some energy?

For instance, the Isotammi Filter+ gramplet shows a timer for filtering. Is there a filter (or custom filter) that would demonstrate the newly optimized difference when run on 5.2 vs. 5.3? 

Posting a capture of Filter+ of the Example.gramps timings would encourage people to archive a similar test now in 5.2 of their real-world data... and again when 5.3 comes out. That might stir up some interest in optimizing the aging internal rules.

Nick-Hall:
@dsblank I'll fix the `Date` in the JSON schema for you tomorrow.

dsblank:
Oh great! I just hit bugs in the tests that I'm working on. Thanks!

dsblank:
Down to two categories of pytest errors and failures:
1. vcard related failures
2. 'Test Case Generator' and 'Check & Repair Database'

The second one seems like a random number change, but still tracking it down.

@Nick-Hall, I think this is all that is needed for Date fix: https://github.com/gramps-project/gramps/pull/1786/commits/d2c301a89f1a732228103fda425d736eb7818f72#diff-de77e4fb6f5c704f314759c3ec0ceddea525f966ddf056e6ee1dee0ba852c5f3L760-L763

But do see that the difference between `null` and `{'format': None, 'calendar': 0, 'modifier': 0, 'quality': 0, 'dateval': [0, 0, 0, False], 'text': '', 'sortval': 0, 'newyear': 0, '_class': 'Date'}` is a big difference in length of json_object.

Nick-Hall:
@dsblank We should also update the JSON schema.  See PR #1789.  There is a unit test that validates the example database against the schema in [schema_test.py](https://github.com/gramps-project/gramps/blob/master/gramps/gen/lib/test/schema_test.py).

The JSON representation of an empty date is longer than null, as you point out.

gramps/gui/views/treemodels/placemodel.py
```diff
@@ -115,86 +117,95 @@ def color_column(self):
         """
         Return the color column.
         """
+        # FIXME: this is a model.get_value() arg
```
dsblank, line 120:
```suggestion
        # This is a model.get_value() arg
```

dsblank:
@Nick-Hall, this is now ready for review. 

There is one test that fails, but fails for me in master too. I suspect a change in the random number generator.

gramps/gen/lib/serialize.py
```diff
@@ -71,3 +73,91 @@ def from_json(data):
...
+    :rtype: dict
+    """
+    return json.loads(to_json(obj))
+
+
+def from_struct(struct):
+    """
+    Convert a struct into a Gramps object.
+
+    :param data: The dictionary to be unserialized.
```
stevenyoungs, line 94: Should this be "struct" rather than "data"?

gramps/gen/lib/serialize.py
```diff
@@ -71,3 +73,91 @@ def from_json(data):
     :rtype: object
     """
     return json.loads(data, object_hook=__object_hook)
+
+
+def to_struct(obj):
```
stevenyoungs, line 78: Is it clearer to call this to_dict given it returns a Python dictionary?

gramps/gen/lib/serialize.py
```diff
@@ -71,3 +73,91 @@ def from_json(data):
...
+
+    :param obj: The object to be serialized.
+    :type obj: object
+    :returns: A dictionary.
+    :rtype: dict
+    """
+    return json.loads(to_json(obj))
+
+
+def from_struct(struct):
```
stevenyoungs, line 90: Is it clearer to call this from_dict?

emyoulation:
Adding some JSON functionality in one place will increase scrutiny in other JSON features.  Could you tweak the [JSON import/export add-ons registrations](https://github.com/gramps-project/addons-source/edit/maintenance/gramps52/JSON/JSON.gpr.py)?


* add a help_url to both:   Addon:JSON_Export_Import
* add export plugin description : "Export Gramps data in JavaScript Object Notation (JSON), "
        "a lightweight data interchange format. "
        "Writes a human-readable, pretty-printed style."
* add import plugin description :  "Import genealogical data in JavaScript Object Notation (JSON), "
        "a lightweight data interchange format."

Perplexity says the import supports 2 styles: Compact, Pretty and Commented but the Export appears to write Pretty style.

dsblank:
@emyoulation, nice catch! Can you turn your above comment into a feature request? I'll handle those issues after this is merged.

gramps/gen/lib/serialize.py
dsblank: Yes, and I did call it that at one point. But I stole some code from the merge/diff code where I had called these "structs" for some reason. A long time ago :) I'll change it throughout.

gramps/gen/lib/serialize.py
dsblank: Do you know about github suggestions (box in edit box with +/-)? Then I could have accepted it :)

gramps/gen/lib/serialize.py
dsblank: Yes, done!

gramps/gen/lib/serialize.py
dsblank: Thanks!

emyoulation:
> Can you turn your above comment into a feature request? I

Done.  Thanks for the direction.
https://gramps-project.org/bugs/view.php?id=13488

gramps/gen/lib/serialize.py
```diff
@@ -71,3 +73,91 @@ def from_json(data):
...
+
+    :param obj: The object to be serialized.
+    :type obj: object
+    :returns: A dictionary.
+    :rtype: dict
+    """
+    return json.loads(to_json(obj))
+
+
+def from_dict(struct):
```
stevenyoungs, line 90:
```suggestion
def from_dict(dict):
```

gramps/gen/lib/serialize.py
```diff
@@ -71,3 +73,91 @@ def from_json(data):
...
+    """
+    Convert a dictionary into a Gramps object.
+
+    :param struct: The dictionary to be unserialized.
+    :type struct: dict
+    :returns: A Gramps object.
+    :rtype: object
+    """
+    return from_json(json.dumps(struct))
```
stevenyoungs, line 99:
```suggestion
    return from_json(json.dumps(dict))
```

gramps/gen/lib/serialize.py
```diff
@@ -71,3 +73,91 @@ def from_json(data):
...
+    :rtype: dict
+    """
+    return json.loads(to_json(obj))
+
+
+def from_dict(struct):
+    """
+    Convert a dictionary into a Gramps object.
+
+    :param struct: The dictionary to be unserialized.
```
stevenyoungs, line 94:
```suggestion
    :param dict: The dictionary to be unserialized.
```

gramps/gen/lib/serialize.py
```diff
@@ -71,3 +73,91 @@ def from_json(data):
...
+    :type struct: dict
```
stevenyoungs, line 95:
```suggestion
    :type dict: dict
```

gramps/gen/lib/serialize.py
stevenyoungs: I do now :-)

gramps/gui/views/treemodels/citationtreemodel.py
```diff
@@ -211,17 +212,23 @@ def add_row2(self, handle, data):
         # add the citation as a child of the source. Otherwise we add the source
         # first (because citations don't have any meaning without the associated
         # source)
-        if self._get_node(data[5]):
+        if self._get_node(data["source_handle"]):
```
stevenyoungs, line 215: Pull "source_handle" out as a constant and use here and line 222 below

gramps/gui/views/treemodels/eventmodel.py
```diff
@@ -137,7 +139,7 @@ def column_description(self, data):
         return data[COLUMN_DESCRIPTION]
 
     def column_participant(self, data):
-        handle = data[0]
+        handle = data["handle"]
```
stevenyoungs, line 142:
```suggestion
        handle = data[COLUMN_HANDLE]
```

gramps/gui/views/treemodels/eventmodel.py
```diff
@@ -148,12 +150,11 @@ def column_participant(self, data):
 
     def column_place(self, data):
         if data[COLUMN_PLACE]:
-            cached, value = self.get_cached_value(data[0], "PLACE")
+            cached, value = self.get_cached_value(data["handle"], "PLACE")
```
stevenyoungs, line 153:
```suggestion
            cached, value = self.get_cached_value(data[COLUMN_HANDLE], "PLACE")
```

gramps/gui/views/treemodels/eventmodel.py
```diff
@@ -148,12 +150,11 @@ def column_participant(self, data):
...
             if not cached:
-                event = Event()
-                event.unserialize(data)
+                event = from_dict(data)
                 value = place_displayer.display_event(self.db, event)
-                self.set_cached_value(data[0], "PLACE", value)
+                self.set_cached_value(data["handle"], "PLACE", value)
```
stevenyoungs, line 157:
```suggestion
                self.set_cached_value(data[COLUMN_HANDLE], "PLACE", value)
```

gramps/gui/views/treemodels/eventmodel.py
```diff
@@ -219,7 +218,7 @@ def column_tag_color(self, data):
         """
         Return the tag color.
         """
-        tag_handle = data[0]
+        tag_handle = data["handle"]
```
stevenyoungs, line 221:
```suggestion
        tag_handle = data[COLUMN_HANDLE]
```

gramps/gui/views/treemodels/familymodel.py
```diff
@@ -116,63 +117,63 @@ def on_get_n_columns(self):
         return len(self.fmap) + 1
 
     def column_father(self, data):
-        handle = data[0]
+        handle = data["handle"]
```
stevenyoungs, line 120: add constants and use instead of raw strings throughout file?

gramps/gui/views/treemodels/citationbasemodel.py
```diff
@@ -54,26 +56,26 @@
...
-COLUMN_HANDLE = 0
-COLUMN_ID = 1
-COLUMN_DATE = 2
-COLUMN_PAGE = 3
-COLUMN_CONFIDENCE = 4
-COLUMN_SOURCE = 5
-COLUMN_CHANGE = 9
-COLUMN_TAGS = 10
-COLUMN_PRIV = 11
+COLUMN_HANDLE = "handle"
```
stevenyoungs, line 59: Would the COLUMN_XXX be clearer if called KEY_XXX since they are now the key in to a Python dictionary?

gramps/gui/views/treemodels/peoplemodel.py
```diff
@@ -79,18 +81,18 @@
...
-COLUMN_DEATH = 5
-COLUMN_BIRTH = 6
-COLUMN_EVENT = 7
-COLUMN_FAMILY = 8
-COLUMN_PARENT = 9
-COLUMN_NOTES = 16
-COLUMN_CHANGE = 17
-COLUMN_TAGS = 18
-COLUMN_PRIV = 19
+COLUMN_ID = "gramps_id"
```
stevenyoungs, line 84:
```suggestion
COLUMN_HANDLE = "handle"
COLUMN_ID = "gramps_id"
```

stevenyoungs: and then use COLUMN_HANDLE within the file

gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -81,6 +83,17 @@ class DBAPI(DbGeneric):
     def _initialize(self, directory, username, password):
         raise NotImplementedError
 
+    def upgrade_table_for_json_access(self, table_name):
+        """
+        A DBAPI level method for upgrading the given table
+        for a json_data field.
```
stevenyoungs, line 89:
```suggestion
        adding a json_data column.
```

gramps/plugins/tool/check.py
```diff
@@ -1065,16 +1056,15 @@ def _fx(value):
...
         change, handle and gramps_id are not compared"""
-        if changepos is not None:
-            return (
-                data[2:changepos] == empty_data[2:changepos]
-                and data[changepos + 1 :] == empty_data[changepos + 1 :]
-            )
-        else:
-            return data[2:] == empty_data[2:]
+        for key in empty_data:
+            if key in ["change", "gramps_id", "handle"]:
```
stevenyoungs, line 1063: Is it worth extracting ["change", "gramps_id", "handle"] and giving it a name to describe why these keys are skipped e.g. internal_data_keys = ["change", "gramps_id", "handle"]

stevenyoungs:
@dsblank apologies for all the nit-picky comments.
Overall looks really good

dsblank:
@stevenyoungs, I appreciate the review! 

For the `COLUMN_` suggestions, I decided to go the other way, and remove all of them. I did that for the following reasons:

1. There are actually more uses of just using the key than having constants, and I'd like the uses to be consistent
2. All of the uses of the `COLUMN_` constants are inside methods of the same name. So, `COLUMN_DATE` was used in a method called `citation_date()`
3. Positions (like in `data["tag_list"][0]`) are not constants.

If others disagree, they make it consistent a different way in a follow-up PR.

Thanks!

gramps/plugins/tool/check.py
dsblank: I think I'll leave it like this as the only note about it is in the comments here, and I think this makes it clear. Also, as far as I saw, it was used only here.

gramps/gui/views/treemodels/citationbasemodel.py
dsblank: See note. But I think using the actual schema name is clearer.

gramps/gui/views/treemodels/familymodel.py
dsblank: See note.

gramps/gui/views/treemodels/eventmodel.py
dsblank: See note.

gramps/gui/views/treemodels/eventmodel.py
dsblank: See note.

gramps/gui/views/treemodels/citationtreemodel.py
dsblank: See note.

dsblank:
@Nick-Hall, all tests are passing, and I have addressed all review comments so far.

gramps/plugins/test/tools_test.py
stevenyoungs, line 1: @dsblank I'm puzzled by this one. Given the nature of your change, I instinctively would not have expected any change in test results. The implication therefore is that the 'Test Case Generator' is generating a different tree for input to the 'Check & Repair Database' command.
It sets a fixed seed in the random number generator and I'm assuming the stream of number generated is then deterministic.

Is the python random number generator deterministic across different python versions and platforms? Although this does not explain why the pipeline tests would differ.
You are not directly generating random numbers in your PR. Is it possible one of the new modules you import is generating random numbers, modifying the stream of random numbers being generated by 'Test Case Generator'?

dsblank: I have no idea why the test seems to have changed. I was going to generate the tree in a database, but I couldn't generate it. I tried different versions of python. My best guess is that there is a call to random (somewhere in a constructor) that isn't called anymore. But I don't know.

Nick-Hall: We occasionally have problems with this test.  I haven't investigated in detail.

As long as the new result is similar to the original result, then fixing a few lines is perfactly acceptable.

Nick-Hall:
> For the COLUMN_ suggestions, I decided to go the other way, and remove all of them.

Yes, I agree that this is the best approach.

stevenyoungs:
> @stevenyoungs, I appreciate the review!
> 
> If others disagree, they can make it consistent a different way in a follow-up PR.
> 
I can make arguments for either approach. Consistency is of greater value. I can't imagine the strings will change frequently.

Nick-Hall:
Have a look at how I do it in the [MongoDB](https://github.com/gramps-project/addons-source/blob/maintenance/gramps52/MongoDB/mongodb.py#L260) backend.  All the tables become collections of JSON documents.  It's only really a proof of concept so feel free to use a different JSON structure.

Getting rid of all the pickled data is probably the way to go.  I don't mind if you continue with this PR.

dsblank:
@Nick-Hall, thanks for the pointer to your MongoDB database implementation. I stole the basic framework.

This was a little tricky as I was using the version to know whether to use JSON or blob, and the version is of course stored in the metadata. But I moved away from using the version number and made a probe to see if the DB supports JSON.

I think that is the last thing I'd want to add to this PR.

gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -83,6 +82,21 @@ class DBAPI(DbGeneric):
     def _initialize(self, directory, username, password):
         raise NotImplementedError
 
+    def supports_json_access(self):
+        """
+        A DBAPI level method for testing if the
+        database support JSON access.
+        """
+        try:
+            self.dbapi.execute("SELECT json_data from person limit 1;")
```
stevenyoungs, line 91: @dsblank are both checks required? Could you just rely on reading from metadata alone?

gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -83,6 +82,21 @@ class DBAPI(DbGeneric):
...
+            self.dbapi.fetchone()
+            self.dbapi.execute("SELECT json_data from metadata limit 1;")
+            self.dbapi.fetchone()
+            self.dbapi.commit()
+            return True
+        except Exception:
```
stevenyoungs, line 97: What if an exception is thrown for a reason other than the json_data column does not exist in the table?
In this scenario, you do not know if the DB support JSON or not. Maybe it is best to let the exception go upwards in this case - there is probably nothing you can do as you cannot determine if the DB is using blob or json.

gramps/plugins/db/dbapi/dbapi.py
dsblank: Yes, you are correct. I'll fix. Thanks!

gramps/plugins/db/dbapi/dbapi.py
dsblank: That is possible, but defeats the purpose of the function to return True or False. There are better ways to check to see if a column exists in specific implementations of SQL (such as using PRAGMA in SQLite), but as far as I know, there is not a DB-API level check. Overrides can do something more specific.

stevenyoungs: Sorry, re-reading my comment I realise it's not very clear. Here's some pseudo-python in which the 3 scenarios are hopefully clearer

```
try:
  self.dbapi.execute("SELECT json_data from metadata limit 1;")
  self.dbapi.fetchone()
  self.dbapi.commit()
  return True
except DbException: # this may not be the correct type of exception to catch
  # handle *only* an exception raised due to the column json_data not existing
  return False
except:
  #  for all other exceptions, re-throw since we don't know if True of False is the correct answer
  raise
```

dsblank: I see. Yes that makes sense. Thanks!

dsblank: The error is a `OperationalError`. However, that is a SQLite error. 

I could require DB-API backends (like SQLite) to implement this method. 

What do you think?

stevenyoungs: We don't want SQLite implementation detail to leak into this layer. 
If we step back, the question is "is the DB we are opening JSON or blob (or conceptually something else in future)"
The current approach is to select up to 1 row from metadata and infer the answer from a lack of error.
What if you change the approach and ask the DB layer if the column 'json_data' exists in the table metadata? This feels more generic and is more directly the question we want to answer. The db layer can handle the True/False/Exception based on DB specific knowledge. The downside is that it extends the db layer and I assume don't immediately have another use for it.

dsblank: I cleaned up this code, and I think it is much better. Everything is at the correct level. It does require DB-API databases to add one new method, `column_exists()` which goes with `table_exists()` that was already present.

dsblank: I think this is resolved. If not, let me know :)

gramps/plugins/db/dbapi/sqlite.py
```diff
@@ -191,6 +191,24 @@ def table_exists(self, table):
         )
         return self.fetchone()[0] != 0
 
+    def column_exists(self, table, column):
+        """
+        Test whether the specified SQL column exists.
+
+        :param table: table name to check.
+        :type table: str
+        :param table: column name to check.
```
stevenyoungs, line 200:
```suggestion
        :param column: column name to check.
```

gramps/plugins/db/dbapi/sqlite.py
```diff
@@ -191,6 +191,24 @@ def table_exists(self, table):
...
+        :type table: str
```
stevenyoungs, line 201:
```suggestion
        :type column: str
```

gramps/plugins/db/dbapi/sqlite.py
```diff
@@ -191,6 +191,24 @@ def table_exists(self, table):
         )
         return self.fetchone()[0] != 0
 
+    def column_exists(self, table, column):
+        """
+        Test whether the specified SQL column exists.
```
stevenyoungs, line 196:
```suggestion
        Test whether the specified SQL column exists in the specified table.
```

gramps/gen/db/generic.py
```diff
@@ -621,6 +623,18 @@ def _initialize(self, directory, username, password):
         """
         raise NotImplementedError
 
+    def upgrade_table_for_json_access(self, table_name):
+        """
+        Overload this method to add JSON access
+        """
+        raise NotImplementedError
+
+    def supports_json_access(self):
```
stevenyoungs, line 632:
```suggestion
    def uses_json_data(self):
```
I wonder if `uses_json_data` is better name than `supports_json_access`?
My thinking: SQLite supports JSON data. Older (blob) gramps databases in SQLite *support* JSON, but don't **use** JSON.
Using similar logic, `upgrade_table_for_json_access` would become `upgrade_table_for_json_data`

gramps/gen/db/generic.py
```diff
@@ -621,6 +623,18 @@ def _initialize(self, directory, username, password):
         """
         raise NotImplementedError
 
+    def upgrade_table_for_json_data(self, table_name):
+        """
+        Overload this method to add JSON access
```
stevenyoungs, line 628:
```suggestion
        Overload this method to upgrade the table to store data in JSON format
```

Nick-Hall (changes requested):
I think that the way you are upgrading the database will cause problems in the future.  Suppose a user is upgrading from schema version 20 to some future version, say 24.  In the upgrade from 20 to 21, an object serialized in version 20 will be unserialized to create a version 24 object, which may possibly result in an error.  Then the object will be converted to version 24 JSON which will be used in the upgrade from 21 to 22.  If a field is added in version 24, then it will appear in the upgrade path before it is expected.  Renamed and deleted object attributes will also cause problems.

We don't instantiate objects in the database upgrades for this reason as well as for better upgrade performance.

Apart from this one issue everything in the PR looks good.

dsblank:
> We don't instantiate objects in the database upgrades for this reason 

I see what you mean. Yes, will change. I'll leave the serializer to read the blob data, but you are right: I'll have to manually convert the unpickled blob data to the current JSON format.

gramps/gen/db/upgrade.py
```diff
@@ -73,36 +75,42 @@ def gramps_upgrade_21(self):
...
             self.set_serializer("json")
-            self._commit_raw(raw, key)
+            # We can use commit_raw as long as json_data
+            # has "handle", "_class", and uses json.dumps()
+            self._commit_raw(json_data, key)
             self.update()
 
     self._txn_commit()
     # Bump up database version. Separate transaction to save metadata.
+    self.set_serializer("json")
```
stevenyoungs, line 113: Is the version 21 update code robust enough to handle a version 20 DB where the conversion to JSON has already taken place?

There is a possible risk here if the preceding transaction completes successfully but either set_serializer or, more likely, _set_metadata calls fails. On disk, we'll have a version 20 DB with JSON data.

gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -81,6 +82,26 @@ class DBAPI(DbGeneric):
...
+    def upgrade_table_for_json_data(self, table_name):
+        """
+        A DBAPI level method for upgrading the given table
+        adding a json_data column.
+        """
+        try:
+            self.dbapi.execute("ALTER TABLE %s ADD COLUMN json_data TEXT;" % table_name)
+            self.dbapi.commit()
+        except Exception:
+            pass
```
stevenyoungs, line 103: Do you want to pass on *all* exceptions or just those thrown because the column json_data already exists?

dsblank: Yeah, that's a good point. I rewrote this without using try/except.

gramps/gen/db/upgrade.py
dsblank: You are correct. I rewrote all of the changes to operate in one transaction. If not committed, then we should be good.

dsblank:
@Nick-Hall, that took a bit of work to convert the blob array directly to JSON, but it is the correct way to do it. And actually found a bug in the to_json() infrastructure: 

https://github.com/gramps-project/gramps/pull/1786/files#diff-15dffcd29856a6b68d1d4e72f80d34ad1146136ade37db4a3bf6c4709b39fa96R72-R90

Surname was missing the object_state. One tricky thing was to make sure to include the defaults that are set the constructors.

I think this is correct. I went through all of the data and checked the conversions and they match on all values.

Now that we aren't creating objects in the upgrade code, we can remove all of the serialize/unserialize code as it isn't used. But I'll do that after this PR.

gramps/gen/db/upgrade.py
```diff
@@ -52,12 +55,66 @@
...
+        count_func = self._get_table_func(key, "count_func")
+        length += count_func()
+
+    self.set_total(length)
+
+    # First, do metadata:
+
+    self._txn_begin()
+    self.set_serializer("blob")
+    self.upgrade_table_for_json_data("metadata")
```
stevenyoungs, line 81:
```suggestion
    self.set_serializer("blob")
    self._txn_begin()
    # First, do metadata:
    self.upgrade_table_for_json_data("metadata")
```

Is there any reason to have the call to set_serializer inside the transaction?
Move the comment closer to the code. Comment may be superfluous, unless you want to expand on why you do the metadata first?

dsblank: > Is there any reason to have the call to set_serializer inside the transaction?

No, not really.

stevenyoungs:
> Surname was missing the object_state.
> 
> One tricky thing was to make sure to include the defaults that are set the constructors.
> 
> I think this is correct. I went through all of the data and checked the conversions and they match on all values.

Is it worth writing a test to roundtrip all objects into and back out of json representation to catch any other, or future, problems?

stevenyoungs:
> that took a bit of work to convert the blob array directly to JSON, but it is the correct way to do it.

👏 that was a tedious bit of code to write 😄

dsblank:
> Is it worth writing a test to roundtrip all objects into and back out of json representation to catch any other, or future, problems?

Sorta... once the pickled blobs are gone, there won't be anything to test. But it is probably worth adding a script for people to run now to verify that this works. I'll do that.

test/blog_to_json_test.py
```diff
@@ -0,0 +1,41 @@
+from gramps.gen.db.utils import open_database
+from gramps.gen.lib.serialize import to_dict, from_dict
+from gramps.gen.db.conversion_tools import convert_21
+
+# 1. Prepare: create a table named "Example" in version 20
```
stevenyoungs, line 5:
```suggestion
# 1. Prepare: create a database named "Example" in version 20
```

test/blog_to_json_test.py
```diff
@@ -0,0 +1,41 @@
+from gramps.gen.db.utils import open_database
```
stevenyoungs, line 1: Update filename to test/blo**b**_to_json_test.py

test/blog_to_json_test.py
dsblank: Done!

dsblank:
@Nick-Hall, there is one part of the conversion we may have to deal with in the future: we used to store a pickled Researcher instance in the metadata. If the Researcher class changes, pickle might not be able to load it, and we would lose the info. We don't have to do anything right now, but we may have to deal with that someday. Other than that, I think this PR is done.

Nick-Hall:
@dsblank OK. Thanks.  I'll have a look at it tomorrow.

Nick-Hall:
After the comment in your other PR #1791 by @Call-Me-Dave about using the MessagePack format, I've been thinking more about the `Serializer` classes.  It would be nice to choose the serialization format in a configuration file and then use that format for all objects.  Other non-object data could be stored in a backend specific format.  This would be useful for testing new formats or perhaps for power users to choose a format in the future.

Updating the `Researcher` object, although not essential, would seem like a good idea.  The PR doesn't quite feel finished yet.

My thoughts about the next release is perhaps it should be a major 6.0 release.   Although most people won't be using the "raw" database API methods, they will change  significantly.  We should also probably limit the release to your PRs on the backend, but otherwise allow code quality changes (such as David's typing PRs) and front-end updates only.  I'll post to the _gramps-devel_ mailing list to get feedback.

dsblank:
@Nick-Hall, please let me know how to proceed with this PR.

Nick-Hall:
@dsblank I should have time to look at this again tomorrow.

Nick-Hall:
I get some "too many values to unpack" errors when I open an old database.  It happens on the get object methods and the iterators, and I see it on the Person and Tag objects so far.  These are errors from gramplets on my dashboard.

```
... gramps/gen/db/generic.py", line 1390, in _get_from_handle
    return self.serializer.data_to_object(obj_class, data)
... gramps/gen/lib/serialize.py", line 119, in data_to_object
    return obj_class.create(data)
... gramps/gen/lib/baseobj.py", line 237, in create
    return cls().unserialize(data)
... gen/lib/tag.py", line 102, in unserialize
    (
ValueError: too many values to unpack (expected 5)
```

I'm not sure why the `unserialize` methods are being called.  We shouldn't need them any more.

When I create a new database and import the example tree I don't get the errors.  Also creating a new example database in the old version and then letting it upgrade seems to work.

I'll investigate further tomorrow.

dsblank:
Right, unserialize would only be called with the blob serializer. But that shouldn't happen.

gramps/gen/lib/surname.py
```diff
@@ -69,6 +69,25 @@ def __init__(self, source=None, data=None):
         if data:
             self.unserialize(data)
 
+    def get_object_state(self):
```
Nick-Hall, line 72: Since `Surname` inherits from `SecondaryObject` which in turn inherits from `BaseObject`, why do we need to define the `get_object_state` and `set_object_state` methods here?

gramps/gen/lib/locationbase.py
```diff
@@ -189,3 +189,44 @@ def set_county(self, data):
     def get_county(self):
         """Return the county name of the LocationBase object."""
         return self.county
+
+    def get_object_state(self):
```
Nick-Hall, line 193: Why is it necessary to define the `get_object_sate` and `set_object_state` methods here?  They should already be inherited by the Primary or Secondary objects.

Nick-Hall:
It is possible that I opened a database with an extra "create date" field that I intended for v5.3 which would also have been schema version 21.  I'll check.

@dsblank It would be worth you grepping the code for "serialize(" and look for occurrences outside of _gen/lib_.  The `ProxyBase`  and ProGen code uses the serialize and/or unserialize methods.  The fixes look easy.

We also compare serialized objects to check for changes in the editors.

dsblank:
Yes, I can do that. Thanks!

dsblank:
Regarding removing all serialize/unserialize from code: I fixed a couple easy ones in libprogen.py.

There are three remaining problems:

1. gramps/gen/db/upgrade.py: `event_type.unserialize(new_event[2])`, `SrcAttributeType(key).serialize()`
2. gramps/gen/db/upgrade.py: `marker.unserialize(marker_field)`, `tag.serialize()`
3. gramps/plugins/db/bsddb/bsddb.py: `data = Researcher().unserialize(new_data)`

I think we will have to:

1. Keep (or rewrite) the event type and source type serialize/unserialize code for version 17 
2. Keep (or rewrite) the marker and tag serialize/unserialize code for version 15 
3. Keep the `Researcher.unserialize()`

As these are preexisting problems ("there should be no object creation in upgrade code"), I think they are outside the scope of this PR, and we can leave that code for now.

gramps/gen/db/upgrade.py
dsblank: I moved (and added) this line to the top of all of the conversion code. It might not be needed, but is a reminder to humans that these are all blob-based upgrades.

dsblank:
> It is possible that I opened a database with an extra "create date" field that I intended for v5.3 which would also have been schema version 21. 

I like the idea of an explicit creation date field (instead having to do math on the handle). If you want to add that later to the new JSON format in version 21, that would be great.

dsblank:
@Nick-Hall, you have any estimate on possible review on this PR (and the #1794 filter fixes)?

I have some available time coming up, and would like to start work on checking the addons for this next version (6.0 or 5.3).

Nick-Hall:
@dsblank I'll make time this weekend, but may be able to start sooner.

dsblank:
@Nick-Hall, if you'd like to meet over Google Meet or Zoom so that I can walk you (and others) through proposed changes, I'd be glad to.

Nick-Hall:
This PR is looking good now.

I agree with you that the remaining serialize/unserialize code in the upgrade path should be left to another PR.  Changes to the upgrades always require extra testing.

Would it be useful to log database upgrades?  Perhaps a version table that could store the dates of database creation and any upgrades.  I'm not suggesting adding to this PR though.

I'll convert my changes to add a create timestamp field to the primary objects so that they work with the new raw JSON format.  Upgrades to the schema should be easier after this PR is merged.  I'm not sure if we'll want to include it in the next release.  We can discuss this later.

dsblank:
Sounds good!

dsblank:
Shall we merge this PR then?

Nick-Hall:
I'm just about to do some final checks, followed by a rebase and merge now.  It was getting late last night.

Nick-Hall (approved):

stevenyoungs (approved):
Good from my point of view

dsblank:
Thank you all for the reviews and comments!

Nick-Hall:
@dsblank Please don't merge PRs until I have done a final review.  I was about to merge this, but noticed that the new `gen.db.conversion_tools` package was not listed in the `setup.py` and the two files it contains are not in the `POTFILES.skip` file.

In your merge a "Co-authored-by: stevenyoungs <steve@youngs.cc>" credit seems to have been lost.  Was this intentional?

Otherwise, I appreciate that you squashed the commits and rebased to maintain a linear history according to our committing policies.

dsblank:
Oh, sorry... how do you want to fix? I did not mean to lose Steve's credit.

Nick-Hall:
I've created PR #1823 with the changes I was going to include.

Unfortunately, we can't go back and add the credit to the commit now.  We can add a copyright line in a file if it hasn't already been done.  I'll make sure that I mention Steve in the release announcement.

PR #1787, opened 2024-10-14
Refactor JSON serialisation

Nick-Hall:
Simplify the JSON serialisation by creating methods to get and set the state of an object.

The method names `get_attrs` and `set_attrs` are not ideal and will probably need changing.

There is some copy and paste of these methods from the `BaseObject` into the `Date` object.  Why are `GrampsType`, `Date`, `StyledText` and `StyledTextTag` not derived from `BaseObject`?

TODO:  Look at removing the checks for empty dates from the `to_json` and `from_json` functions.

### DIFF
--- a/gramps/gen/lib/grampstype.py
+++ b/gramps/gen/lib/grampstype.py
@@ -2,7 +2,7 @@
 # Gramps - a GTK+/GNOME based genealogy program
 #
 # Copyright (C) 2000-2007  Donald N. Allingham
-# Copyright (C) 2017       Nick Hall
+# Copyright (C) 2017,2024  Nick Hall
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -136,6 +136,30 @@ class GrampsType(metaclass=GrampsTypeMeta):
         if value is not None:
             self.set(value)
 
+    def get_object_state(self):
+        """
+        Get the current object state as a dictionary.
+
+        We override this method to handle the `value` and `string` properties.
+        """
+        attr_dict = {"_class": self.__class__.__name__}
+        attr_dict["value"] = self.__value
+        attr_dict["string"] = self.__string
+        return attr_dict
+
+    def set_object_state(self, attr_dict):
+        """
+        Set the current object state using information provided in the given
+        dictionary.
+
+        We override this method to handle the `value` and `string` properties.
+        """
+        self.__value = attr_dict["value"]
+        if self.__value == self._CUSTOM:
+            self.__string = attr_dict["string"]
+        else:
+            self.__string = ""
+
     def __set_tuple(self, value):
         "Set the value/string properties from a tuple."
         val, strg = self._DEFAULT, ""
@@ -225,7 +249,8 @@ class GrampsType(metaclass=GrampsTypeMeta):
             "title": _("Type"),
             "properties": {
                 "_class": {"enum": [cls.__name__]},
-                "string": {"type": "string", "title": _("Type")},
+                "string": {"type": "string", "title": _("Custom type")},
+                "value": {"type": "integer", "title": _("Type code")},
             },
         }
 
--- a/gramps/gen/lib/baseobj.py
+++ b/gramps/gen/lib/baseobj.py
@@ -2,6 +2,7 @@
 # Gramps - a GTK+/GNOME based genealogy program
 #
 # Copyright (C) 2000-2006  Donald N. Allingham
+# Copyright (C) 2024       Nick Hall
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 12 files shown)


dsblank:
> Why are GrampsType, Date, StyledText and StyledTextTag not derived from BaseObject?

I always wondered that too.

DavidMStraub:
By the way, at the moment, the GrampsType values are still dependent on the locale, as we are not using `xml_str`. It would be nice if we could change that in the next db schema change, so the JSON becomes locale independent.

Nick-Hall:
Yes. I've already changed `GrampsType` to store an integer value for pre-defined options and a string for custom values.  This now mirrors the properties in the class.

Nick-Hall:
I've just tried making `BaseObject` the parent class for `GrampsType`, `Date`, `StyledText` and `StyledTextTag`.  The change to `GrampsType` causes errors so I'll probably leave it as it is for now.

Nick-Hall:
Renamed `get_attrs` and `set_attrs` to `get_object_state` and `set_object_state`.

dsblank:
I just tested the conversions (to and from) on all objects in the "Example" Gramps tree, and everything checks out. I added a `from_struct()` in `gramps.gen.merge.diff` for completeness and convenience. I checked the diff, and also the serialization just to be certain.

Here are the tests (which could easily be turned into a pytest):

```python
from gramps.gen.db.utils import open_database
from gramps.gen.merge.diff import diff_items, from_struct, to_struct

db = open_database("Example")

for table in db._get_table_func():
    print(table)
    iter_func = db._get_table_func(table, "iter_func")

    for obj1 in iter_func():
        d1 = to_struct(obj1)
        obj2 = from_struct(d1)
        d2 = to_struct(obj1)
        different = diff_items("Object", d1, d2)
        if different:
            print("different structs!")
        if obj1.serialize() != obj2.serialize():
            print("different serializations!")
            
db.close()
```

Is there anything left to do on this before merging?

Nick-Hall:
Thanks for reminding me.  I'll try to get this merged tomorrow.

Nick-Hall:
@dsblank I think that this is ready to merge now, but it is getting late so I'll leave it until tomorrow morning.

I've added docstrings and made some changes which should improve performance.  I haven't run any performance tests yet though.

We already have some unit tests in the [serialize_test.py](https://github.com/gramps-project/gramps/blob/master/gramps/gen/lib/test/serialize_test.py) file which test a round trip with `to_json` and `from_json`.  We can add to these.  The public method `db.method` is now preferred to the protected table functions.

dsblank:
I can add to those tomorrow.

dsblank:
Actually, the serialize_test.py covers everything that I just tested. Wonder why? I wrote that 11 years ago :)

gramps/gen/lib/grampstype.py
```diff
@@ -225,7 +249,8 @@ def get_schema(cls):
             "title": _("Type"),
             "properties": {
                 "_class": {"enum": [cls.__name__]},
-                "string": {"type": "string", "title": _("Type")},
+                "string": {"type": "string", "title": _("Type String")},
+                "value": {"type": "integer", "title": _("Type Value")},
```
dsblank, line 253: What effect does this have on other parts of Gramps?

dsblank (approved):
Looks good to me! I left one question.

Nick-Hall:
I've just tried creating the objects with `__new__` so that the `__init__` methods are not run.

It works and should give us better performance as @prculley previously suggested.  The only disadvantage is that empty or incomplete JSON classes are no longer possible.  For example `{"_class": "Person"}` or  `{"_class": "Person", "gender": 2}` will now give an error.  This probably isn't an issue though - I  can just remove the corresponding unit tests.

I'll also investigate handling private attributes in a different way.  It's probably neater to mangle/unmangle them automatically.

dsblank:
Looks like tests are finding that the objects don't have "handle" set for some reason.

Nick-Hall:
I just wrote a quick script which I ran in the Python Evaluation gramplet with the example database.

```python
import pickle
import time
from gramps.gen.lib import Person
from gramps.gen.lib.serialize import to_json, from_json

t1 = 0
t2 = 0

for handle in self.dbstate.db.get_person_handles():
    raw = self.dbstate.db.get_raw_person_data(handle)
    data = pickle.dumps(raw)
    c = time.time()
    obj1 = Person.create(pickle.loads(data))
    t = time.time() - c
    t1 += t

    json = to_json(obj1)
    c = time.time()
    obj2 = from_json(json)
    t = time.time() - c
    t2 += t

print (t1, t2, t2/t1)
```

It compares the time difference between creating an object from a JSON string to pickled data.

This PR appears to improve the performance over the existing code, but the JSON is still significantly slower than the pickled data.  I don't include database access times in this test.

Some further investigation is probably needed.

dsblank:
I tweaked your `__object_hook()` a little, and made all of the `set_object_state()` methods return `self`, and now it appears to be **faster** than `Object.create(pickle.load())` version.

```python
def __object_hook(obj_dict):
    _class = obj_dict.pop("_class")
    cls = lib.__dict__[_class]
    obj = cls.__new__(cls)
    return obj.set_object_state(obj_dict)
```

Nick-Hall:
Replacing `getattr` made a big improvement especially for the `Person` objects which I would expect to contain a larger hierarchy of secondary objects.  However, changing the  `set_object_state()` methods to return `self` had little or no effect.

For people I'm getting 24ms for the pickle (4ms unpickle, 20ms serialize) and 30ms for the JSON.  So the new method is still about 25% slower for me.

I've just tested the orjson module compared to the json module.  With all the people in our example database: json.loads takes 18ms compared to 8ms for orjson.loads.  Unfortunately orjson doesn't provide an object hook like json.

dsblank:
I'm not getting times like you report. I'm running Python 3.10.13 running on a laptop (SSD drive) I'm getting:

Pickle + create: 82.6 ms
from_json: 84.6 ms 
Difference: 0.02 % slower

with:

```python
def __object_hook(obj_dict):
    _class = obj_dict.pop("_class")
    cls = lib.__dict__[_class]
    obj = cls.__new__(cls)
    obj.set_object_state(obj_dict)
    return obj
```

dsblank:
On a much bigger, newer machine (Ubuntu 22.04, 64GB RAM, SSD drives, AMD® Ryzen 9 7950x 16-core processor × 32, 64 bit), I get (on the same code above):

Pickle + create: 25.41
from_json: 23.32 
Difference: 8 percent faster

Nick-Hall:
My machine is a modern with an Intel i5 processor, 32GB RAM and SSD drives.  The main difference is that I am running Arch Linux with python 3.12.7.

dsblank:
@Nick-Hall, I think we can go ahead and merge this PR (with above tweak) because:

1. It is about 70% faster than before the PR
2. It fixes a bug
3. It better represents custom Gramps types
4. On the machines I have tested it is +/- 3 percent of pickle+create

We can of course continue to refine and explore, but this is better than before.

dsblank:
@Nick-Hall, I see in one place in the code where it assumes that all Events have a Date. With the new format is seems that the Date may be None. Is that correct that Event dates may now be None?

DavidMStraub:
A great use case for type hints 😉

Nick-Hall:
@dsblank It looks like that empty dates are represented in the raw format as `None`.  They are converted to empty dates in the `unserialize` methods.

We should do the same with the JSON format.  It would be more convenient to store the empty date as a date in the JSON though.  I'll create a PR.

Call-Me-Dave:
Congratulations on this from what I see and understand I'm hoping that a news/blog post could be made about this large change when the time is right 👍 but best while fresh in your minds 🧷   

This is for Gramps 5.3 right, will a test version for windows be made so people can test it or is it too early?

dsblank:
@Call-Me-Dave, this is actually a major update of existing functionality.

But what might come next is the really exciting stuff! With this change, and changes to the DB,  we can query the database directly to get items directly that used to be unreachable inside the pickled-blob.

PR #1789, opened 2024-10-30
Store empty date objects instead of null in the JSON schema

Nick-Hall:


dsblank:
Thank you for the updated representation! Much better than storing the whole date repr.

dsblank (approved):
Looks good to me!

PR #1790, opened 2024-11-03
Proposed code changes for ProbablyAlive code

CameronD73:
This is a squashed merge of development against the gramps master branch (as of 2-Nov-2024, commit 8ca23265 ). The original commits  remain in branch _test-living-again_ on https://github.com/CameronD73/grampsf.git

It addresses several bugs, referenced as related in issue [#13443](https://gramps-project.org/bugs/view.php?id=13443) on mantis, as well as restructuring the probably_alive code and improving the algorithm to determine likely range of birth and death dates. Some of those bugs also have PRs against maintenance/gramps52.

It includes updates to unit test files, to account for changes in output from the new code.

See issue [#13443](https://gramps-project.org/bugs/view.php?id=13443).

### DIFF
--- a/gramps/gen/utils/alive.py
+++ b/gramps/gen/utils/alive.py
@@ -4,6 +4,7 @@
 # Copyright (C) 2000-2007  Donald N. Allingham
 # Copyright (C) 2009       Gary Burton
 # Copyright (C) 2011       Tim G L Lyons
+# Copyright (C) 2024       Cameron Davidson
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -32,8 +33,6 @@ privacy in reports and exports.
 # -------------------------------------------------------------------------
 import logging
 
-LOG = logging.getLogger(".gen.utils.alive")
-
 # -------------------------------------------------------------------------
 #
 # Gramps modules
@@ -41,11 +40,16 @@ LOG = logging.getLogger(".gen.utils.alive")
 # -------------------------------------------------------------------------
 from ..display.name import displayer as name_displayer
 from ..lib.date import Date, Today
+from ..lib.person import Person
 from ..errors import DatabaseError
 from ..const import GRAMPS_LOCALE as glocale
+from ..proxy.proxybase import ProxyDbBase
+
+LOG = logging.getLogger(".gen.utils.alive")
 
 _ = glocale.translation.sgettext
 
+DEBUGLEVEL = 4  # 4 = everything; 3 much detail; 2= minor detail; 1 = summary
 # -------------------------------------------------------------------------
 #
 # Constants from config .ini keys
@@ -58,11 +62,13 @@ try:
     _MAX_AGE_PROB_ALIVE = config.get("behavior.max-age-prob-alive")
     _MAX_SIB_AGE_DIFF = config.get("behavior.max-sib-age-diff")
     _AVG_GENERATION_GAP = config.get("behavior.avg-generation-gap")
+    _MIN_GENERATION_YEARS = config.get("behavior.min-generation-years")
 except ImportError:
     # Utils used as module not part of GRAMPS
     _MAX_AGE_PROB_ALIVE = 110
     _MAX_SIB_AGE_DIFF = 20
     _AVG_GENERATION_GAP = 20
+    _MIN_GENERATION_YEARS = 13
 
 
 # -------------------------------------------------------------------------
@@ -81,6 +87,7 @@ class ProbablyAlive:
         max_sib_age_diff=None,
         max_age_prob_alive=None,
         avg_generation_gap=None,
+        min_generation_years=None,
     ):
         self.db = db
         if max_sib_age_diff is None:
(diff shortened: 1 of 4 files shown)


gramps/gen/utils/alive.py
```diff
@@ -89,234 +96,420 @@ def __init__(
...
+                    death_found = True
+                    dateobj = evnt.get_date_object()
+                    if dateobj and dateobj.is_valid():
+                        death_date = dateobj
+                        explain_death = _("date")
+
+            # at this stage death_date is None or a valid date.
+            # death_found is true if thisperson is known to be dead,
+            #        whether or not a date was found.
+            # If we have no death_date then look for fallback even such as Burial.
```
stevenyoungs, line 193:
```suggestion
            # If we have no death_date then look for fallback event such as Burial.
```

gramps/gen/utils/alive.py
```diff
@@ -89,234 +96,420 @@ def __init__(
...
+            # These fallbacks are fairly good indications that someone's not alive.
+            # If that date itself is not valid, it means we know they are dead
```
stevenyoungs, line 195:
```suggestion
            # If death_date itself is not valid, it means we know they are dead
```
I think this is the date you are talking about.

gramps/gen/utils/alive.py
```diff
@@ -89,234 +96,420 @@ def __init__(
...
+                if min_birth_year:
+                    birth_date.set_yr_mon_day(min_birth_year, 1, 1)
+                    birth_date.set_modifier(Date.MOD_AFTER)
+                elif max_birth_year:
+                    birth_date.set_yr_mon_day(max_birth_year, 12, 31)
+                    birth_date.set_modifier(Date.MOD_BEFORE)
+            birth_date.recalc_sort_value()
+
+        # If we have no death date but we know death has happened then
+        # we set death range somewhere between birth and yesterday.
```
stevenyoungs, line 390: Being pedantic, why do you assume yesterday as the upper bound?
They could have died earlier today

gramps/gen/utils/alive.py
CameronD73: At that stage I am referring to the fallback date being invalid. What I meant to says was...
"If the fallback death event does not have a valid date then..."
The comment seems rather self-evident now, but I initially wrote it in the context that the previous code would scan all fallback events and the last event would apply - potentially overwriting a good date with Today().

stevenyoungs: That comment is clearer and makes sense given the proceeding sentence.

gramps/gen/utils/alive.py
CameronD73: The chance of it happening today and the date being unknown is rather remote, so the choice is not keeping me awake at night. But mainly I chose yesterday so there was no ambiguity whether code used 'less than' or 'less than or equal' against today.

gramps/gen/utils/alive.py
CameronD73: Sorry to be thick, but am I able to change my comment in the code? it is not obvious from this UI view how I would do it.

gramps/gen/utils/alive.py
```diff
@@ -89,234 +96,420 @@ def __init__(
...
+            )
+        explanation += _(", death: ") + explain_death
+        explanation = "2ND + " + explanation
+        if birth_date.is_valid() and death_date.is_valid():
+            return (birth_date, death_date, explanation, person)
+
+        # have finished immediate family, so try spouse, as the
+        # remaining person (probably) of this generation ..
+
+        def spouse_test(passnum=1):
```
stevenyoungs, line 436: Given that `passnum` only appears to be used to determine the value of `immediate_fam_only`, you could have have `immediate_fam_only` as the method parameter. Would this be clearer than `passnum`?

gramps/gen/utils/alive.py
stevenyoungs: I'm not a github expert, but I think you need to edit the file in your branch to make the change.
That's essentially what happened in the background when you accepted my earlier suggestion:
image
https://github.com/CameronD73/grampsf/commits/probablyAlive-0013443/

gramps/gen/utils/alive.py
stevenyoungs: Makes sense. I agree it is a very low probability occurrence and won't be keeping me awake either 😉

gramps/gen/utils/alive.py
CameronD73: Yes, you are right, although I would still use a different variable name because `immediate_fam_only` is a parameter in the outer function. Using the same name in overlapping scopes leaves me a bit uneasy.

gramps/gen/utils/alive.py
CameronD73: OK done that.

Nick-Hall:
Rebased.

emyoulation:
Added as a "new for 6.0" to the Wiki's Fallback since this would probably fall through the WikiContributer cracks.

https://gramps-project.org/wiki/index.php/Events_in_Gramps#Fallback_Events

emyoulation:
The description will be pretty nebulous for the Change log/roadmap too. Can it be edited to mention Probate fallback?

Nick-Hall:
@emyoulation Feel free to update the bug report and roadmap.  The commit message cannot be changed once it has been merged.

gramps/gen/utils/alive.py
```diff
@@ -89,234 +96,421 @@ def __init__(
...
+                    death_date,
+                )
+            return (birth_date, death_date, death_found, explain_birth, explain_death)
+
+        birth_date, death_date, known_to_be_dead, explain_birth_min, explain_death = (
+            get_person_bd(person)
+        )
+
+        explanation = (
+            _("DIRECT birth: ") + explain_birth_min + _(", death: ") + explain_death
```
tosky, line 254: This is marked as translatable, but then DIRECT is used in a string comparison operation, which means that it would fail if translated.

gramps/gen/utils/alive.py
```diff
@@ -656,66 +958,116 @@ def probably_alive(
...
-        name_displayer.display(person),
-        birth,
-        death,
-        explain,
-    )
-    if not birth or not death:
+    elif not current_date.is_valid():
+        current_date = Today()
+
+    if not explain.startswith("DIRECT"):
```
tosky, line 984: This condition compares the initial part of explain with "DIRECT", but "DIRECT" can be translated (see above), so it may not work. Either DIRECT should be removed from the translatable part of the string or (better) it would need to be moved to a different variable.

tosky (commented):
Late to the party, but there is an issue here

gramps/gen/utils/alive.py
CameronD73: And I have just seen more problems due to my poor understanding of the translation mechanism - I will fix.

gramps/gen/utils/alive.py
CameronD73: Have removed passnum along the lines you suggested - will appear in a new PR.

gramps/gen/utils/alive.py
CameronD73: Have made various improvements to messages to assist translation. 
This includes replacing the "DIRECT" with more understandable language and the same with an untranslated "2ND".
The condition `if not explain.startswith("DIRECT"):` was actually not necessary and related to an interim bit of diagnostic logging code.
Will submit PR.

PR #1791, opened 2024-11-16
Select json data from database

dsblank:
This PR adds a generic method for querying the database. It has a pure-python implementation (in case some database backend cannot implement it, or for testing, but not designed for real use) and a DB-API implementation. The pure-python version could be removed from this PR.

## Motivation

As we create an optimized filter API, we will need to create methods for querying the data using the power of the underlying database (such as SQL, or MongoDB). Some of these methods will require "business logic" methods that require JOINS or other more complicated queries.

However, a large number of queries are simple and can be implemented by simply querying the JSON data in a table.

## Example

There are many examples in the 274 filter rules. Here is one that is used often: `_hastag.py`. This is used to determine if an object contains a tag.

```python
    def apply(self, db, obj):
        """
        Apply the rule.  Return True for a match.
        """
        if self.tag_handle is None:
            return False
        return self.tag_handle in obj.get_tag_list()
```

Currently, it is designed to examine every object to see if the tag_handle is in the tag_list.

However, that can be re-written as:

```python
    def prepare(self, db, user):
        self.tag_handle = None
        tag = db.get_tag_from_name(self.list[0])
        if tag is not None:
            self.tag_handle = tag.get_handle()
            results = db.select(self.table, ["$.handle"], ("$.tag_list", "LIKE", f'%"{self.tag_handle}"%'))
            self.map = set([row["handle"] for row in list(results)])
        else:
            self.map = set()

    def apply_to_one(self, db, data):
        return data["handle"] in self.map
```

The key line is:

```python
db.select(self.table, ["$.handle"], ("$.tag_list", "LIKE", f'%"{self.tag_handle}"%'))
```

## Decision

The choice is:

1. Write specific methods for each of these simple selects
2. Add a single method, `db.select()`, that will prevent having to write dozens of one-line methods

We need to make this decision before touching the 274 rules for the filter refactor/optimize PR.

### DIFF
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -91,7 +91,6 @@ from . import (
 from .bookmarks import DbBookmarks
 from .exceptions import DbUpgradeRequiredError, DbVersionError
 from .utils import clear_lock_file, write_lock_file
-from .select_utils import select
 
 _ = glocale.translation.gettext
 
@@ -2741,13 +2740,31 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
 
     def select(self, table, selections=None, where=None, sort_by=None):
         """
-        Generic function that can handle jsonpath items in a list.
+        Generic implementation of select, with where and sort_by clauses.
 
-        table - name of table
-        selections -
-            Example: ("gender", "primary_name.suname_list[0].surname", ...)
-        sort_by - "gender"
-        where - ("and", ("handle", "=", "abc64564346"),
-                        ("gramps_id", "=", ""))
+        :param table: Name of table
+        :type table: str
+        :param selections: List of json-paths
+        :type selections: List[str] or tuple(str)
+        :param where: A single where-expression (see below)
+        :type where: tuple or list
+        :param sort_by: A list of expressions to sort on
+        :type where: tuple or list
+        :returns: Returns selected items from select rows, from iterator
+        :rtype: dict
+
+        Examples:
+
+        Get the gender and surname of all males, sort by gramps_id:
+        ```
+        db.select(
+            "person",
+            ["$.gender", "$.primary_name.surname_list[0].surname"],
+            where=["$.gender", "=", Person.MALE],
+            sort_by=["$.gramps_id"],
+        )
+        ```
         """
+        from gramps.gen.db.select_utils import select
+
         yield from select(self, table, selections, where, sort_by)
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -2738,7 +2738,8 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
         elif serializer_name == "json":
             self.serializer = JSONSerializer
 
-    def select(self, table, selections=None, where=None, sort_by=None):
+    def select(self, table, selections=None, where=None, sort_by=None,
+               page=0, page_size=25):
         """
         Generic implementation of select, with where and sort_by clauses.
 
@@ -2748,8 +2749,10 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
         :type selections: List[str] or tuple(str)
         :param where: A single where-expression (see below)
         :type where: tuple or list
-        :param sort_by: A list of expressions to sort on
-        :type where: tuple or list
+        :param page: The page number to return (zero-based)
+        :type page: int
+        :param page_size: The size of a page in rows; None means ignore
+        :type page: int or None
         :returns: Returns selected items from select rows, from iterator
         :rtype: dict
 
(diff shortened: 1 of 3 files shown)


gramps/gen/db/select_utils.py
```diff
@@ -0,0 +1,315 @@
...
+
+    The where expressions only support the following operators:
+        "=", "!=", "<", "<=", ">", ">=", "in", "not in", "like"
+
+    The `like` operator supports "%" (zero or more characters)
+    and "_" (one character) regular expression matches.
+    """
+    selections = selections if selections else ["$"]
+    if page_size is None:
+        limit = float("+inf")
```
stevenyoungs, line 76:
```suggestion
        limit = sys.maxsize
```
Could you use sys.maxsize instead of converting a string to a float?

gramps/gen/db/select_utils.py
```diff
@@ -0,0 +1,315 @@
...
+            raise Exception("Operator %r is not supported" % where[1])
+    else:
+        raise Exception("Malformed where expression: %r" % where)
+
+
+def compare_where(lhs, op, rhs):
+    """
+    Evaluate the where expression.
+    """
+    if op == "=":
```
stevenyoungs, line 257: would it be clearer to use match \ case instead of if \ elif?

gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -1182,3 +1183,122 @@ def _sql_cast_list(self, values):
...
+        method.
+
+        :param table: Name of table
+        :type table: str
+        :param selections: List of json-paths
+        :type selections: List[str] or tuple(str)
+        :param where: A single where-expression (see below)
+        :type where: tuple or list
+        :param sort_by: A list of expressions to sort on
+        :type where: tuple or list
```
stevenyoungs, line 1245:
```suggestion
        :type sort_by: tuple or list
```

gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -1182,3 +1183,122 @@ def _sql_cast_list(self, values):
...
+        :param page: The page number to return (zero-based)
+        :type page: int
+        :param page_size: The size of a page in rows; None means ignore
+        :type page: int or None
```
stevenyoungs, line 1249:
```suggestion
        :type page_size: int or None
```

gramps/gen/db/generic.py
```diff
@@ -2737,3 +2737,50 @@ def set_serializer(self, serializer_name):
...
+        :param table: Name of table
+        :type table: str
+        :param selections: List of json-paths
+        :type selections: List[str] or tuple(str)
+        :param where: A single where-expression (see below)
+        :type where: tuple or list
+        :param page: The page number to return (zero-based)
+        :type page: int
+        :param page_size: The size of a page in rows; None means ignore
+        :type page: int or None
```
stevenyoungs, line 2756:
```suggestion
        :type page_size: int or None
```

gramps/gen/db/select_utils.py
```diff
@@ -0,0 +1,315 @@
...
+    :param selections: List of json-paths
+    :type selections: List[str] or tuple(str)
+    :param where: A single where-expression (see below)
+    :type where: tuple or list
+    :param sort_by: A list of expressions to sort on
+    :type where: tuple or list
+    :param page: The page number to return (zero-based)
+    :type page: int
+    :param page_size: The size of a page in rows; None means ignore
+    :type page: int or None
```
stevenyoungs, line 47:
```suggestion
    :type page_size: int or None
```

stevenyoungs:
If I read the code correctly, the current implementation of `db.select()` is is limited to a single table.
Is there a way to extend this to support multi-table queries e.g. get all the women who were born in 1850
Your PR will improve performance for this query but we still end up bringing more rows into memory than if we ran the entire query within the DB. 
Without thinking about it too hard, you'd bring rows for all births in 1850 and all women and intersect the two results in python.

Any query would still need transforming from "gramps SQL" into the SQL dialect used by the DB given gramps support of different DBs and their differing syntax, especially for querying JSON.

gramps/gen/db/select_utils.py
QuLogic: That's Python 3.10+.

dsblank:
> If I read the code correctly, the current implementation of db.select() is is limited to a single table.

Yes, that is correct. I thought that this was a compromise to add just a little bit that can be converted into SQL, but also can be run without it if needed (Pure-Python version). There has been a lot of debate over this issue over the years, and I didn't want to wade into that.

> Is there a way to extend this to support multi-table queries e.g. get all the women who were born in 1850

The line I am drawing is that if you need a JOIN, then you should write a "business logic" method.

> Your PR will improve performance for this query but we still end up bringing more rows into memory than if we ran the entire query within the DB. Without thinking about it too hard, you'd bring rows for all births in 1850 and all women and intersect the two results in python.

As long as the selected handles are less than the total rows, then it will be faster, but at the expensive of some memory. We can actually make a decision in the rule.prepare() to decide to not make the `rule.map` if it is too big.

> Any query would still need transforming from "gramps SQL" into the SQL dialect used by the DB given gramps support of different DBs and their differing syntax, especially for querying JSON.

Yes, another reason to keep this simple: it will be overloaded in MongoDb, etc.

Call-Me-Dave:
Would MessagePack be useful?

> It's like JSON.
> but fast and small.
> 
> MessagePack is an efficient binary serialization format. It lets you exchange data among multiple languages like JSON. But it's faster and smaller. Small integers are encoded into a single byte, and typical short strings require only one extra byte in addition to the strings themselves.
> 
> https://msgpack.org/index.html

dsblank:
> Would MessagePack be useful?

Maybe for some functions, but I don't think anything related to the latest work on Gramps. JSON is useful because it is self-documenting, allows the database to be used outside of Gramps, and can be directly queried by SQL.

dsblank:
I'm going to close this for now, as I don't want SQL issues getting in the way of moving forward with the filter fixes.

PR #1792, opened 2024-11-16
Refactor and optimize filters

dsblank:
This PR refactors (and fixes) many issues in the filter system. The initial fix was just unrolling recursive calls to be less expensive.

Next, instead of automatically creating an Gramps object for each line, JSON data is passed to the rules. Some rules will still require an object, but it is only created when needed, and only created once per row. This results in a large speed-up from before (like 30%).

Finally, a set of optimizations are added to make the filters superfast. This uses the conventions already created in Gramps: some expensive filters create a map and use that to see if the object matches. We turn the filter operation inside out, using the maps to narrow down the items considered. This can take operations that took minutes into ones that can be done in a few seconds.

On overview can be read here: https://gramps.discourse.group/t/making-gramps-filters-faster-and-then-superfast/6460/7


dsblank:
I'm going to close this PR and open one not based on any SQL access

PR #1756, opened 2024-08-21
Add rule 'having note of type' for citation, event, family, person, p…

royas:
New rules for custom filters for citation, event, family, person, place and source, matching objects having note of speciied type.

PR #1794, opened 2024-11-22
Refactor, fix, and optimizer filters/rules

dsblank:
This PR does four things:

1. Unrolls the recursive call of filter.apply(), and splits out single checks into filter.apply_to_one()
2. Uses data attributes (`person.gender`) rather than accessor functions (`person.get_gender()`) where possible
3. Adds an optimizer based on `rule.selected_handles` sets
4. Adds typing hints to make sure the right objects are passed into methods

Final comparison of finding those related to home person in 40k person Family Tree, between Gramps 5.2 and master + this PR (Gramps 6.0), in seconds (smaller is better):

Version | Prepare Time | Apply Time | Total Time
--------| -------------:|-----------:|----------:
Gramps 5.2 | 4.5 | 27.7 | 32.2
Gramps 6.0 | 8.0 | 0.5 | 8.5

The above uses the optimizer. Here is a test finding all people with a tag (5 people match):

Version | Prepare Time | Apply Time | Total Time
--------| -------------:|-----------:|----------:
Gramps 5.2 | 0.0 | 5.0 | 5.0
Gramps 6.0 | 0.0 | 1.6 | 1.6

Recall that converting from JSON to objects is a little slower than converting from array BLOBS to objects, so this is a large improvement.


gramps/gen/filters/rules/repository/_hasrepo.py
```diff
@@ -67,10 +67,12 @@ def prepare(self, db, user):
...
             self.rtype.set_from_xml_str(self.list[1])
 
-    def apply(self, _db, obj):
+    def apply_to_one(self, _db, data):
         """
         Apply the rule. Return True on a match.
         """
+        obj = self.get_object(data)
+        # FIXME: probably can do the rest without object:
         if not self.match_substring(0, obj.get_name()):
```
stevenyoungs, line 76:
```suggestion
        if not self.match_substring(0, data["name"]):
```

gramps/gen/filters/rules/person/_nobirthdate.py
```diff
@@ -45,7 +45,9 @@ class NoBirthdate(Rule):
     description = _("Matches people without a known birthdate")
     category = _("General filters")
 
-    def apply(self, db, person):
+    def apply_to_one(self, db, data):
+        person = self.get_object(data)
+        # FIXME: can probably do all of this without object:
```
stevenyoungs, line 50: use data["birth_ref_index"] to avoid contructing the object?

gramps/gen/filters/rules/place/_inlatlonneighborhood.py
```diff
@@ -145,6 +145,8 @@ def apply(self, db, place):
 
         # now we know at least one is given in the filter and is valid
 
+        place = self.get_object(data)
```
stevenyoungs, line 148: Can you use data["lat"] and data["long"] to avoid constructing the object?

dsblank:
Thanks @stevenyoungs for looking this over and the feedback. I'm hoping that @Nick-Hall is open to these three ideas (fixes, using JSON data, and the optimizer) because all three make the filter system so much faster while keeping the API almost the same. Many places in Gramps use the raw data for speed, including tools, views, displayers, and importers. It would be a shame if filters couldn't do the same.

stevenyoungs:
> Thanks @stevenyoungs for looking this over and the feedback. I'm hoping that @Nick-Hall is open to these three ideas (fixes, using JSON data, and the optimizer) because all three make the filter system so much faster while keeping the API almost the same. Many places in Gramps use the raw data for speed, including tools, views, displayers, and importers. It would be a shame if filters couldn't do the same.

No problem. I'm also keen to see the benefits of more standard storage and see what opportunities it unlocks. The recent discussion of GQL is an interesting option. 
I've learnt a lot from following your changes and hopefully my comments have had a minor benefit to the overall quality of the PR.

Nick-Hall:
> hopefully my comments have had a minor benefit to the overall quality of the PR.

Your feedback is very useful.  It always helps for another person to review the code.  I very much appreciate your contributions.

Nick-Hall:
> I'm hoping that @Nick-Hall is open to these three ideas (fixes, using JSON data, and the optimizer) because all three make the filter system so much faster while keeping the API almost the same.

With the new JSON format, I think that we can regard the raw data as providing lightweight objects.  The format mirrors the Gramps objects and is therefore much easier to understand and more resilient to updates.

I've been doing some research today to see if we can make the dict format more object-like.  For example, `data.gender` would perhaps be neater than `data["gender"]`.  The dataclasses introduced in python 3.7 look interesting, but I don't think that they are of much help to us.  Using a JSON parser written in C may be an option in the future, but the ones I was looking at deserialized to dict structures.

I am open to using JSON data in more core code, including filters.  We should probably discuss where the raw format is acceptable.  Third-party code should be careful when using the raw format.  It may change between feature releases, not just major releases, but his has always been the case.

An optimizer seems like a good idea.  I haven't reviewed your code yet though.

stevenyoungs:
The only downside is that we leak the internal representation of the objects into the higher layers of code. But since those layers, in part, already use the raw format, this PR does not change make a material difference.

dsblank:
> The only downside is that we leak the internal representation of the objects into the higher layers of code.

For many attributes of an object, it isn't so much "internal" now that we are switching to JSON. For example, this is the Person object API to get the parent family list:

```python
person.parent_family_list # object attribute
person.get_parent_family_handle_list() # function call
```

With the above `DataDict` (with some additional code) with new JSON, it could look like (where `person` is really JSON data):

```
person.parent_family_list  # JSON access
person.get_parent_family_handle_list() # function call, instantiates object
```

The new version would hide the fact that the function call creates the object (but just once). So because the JSON data mirrors the actual attribute names, it doesn't have to change at all.

dsblank:
Created a PR #1824 to explore attribute access of the JSON dict.

dsblank:
I'll update this PR to use new DataDict attribute access to dict. Which will actually revert most of the data access to what it is in master.

dsblank:
@Nick-Hall, I want to refactor this PR based on the DataDict from #1824. Should I change the base branch to #1824, or wait until it is merged?

Nick-Hall:
I have reviewed and merged PR #1824.  Normally we would wait longer to give people longer to comment, but in this case the PR was related to existing work so I made an exception.

dsblank:
Thank you!

gramps/gen/filters/rules/repository/_hasrepo.py
dsblank: Resolved to `obj.name` with new attr accessor.

gramps/gen/filters/rules/place/_inlatlonneighborhood.py
dsblank: Done!

gramps/gen/filters/rules/person/_nobirthdate.py
dsblank: Replaced with:

```
person.event_ref_list[person.birth_ref_index]
```

gramps/gen/filters/rules/place/_hasnolatorlon.py
```diff
@@ -50,7 +50,7 @@ class HasNoLatOrLon(Rule):
     description = _("Matches places with empty latitude or longitude")
     category = _("Position filters")
 
-    def apply(self, db, place):
-        if place.get_latitude().strip and place.get_longitude().strip():
```
dsblank, line 54: This is a bug fix (strip wasn't called).

gramps/gen/filters/rules/place/_inlatlonneighborhood.py
```diff
@@ -146,10 +146,10 @@ def apply(self, db, place):
         # now we know at least one is given in the filter and is valid
 
         # the place we look at must have lat AND lon entered
-        if not (place.get_latitude().strip and place.get_longitude().strip()):
```
dsblank, line 149: Bug fix.

gramps/gui/views/treemodels/placemodel.py
```diff
@@ -124,41 +123,39 @@ def on_get_n_columns(self):
         return len(self.fmap) + 1
 
     def column_title(self, data):
-        handle = data["handle"]
```
stevenyoungs, line 127: Have you benchmarked `data["handle"]` vs `data.handle`?
For `data["handle"]` is the performance of the DataDict version the same as builtin dict?
Instinctively, I'd expect `data.handle` to be slightly slower. DataDict.__get_attr__ is doing extra work in the normal case:

1. a string comparison `key.startswith("_")`
2. lookup in dict: `key in self`
and then repeat for `self[key]`. Shame we cannot optimise this to avoid the double lookup
3. two x calls to `isinstance()`


I did see your comment in PR #1824 where you reported that `person.get_gender()` was 4.4 times slower than `person.gender`

gramps/gen/filters/_genericfilter.py
```diff
@@ -250,16 +269,44 @@ def apply(self, db, id_list=None, tupleind=None, user=None, tree=False):
                 if id_list not given, all items in the database that
                 match the filter are returned as a list of handles
         """
-        m = self.get_check_func()
+        if user:
+            user.begin_progress(_("Filter"), _("Preparing ..."), len(self.flist) + 1)
+            # Give the progress popup some time to be visible:
+            time.sleep(0.1)
```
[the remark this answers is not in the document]
dsblank: Yes, that is a hack. There must be a bug in the progress bar display. I'll try to fix the bug.

gramps/gui/views/treemodels/placemodel.py
dsblank: I'll do a test, but the cost should be negligible. The get_gender() is 4 times slower because it instantiates an object.  The gain in readability and flexibility should be also weighed.

stevenyoungs: Great, thanks for doing the benchmark.
I'm strongly in favour of the change so apologies if my original comment came across negatively. It was not intended!

dsblank: No, all good points!

gramps/gen/filters/_genericfilter.py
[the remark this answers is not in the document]
dsblank: I removed the hack, but I was unable to find a solution to why most of the time the preparation dialog does not show. I remember wrestling with a similar issue many years ago that required the processing of pending Gtk events. Anyway, this one can be resolved.

gramps/gui/views/treemodels/placemodel.py
dsblank: I think this one can be resolved.

dsblank: Had to instantiate the full object in order to work with proxies.

dsblank:
@Nick-Hall, @stevenyoungs, @DavidMStraub and other reviewers: I believe that this PR is now complete. Over the last couple of days, I removed the use of `DataDict` as that does not respect any proxies, and added typing to the apply methods (this really helps making sure that we are passing in the right kinds of values, and catches other issues). The PR now changes little in the majority of the code. The main change is the restructuring of the main application of the filters, and the addition of the optimizer. 

Timing results are added to the PR description.

dsblank:
Although this PR touches a lot of files, it fixes a serious performance issue that has been in Gramps for a very long time. And it add a performance boost on top of that. It needs a review and testing. What is the plan?

Nick-Hall:
> It needs a review and testing. What is the plan?

I suggest that we ask for volunteers to help test this PR.  If initial testing is successful we could schedule an alpha or beta release to give it wider exposure.

gramps/gen/filters/rules/_changedsincebase.py
```diff
@@ -93,16 +102,16 @@ def time_str_to_sec(self, time_str):
...
         self.since = None
         self.before = None
         if self.list[0]:
             self.since = self.time_str_to_sec(self.list[0])
         if self.list[1]:
             self.before = self.time_str_to_sec(self.list[1])
 
-    def apply(self, db, obj):
-        obj_time = obj.get_change_time()
+    def apply_to_one(self, db: Database, obj: PrimaryObject) -> bool:
```
stevenyoungs, line 113: For my own understanding why use the type `PrimaryObject` here when `TableObject` appears to be sufficient?

gramps/gen/filters/rules/family/_fatherhasidof.py
```diff
@@ -48,5 +56,13 @@ class FatherHasIdOf(RegExpIdBase):
...
-    apply = father_base
+
+    def apply_to_one(self, db: Database, family: Family) -> bool:  # type: ignore[override]
+        father_handle = family.father_handle
+        if father_handle:
+            father = db.get_person_from_handle(father_handle)
+            if father:
+                return super().apply_to_one(db, father)
+            else:
+                return False
```
stevenyoungs, line 67:
```suggestion
            return super().apply_to_one(db, father) if father else False
```
Question: is the shorter ternary style or move verbose if: else: preferred in gramps?

gramps/gen/db/__init__.py
```diff
@@ -71,3 +71,5 @@
 from .undoredo import *
 from .utils import *
 from .generic import *
+
+Database = DbGeneric
```
stevenyoungs, line 75: Is it clearer to use `DbGeneric` rather than introduce a new type alias `Database`? (at least I think that's what you are doing)
Even if I find `Database` easier to read, I'm not sure having two names for the same type helps.

gramps/gen/filters/rules/person/_hasrelationship.py
```diff
@@ -57,31 +66,28 @@ def __init__(self, arg, use_regex=False, use_case=False):
...
+        for handle in obj.family_list:
             family = db.get_family_from_handle(handle)
             if family:
-                total_children += len(family.get_child_ref_list())
-                if (
-                    self.relationship_type
-                    and self.relationship_type == family.get_relationship()
-                ):
+                total_children += len(family.child_ref_list)
+                if self.relationship_type and self.relationship_type == family.type:
```
stevenyoungs, line 90:
```suggestion
                if self.relationship_type and (self.relationship_type == family.type):
```
Does using parenthesis to be explicit about precedence, rather than rely on precedence rules, make it easier for others in future?

gramps/gen/filters/rules/_changedsincebase.py
Nick-Hall: We only have rules for the primary object types.  The `TableObject` includes the `Tag` object that is stored in a table, but is not considered a full primary object.  Table objects do contain a change timestamp though.

gramps/gen/filters/rules/family/_fatherhasidof.py
dsblank: I think it depends on the situation and clarity. But makes sense for this one.

gramps/gen/filters/rules/_changedsincebase.py
dsblank: I didn't spend a huge amount of time (but it took some time to get everything to pass the typing tests), but I also think that there may be a better way. In my tests, DataDict objects have no problem with any of the code. Thus, we might want to have a union type that could include objects with a Tag (in this particular case) or might be a Tag DataDict. I figure we can do lots more with types as we move forward.

gramps/gen/db/__init__.py
dsblank: Yeah, I agree. But I really didn't want to use the `DbGeneric` name when that doesn't convey the right information to the humans. Another item we can wrestle with as we come up with better type infrastructure.

gramps/gen/filters/rules/person/_hasrelationship.py
dsblank: Yes, I think so.

Nick-Hall:
@dsblank Did you update the new 'having note of type' rule added in PR #1756 ?

dsblank:
> Did you update the new 'having note of type' rule added in PR https://github.com/gramps-project/gramps/pull/1756 ?

No, I guess I would have no way of knowing that code was merged after I did the major conversion. Although I would have expected a test to fail. Maybe there were no tests for that rule?

Anyway, I will convert it. Thanks for noticing!

gramps/gen/filters/rules/family/_regexpchildname.py
```diff
@@ -33,7 +33,24 @@
...
+from gramps.gen.db import Database
+
+
+# -------------------------------------------------------------------------
+#
+# Typing modules
+#
+# -------------------------------------------------------------------------
+from gramps.gen.lib import Family
+from gramps.gen.db import Database
```
stevenyoungs, line 53:
```suggestion
```
Duplication of imports

gramps/gen/filters/rules/person/_regexpname.py
```diff
@@ -53,16 +62,16 @@ class RegExpName(Rule):
...
+        for name in [person.primary_name] + person.alternate_names:
             for field in [
                 name.first_name,
-                name.get_surname(),
                 name.suffix,
                 name.title,
                 name.nick,
                 name.famnick,
                 name.call,
+                name.get_surname(),
```
stevenyoungs, line 74: Is there a reason for changing the order in which fields are tested

gramps/gen/filters/rules/person/_isdescendantfamilyof.py
```diff
@@ -72,48 +82,48 @@ def prepare(self, db, user):
             self.exclude()
 
     def reset(self):
-        self.matches = set()
+        self.map = set()
```
stevenyoungs, line 85:
```suggestion
        self.map.clear()
```

gramps/gen/filters/rules/person/_regexpname.py
dsblank: Yes, it is more expensive, so do the others first.

stevenyoungs:
I'm interested in how we can effectively review large PRs. As Doug mentioned this is, out of necessity, a big PR.
My first attempt was considering the PR in its entirety. At the end of the evening, my head was spinning 😵‍💫
My second approach was (in my own repo)
1. rebase this PR on top of master
2. squash all commits from this PR into one commit
3. undo the commit and unstage all files
the changes are now sat in my working tree
4. group related changes into a single commit (e.g. all the lines where type hints are added to `prepare` in one commit, all the typing module imports in another commit, using attributes instead of methods in another etc)
5. keep going until you have no changes left

I'm quite a long way towards step 5; it seems easier. It allowed me to identify a couple of comments yesterday. Is there a better way?

dsblank:
> Is there a better way?

Sorry that this PR is such an evolution. Luckily, we have a good test suite.

dsblank:
@stevenyoungs, this might help: If you want to look at groups of changes by path, you can do this:

```
% git diff master gramps/gen/filters/rules/
```

This example will show all changes between current code and master in just the gramps/gen/filters/rules/ folder.

gramps/gen/filters/rules/place/_withinarea.py
```diff
@@ -106,17 +116,19 @@ def prepare(self, db, user):
...
             return False
         if self.longitude is None:
             return False
         if place:
-            lat = place.get_latitude()
-            lon = place.get_longitude()
-        if lat and lon:
+            lat = place.lat
+            lon = place.long
+        if lat is not None and lon is not None:
```
stevenyoungs, line 131: I assume a bugfix for when `lat` and `lon` are both zero

gramps/gen/filters/rules/person/_hascommonancestorwith.py
```diff
@@ -96,17 +106,22 @@ def add_ancs(self, db, person):
     def reset(self):
         self.ancestor_cache = {}
 
-    def has_common_ancestor(self, other):
+    def has_common_ancestor(self, other: Person):
         for handle in self.with_people:
-            if (handle in self.ancestor_cache and self.ancestor_cache[handle]) & (
+            # Requires bitwise and for some reason:
```
stevenyoungs, line 111: Did you investigate why?

gramps/gen/filters/rules/person/_hascommonancestorwith.py
```diff
@@ -96,17 +106,22 @@ def add_ancs(self, db, person):
...
+            left_and = (
+                handle in self.ancestor_cache and self.ancestor_cache[handle]
+            )  # type: ignore
```
stevenyoungs, line 114: from reading the code, I would have expected this to be type `bool`
I'm assuming it's not since you've ignored the type?

gramps/gen/lib/serialize.py
```diff
@@ -66,7 +66,65 @@ def __str__(self):
             self["_object"] = from_dict(self)
         return str(self["_object"])
 
+    def __int__(self):
+        if "_object" not in self:
+            self["_object"] = from_dict(self)
```
stevenyoungs, line 71:
```suggestion
    def __get_object(self):
        if "_object" not in self:
            self["_object"] = from_dict(self)

    def __int__(self):
        self.__get_object()
```

Given the number of places you repeats these two lines, consider extracting into a private method

gramps/gen/filters/rules/family/_hasnote.py
```diff
@@ -46,3 +46,4 @@ class HasNote(HasNoteBase):
 
     name = _("Families having <count> notes")
     description = _("Matches families having a certain number notes")
+    table = "family"
```
stevenyoungs, line 49: I could not work out what this was for? 
The filter appears to work with this line commented out.

gramps/gen/filters/_genericfilter.py
```diff
@@ -53,7 +58,8 @@
 class GenericFilter:
     """Filter class that consists of several rules."""
 
-    logical_functions = ["or", "and", "xor", "one"]
+    logical_functions = ["and", "or", "one"]
+    table = "person"
```
stevenyoungs, line 62: I could not find where `table` is used?
Also defined in derived classes

stevenyoungs:
@dsblank you'll be glad to know that I've reached the end. Great PR ❤️
In case it helps anyone else reviewing, my re-organisation of the commits is on this [branch](https://github.com/stevenyoungs/gramps/commits/refactor-filter-and-optimize/). Note however that Doug pushed a few new commits after I created this branch.

gramps/gen/filters/rules/place/_withinarea.py
dsblank: I think I should put the logic back... you are right that it could change the result and I didn't mean to "fix" anything.

gramps/gen/filters/rules/person/_hascommonancestorwith.py
dsblank: No, I only verified that it needed to be that way.

gramps/gen/filters/optimizer.py
```diff
@@ -0,0 +1,137 @@
...
+        for item in filter.flist:
+            if hasattr(item, "find_filter"):
+                rule_filter = item.find_filter()
+                if rule_filter is not None:
+                    self.walk_filters(
+                        rule_filter,
+                        current_invert,
+                        result,
+                    )
+            elif hasattr(item, "map"):
```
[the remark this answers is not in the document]
dsblank: Yes, I agree. I want it to be a special interface. What if I just renamed `map` to `selected_handles` (or something like that) for now, and follow-up later with an API?

gramps/gen/filters/_genericfilter.py
dsblank: I think you are correct. I rewrote some generic code, and so this is no longer needed.

gramps/gen/filters/rules/person/_hascommonancestorwith.py
dsblank: So, getting the type right is harder than you might think. To fix, one would need to get the types of the dependent variables correct, and would require a few changes. For example `set()` is not True, but the type is still `Set` (not `bool`), and `person` would need to change to `person is not None`, etc. If it is ok, I'll leave the type wrangling for a later PR. (The only reason the `ignore` is needed is because I added a `Person` type in the function signature, and I still want to pass the mypy checks).

dsblank:
@stevenyoungs, thank you! Please feel free to add your copyright and name to any of these files!

gramps/gen/lib/serialize.py
```diff
@@ -66,7 +66,68 @@ def __str__(self):
...
+    def __int__(self):
+        self.__get_object()
+        return int(self["_object"])
+
+    def __eq__(self, value):
+        if isinstance(value, dict):
+            return dict(self) == value
+
+        if "_object" not in self:
+            self["_object"] = from_dict(self)
```
stevenyoungs, line 82:
```suggestion
        self.__get_object()
```

gramps/gen/lib/serialize.py
```diff
@@ -66,7 +66,68 @@ def __str__(self):
...
+
+        if isinstance(value, DataDict):
+            value = value._object
+
+        return self["_object"].__eq__(value)
+
+    def __ne__(self, value):
+        if "_object" not in self:
+            self["_object"] = from_dict(self)
```
stevenyoungs, line 91:
```suggestion
        self.__get_object()
```

gramps/gen/lib/serialize.py
```diff
@@ -66,7 +66,68 @@ def __str__(self):
...
+
+        if isinstance(value, DataDict):
+            value = value._object
+
+        return self["_object"].__ne__(value)
+
+    def __lt__(self, value):
+        if "_object" not in self:
+            self["_object"] = from_dict(self)
```
stevenyoungs, line 100:
```suggestion
        self.__get_object()
```

gramps/gen/lib/serialize.py
```diff
@@ -66,7 +66,68 @@ def __str__(self):
...
+
+        if isinstance(value, DataDict):
+            value = value._object
+
+        return self["_object"].__lt__(value)
+
+    def __gt__(self, value):
+        if "_object" not in self:
+            self["_object"] = from_dict(self)
```
stevenyoungs, line 109:
```suggestion
        self.__get_object()
```

gramps/gen/lib/serialize.py
```diff
@@ -66,7 +66,68 @@ def __str__(self):
...
+
+        if isinstance(value, DataDict):
+            value = value._object
+
+        return self["_object"].__gt__(value)
+
+    def __le__(self, value):
+        if "_object" not in self:
+            self["_object"] = from_dict(self)
```
stevenyoungs, line 118:
```suggestion
        self.__get_object()
```

gramps/gen/lib/serialize.py
```diff
@@ -66,7 +66,68 @@ def __str__(self):
...
+
+        if isinstance(value, DataDict):
+            value = value._object
+
+        return self["_object"].__le__(value)
+
     def __getattr__(self, key):
+        if key == "_object":
+            if "_object" not in self:
+                self["_object"] = from_dict(self)
```
stevenyoungs, line 128:
```suggestion
            self.__get_object()
```

stevenyoungs:
> @stevenyoungs, thank you! Please feel free to add your copyright and name to any of these files!

Thanks @dsblank, much appreciated. Perhaps just _genericfilter.py? Since I can't suggest the change perhaps you can add a commit:

# Copyright (C) 2024,2025  Steve Youngs <steve@youngs.cc>

gramps/gen/filters/optimizer.py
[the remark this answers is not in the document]
dsblank: Please reopen this if you feel that renaming `map` to `selected_handles` is not enough for now.

dsblank:
@Nick-Hall, I think this ready for some real-use testing.

Nick-Hall:
@dsblank Thanks.  I should get a chance to look at this in the next couple of days.

Then we have a beta release scheduled for 31st Jan.  I'll review the situation tomorrow.

gramps/gen/filters/optimizer.py
```diff
@@ -78,11 +78,11 @@ def walk_filters(self, filter, parent_invert, result):
...
                         result,
                     )
-            elif hasattr(item, "map"):
-                rules.append(set(item.map))
+            elif hasattr(item, "selected_handles"):
+                rules.append(set(item.selected_handles))
         if rules:
             LOG.debug(
-                "filter %s: parent_invert=%s, invert=%s, op=%s, number of maps=%s",
+                "filter %s: parent_invert=%s, invert=%s, op=%s, number of selected_handles=%s",
```
stevenyoungs, line 85: `selected_handles=%s`
You are printing `len(rules)` which does not, at face value, seem right

gramps/gen/filters/rules/person/_islessthannthgenerationancestorofbookmarked.py
```diff
@@ -58,45 +68,42 @@ class IsLessThanNthGenerationAncestorOfBookmarked(Rule):
...
+        if len(bookmarks) != 0:
+            self.bookmarks: Set[str] = set(bookmarks)
             for self.bookmarkhandle in self.bookmarks:
                 self.init_ancestor_list(self.bookmarkhandle, 1)
 
-    def init_ancestor_list(self, handle, gen):
-        #        if p.get_handle() in self.map:
+    def init_ancestor_list(self, handle: str, gen: int):
+        #        if p.get_handle() in self.selected_handles:
         #            loop_error(self.orig,p)
```
stevenyoungs, line 82:
```suggestion
```
Delete commented out code

gramps/gen/filters/rules/person/_islessthannthgenerationancestorofdefaultperson.py
```diff
@@ -52,44 +62,40 @@ class IsLessThanNthGenerationAncestorOfDefaultPerson(Rule):
...
-            self.init_ancestor_list(self.def_handle, 1)
-        else:
-            self.apply = lambda db, p: False
+            self.init_ancestor_list(p.handle, 1)
 
-    def init_ancestor_list(self, handle, gen):
-        #        if p.get_handle() in self.map:
+    def init_ancestor_list(self, handle: str, gen: int):
+        #        if p.get_handle() in self.selected_handles:
         #            loop_error(self.orig,p)
```
stevenyoungs, line 74:
```suggestion
```
Delete commented out code

gramps/gen/filters/optimizer.py
```diff
@@ -0,0 +1,137 @@
...
+        The handles_in are selected if all of the rules are connected with
+        "and" and not inverted. The handles_out are selected if all of the
+        rules are connected with "and" and inverted.
+        """
+        handles_in = None
+        handles_out = set()
+        # Get all positive non-inverted selected_handles
+        for inverted, logical_op, selected_handles in self.all_selected_handles:
+            if logical_op == "and" and not inverted:
+                LOG.debug("optimizer positive match!")
```
stevenyoungs, line 121:
```suggestion
```
Is this message useful enough to keep?

gramps/gen/filters/optimizer.py
```diff
@@ -0,0 +1,137 @@
...
+                if handles_in is None:
+                    handles_in = intersection(selected_handles)
+                else:
+                    handles_in = intersection([handles_in] + selected_handles)
+
+        # Get all inverted selected_handles:
+        for inverted, logical_op, selected_handles in self.all_selected_handles:
+            if logical_op == "and" and inverted:
+                LOG.debug("optimizer inverted match!")
```
stevenyoungs, line 130:
```suggestion
```
Is this message useful enough to keep?

gramps/gen/filters/rules/person/_isdescendantfamilyoffiltermatch.py
```diff
@@ -55,7 +55,7 @@ class IsDescendantFamilyOfFilterMatch(IsDescendantFamilyOf):
 
     def prepare(self, db, user):
         self.db = db
-        self.matches = set()
+        self.selected_handles = set()
```
stevenyoungs, line 58:
```suggestion
        self.selected_handles : Set[str] = set()
```
Add type

gramps/gen/filters/rules/person/_deeprelationshippathbetween.py
```diff
@@ -170,13 +183,15 @@ def prepare(self, db, user):
                 _("Evaluating people"),
                 db.get_number_of_people(),
             )
-        self.__matches = find_deep_relations(db, user, root_person, target_people)
+        self.selected_handles = find_deep_relations(
```
stevenyoungs, line 186:
```suggestion
        self.selected_handles: Set[str] = find_deep_relations(
```
Add type

gramps/gen/filters/optimizer.py
dsblank: Yes, I think so.

gramps/gen/filters/optimizer.py
dsblank: Yes, I think so.

gramps/gen/db/__init__.py
dsblank: I'll try to define a Database type...

gramps/gen/filters/rules/person/_hascommonancestorwith.py
stevenyoungs:
```suggestion
```
this is actually doing intersection between two dicts

stevenyoungs, line 111:
```suggestion
```
[I can't unresolve my previous comment so adding a new one]
This is actually doing intersection between two dicts

stevenyoungs: It might be clearer to use `intersection()` instead of `&`

gramps/gen/filters/rules/person/_hascommonancestorwith.py
```diff
@@ -96,17 +106,22 @@ def add_ancs(self, db, person):
...
+            right_and = (
                 other
                 and other.handle in self.ancestor_cache
                 and self.ancestor_cache[other.handle]
-            ):
+            )  # type: ignore
+            if left_and & right_and:  # type: ignore
```
stevenyoungs, line 120:
```suggestion
            if left_and.intersection(right_and):  # type: ignore
```
use `intersection()` in preference to `&`

gramps/gen/filters/rules/person/_hascommonancestorwith.py
dsblank: Do you want to make a PR on this PR? You can do that here in the UI. Just click on "Edit" of that file, make your changes, and select "branch" if you can.

dsblank: Oh, I understand now! I can fix that.

stevenyoungs: > Do you want to make a PR on this PR? You can do that here in the UI. Just click on "Edit" of that file, make your changes, and select "branch" if you can.

I don't have write access to the repo so I'm not sure if I'll be able to do this

dsblank: No problem, I got this one.

gramps/gen/filters/rules/test/person_rules_test.py
```diff
@@ -1089,6 +1090,37 @@ def test_isdefaultperson(self):
...
+        """
+        Test IsFemale rule.
+        """
+        # Make everyone a female for this test:
+        mock.return_value = True
+        rule = IsFemale([])
+        # too many to list out to test explicitly
+        self.assertEqual(len(self.filter_with_rule(rule)), 2128)
+        # This did not use the optimizer, so it did loop through DB
+        self.assertEqual(mock.call_count, 2128)
```
stevenyoungs, line 1122:
```suggestion
        number_of_people = 2128
        self.assertEqual(len(self.filter_with_rule(rule)), number_of_people)
        # This did not use the optimizer, so it did loop through DB
        self.assertEqual(mock.call_count, number_of_people)
```
Is this a bit clearer? (I am assuming it's one call per person)
It is worth computing the 2128 figure from the number of people in the test set?

dsblank: Yeah, mock was too complicated to do it properly, so I refactor the call_counter into a context manager. I think it could be generally useful, and the optimizer could use more tests. Thanks!

stevenyoungs: Nice ❤️ 
I agree that this feels more generally useful but conscious that this PR really needs to get merged before it takes on a life of its own

Nick-Hall: I plan to review, merge and probably release a beta over the next couple of days.

dsblank:
@Nick-Hall, once this and #1876 are merged, I'll go through the addons and update those.

fxtmtrsine:
Great stuff 👍

> Then we have a beta release scheduled for 31st Jan. I'll review the situation tomorrow.

@Nick-Hall Can I suggest an alpha and few betas with all these ongoing changes still? And suggest maybe realistically target releasing on April 21, 2025 the 23rd anniversary of Gramps first release?

I'd like to see something added to the unit testing that shows automatically (back ported to the 5.2.x series) and empirically the speed boost between the versions and allowing you to detect regressions in speed?

gramps/gen/filters/rules/family/_fatherhasnameof.py
```diff
@@ -47,5 +55,13 @@ class FatherHasNameOf(HasNameOf):
...
-    apply = father_base
+
+    def apply_to_one(self, db: Database, family: Family) -> bool:  # type: ignore[override]
+        father_handle = family.father_handle
+        if father_handle:
+            father = db.get_person_from_handle(father_handle)
+            if father:
+                return super().apply_to_one(db, father)
+            else:
+                return False
```
cdhorn, line 66:
```suggestion
            return False
```
Unnecessary else.

gramps/gen/filters/rules/family/_hastwins.py
```diff
@@ -48,18 +57,19 @@ class HasTwins(Rule):
...
+                if 0 <= child.birth_ref_index < len(child.event_ref_list):
+                    birthref = child.event_ref_list[child.birth_ref_index]
+                    if birthref:
+                        birth = db.get_event_from_handle(birthref.ref)
+                        sortval = birth.date.sortval
+                        if sortval != 0:
+                            if sortval in date_list:
+                                return True
+                            else:
+                                date_list.append(sortval)
```
cdhorn, line 74:
```suggestion
                            date_list.append(sortval)
```
Unnecessary else.

gramps/gen/filters/rules/family/_motherhasidof.py
```diff
@@ -48,5 +56,13 @@ class MotherHasIdOf(RegExpIdBase):
...
-    apply = mother_base
+
+    def apply_to_one(self, db: Database, family: Family) -> bool:  # type: ignore[override]
+        mother_handle = family.mother_handle
+        if mother_handle:
+            mother = db.get_person_from_handle(mother_handle)
+            if mother:
+                return super().apply_to_one(db, mother)
+            else:
+                return False
```
cdhorn, line 67:
```suggestion
```
Unnecessary else.

gramps/gen/filters/rules/family/_motherhasnameof.py
```diff
@@ -47,5 +55,13 @@ class MotherHasNameOf(HasNameOf):
...
-    apply = mother_base
+
+    def apply_to_one(self, db: Database, family: Family) -> bool:  # type: ignore[override]
+        mother_handle = family.mother_handle
+        if mother_handle:
+            mother = db.get_person_from_handle(mother_handle)
+            if mother:
+                return super().apply_to_one(db, mother)
+            else:
+                return False
```
cdhorn, line 66:
```suggestion
```
Unnecessary else.

gramps/gen/filters/rules/family/_regexpfathername.py
```diff
@@ -50,5 +58,13 @@ class RegExpFatherName(RegExpName):
...
-    apply = father_base
+
+    def apply_to_one(self, db: Database, family: Family) -> bool:  # type: ignore[override]
+        father_handle = family.father_handle
+        if father_handle:
+            father = db.get_person_from_handle(father_handle)
+            if father:
+                return super().apply_to_one(db, father)
+            else:
+                return False
```
cdhorn, line 69:
```suggestion
```
Unnecessary else.

gramps/gen/filters/rules/family/_regexpmothername.py
```diff
@@ -50,5 +58,13 @@ class RegExpMotherName(RegExpName):
...
-    apply = mother_base
+
+    def apply_to_one(self, db: Database, family: Family) -> bool:  # type: ignore[override]
+        mother_handle = family.mother_handle
+        if mother_handle:
+            mother = db.get_person_from_handle(mother_handle)
+            if mother:
+                return super().apply_to_one(db, mother)
+            else:
+                return False
```
cdhorn, line 69:
```suggestion
```
Unnecessary else.

gramps/gen/filters/optimizer.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def intersection(sets):
+    if sets:
+        result = sets[0]
+        for s in sets[1:]:
+            result = result.intersection(s)
+        return result
+    else:
+        return set()
```
cdhorn, line 33:
```suggestion
    return set()
```
Unnecessary else.

gramps/gen/filters/optimizer.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def union(sets):
+    if sets:
+        result = sets[0]
+        for s in sets[1:]:
+            result = result.union(s)
+        return result
+    else:
+        return set()
```
cdhorn, line 43:
```suggestion
    return set()
```
Unnecessary else.

gramps/gen/filters/rules/citation/_hassource.py
```diff
@@ -52,8 +61,8 @@ class HasSource(HasSourceBase):
...
     category = _("Source filters")
 
-    def apply(self, dbase, citation):
-        source = dbase.get_source_from_handle(citation.get_reference_handle())
-        if HasSourceBase.apply(self, dbase, source):
+    def apply_to_one(self, dbase: Database, citation: Any) -> bool:
+        source = dbase.get_source_from_handle(citation.ref)
+        if HasSourceBase.apply_to_one(self, dbase, source):
             return True
         return False
```
cdhorn, line 68:
```suggestion
        return HasSourceBase.apply_to_one(self, dbase, source)
```
Unnecessary if.

gramps/gen/filters/rules/citation/_hassourceidof.py
```diff
@@ -49,8 +58,8 @@ class HasSourceIdOf(HasGrampsId):
...
     category = _("Source filters")
 
-    def apply(self, dbase, citation):
-        source = dbase.get_source_from_handle(citation.get_reference_handle())
-        if HasGrampsId.apply(self, dbase, source):
+    def apply_to_one(self, dbase: Database, citation: RefBase) -> bool:  # type: ignore[override]
+        source = dbase.get_source_from_handle(citation.ref)
+        if HasGrampsId.apply_to_one(self, dbase, source):  # type: ignore[override]
             return True
         return False
```
cdhorn, line 65:
```suggestion
        return HasGrampsId.apply_to_one(self, dbase, source)  # type: ignore[override]
```
Unnecessary if.

gramps/gen/filters/rules/citation/_hassourcenoteregexp.py
```diff
@@ -58,8 +67,8 @@ class HasSourceNoteRegexp(HasNoteRegexBase):
...
     category = _("Source filters")
 
-    def apply(self, db, citation):
-        source = db.get_source_from_handle(citation.get_reference_handle())
-        if HasNoteRegexBase.apply(self, db, source):
+    def apply_to_one(self, db: Database, citation: RefBase) -> bool:  # type: ignore[override]
+        source = db.get_source_from_handle(citation.ref)
+        if HasNoteRegexBase.apply_to_one(self, db, source):
             return True
         return False
```
cdhorn, line 74:
```suggestion
        return HasNoteRegexBase.apply_to_one(self, db, source)
```
Unnecessary if.

gramps/gen/filters/rules/citation/_matchessourcefilter.py
```diff
@@ -57,15 +66,16 @@ class MatchesSourceFilter(MatchesFilterBase):
...
+    def apply_to_one(self, db: Database, object: Citation) -> bool:
         if self.MRF_filt is None:
             return False
 
         source_handle = object.source_handle
-        if self.MRF_filt.check(db, source_handle):
+        source = db.get_source_from_handle(source_handle)
+        if self.MRF_filt.apply_to_one(db, source):
             return True
         return False
```
cdhorn, line 81:
```suggestion
        return self.MRF_filt.apply_to_one(db, source)
```
Unnecessary if.

gramps/gen/filters/rules/citation/_regexpsourceidof.py
```diff
@@ -54,8 +63,8 @@ class RegExpSourceIdOf(RegExpIdBase):
...
     category = _("Source filters")
 
-    def apply(self, dbase, citation):
-        source = dbase.get_source_from_handle(citation.get_reference_handle())
-        if RegExpIdBase.apply(self, dbase, source):
+    def apply_to_one(self, dbase: Database, citation: RefBase) -> bool:  # type: ignore[override]
+        source = dbase.get_source_from_handle(citation.ref)
+        if RegExpIdBase.apply_to_one(self, dbase, source):
             return True
         return False
```
cdhorn, line 70:
```suggestion
        return RegExpIdBase.apply_to_one(self, dbase, source)
```
Unnecessary if.

gramps/gen/filters/rules/event/_hasdayofweek.py
```diff
@@ -42,9 +51,11 @@ class HasDayOfWeek(Rule):
...
+    def apply_to_one(self, db: Database, event: Event) -> bool:
         if not self.list[0]:
             return False
         else:
-            dow = event.get_date_object().get_dow()
-            return dow == int(self.list[0])
+            if event.date:
+                dow = event.date.get_dow()
+                return dow == int(self.list[0])
+            return False
```
cdhorn, line 61: Unnecessary else here as well. Can not apply a suggestion here.

gramps/gen/filters/rules/family/_fatherhasidof.py
```diff
@@ -48,5 +56,10 @@ class FatherHasIdOf(RegExpIdBase):
...
     description = _("Matches families whose father has a specified " "Gramps ID")
     category = _("Father filters")
-    base_class = RegExpIdBase
-    apply = father_base
+
+    def apply_to_one(self, db: Database, family: Family) -> bool:  # type: ignore[override]
+        father_handle = family.father_handle
+        if father_handle:
+            father = db.get_person_from_handle(father_handle)
+            return super().apply_to_one(db, father) if father else False
```
cdhorn, line 64:
```suggestion
            if father:
                return super().apply_to_one(db, father)
```
I think this is slightly clearer maybe?

gramps/gen/filters/rules/family/_searchfathername.py
```diff
@@ -47,5 +55,13 @@ class SearchFatherName(SearchName):
...
-    apply = father_base
+
+    def apply_to_one(self, db: Database, family: Family) -> bool:  # type: ignore[override]
+        father_handle = family.father_handle
+        if father_handle:
+            father = db.get_person_from_handle(father_handle)
+            if father:
+                return super().apply_to_one(db, father)
+            else:
+                return False
```
cdhorn, line 66:
```suggestion
```
Unnecessary else.

gramps/gen/filters/rules/family/_searchmothername.py
```diff
@@ -47,5 +55,13 @@ class SearchMotherName(SearchName):
...
-    apply = mother_base
+
+    def apply_to_one(self, db: Database, family: Family) -> bool:  # type: ignore[override]
+        mother_handle = family.mother_handle
+        if mother_handle:
+            mother = db.get_person_from_handle(mother_handle)
+            if mother:
+                return super().apply_to_one(db, mother)
+            else:
+                return False
```
cdhorn, line 66:
```suggestion
```
Unnecessary else.

gramps/gen/filters/rules/note/_matchesregexpof.py
```diff
@@ -49,8 +58,8 @@ class MatchesRegexpOf(Rule):
     category = _("General filters")
     allow_regex = True
 
-    def apply(self, db, note):
+    def apply_to_one(self, db: Database, note: Note) -> bool:
         """Apply the filter"""
-        if self.match_substring(0, note.get()):
+        if self.match_substring(0, str(note.text)):
             return True
         return False
```
cdhorn, line 65:
```suggestion
        return self.match_substring(0, str(note.text))
```
Unnecessary if.

gramps/gen/filters/rules/person/_nobirthdate.py
```diff
@@ -45,15 +54,18 @@ class NoBirthdate(Rule):
...
+            birth = db.get_event_from_handle(birth_ref.ref)
+            if birth:
+                birth_obj = birth.date
+                if not birth_obj:
+                    return True
+                if birth_obj.sortval == 0:
+                    return True
+            return False
+        else:
+            return True
```
cdhorn, line 71:
```suggestion
        return True
```
Unnecessary else.

gramps/gen/filters/rules/person/_nodeathdate.py
```diff
@@ -45,15 +54,18 @@ class NoDeathdate(Rule):
...
+            death = db.get_event_from_handle(death_ref.ref)
+            if death:
+                death_obj = death.date
+                if not death_obj:
+                    return True
+                if death_obj.sortval == 0:
+                    return True
+            return False
+        else:
+            return True
```
cdhorn, line 71:
```suggestion
        return True
```
Unnecessary else.

gramps/gen/filters/rules/person/_relationshippathbetweenbookmarks.py
```diff
@@ -58,56 +68,57 @@ class RelationshipPathBetweenBookmarks(Rule):
...
         if person is None:
             return None
-        try:
-            name = person.get_primary_name().get_name()
-        except:
+
+        if person.primary_name:
+            return person.primary_name.name
+        else:
             return None
```
cdhorn, line 97:
```suggestion
        return None
```
Unnecessary else.

gramps/gen/filters/rules/family/_fatherhasnameof.py
dsblank: I think it is more explicit, and "Explicit is better than implicit."

gramps/gen/filters/rules/family/_hastwins.py
dsblank: I think it is more explicit, and "Explicit is better than implicit."

gramps/gen/filters/rules/family/_motherhasidof.py
dsblank: I think it is more explicit, and "Explicit is better than implicit."

gramps/gen/filters/rules/family/_motherhasnameof.py
dsblank: I think it is more explicit, and "Explicit is better than implicit."

gramps/gen/filters/rules/family/_regexpfathername.py
dsblank: I think it is more explicit, and "Explicit is better than implicit."

gramps/gen/filters/rules/family/_regexpmothername.py
dsblank: I think it is more explicit, and "Explicit is better than implicit."

gramps/gen/filters/optimizer.py
dsblank: I think it is more explicit, and "Explicit is better than implicit."

gramps/gen/filters/optimizer.py
dsblank: I think it is more explicit, and "Explicit is better than implicit."

dsblank:
Thanks @cdhorn for the reviews! I'm not going to change the if/else constructs as I believe they are more clear and explicit.

cdhorn:
> Thanks @cdhorn for the reviews! I'm not going to change the if/else constructs as I believe they are more clear and explicit.

You're welcome. I always tend to fix these out of habit since pylint and a Sonarcube scan will call them out for refactoring.

Is nice seeing all the great work you, Steven and Nick have been doing.

stevenyoungs:
>I always tend to fix these out of habit since pylint and a Sonarcube scan will call them out for refactoring.

Maybe this is something that should incorporated into the pipeline.

dsblank:
> >I always tend to fix these out of habit since pylint and a Sonarcube scan will call them out for refactoring.
> 
> Maybe this is something that should incorporated into the pipeline.
> 
> 
> 

Please no. There are lots of styles, but not this one.

Nick-Hall:
@dsblank I tried to fix a couple of conflicts, but now the unit tests are failing.  It looks like the problem is due to the new type checking.  Feel free to check or revert my merge.

dsblank:
Ok, I'll take a look

dsblank:
@Nick-Hall, types now adjusted. And a bug found via typing checks (a method was spelled wrong). I had to use a few `# type: ignore` as I couldn't figure out the correct way to handle conflicts.

This probably going to be difficult to merge in a few places.

Nick-Hall:
@dsblank Is this ready for me to do a final review?

dsblank:
Yes, all ready!

dsblank:
Do you want me to resolve conflicts?

Nick-Hall:
> Do you want me to resolve conflicts?

Yes, Please.

dsblank:
@Nick-Hall, I corrected the diffs in the webUI and now everything passes tests, but now it wants to do a rebase via command line. But `git rebase master` seems like it wants to reply every commit. Any suggestions?

Nick-Hall (changes requested):
This PR looks good. I clearly have to thank previous reviewers for that.

Just a couple of things I noticed:

1. The `po/POTFILES.skip` file needs updating.
* Add `gramps/gen/filters/optimizer.py`
* Remove `gramps/gen/filters/rules/family/_memberbase.py`

2. The typing modules in `gramps/gen/filters/rules` are imported using absolute imports when they should be using relative imports.

Nick-Hall:
> I corrected the diffs in the webUI and now everything passes tests, but now it wants to do a rebase via command line. But git rebase master seems like it wants to reply every commit. Any suggestions?

The webUI creates branch merges which we don't want in our commit history.  The workflow that I recommend is to rebase from the command line on a regular basis and force push back to the branch.

However, your workflow doesn't actually cause a problem.  To merge, I would do a `git merge --squash` followed by a `git commit`.  I would tend to use the PR title and contents of the first comment for the commit message.  I don't mind doing this for you.

If you wanted to have several commits in a PR then, you would need to avoid the branch merges.  With multiple commits, I still rebase, but then I commit with the `--no-ff` option to group the commits, but preserve a linear history.  A linear history helps if we need to bisect to find a commit that introduced a bug.

We also need to remember that the commit messages are used to create the ChangeLog and the first and last lines (and possibly more) become part of the NEWS file and release notes.  I often have to remind developers about this.

Finally, I was going to mention that we normally work in our own forks now rather than creating branches in origin.  Again, I doesn't actually cause a problem, so I didn't think it was worth mentioning earlier.

dsblank:
> I don't mind doing this for you.

Thanks, I would appreciate that!

> I clearly have to thank previous reviewers for that.

Absolutely. And they aren't too numerous to name: @cdhorn, @stevenyoungs, and @fxtmtrsine.

dsblank:
And thanks to @emyoulation and others for emoji encouragement and positive feedback :)

Nick-Hall (approved):

dsblank:
@Nick-Hall, you can take it from here? Let me know if there is still resolving to be done.

Nick-Hall:
Squashed and rebased.

QuLogic:
> A linear history helps if we need to bisect to find a commit that introduced a bug.

`git bisect` is quite happy to work without linear histories; that's not directly a reason to require linear history. What makes bisection difficult is non-atomic commits (i.e., ones that are broken and fixed up later), which squash merging can help with but has other trade-offs.

azrdev:
All subclasses of Rule now have to rename their implementations of `apply` to `apply_to_one`, and change any invocation on `MatchesFilter.apply` likewise (which silently returns always true, apparently)

This is not documented anywhere, or is it?  Because it's relevant for any external (out of source tree) filter plugins

dsblank:
> All subclasses of Rule now have to rename their implementations of apply to apply_to_one, and change any invocation on MatchesFilter.apply likewise (which silently returns always true, apparently)

That is true: Gramps 6.0 renamed some methods and fixed many issues. 

> This is not documented anywhere, or is it? Because it's relevant for any external (out of source tree) filter plugins

Good point! This has been an evolving standard, and it is time to document it.

PR #1801, opened 2024-11-28
Update descendtree.py

dave-khuon:
Add an option to use color for the person's box based on the gender.

### DIFF
--- a/gramps/plugins/drawreport/descendtree.py
+++ b/gramps/plugins/drawreport/descendtree.py
@@ -6,6 +6,7 @@
 # Copyright (C) 2010       Jakim Friant
 # Copyright (C) 2009-2010  Craig J. Anderson
 # Copyright (C) 2014       Paul Franklin
+# Copyright (C) 2014       Dave Khuon - option to color the box by gender
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


Call-Me-Dave:
related to https://gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Reports_-_part_4#Descendant_Tree

gramps/plugins/drawreport/descendtree.py
```diff
@@ -6,6 +6,7 @@
 # Copyright (C) 2010       Jakim Friant
 # Copyright (C) 2009-2010  Craig J. Anderson
 # Copyright (C) 2014       Paul Franklin
+# Copyright (C) 2014       Dave Khuon - option to color the box by gender
```
Call-Me-Dave, line 9: Use only your name please

gramps/plugins/drawreport/descendtree.py
```diff
@@ -71,7 +72,11 @@
 _MARR = (_("m.", "marriage abbreviation"),)
 
 _RPT_NAME = "descend_chart"
-
+_BOXCOLOR_DEFAULT = (255,255,255) # default dark
```
Call-Me-Dave, line 75: Can you please use the users default colors already defined in the Colors preference tab see

https://gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Settings#Colors

PR #1804, opened 2024-11-30
Gramps version output now reports OS rather than Platform

hgohel:
Fixes [#12285](https://gramps-project.org/bugs/view.php?id=12285)

Currently `gramps -v` reports the platform instead of O.S. which is confusing to users. Now Gramps will report the operating system name. Reference: [platform.system()](https://docs.python.org/3/library/platform.html#platform.system). There's a difference in capitalization between platform and system, so `linux` is reported as `Linux` and likewise for Darwin.

Current:
```
Gramps Settings:
----------------
gramps    : AIO64-5.2.2-r1-f905d14
o.s.      : win32

```
With this change:
```
Gramps Settings:
----------------
gramps    : AIO64-5.2.2-r1-f905d14
o.s.      : Windows
```

### DIFF
--- a/gramps/grampsapp.py
+++ b/gramps/grampsapp.py
@@ -29,8 +29,8 @@
 # -------------------------------------------------------------------------
 import sys
 import os
+import platform
 import signal
-
 import logging
 
 LOG = logging.getLogger(".")
@@ -534,7 +534,7 @@ def show_settings():
     print("Gramps Settings:")
     print("----------------")
     print(" gramps    : %s" % gramps_str)
-    print(" o.s.      : %s" % sys.platform)
+    print(" o.s.      : %s" % platform.system())
     if kernel:
         print(" kernel    : %s" % kernel)
     print("")


emyoulation:
If the processor information is eliminated fromt eh OS, perhaps there should be a separate "Processor:" line?  

In current Windows, it needs to differentiate between 32bit, 64bit Intel/AMD and 64bit ARM architectures. Since those will require different installers.

hgohel:
Gramps did not have a `Processor` field, and this PR doesn't remove anything. It corrects the fact that `O.S.` was not reporting the OS, but rather the platform which you filed a bug report for :) All other locations in Gramps were modified, but this was missed.

With ARM processors becoming available I can see how knowing that may be useful. That change could be rolled into an initiative to support ARM (GEP?).

jralls:
> That change could be rolled into an initiative to support a native ARM build (GEP?).

Gramps already supports ARM builds, no action required.

emyoulation:
But those are macOS ARM rather than the new Windows ARM, right

jralls:
Of course, but that misses the point. It's just a matter of packaging up Arm builds of Python and the compiled libraries. There are no code changes required.

hgohel:
My limited understanding of Windows on ARM (at this point) is that Windows x86_64 binaries run in emulation mode on Snapdragon processors and for best performance ought to be compiled with an ARM architecture flag.

Happy to take this to the mailing list if we need to discuss further.

emyoulation:
Does that mean that the person on reddit.com/r/gramps who is asking whether Gramps will run natively on their Windows Copilot + AI PC with a Snapdragon processor can test with the  Gramps-Arm-5.2.3-1.dmg  file?

https://github.com/gramps-project/gramps/releases/tag/v5.2.3
https://www.reddit.com/r/gramps/comments/1h13jrc/compatibility_with_windows_on_arm/

jralls:
> Does that mean that the person on reddit.com/r/gramps who is asking whether Gramps will run natively on their Windows Copilot + AI PC with a Snapdragon processor can test with the Gramps-Arm-5.2.3-1.dmg file?

Of course not. macOS binaries don't run on Windows and vice-versa. 

Both macOS and Windows provide opcode translation servers on their Arm OSes that enable users to run x86_64 binaries. I haven't tested it extensively on Windows (a Win11 VM on my Apple Silicon Mac) but on macOS the performance is indistinguishable from an Arm build.

emyoulation:
I did not think so but wanted to double-check. (Remember the PowerPC win/mac cross-over debacle? back in the '90s)

Where should I place a request for a .dmg ARM filename change for 5.3? If there will be macOS and Windows ARM, then the future filenames will need to differentiate.

jralls:
> I did not think so but wanted to double-check. (Remember the PowerPC win/mac cross-over debacle? back in the '90s)

That was the first architecture change, from Moto 68000 to PPC. Apple learned from that and the second one, from PPC to x86_64 in 2006 had a similar opcode translation feature and a pretty smooth transition.

> Where should I place a request for a .dmg ARM filename change for 5.3? If there will be macOS and Windows ARM, then the future filenames will need to differentiate.

What's wrong with the current .dmg names?

When somebody gets around to packaging Gramps for Arm on Windows then they'll have to create separate Gramps-foo-setup.exe just like I do for macOS. I don't see why you'd do anything.

Nick-Hall:
We already use `platform.system()` in the "About" dialog and error report assistant, so this change seems reasonable.

PR #1813, opened 2024-12-05
Notes gramplet update

stevenyoungs:
Several updates to the family of Notes gramplets

- refresh the display when relevant objects are deleted
- refresh the display when notes are added, updated or deleted
- refactor the code to reduce duplication between the various Notes classes
- add the ability to edit and remove the currently displayed note
- add the capability to attach a new note to the end of the note list of the active object

image

### DIFF
--- a/gramps/plugins/gramplet/notes.py
+++ b/gramps/plugins/gramplet/notes.py
@@ -3,6 +3,7 @@
 # Copyright (C) 2011 Nick Hall
 # Copyright (C) 2011 Tim G L Lyons
 # Copyright (C) 2020 Matthias Kemmer
+# Copyright (C) 2024 Steve Youngs
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -32,9 +33,12 @@ from gi.repository import Gtk
 #
 # ------------------------------------------------------------------------
 from gramps.gen.plug import Gramplet
+from gramps.gui.editors import EditNote
 from gramps.gui.widgets.styledtexteditor import StyledTextEditor
 from gramps.gui.widgets import SimpleButton
-from gramps.gen.lib import StyledText
+from gramps.gen.db import DbTxn
+from gramps.gen.errors import WindowActiveError
+from gramps.gen.lib import Note, StyledText
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 
 _ = glocale.translation.gettext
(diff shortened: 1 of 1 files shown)


Nick-Hall:
You need to run `black` on the files you modified.

stevenyoungs:
> You need to run `black` on the files you modified.

Thanks for the pointer @Nick-Hall. PR updated and now passes

Nick-Hall:
@stevenyoungs My only concern is that you are updating the note list outside of an editor.  We don't normally allow this in core code because it may cause a conflict if an editor for the active object is open.

Nick-Hall:
> If I understood you correctly, don't we already have this scenario, at least for deleting a note from a note list?

Yes, and that scenario is potentially more serious leading to a broken link, but I seem to remember that we trap that in the database layer now.

The Forms addon has a similar issue.  We edit attributes in event reference objects outside of the person editor.  This could lead to the loss of attribute data if the user has a person editor open at the same time as the Form editor.  It's one of the reasons that I was hesitant about promoting the Forms addon into core Grmaps.

> Is the Edit function OK? This seems no different to opening the edit note dialog from the Notes view. 

Yes.  You invoke the note editor which is fine.

> If so, I can remove the Add and Delete buttons.

If you remove the Add and Delete buttons then I'll merge this.

stevenyoungs:
> > If I understood you correctly, don't we already have this scenario, at least for deleting a note from a note list?
> 
> Yes, and that scenario is potentially more serious leading to a broken link, but I seem to remember that we trap that in the database layer now.

Let me look into this (I got sidetrack by the redo bug this evening). You _can_ delete the note, from Notes view, whilst the person and note editors are displayed. No error is shown. The edit note dialog automatically closes. The edit person dialog automatically updates. The db still passes "check & repair database".

Worst case, I'll remove the Add and Delete options so the PR can proceed.

Nick-Hall:
One solution I tried in the past was to modify the save method of the editors.  I took the JSON of the object before the edit, and then created a patch from the JSON after the edit.  The resulting JSON patch provided useful debug information on its own.  Then I took a new copy of the object from the database and applied the patch before saving.  Even before our latest change to the raw data structure this approach worked quite well and was surprisingly fast.  Perhaps we should revisit the idea.  I'll look out the code tomorrow.

stevenyoungs:
The entry point for remove in the views is [remove_selected_objects](https://github.com/stevenyoungs/gramps/blob/ee585393ecc5bbb1ac3b2054f84323703939fafe/gramps/gui/views/listview.py#L593) which after after prompting the user calls [delete_object_response](https://github.com/stevenyoungs/gramps/blob/ee585393ecc5bbb1ac3b2054f84323703939fafe/gramps/gui/views/listview.py#L646) which calls [remove_object_from_handle](https://github.com/stevenyoungs/gramps/blob/ee585393ecc5bbb1ac3b2054f84323703939fafe/gramps/gui/views/listview.py#L690)
Replicating this behaviour in this PR looks straightforward if you are prepared to accept that?

I can break the current PR code by deleting a note where the edit note dialog is also open. So this PR should not be merged as-is.

Nick-Hall:
The "+" and "-" buttons are not consistent with other gramplets.  I suggest that we remove this functionality for now.  We can consider it again as part of a more general discussion about editing objects outside of editors.

stevenyoungs:
@Nick-Hall I've removed the Add and Delete note functionality as requested so that this can hopefully make the 6.0 release

Nick-Hall:
@stevenyoungs Thanks.  I plan to look at all your pull requests tomorrow.

PR #1823, opened 2024-12-08
Add gen.db.conversion_tools from PR #1786

Nick-Hall:
The `gen.db.conversion_tools` package is missing from `setup.py` and its files from `po/POTFILES.skip`.


dsblank (approved):
Looks good to me.

stevenyoungs:
Looks OK to me. 
I quickly looked at the AIO build and I don't see anywhere that requires a corresponding change. It would be good if someone with more knowledge of the Windows build could confirm.

PR #1824, opened 2024-12-11
Added a dict wrapper that acts like an object

dsblank:
@Nick-Hall had [an interesting idea](https://github.com/gramps-project/gramps/pull/1794#issuecomment-2510311435) to make the raw dict from JSON act like an object. In that way, instead of writing:

```
if data["gender"] == Person.MALE:
```

you could write:

```
if data.gender == Person.MALE:
```

And if you rename `data` to `person` it looks like regular gramps for all of the attributes of all of the Gramps objects:

```
if person.gender == Person.MALE:
```

This has all of the benefits of quick access (no object creation) with a convenient syntax.

However that could create some confusion because methods can't be called as it isn't a real `Person`. With a couple of additional lines, it can act as a real object:

```
if person.get_gender() == Person.MALE:
```

This allows us to use the full object attribute and methods, only creating an object when necessary. Sometimes it is necessary, such as:

```
person.person_ref_list[0].get_referenced_handles()
```

(Although this only instantiates a PersonRef(), much cheaper than instantiating the whole Person)

With this PR applied, I found that `person.get_gender()` was 4.4 times slower than `person.gender`. The only downside with this PR is that those writing code (filter rules, for example) might not realize that they are creating an object when it is not necessary.

### DIFF
--- a/gramps/gen/lib/serialize.py
+++ b/gramps/gen/lib/serialize.py
@@ -42,6 +42,51 @@ import gramps.gen.lib as lib
 LOG = logging.getLogger(".serialize")
 
 
+class DataDict(dict):
+    """
+    A wrapper around a data dict that also provides an
+    object interface.
+    """
+
+    def __str__(self):
+        if "_object" not in self:
+            self["_object"] = from_dict(self)
+        return str(self["_object"])
+
+    def __getattr__(self, key):
+        if key.startswith("_"):
+            raise AttributeError("can't use this API to access hidden attributes")
+
+        if key in self:
+            value = self[key]
+        else:
+            if "_object" not in self:
+                self["_object"] = from_dict(self)
+            return getattr(self["_object"], key)
+
+        if isinstance(value, dict):
+            return DataDict(value)
+        elif isinstance(value, list):
+            return DataList(value)
+        else:
+            return value
+
+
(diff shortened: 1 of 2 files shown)


dsblank:
@stevenyoungs no data leakage! :)

dsblank:
@Nick-Hall, some more decisions to make:

1. Do we want to allow attribute syntax (`data.gender`) instead of dictionary syntax (`data["gender"]`)? I think it is a nice enhancement, and is a very minimal change and cost.
2. Do we want to allow method-calling (`data.get_gender()`)? Although it hides object creation, it eliminates confusion, is sometimes needed, and is easy to implement.
3. Shall we go with this implementation? I can't think of a better method to accomplish the above.

Let me know what you think, and I'll go with the decision on the next set of changes.

Nick-Hall:
@dsblank I'm happy with all of this.  The attribute syntax is neat and concise, method-calling could be useful, and I don't have any objections to your implementation.

I was waiting for other comments before adding my opinion.

dsblank:
I'll work on adding tests, and prepare for moving from draft status.

dsblank:
Tests added, ready for review.

Nick-Hall (approved):

PR #1825, opened 2024-12-12
Teach GrampsType.set() to work with a dict

stevenyoungs:
Teach GrampsType.set() how to work when value is of `dict` type
This fixes a bug introduced by PR #1786 

To reproduce:
1. create a new, empty, tree
2. switch to event view
3. create a new event, changing the event type to Death
4. observe that the event type is shown as `Birth` in the event view

### DIFF
--- a/gramps/gen/lib/grampstype.py
+++ b/gramps/gen/lib/grampstype.py
@@ -151,6 +151,10 @@ class GrampsType(metaclass=GrampsTypeMeta):
         else:
             self.__string = ""
 
+    def __set_dict(self, value):
+        "Set the value/string properties from a dict."
+        self.__set_tuple((value["value"], value["string"]))
+
     def __set_tuple(self, value):
         "Set the value/string properties from a tuple."
         val, strg = self._DEFAULT, ""
@@ -184,7 +188,9 @@ class GrampsType(metaclass=GrampsTypeMeta):
 
     def set(self, value):
         "Set the value/string properties from the passed in value."
-        if isinstance(value, tuple):
+        if isinstance(value, dict):
+            self.__set_dict(value)
+        elif isinstance(value, tuple):
             self.__set_tuple(value)
         elif isinstance(value, int):
             self.__set_int(value)
@@ -228,6 +234,25 @@ class GrampsType(metaclass=GrampsTypeMeta):
         return (self.__value, self.__string)
 
     @classmethod
+    def __get_str(cls, value, string):
+        if value == cls._CUSTOM:
+            return string
+        else:
+            return cls._I2SMAP.get(value, _UNKNOWN)
+
+    @classmethod
+    def get_str(cls, data):
+        """
+        Return the translated string corresponding to the type.
+
+        :param data: A dict representation of the type.
+        :type data: dict
+        :returns: The translated string corresponding to the type.
+        :rtype: str
+        """
+        return cls.__get_str(data.value, data.string)
+
+    @classmethod
     def get_schema(cls):
         """
         Returns the JSON Schema for this class.
@@ -253,9 +278,7 @@ class GrampsType(metaclass=GrampsTypeMeta):
         return self
 
     def __str__(self):
-        if self.__value == self._CUSTOM:
-            return self.__string
-        return self._I2SMAP.get(self.__value, _UNKNOWN)
+        return type(self).__get_str(self.__value, self.__string)
 
     def __int__(self):
         return self.__value
--- a/gramps/gui/views/treemodels/eventmodel.py
+++ b/gramps/gui/views/treemodels/eventmodel.py
@@ -145,7 +145,7 @@ class EventModel(FlatBaseModel):
             return ""
 
     def column_type(self, data):
-        return str(EventType(data["type"]))
+        return EventType.get_str(data.type)
 
     def column_id(self, data):
         return data["gramps_id"]
--- a/gramps/gui/views/treemodels/repomodel.py
+++ b/gramps/gui/views/treemodels/repomodel.py
@@ -134,7 +134,7 @@ class RepositoryModel(FlatBaseModel):
         return data["gramps_id"]
 
     def column_type(self, data):
-        return str(RepositoryType(data["type"]))
+        return RepositoryType.get_str(data.type)
 
     def column_name(self, data):
         return data["name"]
--- a/gramps/gui/views/treemodels/familymodel.py
+++ b/gramps/gui/views/treemodels/familymodel.py
@@ -165,7 +165,7 @@ class FamilyModel(FlatBaseModel):
         return value
 
     def column_type(self, data):
-        return data["type"]["string"]
+        return FamilyRelType.get_str(data.type)
 
     def column_marriage(self, data):
         handle = data["handle"]
--- a/gramps/gui/views/treemodels/notemodel.py
+++ b/gramps/gui/views/treemodels/notemodel.py
@@ -118,9 +118,7 @@ class NoteModel(FlatBaseModel):
 
     def column_type(self, data):
         """Return the type of the Note in readable format."""
-        temp = NoteType()
-        temp.set(data["type"])
-        return str(temp)
+        return NoteType.get_str(data.type)
 
     def column_preview(self, data):
         """Return a shortend version of the Note's text."""
--- a/gramps/gui/views/treemodels/placemodel.py
+++ b/gramps/gui/views/treemodels/placemodel.py
@@ -188,8 +188,7 @@ class PlaceBaseModel:
         return data["gramps_id"]
 
     def column_type(self, data):
-        pt = from_dict(data["place_type"])
-        return str(pt)
+        return PlaceType.get_str(data.place_type)
 
     def column_code(self, data):
         return data["code"]


stevenyoungs:
@dsblank I'd appreciate your review of this
I believe the problem stems from the most recent change to eventmodel.py
`return str(EventType(data[COLUMN_TYPE]))`
was changed to 
`return str(EventType(data["type"]))`

`data[COLUMN_TYPE]` is a `tuple` whilst `data["type"]` is a `dict`.
GrampsType.set() does not know how to set for a `dict`. I'm assuming there will be other instances of this problem in the code base - but I've not verified this statement!

stevenyoungs:
The same problem exists with notes and repositories views.

dsblank:
@stevenyoungs good catch! Yes, you are correct... another place where the array/blob representation had leaked into other parts of the code.

What would be better, though, would be to not create an object if it isn't necessary, and just get the results needed. That might be something like:

```python
def get_event_type_from_raw_data(data):
    return what_ever_is_needed
````

**Comment by dsblank** [comment]
But if that isn't possible, yes, let's go with your approach.

**Comment by dsblank** [comment]
Can't we just `from_dict(data["type"])` ?

**Comment by stevenyoungs** [comment]
> @stevenyoungs good catch! Yes, you are correct... another place where the array/blob representation had leaked into other parts of the code.
> 
> What would be better, though, would be to not create an object if it isn't necessary, and just get the results needed. That might be something like:
> 
> ```python
> def get_event_type_from_raw_data(data):
>     return what_ever_is_needed
> ```

How about this, which allows us to avoid building the object, and also hides the detail inside a static method in EventType

```
    @staticmethod
    def get_str(value):
        if value["value"] == EventType._CUSTOM:
            return value["string"]
        else:
            return EventType._I2SMAP.get(value["value"], EventType.UNKNOWN)
```

and then it becomes

```
    def column_type(self, data):
        return EventType.get_str(data["type"])
```

Thoughts?
Is it worth leaving the existing change in grampstype in case other (external) code has the same problem?

**Comment by dsblank** [comment]
I'm going to do some timing tests between turning just the data["type"] into an instance, versus your static method. If they are not that significantly different, I think we'd just create an object, rather than having to have yet another method to access the data.

**Comment by dsblank** [comment]
Timing tests comparing converting to an object (using an updated #1824) vs static method (above) with Example tree:

Setup:

```python
from gramps.gen.db.utils import open_database
from gramps.gen.lib.eventtype import EventType
db = open_database("Example")
```

Turning into an object (uses old, standard Gramps code; no changes to code base once #1824 is merged):

```ipython
In [16]: %%timeit
    ...: for handle, event in db._iter_raw_event_data():
    ...:     x = str(event.type)
    ...: 
65.8 ms ± 1.02 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Using the static method (requires an import):

```ipython
In [20]: %%timeit
    ...: for handle, event in db._iter_raw_event_data():
    ...:     x = EventType.get_str(event.type)
    ...: 
    ...: 
32.2 ms ± 573 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

So, the static method is twice as fast (saves 30 ms over 3,432 events), but at the cost of having to know about a special static method.

Personally, I don't think it is worth it, but I support you in whatever direction you want to go.

dsblank:
@stevenyoungs, what do you want to do to fix the bugs that you found?

stevenyoungs:
@dsblank  I'll hopefully get some time over the coming days to come back to this.
My current thinking was to provide both the current as well as the staticmethod approach.
For a large tree, I'm assuming that the standard views want the type frequently enough to justify providing both approaches. Let me know if you have any concerns about this.

dsblank:
> For a large tree, I'm assuming that the standard views want the type frequently enough to justify providing both approaches.

That sounds fine to me! It is there if one wants to use it. #1824 has been merged, so either of the code examples above will work.

stevenyoungs:
@dsblank I revised it to an @classmethod and implemented in GrampsType. This way it works for all types, not just EventType
Relevant model classes updated and their column_type methods made consistent
I'd appreciate your review.

gramps/gen/lib/grampstype.py
```diff
@@ -227,6 +239,17 @@ def serialize(self):
...
+    @classmethod
+    def __get_str(cls, value, string):
+        if value == cls._CUSTOM:
+            return string
+        else:
+            return cls._I2SMAP.get(value, _UNKNOWN)
+
+    @classmethod
+    def get_str(cls, dict):
+        return cls.__get_str(dict["value"], dict["string"])
```
dsblank, line 251: It is usually good to avoid using Python names as parameter names. We've been calling the dict representation "data" in other places. How about:

```suggestion
    def get_str(cls, data):
        return cls.__get_str(data["value"], data["string"])
```

Or even:

```suggestion
    def get_str(cls, data):
        return cls.__get_str(data.value, data.string)
```
with the new DataDict wrapper.

gramps/gen/lib/grampstype.py
```diff
@@ -151,6 +151,16 @@ def set_object_state(self, attr_dict):
         else:
             self.__string = ""
 
+    def __set_dict(self, value):
```
dsblank, line 154: Why use the double underscore? It isn't required.

gramps/gen/lib/grampstype.py
```diff
@@ -227,6 +239,17 @@ def serialize(self):
         """Convert the object to a serialized tuple of data."""
         return (self.__value, self.__string)
 
+    @classmethod
+    def __get_str(cls, value, string):
+        if value == cls._CUSTOM:
+            return string
+        else:
+            return cls._I2SMAP.get(value, _UNKNOWN)
```
dsblank, line 247: This core code could also live in a utility function, and called from here.

gramps/gui/views/treemodels/eventmodel.py
```diff
@@ -145,7 +145,7 @@ def column_place(self, data):
             return ""
 
     def column_type(self, data):
-        return str(EventType(data["type"]))
+        return EventType.get_str(data["type"])
```
dsblank, line 148: Why does it have to be `EventType.get_str()` rather than `GrampsType.get_str()`? Is it because each type has a `_CUSTOM`? But isn't it always the same value?

dsblank (approved):
I left some questions/comments, but it is better than before, and also fixes a bug or two. Thanks!

gramps/gen/lib/grampstype.py
stevenyoungs: I was maintaining consistency with the existing `__set_*()` methods.
From the naming I assume that the original design intention was to call users to call `set()`, even where the type of `value` was already know. I'll address this as a separate PR

gramps/gui/views/treemodels/eventmodel.py
stevenyoungs: I believe _CUSTOM is common for all types.
`GrampsType.__get_str`, which this call ultimately ends up in, needs to types `_I2SMAP`. `_I2SMAP` differs for each type.

gramps/gen/lib/grampstype.py
stevenyoungs: See PR #1830

gramps/gen/lib/grampstype.py
dsblank: I don't feel strongly.

gramps/gui/views/treemodels/eventmodel.py
```diff
@@ -145,7 +145,7 @@ def column_place(self, data):
             return ""
 
     def column_type(self, data):
-        return str(EventType(data["type"]))
+        return EventType.get_str(data.type)
```
stevenyoungs, line 148: Which style of attribute access is going to recommend going forwards?
DataDict as in this line `data.type` or standard dict[key] style as found a couple of lines below `data["gramps_id"]`

dsblank: In the PR to revise the filter system, I changed all to `data.attribute` which works for whatever you pass in (obj or data).

Nick-Hall: Yes.  I agree that the `data.attribute` style is preferable.

stevenyoungs:
This is a bug fix for #1786 (which was an enhancement). Can someone add the bug label please?

dsblank:
Ready to merge, IMO.

gramps/gen/lib/grampstype.py
Nick-Hall, line 154: Why doesn't `__set_dict` call `__set_tuple`?

The method would then only need to consist of a single line:
```
self.__set_tuple((value.value, value.string))
```

gramps/gui/views/treemodels/repomodel.py
```diff
@@ -134,7 +134,7 @@ def column_id(self, data):
         return data["gramps_id"]
 
     def column_type(self, data):
-        return str(RepositoryType(data["type"]))
+        return RepositoryType.get_str(data.type)
```
Nick-Hall, line 137: I would prefer to see `return str(RepositoryType(data.type))` here, but I can understand why you don't want to instantiate an object.

Nick-Hall (changes requested):
Only one real issue.

gramps/gui/views/treemodels/repomodel.py
stevenyoungs: Agreed, just trying to avoid instantiating objects.
The internal implementation is common. `get_str` calls the class method `__get_str`. `__str__` also calls `__get_str`.

stevenyoungs:
@Nick-Hall this is ready for re-review. Thanks.

Nick-Hall (approved):

Nick-Hall:
Squashed commits and added a docstring.

PR #1821, opened 2024-12-07
WIP: AIO build updates

hgohel:
1. Changes required due to updating MSYS2 environment (e.g. gspell version, cx-freeze etc.)
2. Remove references to lib2to3 (deprecated)
3. Reduce log output from build.sh so errors and warnings are easier to spot
4. Cleanup README

This is **not ready to merge**, because the AIO build doesn't complete without warnings/errors. For now I've marked it as a draft PR so others can continue working on it until the changes are complete.

FYI @ennoborg

PR #1826, opened 2024-12-13
Msys2 update changes for aio

stevenyoungs:
This PR builds upon PR #1821 
aio/build.sh now completes and produces an AIO installer.
The installer works and gramps is installed.
gramps "starts" but fails during initialisation of the gui:

```python
2024-12-13 12:42:42.967: ERROR: grampsapp.py: line 188: Unhandled exception
  File "AIO/gramps/gui/grampsgui.py", line 768, in do_activate
  File "AIO/gramps/gui/grampsgui.py", line 581, in __init__
  File "AIO/gramps/gui/viewmanager.py", line 576, in init_interface
  File "AIO/gramps/gui/viewmanager.py", line 858, in goto_page
IndexError: list assignment index out of range
``` 
The GUI is at least running - so it's close

image

### DIFF
--- a/aio/README.md
+++ b/aio/README.md
@@ -1,25 +1,30 @@
-The files in this directory are used to build the Windows AIO (All In One) installer.
+The files in this directory are used to build the Gramps Windows AIO (All In One) installer.
 
 To build AIO for the master branch :
 
-1. install msys2
-* download msys2 from <https://www.msys2.org/> .
-* install with default options. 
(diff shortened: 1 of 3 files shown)


stevenyoungs:
Current status: AIO installer builds and runs. Gramps starts and all views at least display.
The example.gramps file can be successfully imported.

It would be useful to know if anyone else can build the AIO installer now.

PR #1827, opened 2024-12-17
Add Isotammi addons to Project Manager defaults

kkujansuu:


emyoulation:
@Nick-Hall 
When I tried to add to the default Projects earlier, this added the Isotammi project during a clean install. But the Reset button in the Addon Manager still only included the main gramps-project collection.

How do we make the reset button use the same defaults?

Call-Me-Dave:
As much as I like this personally I believe that the [Restore project defaults](https://gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Navigation#Restore_project_defaults) button in the Addon Manager  should only restore the **default** managed by Gramps project addon repository, all other repositories like Isotammi  etc should be left as entries but be unchecked when **restore (Gramps) project defaults** is pressed , instead of the current behavior that deletes all my added external repositories ! 

Question should the Gramps project really have external addon urls that can not be verified and would only really be for advanced users!

By changing the behavior of the **restore project default**  option as described above we get around the issue of losing those entries!  

Another option is to be able to load a file that has those optional repositories listed?

emyoulation:
> Question should the Gramps project really have external addon urls that can not be verified and would only really be for advanced users!

Yes, I think it should. But only after the URL has been validated by someone in the Gramps team and with it deselected by default.

We need an effective example of a 2nd addon collection to demonstrate the value of the Addon Manager... and the Isotammi project demonstates incredible value and how a Genealogical Society can leverage Gramps towards their own needs.

I agree about not wanting to purge my custom Projects. But there is a distinct need for a reset when novices mess up their addon manager

wroldwiedbwe:
@Call-Me-Dave Yes I agreed change the how restore works and only keep the default from the project.  I like the idea of having a file to load all the other external repositories.

> Yes, I think it should. But only after the URL has been validated by someone in the Gramps team and with it deselected by default.

The problem is what happens when the repository is malicious, sure you verified it , but what happens 10 minutes later, you can really only verify Gramps projects repository the aim should be to encourage other developers to eventually add what they have created to the main repository for normal users!  Look at all the troubles with other projects like pythons pypi repository or even ruby's RubyGems repository, event androids store and apples store with malicious apps etc! 

> We need an effective example of a 2nd addon collection to demonstrate the value of the Addon Manager... and the Isotammi project demonstates incredible value and how a Genealogical Society can leverage Gramps towards their own needs.

I disagree with that false logic! Even that project is a reason why you should not as didn't they announce they disbanded the project, which if anything makes them a good candidate to roll it into the main repository especially for discovery?

> I agree about not wanting to purge my custom Projects. But there is a distinct need for a reset when novices mess up their addon manager

Maybe the restore option could be two level ,  first option only restores gramps repository and disables all the other respositories but keeps them and the full reset option deletes everything else but keeps Gramps.   

A lot more though should be given to this before going any further and this should be discussed else where.

emyoulation:
> didn't they announce they disbanded the project

The Isotammi announced discontinuation of hosting their Server.  (Used as a self-service uploader pipeline to their curated Finnish genealogical data.) It wasn't getting sufficient traffic to justify the expense.

Meanwhile, the addon collection on GitHub does not has a cost to be justified.

giotodibondone:
> Another option is to be able to load a file that has those optional repositories listed?

@Call-Me-Dave  Why not just package the files normal gramps addon library file that can be installed via the addon manager and if the addon manager detects the json file or some setting from the addon it shows them in the addon project list.

This would have the benefit of the Gramps project managing that list separately from the distribution of the program and being able to update the list as needed?

emyoulation:
> Why not just package the files normal gramps addon library file that can be installed via the addon manager

There's a reasonable justification for this approach.

However, the distribution a Rules for filters is a similar concept and hasn't  worked smoothly. 

Moreover, distribution of Custom Filters, books of Reports, and packets of sources/places/SuperTool_scripts/HistoicalContext content are other features that could use an elegant sharing functionality. 

So that needs discussion.

The expandable Projects feature needs exercising to see if it works in the field as  envisioned. I strongly suspect that a broad field test will reveal usability will be less than hoped.

PQYPLZXHGF:
> However, the distribution of Rules for filters is a similar concept and hasn't worked smoothly. (We've had 2 rule packs and the additional rules are not easily discovered.)

Nothing like what is being suggested here, as with the rules they need a bit of refinement in the way Gramps treats them when you have created a rule/filter that references the addon rule but it is not installed, gramps should recognize ( have some type of placeholder) that says hey that filter needs the rule pack installed to work before running!

> Why not just package the files normal gramps addon library file that can be installed via the addon manager and 

The addon could simply be a plain (*.gpr.py) file packaged with a **externalrepositories.json** file. 

Hell you could even have a second *json file (say **update.json**) that has details of the current version of Gramps and if updated to a newer version of Gramps  , have a popup in Gramps that mentions a new version of Gramps is out ?

wroldwiedbwe:
> The addon could simply be a plain (*.gpr.py) file packaged with a externalrepositories.json file.
> 
> Hell you could even have a second *json file (say update.json) that has details of the current version of Gramps and if updated to a newer version of Gramps , have a popup in Gramps that mentions a new version of Gramps is out ?

@emyoulation Just extend what @PQYPLZXHGF suggested to handle all that link rot!   Have various *.json" or whatever files that overide the builtin ones from the addon or Gramps itself for the mapservices etc   Or distribute the addons without the link rotting  urls etc  and have those addons require @PQYPLZXHGF suggested addon as a single place to collect those links for update  (including exiting SuperTool scripts/collections if needed )?

Call-Me-Dave:
I guess no discussion from leader @Nick-Hall was the way to go!

Nick-Hall:
@Call-Me-Dave Is anyone objecting to adding the Isotammi entry to the list of projects?  I can easily revert this.

wroldwiedbwe:
> @kkujansuu this pr should target gramps master

I'm guessing this from the comment above, and I kind of agree as you are technically adding a feature to an old release instead of to the new release!

Adding this basically blesses the Isotammi entry as approved and run by Gramps project as users are not going to see the difference especially given the behaviors for the restore button, if any thing [the available external projects list](https://gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Navigation#Other_Projects) should be downloaded as part of the addon list as mentioned above!

Some more work is needed.

I can understand the snippy comment, as why have a discussion, just to be ignored!

Nick-Hall:
I'm sorry.  This PR just wanted to add Isotammi to the projects list.  The discussion appeared to be about enhancements intended for the _master_ branch.

emyoulation:
The "Projects" feature is all about having options outside the glacial "blessing" of the gramps‐project addons. 

And if the Isotammi group hasn't proven themselves to be a trustworthy asset to this open source project, then I fear that doing so is utterly impossible.

As I posted earlier, without a built‐in example of an alternate repository, this feature will not be exercised enough to give broad feedback. 

I agree that it needs additional work. (Including the reset, reducing the redundant hard-coded entries, feedback that the project connection finds a addonlist-xx.json and in which language, a local addonlist-en.json for the built-in plugins for search and help, and more.)

But how will those gain advocates without sample data to give a richer exposure to the features?

We need the innovation of Isotammi project more than they need us.

PQYPLZXHGF:
> @Call-Me-Dave Is anyone objecting to adding the Isotammi entry to the list of projects? I can easily revert this.

@Nick-Hall Yes revert, add it to the next major release, and I suggest adding each of the other known projects also instead of showing a single preference!

emyoulation:
Think of the timeline...

If this is pushed to the  next major release, then there will not be more than the wishy-washy feedback that has been filed since 5.2.0, a year ago. 

And since there's no agreement about the feature other than adding an example project, there will only be superficial changes. And the Projects becoming truly useful will be pushed to 6.1... probably in 2026. 

Although more likely further than that because this is a UX enhancement. UX enhancements are always pushed to end of the development cycle when the functional base is solid and the GUI needs are defined. And NotYetImplemented UX features are the first to be cut when people are clamoring for the new functionalities.

daleathan:
@Nick-Hall Revert it, makes no sense tacking it on to the last 5.2.x release!

Nick-Hall:
Reverted in commit 2a92670694c0ed42a37885b29f2ecabac2d76763.

PR #1828, opened 2024-12-21
Added db.dbi.select using ast

dsblank:
Work in progress

### DIFF
--- a/gramps/plugins/db/dbapi/dbapi.py
+++ b/gramps/plugins/db/dbapi/dbapi.py
@@ -65,6 +65,8 @@ from gramps.gen.lib.serialize import from_dict, to_dict
 from gramps.gen.lib.genderstats import GenderStats
 from gramps.gen.updatecallback import UpdateCallback
 
+from .select import Evaluator
+
 LOG = logging.getLogger(".dbapi")
 _LOG = logging.getLogger(DBLOGNAME)
 
@@ -1183,3 +1185,17 @@ class DBAPI(DbGeneric):
         in the appropriate type.
         """
         return [v if not isinstance(v, bool) else int(v) for v in values]
+
+    def select(self, table_name, where, order_by=[], env=None):
+        """
+        Select json_data from table_name where (string).
+        """
+        evaluator = Evaluator(
+            table_name,
+            'json_data->>"$.{attr}"',
+            env if env is not None else globals()
+        )
+        where_expr = str(evaluator.convert(where))
+        order_by_expr = evaluator.get_order_by(order_by)
+        self.dbapi.execute(f"SELECT json_data from {table_name} WHERE {where_expr}{order_by_expr};")
+        return [self.serializer.string_to_data(row[0]) for row in self.dbapi.fetchall()]
--- a/gramps/gen/lib/serialize.py
+++ b/gramps/gen/lib/serialize.py
@@ -47,6 +47,10 @@ class DataDict(dict):
     A wrapper around a data dict that also provides an
     object interface.
     """
+    def __str__(self):
+        if "_object" not in self:
+            self["_object"] = from_dict(self)
+        return str(self["_object"])
 
     def __getattr__(self, key):
         if key.startswith("_"):
--- a/gramps/gen/lib/serialize.py
+++ b/gramps/gen/lib/serialize.py
@@ -132,7 +132,7 @@ def to_dict(obj):
     :returns: A dictionary.
     :rtype: dict
     """
-    return json.loads(to_json(obj))
+    return DataDict(json.loads(to_json(obj)))
 
 
 def from_dict(dict):
--- a/gramps/gen/lib/serialize.py
+++ b/gramps/gen/lib/serialize.py
@@ -47,6 +47,7 @@ class DataDict(dict):
     A wrapper around a data dict that also provides an
     object interface.
     """
+
     def __str__(self):
         if "_object" not in self:
             self["_object"] = from_dict(self)
--- a/gramps/gen/lib/serialize.py
+++ b/gramps/gen/lib/serialize.py
@@ -71,6 +71,7 @@ class DataList(list):
     """
     A wrapper around a data list.
     """
+
     def __getitem__(self, position):
         value = super().__getitem__(position)
         if isinstance(value, dict):
--- a/gramps/gen/lib/serialize.py
+++ b/gramps/gen/lib/serialize.py
@@ -42,6 +42,45 @@ import gramps.gen.lib as lib
 LOG = logging.getLogger(".serialize")
 
 
+class DataDict(dict):
+    """
+    A wrapper around a data dict that also provides an
+    object interface.
+    """
+
+    def __getattr__(self, key):
+        if key.startswith("_"):
+            raise AttributeError("can't use this API to access hidden attributes")
+
+        if key in self:
+            value = self[key]
+        else:
+            if "_object" not in self:
+                self["_object"] = from_dict(self)
+            return getattr(self["_object"], key)
+
+        if isinstance(value, dict):
+            return DataDict(value)
+        elif isinstance(value, list):
+            return DataList(value)
+        else:
+            return value
+
+
+class DataList(list):
+    """
+    A wrapper around a data list.
+    """
+    def __getitem__(self, position):
+        value = super().__getitem__(position)
+        if isinstance(value, dict):
+            return DataDict(value)
+        elif isinstance(value, list):
+            return DataList(value)
+        else:
+            return value
+
+
 def __object_hook(obj_dict):
     _class = obj_dict.pop("_class")
     cls = lib.__dict__[_class]
@@ -174,7 +213,7 @@ class JSONSerializer:
     @staticmethod
     def string_to_data(string):
         LOG.debug("json, string_to_data: %r...", string[:65])
-        return json.loads(string)
+        return DataDict(json.loads(string))
 
     @staticmethod
     def object_to_string(obj):
--- a/gramps/gen/lib/test/serialize_test.py
+++ b/gramps/gen/lib/test/serialize_test.py
@@ -115,23 +115,39 @@ class DatabaseCheck(unittest.TestCase):
     maxDiff = None
 
 
-def generate_case(obj):
+def generate_cases(obj, data):
     """
     Dynamically generate tests and attach to DatabaseCheck.
     """
-    data = to_json(obj)
-    obj2 = from_json(data)
+    json_data = to_json(obj)
+    obj2 = from_json(json_data)
 
     def test(self):
         self.assertEqual(obj.serialize(), obj2.serialize())
 
     name = "test_serialize_%s_%s" % (obj.__class__.__name__, obj.handle)
     setattr(DatabaseCheck, name, test)
-    ####
-    # def test2(self):
-    #    self.assertEqual(obj.serialize(), from_dict(struct).serialize())
-    # name = "test_create_%s_%s" % (obj.__class__.__name__, obj.handle)
-    # setattr(DatabaseCheck, name, test2)
+
+    def test_data(self):
+        class_name = obj.__class__.__name__
+        assert isinstance(data, dict), "Ensure that the data is a dict"
+        assert data.handle == data["handle"], "Test attribute access"
+
+        if class_name == "Person":
+            if (len(data.parent_family_list)) > 0:
+                # Get a handle:
+                assert isinstance(data.parent_family_list[0], str), "Test list access"
+
+        assert "_object" not in data.keys(), "Object not created"
+        assert data.get_handle() == data["handle"], "Test method call"
+        assert "_object" in data.keys(), "Object created"
+        assert data["_object"].handle == data["handle"], "Object is correct"
+        assert (
+            data["_object"].__class__.__name__ == class_name
+        ), "Object is correct type"
+
+    name = "test_data_%s_%s" % (obj.__class__.__name__, obj.handle)
+    setattr(DatabaseCheck, name, test_data)
 
 
 db = import_as_dict(EXAMPLE, User())
@@ -148,7 +164,8 @@ for obj_class in (
 ):
     for handle in db.method("get_%s_handles", obj_class)():
         obj = db.method("get_%s_from_handle", obj_class)(handle)
-        generate_case(obj)
+        data = db.method("get_raw_%s_data", obj_class)(handle)
+        generate_cases(obj, data)
 
 if __name__ == "__main__":
     unittest.main()
--- a/gramps/plugins/db/dbapi/select.py
+++ b/gramps/plugins/db/dbapi/select.py
@@ -1,0 +1,199 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2024 Douglas S. Blank <doug.blank@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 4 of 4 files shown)


gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -1183,3 +1185,17 @@ def _sql_cast_list(self, values):
...
+        """
+        Select json_data from table_name where (string).
+        """
+        evaluator = Evaluator(
+            table_name, 'json_data->>"$.{attr}"', env if env is not None else globals()
+        )
+        where_expr = str(evaluator.convert(where))
+        order_by_expr = evaluator.get_order_by(order_by)
+        self.dbapi.execute(
+            f"SELECT json_data from {table_name} WHERE {where_expr}{order_by_expr};"
```
stevenyoungs, line 1199: If you could make the where_expr optional, if would give a convenient way to retrieve all records from a table
`DBAPI.select("person")` would return all records from person

Discussion originated [here](https://github.com/gramps-project/gramps/pull/1829#issuecomment-2562769708)

dsblank: Indeed. It wasn't clear from this interface, but where="" makes it optional. I changed the interface to make this clear.

dsblank:
I was going to not add the `what` parameter, but it takes a significant amount of time to return, json.loads() the entire object, rather than just have SQL select parts of it. Here we see that just getting the handle is 4 times faster than getting the whole JSON data.

```
In [8]: %%timeit
   ...: for data in db.select_from_person():
   ...:     pass
   ...: 
39.1 ms ± 438 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [9]: %%timeit
   ...: for data in db.select_from_person("person.handle"):
   ...:     pass
   ...: 
10.4 ms ± 305 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

gramps/gen/db/generic.py
```diff
@@ -2739,3 +2739,73 @@ def set_serializer(self, serializer_name):
             self.serializer = BlobSerializer
         elif serializer_name == "json":
             self.serializer = JSONSerializer
+
+    def select_from_table(
+        self, table_name, what=None, where=None, order_by=None, env=None
```
stevenyoungs, line 2744: I know that currently you only pass a single table_name but if you were to make this low level method take an array of table_names then I think everything exists to do cross table selects.
With that it would be possible to add a generic filter UI that allows the user to enter a query defined in terms of tables, what, where and order_by strings. I readily acknowledge that this might be a more "advanced user" feature but it would be *really* powerful.
It will also allow code to execute more complex queries directly within the DB ❤️

dsblank: @stevenyoungs, I like this idea! First, let's make sure that we can get this type of method added, and then think about expanding it.

(I'd like to see an idea of how you would represent a JOIN, but I don't want you to spend too much time if this PR gets rejected.)

dsblank:
@Nick-Hall and other reviewers, I've been thinking for a very long time about how to add a select-style method in Gramps, while keeping with a Python interface. This PR represents the best that I can come up with.

Note that the syntax for the `SELECT` fields, `WHERE` clause and `ORDER BY` fields are all represented as strings of Python expressions. (I was hoping that they wouldn't be strings. I tried lambda functions, but couldn't reliably get the AST from the body of the lambda. I didn't want to use functions for all of the fields and clauses, because that would be a lot of overhead for the user. So, this is my best solution given the constraints.)

The strings are parsed by Python into an Abstract Syntax Tree (ast) that is then used to generate the SQL syntax. I wrote the Evaluator with different DB engines in mind, in case they use different syntax for JSON extraction, etc. The code is fairly minimal and with low complexity to make it easy to maintain and extend.

Let me know if you have concerns or ideas for improvement.

dsblank:
@DavidMStraub this could serve as a replacement in gramps-web for both gramps-ql and object-ql as it is converted into SQL. (It doesn't yet allow everything that the other do though).

dsblank:
One thing that I realize that this doesn't respect is filters and proxies. But I think that can be fixed. Some options:

1. Fallback to using standard gramps access if there is a proxy or filter
2. After selection, see if object is accessible in filter/proxy

Other ideas?

dsblank:
@Nick-Hall, actually, I'm realizing that we have a bigger issue: if you have a proxy/filter in place, then you might not be able to access all of the items in the JSON data. 

That means that:

```python
person_data.family_list != person_object.family_list
```

if a family does not appear in the filter/proxy.

It could be that if we have a filter or proxy, we must force the DataDict to generate the object through methods like `db.get_person_from_handle()`.

DavidMStraub:
If I may ... I think it's really great that so much refactoring and improvement is happening, but I find it a bit strange that so many things are merged so quickly without (sorry - at least my impression) considering all the implications (I was triggered by the example with proxies and filters), while at the same time my simple PR which does nothing but enable static type checking has been open for half a year. Static type checking would make the refactoring less dangerous.

dsblank:
@DavidMStraub, nothing has been merged yet that has any effect on the implications I have raised above. The implication is for the things being considered for merging. It would be great if we had more developers (like yourself) that would be able to comment on such implications. So, no, things aren't being merged "too quickly" and without thinking about consequences.

Working on the what is next gives us insight into complex issues. So no need to get triggered by such a realization. 

Regarding type checking: yes, I would have merged that PR many months ago because I am very familiar with the benefits of typing, and realize there are no down sides. 

But also, the implication above is the realization that a "type" (eg, `Person`) doesn't capture the essence of the issue. A `Person` object is an API to possibly altered and hidden properties. That is hidden in the API. So a `Person` created directly from the data isn't the same kind of `Person` we get from db methods. We might be able to create different types to catch such errors, but we don't even have the concepts for such types yet.

In any event, we need to refactor this PR, and the filter refactor PR. And probably adjust the `DataDict` class to make sure we don't access items that we shouldn't.

stevenyoungs:
One of your optimisations is to keep the data in a `DataDict` unless the true object is required - but at least you have the full set of data in memory. With this db.select_from_TABLE PR, you only have partial object data data in memory. Therefore you have insufficient data to determine in all cases if a record should be returned \ "sanitized" by the proxy db.
From a quick scan, the set of `ProxyDbBase` classes are used in report \ export scenarios where performance is (perhaps) less critical? Perhaps then these proxy db classes force db.select_from_TABLE to read all data that is required to correctly run the filter?
This would hopefully retain the performance gains in scenarios where no proxy is in use.

Nick-Hall:
@dsblank This PR reminds me of the [db.collcetion.find](https://www.mongodb.com/docs/manual/reference/method/db.collection.find/) method in MongoDB.  It may be worth a quick look if you are unfamiliar with it.  You may get some ideas.

I like how you have made the query pythonic.  This is better than previous SQL-like designs and the JSON queries of MongoDB.

@DavidMStraub We seem to have been discussing this on and off for about 7 or 8 years now, so I don't think that the progress is too fast.  There have also been a couple of prototypes.  The static type checking PR makes changes to 51 files. I tend to leave this type of change until fairly close to release in order to avoid potential conflicts when merging up fixes from the maintenance branch.  Also the smaller changes tend to be easier to fit in when I have time available.  Your PR is on my schedule though.

@stevenyoungs Yes.  Proxies are mainly used in the report and export code.  I don't mind if these are not optimised to use the new code, but we must make sure that they don't run significantly slower than at present.  Some people already have to wait a long time for certain reports to run.

I don't regard this PR as essential for the next release, but it may be worth continuing to investigate our options.

DavidMStraub:
>  I tend to leave this type of change until fairly close to release in order to avoid potential conflicts when merging up fixes from the maintenance branch

That is understable, but that PR is blocking adding meaningful type hints to the code base (which would be the main benefit), as this is not done in the PR itself - it only introduces the CI check and makes some minor adjustments to make the check pass. So merging it shortly before release means we won't have actual type checks until the next-to-next feature release.

Sorry, off topic for this PR.

dsblank:
Here is my plan:

1. Add `DbGeneric.select` methods so there is the functionality for proxies and non-DB-API databases
2. Add an argument to the select methods that you must manually set that will allow the select in a proxied database (to prevent accidental usage that would slow the system down, like in a filter)
3. If the argument is set (eg, `allow_use_on_proxy=True`) then it will fallback to DbGeneric method if needed. Else, an exception is raised (if arg is True and on a proxy).
4. The system can decide to use the select method or not, depending on if there is a proxy

This is a little bit complicated to explain, but I'll make an example to demonstrate.

dsblank:
> This PR reminds me of the [db.collcetion.find](https://www.mongodb.com/docs/manual/reference/method/db.collection.find/) method in MongoDB. It may be worth a quick look if you are unfamiliar with it. You may get some ideas.

Thanks for the link, @Nick-Hall. I looked at the mongodb method. It is very similar to the Django methods. But from earlier conversations, Gramps developers did not like such APIs as they are so different from Python. I think Python strings will be more easily understood and used.

gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -1183,3 +1185,43 @@ def _sql_cast_list(self, values):
...
+        else:
+            where_expr = " WHERE %s" % evaluator.convert(where)
+
+        if order_by is None:
+            order_by_expr = ""
+        else:
+            order_by_expr = evaluator.get_order_by(order_by)
+
+        self.dbapi.execute(
+            f"SELECT {what_expr} from {table_name}{where_expr}{order_by_expr};"
```
DavidMStraub, line 1217: This looks very dangerous to me. `table_name="person; DROP TABLE person; SELECT * FROM person"` :boom: 

Of course you could say we leave the responsibility of sanitization to the downstream code like addons etc., but I would very much prefer if we could avoid template strings as much as possible!

Or am I missing something?

DavidMStraub:
I am not sure what the intented use of `DataDict` is, but I recommend using Primary object types rather than `DataDicts` wherever possible (as return types and arguments of functions), because otherwise I suspect we'll be seeing a lot of `AttributeError`s or alternatively have to sprinkle the code with `assert hasattr`. (And it will be impossible to type check.)

gramps/plugins/db/dbapi/dbapi.py
dsblank: Yes, I think you are missing something :) All of the expressions are parsed, and only specific items can be evaluated.

dsblank:
> I am not sure what the intented use of DataDict is, but I recommend using Primary object types rather than DataDicts wherever possible (as return types and arguments of functions), because otherwise I suspect we'll be seeing a lot of AttributeErrors or alternatively have to sprinkle the code with assert hasattr. (And it will be impossible to type check.)

Yeah, we could return Primary Objects. The methods were originally designed to mirror the raw_data methods, but I like this idea! Note that they'll only return Primary Objects when specific fields aren't selected (like "person.gramps_id"). Thanks for the idea; I'll make that change.

dsblank:
> because otherwise I suspect we'll be seeing a lot of AttributeErrors or alternatively have to sprinkle the code with assert hasattr

BTW, `DataDict` is a drop-in replacement in terms of attributes. They have the same attribute fields as the Primary Objects.

gramps/plugins/db/dbapi/dbapi.py
DavidMStraub: Thanks, can you elaborate? Where is `table_name` parsed?

DavidMStraub: >  I'll add an underscore to the name to make that more clear.

Yes, that's perfect, thanks!

dsblank:
Here is an example of where we would need to be careful about falling back to a regular loop through the data in the case of a proxy:

If #1794 is merged, it has an optimization for looking for `rule.map` sets (in a particular scenario). Consider the `HasTag` rule. Currently, it requires looping over all objects O(N), even if only 5 out of a million people are tagged.

But we could add code to the `prepare()` method of the rule like:

```
    rows = db.select_from_person(
        what="person.handle", 
        where=f"'{self.tag_handle}' in person.tag_list"
    )
    self.map = set([handle for handle in rows])
```

But, if `db` is a proxy, that would fallback to another loop around all objects, making in O(2N). To prevent that, the above would raise an exception. 

A fix is to add:

```
if not db.is_proxy():
    rows = db.select_from_person(
        what="person.handle", 
        where=f"'{self.tag_handle}' in person.tag_list",
    )
    self.map = set([handle for handle in rows])
```

then the rule.map is not created, and the standard rule.apply_to_one() would do the regular check.

Here are the time comparisons (seconds) without and with the select/map:

Scenario | Prepare Time | Apply Time | Total Time
-------- | -----------: | ---------: | ----------:
Gramps 5.2 | 0.00 | 1.47 | 1.47
Gramps 6.0 + select + optimizer | 0.21 | 0.02 | 0.23


Finally, if you wanted to force a loop in a proxy, you could still do this:

```
    rows = db.select_from_person(
        what="person.handle", 
        where=f"'{self.tag_handle}' in person.tag_list"
        allow_use_in_proxy=True
    )
```

dsblank:
Bah! @stevenyoungs pointed out that the proxies properly process raw data. All my worrying above, and some of my comments about `DataDict` and raw data are wrong. `DataDict` should be perfectly fine to use throughout gramps to save a bit of time.

stevenyoungs:
> Bah! @stevenyoungs pointed out that the proxies properly process raw data. All my worrying above, and some of my comments about `DataDict` and raw data are wrong. `DataDict` should be perfectly fine to use throughout gramps to save a bit of time. (Well, except in proxies, they probably take a bit longer... not sure).

the `get_raw_*` functions in a proxy look like they create the true object in order to do the filtering. So I agree, not efficient, but should give the correct result.

dsblank:
> the get_raw_* functions in a proxy look like they create the true object in order to do the filtering. So I agree, not efficient, but should give the correct result.

#1839 will allow efficient get_raw_* functions in proxies.

gramps/gen/db/generic.py
dsblank: I think we'll leave JOIN as a future possibility for now.

dsblank:
Asked for comments on the gramps-dev mailing list.

Nick-Hall:
The mailing list thread is called [A DB select method that can be engine-optimized](https://sourceforge.net/p/gramps/mailman/message/59122276/).  I'll start the discussion if nobody replies, but I didn't want to influence people by posting my opinions first.

kulath:
> I am not sure what the intented use of `DataDict` is, but I recommend using Primary object types rather than `DataDicts` wherever possible (as return types and arguments of functions), because otherwise I suspect we'll be seeing a lot of `AttributeError`s or alternatively have to sprinkle the code with `assert hasattr`. (And it will be impossible to type check.)

Off topic, but I think this also applies to the Switch from pickled blobs to JSON data #1786 change https://github.com/gramps-project/gramps/pull/1786/files

dsblank:
> I'll start the discussion if nobody replies,

That would be great.

dsblank:
@Nick-Hall, IMO no serious objections raised on gramps dev mailing list. This PR would probably make a good candidate for 6.1. If accepted, I could update some filters to make some serious speedups in those rules that can be optimized.

Nick-Hall:
@dsblank I'm not sure that this PR is the right way to go.  However, we can discuss it further after v6.0 is released.

dsblank:
@Nick-Hall, I'm starting to work on some parts for Gramps 6.1 (the proxy refactor #1952 is actually ready for review.)

I know that you are quite busy getting 6.0 beta versions out, and preparing for its release, but it would be great to get a hint of what you are thinking re: this PR.

dsblank:
Closing this PR in favor of a new version.

PR #1830, opened 2024-12-23
Grampstype set

stevenyoungs:
Make the `GrampsType.__set` methods public

Based on a comment by @dsblank in [PR #1825](https://github.com/gramps-project/gramps/pull/1825#discussion_r1895209505) 
This PR builds on PR #1825

No attempt has been made to convert calling code. That could be tackled if there is agreement on the approach. 

CC @Nick-Hall for comment on the approach

### DIFF
--- a/gramps/gen/lib/grampstype.py
+++ b/gramps/gen/lib/grampstype.py
@@ -151,6 +151,16 @@ class GrampsType(metaclass=GrampsTypeMeta):
         else:
             self.__string = ""
 
+    def __set_dict(self, value):
+        "Set the value/string properties from a dict."
+        val, strg = self._DEFAULT, ""
+        if value:
+            val = value["value"]
+            if len(value) > 1 and val == self._CUSTOM:
+                strg = value["string"]
+        self.__value = val
+        self.__string = strg
+
     def __set_tuple(self, value):
         "Set the value/string properties from a tuple."
         val, strg = self._DEFAULT, ""
@@ -184,7 +194,9 @@ class GrampsType(metaclass=GrampsTypeMeta):
 
     def set(self, value):
         "Set the value/string properties from the passed in value."
-        if isinstance(value, tuple):
+        if isinstance(value, dict):
+            self.__set_dict(value)
+        elif isinstance(value, tuple):
             self.__set_tuple(value)
         elif isinstance(value, int):
             self.__set_int(value)
--- a/gramps/gen/lib/grampstype.py
+++ b/gramps/gen/lib/grampstype.py
@@ -244,7 +244,7 @@ class GrampsType(metaclass=GrampsTypeMeta):
         if string == cls._CUSTOM:
             return string
         else:
-            return cls._I2SMAP.get(value, cls.UNKNOWN)
+            return cls._I2SMAP.get(value, _UNKNOWN)
 
     @classmethod
     def get_str(cls, dict):
--- a/gramps/gen/lib/grampstype.py
+++ b/gramps/gen/lib/grampstype.py
@@ -240,6 +240,17 @@ class GrampsType(metaclass=GrampsTypeMeta):
         return (self.__value, self.__string)
 
     @classmethod
+    def __get_str(cls, value, string):
+        if string == cls._CUSTOM:
+            return string
+        else:
+            return cls._I2SMAP.get(value, cls.UNKNOWN)
+
+    @classmethod
+    def get_str(cls, dict):
+        return cls.__get_str(dict["value"], dict["string"])
+
+    @classmethod
     def get_schema(cls):
         """
         Returns the JSON Schema for this class.
@@ -265,9 +276,7 @@ class GrampsType(metaclass=GrampsTypeMeta):
         return self
 
     def __str__(self):
-        if self.__value == self._CUSTOM:
-            return self.__string
-        return self._I2SMAP.get(self.__value, _UNKNOWN)
+        return type(self).__get_str(self.__value, self.__string)
 
     def __int__(self):
         return self.__value
--- a/gramps/gen/lib/grampstype.py
+++ b/gramps/gen/lib/grampstype.py
@@ -247,8 +247,8 @@ class GrampsType(metaclass=GrampsTypeMeta):
             return cls._I2SMAP.get(value, _UNKNOWN)
 
     @classmethod
-    def get_str(cls, dict):
-        return cls.__get_str(dict["value"], dict["string"])
+    def get_str(cls, data):
+        return cls.__get_str(data.value, data.string)
 
     @classmethod
     def get_schema(cls):
--- a/gramps/gen/lib/grampstype.py
+++ b/gramps/gen/lib/grampstype.py
@@ -151,7 +151,7 @@ class GrampsType(metaclass=GrampsTypeMeta):
         else:
             self.__string = ""
 
-    def __set_dict(self, value):
+    def set_dict(self, value):
         "Set the value/string properties from a dict."
         val, strg = self._DEFAULT, ""
         if value:
@@ -161,7 +161,7 @@ class GrampsType(metaclass=GrampsTypeMeta):
         self.__value = val
         self.__string = strg
 
-    def __set_tuple(self, value):
+    def set_tuple(self, value):
         "Set the value/string properties from a tuple."
         val, strg = self._DEFAULT, ""
         if value:
@@ -171,12 +171,12 @@ class GrampsType(metaclass=GrampsTypeMeta):
         self.__value = val
         self.__string = strg
 
-    def __set_int(self, value):
+    def set_int(self, value):
         "Set the value/string properties from an integer."
         self.__value = value
         self.__string = ""
 
-    def __set_instance(self, value):
+    def set_instance(self, value):
         "Set the value/string properties from another grampstype."
         self.__value = value.value
         if self.__value == self._CUSTOM:
@@ -184,7 +184,7 @@ class GrampsType(metaclass=GrampsTypeMeta):
         else:
             self.__string = ""
 
-    def __set_str(self, value):
+    def set_str(self, value):
         "Set the value/string properties from a string."
         self.__value = self._S2IMAP.get(value, self._CUSTOM)
         if self.__value == self._CUSTOM:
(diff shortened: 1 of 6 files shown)


dsblank:
@Nick-Hall, @stevenyoungs has made a couple of PRs to choose from: this one, or #1825. I think this one is better, but not sure if there is a difference in functionality of a private static class method.

stevenyoungs:
> @Nick-Hall, @stevenyoungs has made a couple of PRs to choose from: this one, or #1825. I think this one is better, but not sure if there is a difference in functionality of a private static class method.

#1825 contains a bug fix for #1726. 
This PR, #1830, adds one additional [commit](https://github.com/gramps-project/gramps/pull/1830/commits/5c44298030c5f992114313b6f30c5e744c731ddb) on top of #1825, This additional commit makes the `GrampsType.__set_*` methods public.

dsblank:
Ready to merge, IMO.

Nick-Hall:
I don't see the point of the PR.  Why do you want to make these methods public?

Nick-Hall:
The original code uses a standard technique for overloading methods in python.  Four methods (now five) were combined into a single `set` method to make the public interface simpler.  This does come with a slight loss in performance, but that should be negligible.

Note:  The python typing module includes the [overload](https://docs.python.org/3/library/typing.html#typing.overload) decorator for such methods.

QuLogic:
> Note: The python typing module includes the [overload](https://docs.python.org/3/library/typing.html#typing.overload) decorator for such methods.

This is purely for describing the types of the function; it doesn't do anything at runtime. For runtime dispatch, that's [`functools.singledispatch`](https://docs.python.org/3/library/functools.html#functools.singledispatch).

stevenyoungs:
@QuLogic thanks for the information on `singledispatch`. I've refactored my PR to use this technique. I would appreciate if you can check to see if my code can be improved. I did not particularly like having the `def set_instance(self, value: GrampsType)` function outside the class but I could not find any other way. I then had to have a `set_instance` method in the class in order to set the private properties. Introducing a new super class seemed overkill.

@Nick-Hall I don't think `singledispatch` \ `singledispatchmethod` are currently used in the gramps codebase. It allows us to remove the
```
is isinstance()
elif is instance()
elif is instance()
```
construct. This seems like a positive change. Any comments?

gramps/gen/lib/grampstype.py
```diff
@@ -151,11 +158,18 @@ def set_object_state(self, attr_dict):
...
             self.__string = ""
 
-    def __set_dict(self, value):
+    @singledispatchmethod
+    def set(self, value):
+        self.__value = self._DEFAULT
+        self.__string = ""
+
+    @set.register
+    def set_dict(self, value: dict):
```
Nick-Hall, line 167: This should be private.

gramps/gen/lib/grampstype.py
```diff
@@ -151,11 +158,18 @@ def set_object_state(self, attr_dict):
...
         "Set the value/string properties from a dict."
         self.__set_tuple((value["value"], value["string"]))
 
-    def __set_tuple(self, value):
+    @set.register
+    def set_tuple(self, value: tuple):
```
Nick-Hall, line 172: This should be private.

gramps/gen/lib/grampstype.py
```diff
@@ -165,41 +179,27 @@ def __set_tuple(self, value):
         self.__value = val
         self.__string = strg
 
-    def __set_int(self, value):
+    @set.register
+    def set_int(self, value: int):
```
Nick-Hall, line 183: This should be private.

gramps/gen/lib/grampstype.py
```diff
@@ -165,41 +179,27 @@ def __set_tuple(self, value):
...
-        "Set the value/string properties from another grampstype."
-        self.__value = value.value
-        if self.__value == self._CUSTOM:
-            self.__string = value.string
-        else:
-            self.__string = ""
-
-    def __set_str(self, value):
+    @set.register
+    def set_str(self, value: str):
```
Nick-Hall, line 189: This should be private.

gramps/gen/lib/grampstype.py
```diff
@@ -165,41 +179,27 @@ def __set_tuple(self, value):
...
-            self.__set_dict(value)
-        elif isinstance(value, tuple):
-            self.__set_tuple(value)
-        elif isinstance(value, int):
-            self.__set_int(value)
-        elif isinstance(value, self.__class__):
-            self.__set_instance(value)
-        elif isinstance(value, str):
-            self.__set_str(value)
+    def set_instance(self, value):
```
Nick-Hall, line 197: This should be private.

gramps/gen/lib/grampstype.py
```diff
@@ -345,5 +345,11 @@ def __eq__(self, value):
     def __ne__(self, value):
         return not self.__eq__(value)
 
-    value = property(__int__, __set_int, None, "Returns or sets integer value")
-    string = property(__str__, __set_str, None, "Returns or sets string value")
+    value = property(__int__, set_int, None, "Returns or sets integer value")
+    string = property(__str__, set_str, None, "Returns or sets string value")
+
+
+@GrampsType.set.register
```
Nick-Hall, line 352: This should be private and ideally inside the class.

Nick-Hall (changes requested):

gramps/gen/lib/grampstype.py
stevenyoungs: I had not expected it to work if I made `set_dict` private but having tested it, it does.
I'll make the it private

gramps/gen/lib/grampstype.py
stevenyoungs: "decorators apply to the function objects before the class object is created." [Source](https://stackoverflow.com/questions/24063788/python3-singledispatch-in-class-how-to-dispatch-self-type) so I think I have to keep this as a function.
Given then that this is a function, I have to call a helper method `GrampsType.set_instance` to set the private properties of the class. If I make the method `GrampsType.set_instance` private, I can't call it from outside the class.

gramps/gen/lib/grampstype.py
stevenyoungs: See discussion below on [`def set_instance(self, value: GrampsType):`](https://github.com/gramps-project/gramps/pull/1830#discussion_r1912466188)

gramps/gen/lib/grampstype.py
stevenyoungs: This has been addressed in the latest commit

gramps/gen/lib/grampstype.py
stevenyoungs: This has been addressed in the latest commit

gramps/gen/lib/grampstype.py
stevenyoungs: This has been addressed in the latest commit

gramps/gen/lib/grampstype.py
stevenyoungs: This has been addressed in the latest commit

gramps/gen/lib/grampstype.py
QuLogic: It still doesn't need to be public; it's only used by something that is added on to the class. Note however, if you use double underscores, you will need to use the mangled name: `_GrampsType__set_instance`.

gramps/gen/lib/grampstype.py
QuLogic: This also can remain private.

gramps/gen/lib/grampstype.py
stevenyoungs: This is addressed in the most recent commit

gramps/gen/lib/grampstype.py
stevenyoungs: This is addressed in the most recent commit

stevenyoungs:
All review comments are addressed. Many thanks to @QuLogic for his guidance.

Nick-Hall (approved):

Nick-Hall:
@stevenyoungs @QuLogic I've squashed the commits, refactored the `__set_instance` method back into the class, and changed how the method is registered.  If you have no further suggestions then I think that this is ready to merge.

stevenyoungs:
I like your solution to moving `__set_instance` back into the class.
Good to merge from my point of view.

gramps/gen/lib/grampstype.py
```diff
@@ -165,11 +179,13 @@ def __set_tuple(self, value):
...
 
-    def __set_int(self, value):
+    @set.register
+    def __set_int(self, value: int):
         "Set the value/string properties from an integer."
         self.__value = value
         self.__string = ""
 
+    # This method needs to be registered outside of the class.
     def __set_instance(self, value):
```
QuLogic, line 189: I would still type hint this as well. Note, if you want to use `GrampsType` directly, you would have to add `from __future__ import annotations`, though you can also quote it instead.

stevenyoungs: I've added the type hint back using this approach.
I could not find a way to move the register back into the class so it is left at the end of the file

prculley:
I am sorry for the late comment; been busy with other things.
I just encountered this change when I tried to debug some of my addon code; I was initializing a PlaceType with the DataDict.
I was shocked to see that it ran through what seemed like hundreds of lines of code in functools, as opposed to the original if/then/else style from GrampsType.  This seems like it might be a considerable slowdown for code that sets/writes new data, like imports (which I have already observed to be very slow).

Has anyone done a performance comparison?

dsblank:
I don't know if we did a specific performance test on `singledispatch`.  It sounds like a good idea since we are considering the decorator for other uses in the codebase.

prculley:
I just got around to making some tests on performance.  I made a test file that instantiated all of our GrampsTypes with each of the potential 'set' values (dict, tuple, int, string, and type).  I ran this test 100000 times.
with this PR (normal Gramps60 branch), it took 59.5 seconds.
reverting this PR, it took 15.1 seconds.
So about 4 times linger with this PR.

For a more real life test, I thought that doing some imports would involve GrampsType instantiations.
I enhanced the import_test by adding example.gramps for a larger file and then ran the tests.
with PR, it took 26.3 seconds
without, it took 25.8 seconds
or about a 2% slowdown.
Just for fun I also ran import_test this way on gramps52 branch, about 22.8 seconds, which is about 15% faster than gramps60.

PR #1832, opened 2024-12-26
Improve the backlinks (References) gramplets

stevenyoungs:
The backlinks gramplet (called References in the UI) is great for finding linked objects. It also allows direct editing of the object by double clicking. However, if you want to make the linked object the active object, you have to go to the relevant object-type view (e.g. Families), and re-find the object.
This enhancement allows the active object to be set directly from the References gramplet. This is currently a two-step process

1. select the row
image

2. press the middle mouse button (anywhere within the gramplet)
The selected object will be made the active object of that type.
In the above example, Family F0403 will be made the active family.
It works for all object types, not just families
image

### DIFF
--- a/gramps/gui/clipboard.py
+++ b/gramps/gui/clipboard.py
@@ -81,25 +81,32 @@ clipdb = None  # current db to avoid different transient dbs during db change
 
 theme = Gtk.IconTheme.get_default()
 LINK_PIC = theme.load_icon("stock_link", 16, 0)
+OBJ2ICON = {
+    "media": "gramps-media",
+    "note": "gramps-notes",
+    "person": "gramps-person",
+    "place": "gramps-place",
+    "address": "gramps-address",
+    "attribute": "gramps-attribute",
+    "event": "gramps-event",
+    "family": "gramps-family",
+    "location": "geo-place-link",
+    "media": "gramps-media",
+    "name": "geo-show-person",
+    "repository": "gramps-repository",
+    "source": "gramps-source",
+    "citation": "gramps-citation",
+    "text": "gramps-font",
+    "url": "gramps-geo",
+}
+
+
+def obj2icon(target):
+    return OBJ2ICON[target] if target in OBJ2ICON else None
+
+
 ICONS = {}
-for name, icon in (
-    ("media", "gramps-media"),
-    ("note", "gramps-notes"),
-    ("person", "gramps-person"),
-    ("place", "gramps-place"),
-    ("address", "gramps-address"),
-    ("attribute", "gramps-attribute"),
-    ("event", "gramps-event"),
-    ("family", "gramps-family"),
-    ("location", "geo-place-link"),
-    ("media", "gramps-media"),
-    ("name", "geo-show-person"),
-    ("repository", "gramps-repository"),
-    ("source", "gramps-source"),
-    ("citation", "gramps-citation"),
-    ("text", "gramps-font"),
-    ("url", "gramps-geo"),
-):
+for name, icon in OBJ2ICON.items():
     ICONS[name] = theme.load_icon(icon, 16, 0)
 
 
@@ -1257,7 +1264,7 @@ class ClipboardListView:
                 _ob._dbname = dbname
 
             # If this was a ClipPersonLink with a list for a handle, we don't
-            # want it.  Can happen in People Treeview with attemp to drag last
+            # want it.  Can happen in People Treeview with attempt to drag last
             # name group.
             if isinstance(_ob, ClipHandleWrapper) and isinstance(_ob._handle, list):
                 return None
--- a/gramps/gui/listmodel.py
+++ b/gramps/gui/listmodel.py
@@ -77,6 +77,7 @@ class ListModel:
     mode:           Selection mode for TreeView.  See Gtk documentation.
     list_mode:      "list" or "tree"
     right_click:    Function called when the user right-clicks on a row.
+    middle_click:   Function called when the user middle-clicks on a row.
     """
 
     def __init__(
@@ -88,6 +89,7 @@ class ListModel:
         mode=Gtk.SelectionMode.SINGLE,
         list_mode="list",
         right_click=None,
+        middle_click=None,
     ):
         self.tree = tree
         self.tree.set_fixed_height_mode(True)
@@ -97,6 +99,7 @@ class ListModel:
         self.list_mode = list_mode  # "list", or "tree"
         self.double_click = None
         self.right_click = None
+        self.middle_click = None
 
         for info in dlist:
             col_type = TEXT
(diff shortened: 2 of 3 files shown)


stevenyoungs:
I'd welcome suggestions on how to make this feature more discoverable. 
There is no indication to users that this capability exists. The same criticism can be levelled at the Family editor dialog, where middle click on a child, brings up the person editor. It took me a few years to discover this!

dsblank:
Very nice!

Regarding making the functionality more discoverable, some ideas:

* underline the items
* use a different color
* use a different  font
* give message on hover

emyoulation:
In some other features, there is a left selecting action, a left double-click action, a center-click action and a right-click context menu. 

And the context menu normally includes the left and center click actions.

https://gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Keybindings#Context_menu

The experimental CardView addons included not only extensive use of the center mouse button but also preferences GUI for swapping the actions for mouse buttons.

emyoulation:
> give message on hover

Given that some hover tooltip message boxes can be quite large and obscure the click or drag handles, there needs to be a way to suppress the tooltips too.

Tooltips popping up can really impede expert users ability to see the genealogical data and interact with their mouse GUI.

stevenyoungs:
I'll add a simple context menu with Edit and Make Active options

emyoulation:
Note that one of the big uses of the References is drag'n'drop to the clipboard. (Although the copy keybinding is not supported.)

Can anyone point me towards where a context menu for the 'ref' clipboard objects should be added to the clipboard code? The objects have context menus (which include options like 'make active' or make a custom filter) but the ref objects are lacking that. 

So a context menu that merely added the referenced object to the Clipboard would be a workaround.

emyoulation:
> I'll add a simple context menu with Edit and Make Active options

I'll add mention of the button actions to the wiki. And link to it from the Gramps Glossary.

Maybe a 'Help' context menu item pointing to the Glossary item makes sense?

https://gramps-project.org/wiki/index.php/Gramps_Glossary#references_tab

stevenyoungs:
I've added a simple context menu to help with discoverability.

image

stevenyoungs:
> Note that one of the big uses of the References is drag'n'drop to the clipboard. (Although the copy keybinding is not supported.)
> 
> Can anyone point me towards where a context menu for the 'ref' clipboard objects should be added to the clipboard code? The objects have context menus (which include options like 'make active' or make a custom filter) but the ref objects are lacking that.
> 
I've added drag support to the References gramplet. You can now drag a row from this gramplet and drop it onto the clipboard.
Once the object is on the clipboard you can then drag it as usual. The video below demonstrates by dragging a Person reference from a Family object onto the clipboard


https://github.com/user-attachments/assets/329f517f-483e-45b9-b2a9-6b1983f1d113


What doesn't work, and I don't know why yet, is dragging from the Reference gramplet directly to an editor window. For example, drag the same person as above, and drop it onto a new Family editor window and the person is not added to the family. Drag the person from the clipboard into the same family editor window and it works as expected.

stevenyoungs:
> What doesn't work, and I don't know why yet, is dragging from the Reference gramplet directly to an editor window. For example, drag the same person as above, and drop it onto a new Family editor window and the person is not added to the family. Drag the person from the clipboard into the same family editor window and it works as expected.

I think I can answer my own question: I am providing data in the format `DdTargets.HANDLE_LIST` whereas I suspect the EditFamily dialog wants data in the format `DdTargets.PERSON_LINK`

Is there any documentation on the workings of drag and drop within gramps?

emyoulation:
In her Forms addon redevelopment, Renée asked a similar and was referred to the CardView addons since drag'n'drop is implemented so extensively there. (Including the ability to drop citations on closed object to cite that source.)

She found it well organized as a reference.

https://gramps.discourse.group/t/plans-to-update-forms-addon-over-the-2024-summer/5274/153

stevenyoungs:
Drag of objects from the References gramplet is now working as expected for all windows, not just the clipboard.
This required passing a different target data type for each object type.
I believe I've added support all object types that can be returned by `db.find_backlink_handles`. If I missed any, they can easily be added to `Backlinks.__obj2target`

emyoulation:
A thread in Discourse discusses an edge case for citation shared References.

https://gramps.discourse.group/t/reverse-lookups-of-citation-references/4177

In this edge case, it is difficult (or impossible) to determine which non-primary object the Citation was referencing from within the GUI.

Does this ambiguity affect the drag'n'drop ref object drop to the clipboard?

stevenyoungs:
> Does this ambiguity affect the drag'n'drop ref object drop to the clipboard?

Thanks for finding that thread @emyoulation.
I just tested the scenario:
1. start a new database
2. create a new citation
3. create a new person and add the citation from step 1
4. add an Age attribute to the person and add the citation from step 1 to the attribute
5. switch to the citation view and select the citation from step 1
6. add the references gramplet

The only reference listed in the References gramplet is the Person created in step 2. The Age attribute from step 4 is not listed and therefore does not cause any problem for the drag and drop code proposed in this PR.

The References gramplet relies on `self.dbstate.db.find_backlink_handles` to discover the references to display.

Nick-Hall:
@stevenyoungs This looks good.  The only problem I found is that I cannot drag & drop a Note.

stevenyoungs:
> @stevenyoungs This looks good. The only problem I found is that I cannot drag & drop a Note.

How do I get a Note object to appear in the backlinks gramplet? Wouldn't that require a Note object to reference another object. I can't see a way to add a reference **from** a Note object (obviously I can add a reference **to** a Note object from other object types). Did I misunderstand?

stevenyoungs:
Added support for dragging a media object from the backlinks gramplet.
The substantive change was to add this line
https://github.com/gramps-project/gramps/blob/294466c9dfe51b2d721912cc0b8cfdb9d18037a1/gramps/plugins/gramplet/backlinks.py#L67

stevenyoungs:
Unrelated to this PR, could the `DdTargets` properties be more consistently named?
The ones used by this gramplet are mostly `*_LINK`. `EVENT` and `MEDIAOBJ` do not follow the same naming scheme
https://github.com/gramps-project/gramps/blob/294466c9dfe51b2d721912cc0b8cfdb9d18037a1/gramps/plugins/gramplet/backlinks.py#L60-L67

Nick-Hall:
> How do I get a Note object to appear in the backlinks gramplet?

Create a note with a link to another object.  The "Garner von Zieliński, Lewis Anderson Sr" person in the example database contains backlinks of this type.

stevenyoungs:
> Create a note with a link to another object

I didn't think of putting the link in the actual note text!
PR updated.
Hopefully all primary object types are covered now....

PR #1833, opened 2024-12-27
Set initial focus in EditPlaceRef dialog

stevenyoungs:
Currently, in the Place Reference Editor dialog `EditPlaceRef`, the default focus is set to the the top tab control, "General" tab. Given that there is only a single tab, this is not useful. The first action that all users have to do is to move the focus.
With this change, the default focus is set to the name control of the place, allowing the user to immediately start typing \ editing the place name. The Name field was chosen as I felt it was more likely to be edited than the Date field.

image


Nick-Hall:
I think that the focus is set to the notebook control when we add tabs in the setup.  This is why you proably have to set the focus in the code rather than in the ui definition (glade) file.

PR #1836, opened 2025-01-01
Enhancements to Selector Dialogs

stevenyoungs:
A set of enhancements to the selector dialogs:
1. Add a Path column to the Select Media Object dialog (`SelectObject`)
See image 2 below
2. Add the standard search bar to the Select Person dialog (`SelectPerson`)
See image 1 below
3. Add (optional) support for multiple selection in the `BaseSelector` class
This includes unifying the two different `MultiTreeView` classes in gramps
4. Add (optional) support for multiple selection to the Select Media Object dialog
The existing single thumbnail is replaced with a `ScrolledWindow` containing an `IconView`
See image 2 below
5. Use multiple selection when sharing media in the `GalleryTab`

If this PR is accepted, a follow on PR can add support for multiple selection to other selector dialogs. Examples could include: sharing multiple children to a family, sharing multiple events to a person

Currently, where only a single media object can be selected, the Edit Media Ref dialog is shown immediately after selection. The media ref is only added to the primary object when OK is selected from the Edit Media Ref dialog. This PR preserves this behaviour if the user only selects a single object. If _multiple_ media objects are selected by the user, the media refs are immediately created and added. No Edit Media Ref dialogs are shown. The user can manually edit the media ref, using existing GalleryTab functionality (e.g. double click the thumbnail) as required. This avoids a poor user experience of multiple Edit Media Ref dialogs appearing all at once, or in a sequence. It also mirrors the behaviour of drag and drop of multiple files on to the gallery tab.

**Image 1**
image
**Image 2**
image

### DIFF
--- a/gramps/gui/clipboard.py
+++ b/gramps/gui/clipboard.py
@@ -60,7 +60,7 @@ from .makefilter import make_filter
 from .utils import is_right_click, no_match_primary_mask
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 from gramps.gen.config import config
-from gramps.gui.widgets.persistenttreeview import PersistentTreeView
+from gramps.gui.widgets.multitreeview import MultiTreeView
 
 _ = glocale.translation.sgettext
 
@@ -1389,7 +1389,7 @@ class ClipboardWindow(ManagedWindow):
         self.clear_all_btn = self.top.get_object("btn_clear_all")
         self.clear_btn = self.top.get_object("btn_clear")
         objectlist = self.top.get_object("objectlist")
-        mtv = MultiTreeView(self.dbstate, self.uistate, _("Clipboard"))
+        mtv = ClipboardMultiTreeView(self.dbstate, self.uistate, _("Clipboard"))
         scrolledwindow = self.top.get_object("scrolledwindow86")
         scrolledwindow.remove(objectlist)
         scrolledwindow.add(mtv)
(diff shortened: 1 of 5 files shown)


stevenyoungs:
Each commit in the PR handles a single problem but I'm happy to split into multiple PRs if that is preferred.
Can someone add the Enhancement label please?

stevenyoungs:
Adding the search bar to the Select Person Dialog seems like an obvious thing to do.
I'm slightly concerned that there is a good reason that this was not already done!

stevenyoungs:
Addresses mantis feature request [683](https://gramps-project.org/bugs/view.php?id=683)
Related discussion: [Enhancing Selector Dialogs](https://gramps.discourse.group/t/enhancing-selector-dialogs/6684)

Nick-Hall:
> Each commit in the PR handles a single problem but I'm happy to split into multiple PRs if that is preferred.

A single PR is acceptable, but we will need to reword the commit messages so that they make more sense in the change log.

I'll try to find out why we disabled the search bar in the person selection dialog.

Nick-Hall:
@stevenyoungs I've removed the commit that added a search bar to the person selector.  If you discuss this on the forum and there is agreement, we can accept a simple one line change for inclusion in v6.0.

Nick-Hall:
Reworded commit messages to make the change log clearer.

stevenyoungs:
> @stevenyoungs I've removed the commit that added a search bar to the person selector.  If you discuss this on the forum and there is agreement, we can accept a simple one line change for inclusion in v6.0.

Will do. Can you expand on your concerns? Are they technical, usability, both, something else?

stevenyoungs:
> If you discuss this on the forum and there is agreement, we can accept a simple one line change for inclusion in v6.0.

Discussion started [here](https://gramps.discourse.group/t/adding-search-bar-to-select-person-dialog/6845) to gather gramps user comments. There was some previous discussion in 2021 [here](https://gramps.discourse.group/t/select-person-dialog-with-a-search/2009)

PR #1837, opened 2025-01-02
Fixes proxy serialize

dsblank:
This PR:

1. Removes serialize from proxy base
2. Fixes xml parser to get values for booleans
3. Adds a private person to data/tests/example.gramps
4. Makes to_dict() return a DataDict
5. Adds tests for proxies (there were none)


gramps/gen/lib/serialize.py
```diff
@@ -132,7 +132,7 @@ def to_dict(obj):
     :returns: A dictionary.
     :rtype: dict
     """
-    return json.loads(to_json(obj))
+    return DataDict(json.loads(to_json(obj)))
```
stevenyoungs, line 135: Given you have `obj` already, an optimisation would be to use it set to `DataDict._object` during init

In fact do you even need to initialise the `DataDict` super class `dict` if _object is set? 
Just leave the dict empty, except for `_object`
I think `DataDict.__getattr__` will end up calling `return getattr(self["_object"], key)`.

If the above is right, it avoids the `json.loads(to_json())` inefficiency

stevenyoungs: A `DataDict` then has 3 possible states:
1. just the object attributes in the dict
2. the object attributes in the dict _and_ the true object in `dict["_object"]`
3. just the true object in `dict["_object"]`

dsblank: Interesting! I didn't about defaulting to the _object attributes. Very clever!

Let me see if I can create an interface for that...

dsblank: Actually, since `DataDict` mirrors the object now, we can just return the object, right?

Changing:
```
return DataDict(json.loads(to_json(obj)))
```
to just:
```
return obj
```

dsblank: (It would be the wrong type, but would function correctly)

stevenyoungs: well it's the pythonic way, even if my C++ soul is crying 😉
If we accept that, you don't need to call `to_dict` at all, since it just returns it's argument, unaltered.

gramps/plugins/importer/importxml.py
```diff
@@ -156,6 +157,17 @@
 INSTANTIATED = 1
 
 
+def to_bool(value):
```
stevenyoungs, line 160: Why did you choose to roll your own `to_bool` rather than use `attrs.converters.to_bool()`?

gramps/plugins/importer/importxml.py
```diff
@@ -156,6 +157,17 @@
...
+    """
+    Utility function to convert attr value to boolean.
+    """
+    if isinstance(value, str):
+        value = value.lower()
+    if value in (False, "false", "f", "no", "n", "off", "0", 0):
+        return False
+    return True
```
stevenyoungs, line 168: Is it wise to assume that any value not `False` is `True`?
`to_bool("Doug")` would return True. In the case of private, I agree it defaults safe by doing so. However if future someone wants to use `to_bool` in a different context, they may prefer it to defaulted to False.
I'd rather it threw an error if the data is not clearly one of the false or true values.

gramps/gen/filters/rules/test/person_rules_test.py
```diff
@@ -1204,17 +1204,17 @@ def test_peopleprivate(self):
         """
         Test PeoplePrivate rule.
         """
-        # TODO: example.gramps has no people marked private
+        # TODO: example.gramps has one person marked private
```
stevenyoungs, line 1207: consider removing the comment entirely since you've now added a private person
```suggestion        
```

giotodibondone: Or mention which Person ID is private?

gramps/plugins/importer/importxml.py
dsblank: Well, attrs is a dependency, and currently `gramps` doesn't have any (and I couldn't figure out an easy way to make `python setup.py build` install it with our custom setup).

gramps/gen/lib/serialize.py
dsblank: I think I'll make an DataDict wrapper for objects.

gramps/plugins/importer/importxml.py
dsblank: In regular Python, anything other than empty sequences (`""`, `[]`, `{}`), `False`, `0`, and `None` will return `True`:

```
if "Steven":
    print("True!")
```

`if` (hardly) ever would raise an exception. (Although numpy does). 

I'd rather parse it into `True` or `False`, rather than crashing. I didn't put the function in a general module, as it is specific to this use.

gramps/gen/lib/serialize.py
dsblank: Added `DataDict(obj)` in #1839

gramps/gen/filters/rules/test/person_rules_test.py
dsblank: Thanks for the comments. I just removed it; the person handle is in the test.

gramps/plugins/importer/importxml.py
dsblank: Refined `to_bool()` and added more docs.

dsblank:
This PR is dependent on #1839

setup.py
```diff
@@ -450,6 +450,7 @@ def run(self):
     cmdclass={"build": build, "test": test},
     packages=packages,
     package_data={"gramps": package_data},
+    install_requires=["attrs"],
```
stevenyoungs, line 453: Is this needed given your comment [above](https://github.com/gramps-project/gramps/pull/1837/files#r1901429036)?
Personally I'd still prefer to see `attrs.converters.to_bool()` used!

dsblank: No it isn't needed; I'll fix. If you can figure out how to add requirements to gramps, I'll let you change it, either here or in a follow-up PR.

gramps/gen/lib/serialize.py
dsblank: I'll update this code after #1839 is merged.

setup.py
stevenyoungs: Let's work on the basis of a separate PR given the planned release schedule. Then if I get anything done before this PR is merged, it's a bonus.

dsblank: Sounds good.

gramps/gen/proxy/test/test_proxies.py
```diff
@@ -0,0 +1,108 @@
...
+            years_after_death=10,
+        )
+
+    def test_person_count(self):
+        count = self.db.get_number_of_people()
+        self.assertEqual(count, 1316)
+
+    def test_person_raw(self):
+        data = self.db._get_raw_person_from_id_data("I0552")
+        assert isinstance(data, dict)
```
Nick-Hall, line 82:
```suggestion
        self.assertIsInstance(data, dict)
```

gramps/gen/proxy/test/test_proxies.py
```diff
@@ -0,0 +1,108 @@
...
+            years_after_death=10,
+        )
+
+    def test_person_count(self):
+        count = self.db.get_number_of_people()
+        self.assertEqual(count, 1315)
+
+    def test_person_raw(self):
+        data = self.db._get_raw_person_from_id_data("I0552")
+        assert isinstance(data, dict)
```
Nick-Hall, line 108:
```suggestion
        self.assertIsInstance(data, dict)
```

dsblank:
Thanks @Nick-Hall for the "Suggestions" ... very nice!

dsblank:
I'm going to close this PR and break into smaller pieces. Using `DataGrid(object)` is more complicated than I thought, and I'll deal with that separately.

PR #1838, opened 2025-01-02
Fix scroll direction in pedigree view

Nick-Hall:
Fix the mouse scroll direction in the pedigree view when the Left <-> Right option is selected in the context menu.


QuLogic (approved):
I tried this, and it worked, as did holding Shift to switch directions.

PR #1839, opened 2025-01-03
Create DataDict from object; Get object with from_dict()

dsblank:
This PR:

1. Allows construction of a `DataDict` from an instance
2. Cleans up `DataDict` a bit
3. Returns object instance `from_dict()` if it exists
4. Adds tests

### DIFF
--- a/gramps/gen/lib/serialize.py
+++ b/gramps/gen/lib/serialize.py
@@ -48,6 +48,19 @@ class DataDict(dict):
     object interface.
     """
 
+    def __init__(self, data=None):
+        """
+        Wrap a data dict (raw data) or object
+        with an attribute API. If data is an
+        object, we use it to get the attributes.
+        """
+        if isinstance(data, dict):
+            super().__init__(data)
+        else:
+            super().__init__()
+            if data:
+                self["_object"] = data
+
     def __str__(self):
         if "_object" not in self:
             self["_object"] = from_dict(self)
@@ -55,21 +68,23 @@ class DataDict(dict):
 
     def __getattr__(self, key):
         if key.startswith("_"):
-            raise AttributeError("can't use this API to access hidden attributes")
+            raise AttributeError(
+                "this method cannot be used to access hidden attributes"
+            )
 
-        if key in self:
-            value = self[key]
-        else:
-            if "_object" not in self:
-                self["_object"] = from_dict(self)
+        if "_object" in self:
             return getattr(self["_object"], key)
-
-        if isinstance(value, dict):
-            return DataDict(value)
-        elif isinstance(value, list):
-            return DataList(value)
+        elif key in self:
+            value = self[key]
+            if isinstance(value, dict):
+                return DataDict(value)
+            elif isinstance(value, list):
+                return DataList(value)
+            else:
+                return value
         else:
-            return value
+            self["_object"] = from_dict(self)
+            return getattr(self["_object"], key)
 
 
 class DataList(list):
@@ -144,7 +159,10 @@ def from_dict(dict):
     :returns: A Gramps object.
     :rtype: object
     """
-    return from_json(json.dumps(dict))
+    if dict is not None and "_object" in dict:
+        return dict["_object"]
+    else:
+        return from_json(json.dumps(dict))
 
 
 class BlobSerializer:
@@ -170,7 +188,7 @@ class BlobSerializer:
 
     @staticmethod
     def string_to_data(bytes):
-        LOG.debug("blob, string_to_object: %r...", bytes[:35])
+        LOG.debug("blob, string_to_data: %r...", bytes[:35])
         return pickle.loads(bytes)
 
     @staticmethod
--- a/gramps/gen/lib/test/datadict_test.py
+++ b/gramps/gen/lib/test/datadict_test.py
@@ -1,0 +1,121 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2025      Doug Blank
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+""" Unittest for DataDict """
+
+# See also tests in ./serialize_test.py
+
+import unittest
+
+from gramps.gen.lib.serialize import DataDict, DataList, to_dict, from_dict
+from gramps.gen.lib import (
+    Person,
+    Family,
+)
+
+
+class DataDictTest(unittest.TestCase):
+    def test_person(self):
+        p = Person()
+        d = DataDict(p)
+        # Get the object from the dict:
+        self.assertEqual(id(p), id(from_dict(d)))
+
+    def test_person_to_dict(self):
+        p = Person()
+        d = to_dict(p)
+        self.assertFalse(hasattr(d, "_object"))
+
+    def test_family(self):
+        p = Family()
+        d = DataDict(p)
+        # Get the object from the dict:
+        self.assertEqual(id(p), id(from_dict(d)))
+
+    def test_family_to_dict(self):
+        p = Family()
+        d = to_dict(p)
+        self.assertFalse(hasattr(d, "_object"))
+
+
+class DataListTest(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


dsblank:
CC: @stevenyoungs (for some reason I can't invite you to review)

gramps/gen/lib/serialize.py
```diff
@@ -48,6 +48,20 @@ class DataDict(dict):
...
+        with an attribute API. If data is an
+        object, we use it to get the attributes.
+        """
+        if data is None:
+            super().__init__()
+        elif isinstance(data, dict):
+            super().__init__(data)
+        else:
+            super().__init__()
+            self["_object"] = data
```
stevenyoungs, line 63: You are initialising _object without knowing it really is a gramps object. All you know is it’s not None or a dict. Is it worth adding a check or do we trust all clients?

gramps/gen/lib/serialize.py
```diff
@@ -57,20 +71,20 @@ def __getattr__(self, key):
         if key.startswith("_"):
             raise AttributeError("can't use this API to access hidden attributes")
 
-        if key in self:
+        if "_object" in self:
```
stevenyoungs, line 74: If performance is ever an issue you could make _object a direct attribute. This would avoid having to lookup _object each time in the dictionary.

gramps/gen/lib/serialize.py
dsblank: I don't think it would be worth it. But it would be worth adding some type checking when we get to that.

gramps/gen/lib/serialize.py
dsblank: Probably a minor impact... but it is true that the data could already have an object in it, `{"_object": object}`. Changing to an attribute would require some refactoring.

dsblank:
Other PRs waiting on this one.

Nick-Hall:
OK.  I'll review this after #1717.

gramps/gen/lib/serialize.py
stevenyoungs, line 63:
```suggestion
        if isinstance(data, dict):
            super().__init__(data)
        else:
            super().__init__()
            if data:
                self["_object"] = data
```
Possibly simpler formulation?

gramps/gen/lib/serialize.py
```diff
@@ -57,20 +71,20 @@ def __getattr__(self, key):
         if key.startswith("_"):
             raise AttributeError("can't use this API to access hidden attributes")
```
stevenyoungs, line 72:
```suggestion
            raise AttributeError("this method cannot be used to access hidden attributes")
```
Improve message?

gramps/gen/lib/test/datadict_test.py
```diff
@@ -0,0 +1,55 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+""" Unittest for DataDict API """
```
stevenyoungs, line 21:
```suggestion
""" Unittest for DataDict """
```

gramps/gen/lib/test/datadict_test.py
```diff
@@ -0,0 +1,55 @@
...
+import unittest
+
+from gramps.gen.lib.serialize import DataDict, to_dict, from_dict
+from gramps.gen.lib import (
+    Person,
+    Family,
+)
+
+
+class DataDictTest(unittest.TestCase):
```
stevenyoungs, line 34: Add unittests for DataList as well?

Nick-Hall:
This looks OK to me, but we usually wait 3 weeks to give all developers the opportunity to comment on a PR.

gramps/gen/lib/test/datadict_test.py
dsblank: `DataList` is checked implicitly inside the `DataDict` tests. But, I can add some specific one too.

dsblank: Done!

gramps/gen/lib/test/datadict_test.py
```diff
@@ -53,3 +54,68 @@ def test_family_to_dict(self):
...
+        assert len(dl) == 0
+
+    def test_value_1(self):
+        dl = DataList([42])
+        assert isinstance(dl, DataList)
+        assert isinstance(dl[0], int)
+        assert dl[0] == 42
+
+    def test_value_exception(self):
+        with pytest.raises(TypeError, match="'int' object is not iterable"):
```
Nick-Hall, line 77: No need to use pytest here.  Use `self.assertRaises`.

stevenyoungs:
This PR looks good to me now

gramps/gen/lib/test/datadict_test.py
```diff
@@ -0,0 +1,121 @@
...
+    Family,
+)
+
+
+class DataDictTest(unittest.TestCase):
+    def test_person(self):
+        p = Person()
+        d = DataDict(p)
+        # Get the object from the dict:
+        assert id(p) == id(from_dict(d))
```
Nick-Hall, line 39: Use `self.assertEqual` instead.

gramps/gen/lib/test/datadict_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def test_person_to_dict(self):
+        p = Person()
+        d = to_dict(p)
+        assert not hasattr(d, "_object")
```
Nick-Hall, line 44: Use `self.assertFalse` here.

gramps/gen/lib/test/datadict_test.py
```diff
@@ -0,0 +1,121 @@
...
+    def test_family_to_dict(self):
+        p = Family()
+        d = to_dict(p)
+        assert not hasattr(d, "_object")
+
+
+class DataListTest(unittest.TestCase):
+    def test_empty_1(self):
+        dl = DataList()
+        assert isinstance(dl, DataList)
```
Nick-Hall, line 61: Use `self.assertIsInstance` here.

Nick-Hall (changes requested):
We shouldn't be using the python `assert` statement in our unit tests.  The `TestCase` assert methods should be used instead.

gramps/gen/lib/test/datadict_test.py
dsblank: Sure, I can make those self.assert changes. Do you know about the "Add a suggestion" button when making a comment?

Nick-Hall: No.  I haven't used that before.

dsblank: Here is a short description with screen shot: https://haacked.com/archive/2019/06/03/suggested-changes/

stevenyoungs: You can do multi-line suggestions as well.
One gotcha is that the suggestion may fall foul of our linting rules. This happened elsewhere in this PR where I suggested a longer message which then failed linting. Doug had to go back and make a new commit to correct. Sorry @dsblank

dsblank:
orjson for regular json loads/dumps is a significant speed increase.

Nick-Hall:
> orjson for regular json loads/dumps is a significant speed increase.

@dsblank Please don't add new features to an existing PR.  Although we are in a feature freeze this is not absolute.  We cab consider new PRs on a case by case basis.

dsblank:
@Nick-Hall my mistake. But please consider this commit as it can have a huge impact on the reception of Gramps 6.0.

Nick-Hall:
@dsblank I have no problem accepting this commit.  Just let me know if you want to add anything else.  I'd really like to start focusing on a release now.

dsblank:
Is this release going to be a pre-release? What is the next release, and what is the target date? Sorry if this is posted info.

I don't have anything critical, but the #1794 is also going to have a large impact.

dsblank:
I reverted the orjson changes, so as soon as this passes, nothing more from me.

Nick-Hall:
> What is the next release, and what is the target date? Sorry if this is posted info.

I was planning a beta release at the end of January and a final release at the end of February.  I posted to the list and sent you a private email.  Perhaps you didn't receive it?  My "gramps-project.org" email address has been unreliable recently.

Nick-Hall:
> I reverted the orjson changes

Any idea what the problem was?

dsblank:
I was being to aggressive in application. I'll take my time and add a new PR after this is merged.

emyoulation:
> I was planning a beta release at the end of January and a final release at the end of February. 

The beta cycle for 5.2 fractured the release notes and Road Map. 

Is there any way the final release will be able to have a consolidated Release Note (including all beta version change log items in addition to the differences between final release candidate and the public release) on this cycle?

dsblank:
All working. Release the hounds!

dsblank:
Added tested orjson use in #1861

Nick-Hall (approved):

Nick-Hall:
Rebased.

PR #1840, opened 2025-01-03
Update build_deb.yml for Ubuntu 24.04 deb

dasilvatek:
Ubuntu 20.04 goes end of life April 2025 so we should update to include 24.04

### DIFF
--- a/.github/workflows/build_deb.yml
+++ b/.github/workflows/build_deb.yml
@@ -34,7 +34,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04]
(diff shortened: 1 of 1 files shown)


dasilvatek:
Tested locally and installed the new 24.04 deb successfully. No issues with the application or dependencies.

I did have an issue opening the prior Gramps 5.1.6 (default Ubuntu package) Family Tree I had stored locally which failed due to incompatibility with 5.2.3. Luckily, I'm also running Gramps Web 5.2.3 on another box where I had imported my family tree - was able to export it out of there and import it locally, so I'm back up running.

My secondary goal with this update is to [Synchronize Gramps Web and Gramps Desktop](https://www.grampsweb.org/administration/sync/)

dasilvatek:
@Nick-Hall I don't see Ubuntu/Deb assets in the latest 6.0.1 release - is something broken?

PR #1844, opened 2025-01-04
Fix Citations gramplet to recognize event reference citations

DaveSch-gramps:
Fixes Citations gramplet to recognize citations attached to an event's reference section


DaveSch-gramps:
Bug #0013555

wroldwiedbwe:
@Nick-Hall Same with this one https://gramps-project.org/bugs/view.php?id=13555  the bug is actually a feature enhancement for the next release and not a bug fix!

PR #1845, opened 2025-01-09
Fix German date handler for reports

Nick-Hall:
Date displayers have an associated locale which can cause problems if more then one locale uses the same handler.

A similar fix may be needed for pt_PT and pt_BR and also zh_TW and zh_HK which also share handlers.

Fixes [#13312](https://gramps-project.org/bugs/view.php?id=13312).


dsblank (approved):
Make sense. 

(If we had git hub issues enabled, I'd link to this in an issue so we don't forget to check the others mentioned.)

Nick-Hall:
There is a thread on the _gramps_devel_ mailing list to discuss a [Proposal to use GitHub Issues](https://sourceforge.net/p/gramps/mailman/message/59113680/).

PR #1846, opened 2025-01-12
Allow importing and exporting nicknames to CSV

thcrt:
This allows nicknames to be exported and imported to and from CSV files. They are serialised and deserialised with the header "nickname" or "nick".

Resolves [#13566](https://gramps-project.org/bugs/view.php?id=13566).


emyoulation:
This is limited to the nickname of the Preferred Name (but not alternative names), correct?

Nick-Hall:
You will also need to update the associated unit test.

emyoulation:
The exporter uses a camel-case "Nick" literal string or translated "Nickname" in the CSV.  

But the Importer only has the lower-case in its untranslated string list. The "Nick" is in the weblate strings.

Does this ensure that an exported "Nick" header CSV will ALWAYS be recognized when importing on a non-english Gramps?

thcrt:
@Nick-Hall Just to be sure I understand: I should have to edit files in `data/tests`, i.e. `exp_sample_csv.csv` and `imp_sample_csv.csv`? It doesn't look like the test code in `plugins/tests/imports_test.py` actually specifies the fields to be imported and exported.

This is my first time ever looking at the Gramps code base so please let me know if I've misunderstood something :)

thcrt:
@emyoulation I really don't know. I don't understand how weblate is used with Gramps. To tell you the truth, I just guessed how to structure it based on what was already there.

It definitely looks like `exportcsv.py` should use `"Nickname"` as both a literal string and translation key. That's a mistake on my part.

The other attributes handled by the importer have all-lowercase literals and only use capitalisation for translation keys, FWIW.

emyoulation:
> The other attributes handled by the importer have all-lowercase literals and only use capitalisation for translation keys, FWIW.

Yes. That CSVs exported from Gramps already cannot import without editing the Titles is bad. Let' us not add to the incompatibility for end-users who have existing scripts to handle the other incompatibilities.

You may want to consider adding a Nickname column to the Configuation options for the 2 core People view modes.

emyoulation:
> The other attributes handled by the importer have all-lowercase literals and only use capitalisation for translation keys, FWIW.

Yes. That CSVs exported from Gramps already cannot import without editing the Titles is bad. Let' us not add to the incompatibility for end-users who have existing scripts to handle the other incompatibilities.

You may want to consider adding a Nickname column to the Configuation options for the 2 core People view modes.

emyoulation:
BTW, have you noticed the [CSV Dialect](https://www.gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Settings#CSV_Dialect) delimiter Configuration option for the People view's? The Gramps default is for comma separated values. But if you enable the Tab delimiter, you can copy'n'paste directly from a spreadsheet into the [Import Text](https://www.gramps-project.org/wiki/index.php/Addon:Import_Text_Gramplet) gramplet.

You will still have to import the Gramps CSV file into your spreadsheet. But then you can use a simplified workflow with copy'n'paste.

thcrt:
> Let us not add to the incompatibility for end-users who have existing scripts to handle the other incompatibilities.

Done in 561d98e928c55dac92b0cda3b20eec8de64b450f.

> You may want to consider adding a Nickname column to the Configuation options for the 2 core People view modes.

I've just looked into that, and it's pretty far over my head. I'm primarily interested in being able to import nicknames, since that's the functionality I need for my own use case :^)

> BTW, have you noticed the [CSV Dialect](https://www.gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Settings#CSV_Dialect) delimiter Configuration option for the People view's?

Yeah, I know I can adjust the delimiter, but that doesn't mean anything for my specific use case. I'm importing CSVs that are generated by code, not just directly exported from a spreadsheet. 

See: https://github.com/thcrt/poppa/blob/5fd1e276fadfda2cb3a6d296ce1bb30bb2a63ff2/src/poppa/export.py#L69-L76

Nick-Hall:
> Just to be sure I understand: I should have to edit files in data/tests, i.e. exp_sample_csv.csv and imp_sample_csv.csv?

Yes.  You should modify the test data so that it tests your new feature.  Just check that the current failure of the export test is expected and not a problem that you have caused.

thcrt:
Tests have been updated and should now run fine if my understanding is correct :)

dsblank (approved):
LGTM! I'm not sure that both "nick" and "nickname" need be allowed, but since you already did it, that's fine. 

M-Dawg :) 

Thanks for this!

Nick-Hall:
Rebased.

thcrt:
Nice! Thanks everyone for your patience dealing with a first-time contributor.

PR #1819, opened 2024-12-07
Add gender symbol option to the detailed descendant report

dave-khuon:
Add an option to insert a gender symbol before the person's name

PR #1818, opened 2024-12-07
Add gender symbol option to the detailed ancestral report

dave-khuon:
Add an option to insert a gender symbol before the person's name

PR #1857, opened 2025-01-18
Improved descend report

dave-khuon:


dave-khuon:
Replacing PR #1820.  Refactoring get_gender_symbol to utils.py per Nick's suggestion.

Nick-Hall:
You probably don't want to instantiate a new `Symbols` class for each call.  Some of our existing code creates a new instance in response to font changes.  Does this give us a problem in reports where the system font isn't used and a font can be defined at the paragraph level?

It is also worth noting that the gender field cannot be `None`, so your function reduces to a single line.  Perhaps it is better not to use a function at all.  This needs more thought.

Nick-Hall:
Actually, we probably don't have a problem with reports.  The font issue is related to fallback symbols which are not defined for genders.  If a chosen report font doesn't support the gender characters, then there isn't much we can do about it.

gramps/gen/plug/report/utils.py
```diff
@@ -46,6 +47,8 @@
 from ...utils.file import media_path_full
 from ..docgen import IndexMark, INDEX_TYPE_ALP
 
+from gramps.gen.utils.symbols import Symbols
```
Nick-Hall, line 50: This import should be made relative and moved next to the `media_path_full` import.

gramps/gen/plug/report/utils.py
```diff
@@ -423,3 +426,18 @@ def get_family_filters(database, family, include_single=True, name_format=None):
...
     the_filters.extend(CustomFilters.get_filters("Family"))
     return the_filters
+
+
+# -------------------------------------------------------------------------
+#
+# Get Gender Symbol of the person
+#
+# -------------------------------------------------------------------------
+def get_gender_symbol(person):
```
Nick-Hall, line 436: This whole function could be reduced to:
```
return SYMBOLS.get_symbol_for_string(person.gender)
```
It's better not to include a space in the utility function.  You don't know how it will be used.

dave-khuon: I don't know why!  I kept getting errors when starting the app.
The only way for me to make it work is to invoke the function this way:
  return Symbols().get_symbol_for_string(person.gender)
  where Symbols is the name of the import.

Nick-Hall: I'll investigate.

Nick-Hall: Fixed.

dave-khuon:
Hi Nick,

Since this PR# has been successfully merged, is it ok for me to push similar changes for the other 2 reports (PRs #1819, #1818) that are dependent on the new utils.py to benefit from the same feature benefits?  Please let me know.

Nick-Hall:
Yes, please go ahead.  I'll merge them when you're ready.

PR #1859, opened 2025-01-19
Use Ubuntu 22.04 for CI

DavidMStraub:
In https://github.com/gramps-project/gramps/pull/1858, I realized the CI currently runs on Ubuntu 20.04, which uses Python 3.8, that we are not supporting anymore. This PR switches conservatively to 22.04 LTS, which uses Python 3.10. One package had to be removed as well to make it run, but I believe this is anyway an old Python 2 leftover.

### DIFF
--- a/.github/workflows/gramps-ci.yml
+++ b/.github/workflows/gramps-ci.yml
@@ -28,7 +28,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Thew new package name for `python-libxml2` is `python3-libxml2`.

dsblank (approved):
LGTM!

Nick-Hall:
56 unit tests are being skipped.

DavidMStraub:
> 56 unit tests are being skipped.

But they are currently skipped everywhere, also on master.

> Thew new package name for python-libxml2 is python3-libxml2.

Thanks, added back.

Nick-Hall:
The tests that are skipped are from `gramps.gen.merge.test.merge_ref_test` and the message is:

```
skipped 'These tests need the 3rd-party plugins "CliMerge" and "ExportRaw".'
```

The _gramps60_ branch is not yet available for the addons, so this is understandable.

These are also the tests that require the lxml functionality.

DavidMStraub:
I'm confused, nothing added in this PR changes the number of tests skipped compared to main, right?

Nick-Hall:
We are trying to load the v5.2 version of CliMerge and ExportRaw which is going to fail.  The tests are then skipped.

Everything is working in the _gramps52_ branch.  I checked the unit tests as part of the v5.2.4 release process.

Nick-Hall:
Rebased.

PR #1861, opened 2025-01-21
Use orjson, if available

dsblank:
This PR uses the orjson fast loads/dumps for two specific places:

1. converting from string to data (dict)
2. converting data (dict) to string

It cannot be used in other places as it does not support object creation.

Using orjson has a large increase in performance where it can be used, namely in places that use `db.get_raw_OBJECT_data()` and `db._iter_raw_OBJECT_data()`. Those have already been used throughout the Gramps codebase.

Test code for performance:

```python
>>> from gramps.gen.db.utils import open_database
>>> db = open_database("Example")

>>> %%timeit
>>> count = 0
>>> for person in db._iter_raw_person_data():
...     count += 1

>>> db.close()
```

Scenario | Time in millseconds
-------- | --------------:
PR       | 43.7
PR + orjson installed         | 18.8

### DIFF
--- a/aio/build.sh
+++ b/aio/build.sh
@@ -41,6 +41,7 @@ pacman -S --needed --noconfirm \
     mingw-w64-x86_64-python-requests \
     mingw-w64-x86_64-python-setuptools \
     mingw-w64-x86_64-python-wheel \
+    mingw-w64-x86_64-rust \
     mingw-w64-x86_64-toolchain \
     perl-XML-Parser \
     subversion \
@@ -57,7 +58,7 @@ source $pythonvenv/bin/activate
 
 ## prerequisites in pip packages
 python -m pip install --upgrade pip
-pip install --upgrade pygraphviz pydot pydotplus requests asyncio
+pip install --upgrade asyncio orjson pydot pydotplus pygraphviz requests
 
 ## download dictionaries
 mkdir -p /mingw64/share/enchant/hunspell
--- a/README.md
+++ b/README.md
@@ -41,6 +41,11 @@ to your language:
 
 The following packages are **STRONGLY RECOMMENDED** to be installed:
 --------------------------------------------------------------------
+*  [**orjson**](https://pypi.org/project/orjson/)
+
+  A fast JSON library for Python. Used to increase the performance of Gramps.
+  The package name is usually orjson, python-orjson or python3-orjson.
+
 *  [**osmgpsmap**](https://nzjrs.github.io/osm-gps-map/)
 
  Used to show maps in the [Geography Category](https://gramps-project.org/wiki/index.php?title=Gramps_5.1_Wiki_Manual_-_Categories#Geography_Category).
--- a/aio/setup.py
+++ b/aio/setup.py
@@ -88,6 +88,7 @@ PACKAGES = [
     "pydotplus",
     "pygraphviz",
     "pydot",
+    "orjson",
 ]
 EXCLUDES = [
     "tkinter",
--- a/gramps/gen/lib/serialize.py
+++ b/gramps/gen/lib/serialize.py
@@ -32,14 +32,26 @@ import json
 import pickle
 import logging
 
+LOG = logging.getLogger(".serialize")
+
+try:
+    import orjson
+
+    json_dumps = orjson.dumps
+    json_loads = orjson.loads
+except ImportError:
+    LOG.warning(
+        "The Python package 'orjson' is not installed; performance may be degraded."
+    )
+    json_dumps = json.dumps
+    json_loads = json.loads
+
 # ------------------------------------------------------------------------
 #
 # Gramps modules
 #
 # ------------------------------------------------------------------------
 import gramps.gen.lib as lib
-
-LOG = logging.getLogger(".serialize")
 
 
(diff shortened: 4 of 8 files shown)


Nick-Hall:
We should add `orjson` to the strongly recommended section in the [README](https://github.com/gramps-project/gramps/blob/master/README.md#the-following-packages-are-strongly-recommended-to-be-installed).

stevenyoungs:
Are there platforms where orjson is not supported and gramps is?
In other words could orjson be mandatory

dsblank:
> could orjson be mandatory

That is a tough question, and applies to all future uses of Gramps too. So, I think to be safe, no. But we should make sure all of the standard Gramps packages include it. (I wish we supported pip install and requirements).

But maybe a warning is in order if it is not installed.

Nick-Hall:
I can't find a package in the [MSYS2](https://packages.msys2.org) repository, so we may have to investigate building a package ourselves for Windows.

They have `ujson`, but not `orjson`.  This is surprising since the `ujson` site recommends that people upgrade to `orjson`.  I thought that `orjson` was available.

stevenyoungs:
Orjson has minor differences compared to json. They seem like very edge cases for gramps. I can't imagine they will ever be a problem but worth noting. 

https://github.com/ijl/orjson?tab=readme-ov-file#str

Theoretically it sounds like you could craft a JSON string that loaded with json but failed with orjson. How you'd actually get that bad UTF16 into the gramps db in the first place is another challenge!

Nick-Hall:
@jralls Is the `python-orjson` package easy to include in the Mac bundle?

jralls:
@Nick-Hall no harder than any other Pypi package.

jralls:
Doug, I have a commit to add orjson to the mac build. May I push it to this branch?

dsblank:
It's ok by me. Nick likes to wait some time, but this seems like a no-brainer.

jralls:
It adds to the PR. It's still up to Nick about when to merge it. Speaking of which

> I can't find a package in the [MSYS2](https://packages.msys2.org/) repository, so we may have to investigate building a package ourselves for Windows.

@Nick-Hall can you add a `pip install orjson` to the Win32 build script? It does build easily enough with python build though it also wants Rust.

stevenyoungs:
I don't have permission to push to the repo and I couldn't find a way to make a suggestion on an unmodified file in a PR.
This diff should be sufficient to install orjson on Windows AIO agents.
There may be some changes required in `aio/setup.py` as well. I'll look later

```suggestion 
diff --git a/aio/build.sh b/aio/build.sh
--- a/aio/build.sh
+++ b/aio/build.sh
@@ -41,6 +41,7 @@ pacman -S --needed --noconfirm \
     mingw-w64-x86_64-python-requests \
     mingw-w64-x86_64-python-setuptools \
     mingw-w64-x86_64-python-wheel \
+    mingw-w64-x86_64-rust \
     mingw-w64-x86_64-toolchain \
     perl-XML-Parser \
     subversion \
@@ -57,7 +58,7 @@ source $pythonvenv/bin/activate

 ## prerequisites in pip packages
 python -m pip install --upgrade pip
-pip install --upgrade pygraphviz pydot pydotplus requests asyncio
+pip install --upgrade pygraphviz pydot pydotplus requests asyncio orjson

 ## download dictionaries
 mkdir -p /mingw64/share/enchant/hunspell
```

stevenyoungs:
Here's the change required for Windows AIO to install orjson
```suggestion
diff --git a/aio/setup.py b/aio/setup.py
--- a/aio/setup.py
+++ b/aio/setup.py
@@ -88,6 +88,7 @@ PACKAGES = [
     "pydotplus",
     "pygraphviz",
     "pydot",
+    "orjson",
 ]
 EXCLUDES = [
     "tkinter",
```

`lib/orjson.orjson.cp312-mingw_x86_64_msvcrt_gnu.pyd` is then present on an installed version of Gramps AIO 6.0
After adding `import orjson` to serialize.py (without `try` `catch`), Gramps still starts. I've not tested the full PR on Windows.

aio/build.sh
```diff
@@ -57,7 +58,7 @@ source $pythonvenv/bin/activate
 
 ## prerequisites in pip packages
 python -m pip install --upgrade pip
-pip install --upgrade pygraphviz pydot pydotplus requests asyncio
+pip install --upgrade pygraphviz pydot pydotplus requests asyncio orjson
```
stevenyoungs, line 61:
```suggestion
pip install --upgrade asyncio orjson pydot pydotplus pygraphviz requests
```
list the packages in alphabetical order for easier maintenance

Nick-Hall:
I've just created a [Windows AIO](https://github.com/gramps-project/gramps/actions/runs/12967102188) from this branch.  It would be helpful if someone with Windows installed it and checked that orjson is listed in the About dialog box.

stevenyoungs:
> I've just created a [Windows AIO](https://github.com/gramps-project/gramps/actions/runs/12967102188) from this branch. It would be helpful if someone with Windows installed it and checked that orjson is listed in the About dialog box.

Looks good:
image

Nick-Hall:
Rebased.

PR #1865, opened 2025-01-23
Add the standard search bar to the SelectPerson dialog

stevenyoungs:
Add the standard search bar to the SelectPerson dialog

PR #1869, opened 2025-01-24
Fix exactsearch filters

stevenyoungs:
Some filters require exact search, "is", rather than "contains". This detail is given in the second value of the tuple, pair[1].
Correct the code so that it uses pair[1]

### DIFF
--- a/gramps/gui/selectors/baseselector.py
+++ b/gramps/gui/selectors/baseselector.py
@@ -315,8 +315,9 @@ class BaseSelector(ManagedWindow):
         """
         Builds the selection people see in the Selector
         """
+        # any search_bar filter takes precedence over self.filter
         search_bar_filter = self.search_bar.get_value()
-        if (not self.filter[1]) or (search_bar_filter[1]):
+        if search_bar_filter[1]:
             filter_info = (
                 0,
                 search_bar_filter,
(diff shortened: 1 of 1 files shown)


stevenyoungs:
This PR needs the bug label applying.

stevenyoungs:
To test either apply PR #1865 or change this line to `True`
https://github.com/gramps-project/gramps/blob/4e18da0a5aab7084ef6bdcad5d50d7ae5c0b8aff/gramps/gui/selectors/selectperson.py#L63

gramps/gui/selectors/baseselector.py
```diff
@@ -315,8 +315,14 @@ def build_tree(self):
         """
         Builds the selection people see in the Selector
         """
-        if not self.filter[1]:
-            filter_info = (False, self.search_bar.get_value(), False)
+        # any search_bar filter takes precedence over self.filter
```
Nick-Hall, line 318: I'm not sure if we want this.  Why should a search take precedence over a filter?

stevenyoungs: I agree. PR #1865 does just that.
I'm tempted to cherry-pick the first commit (which is a bug fix) into PR #1865 and abandon this one.
Can you let me know if you agree?

stevenyoungs:
Closing in favour of #1865

gramps/gui/selectors/baseselector.py
Nick-Hall: Yes.  We can still keep it as a separate commit.

PR #1870, opened 2025-01-25
New feature

dave-khuon:
Allow user to fill the person and marriage boxes with color.  Add new named boxes to control their BG color.


dave-khuon:
withdrawn

PR #1872, opened 2025-01-25
Use the correct variable in baseselector to prevent error

stevenyoungs:
PR #1836 introduced a bug when selecting a single object.
This PR fixes it.

To reproduce:
1. using a db with at least one person
2. Shift+Alt+F to add a new family
3. Click the select child button
4. Select a child and click OK

An error similar to the following is generated
```
18794: ERROR: grampsapp.py: line 188: Unhandled exception
  File "C:\msys64\home\steve\gramps\gramps\gui\editors\editfamily.py", line 303, in share_button_clicked
    person = sel.run()
             ^^^^^^^^^
  File "C:\msys64\home\steve\gramps\gramps\gui\selectors\baseselector.py", line 248, in run
    result = self.get_from_handle_func()(id_list[0])
                                         ^^^^^^^
NameError: name 'id_list' is not defined
```

### DIFF
--- a/gramps/gui/editors/editfamily.py
+++ b/gramps/gui/editors/editfamily.py
@@ -412,24 +412,6 @@ class ChildEmbedList(DbGUIElement, EmbeddedList):
             return name
 
 
-class FastMaleFilter:
-    def __init__(self, db):
-        self.db = db
-
-    def match(self, handle, db):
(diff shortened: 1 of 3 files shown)


Nick-Hall:
It looks like this fix was already included in PR #1865.

PR #1875, opened 2025-01-25
Make cleanup optional after an AIO build

stevenyoungs:
If the first argument passed to build.sh is "true", cleanup the python virtual environment at the end of the script.

For local development it is useful to keep the python virtual environment after the workflow completes.
For cloud runners, there is no need to cleanup as a new runner is provisioned for each workflow run
For locally testing build scripts it is helpful to cleanup the virtual environment. Since this is the less common scenario, default to not cleaning the environment.

For Windows developers it's then a case of install MSYS2, clone gramps sources, `cd aio; ./build.sh`
The leftover python virtual env (`c:\msys64\tmp\grampspythonenv\bin\python.exe` by default) can then be configured in Visual Studio Code and used to debug etc.

### DIFF
--- a/.github/workflows/windows-aio.yml
+++ b/.github/workflows/windows-aio.yml
@@ -19,7 +19,14 @@
 
 name: Windows AIO
 
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+    inputs:
+      cleanup:
+        description: 'Cleanup python environment on completion?'
+        type: boolean
+        required: false
+        default: false
 
 jobs:
   build:
@@ -38,7 +45,7 @@ jobs:
     - name: Build
       run: |
         cd aio
-        ./build.sh
+        ./build.sh ${{ inputs.cleanup }}
     - uses: actions/upload-artifact@v4
       with:
         name: GrampsAIO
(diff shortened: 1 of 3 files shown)


stevenyoungs:
I don't yet understand how my AIO only change broke unit testing on Ubuntu but this message in the logs caught my attention

```
The Python package 'orjson' is not installed; performance may be degraded.
```

orjson is now strongly recommended (and default in AIO and Mac). Our unit test runner does not have orjson installed. Therefore we are unit testing a configuration that is likely to be used by a minority of users.

stevenyoungs:
No code changes and now the unit tests are passing. 🤷

hgohel:
Good change. The last sentence of your description would be a useful addition to the AIO build [README](https://github.com/gramps-project/gramps/blob/master/aio/README.md):

_The leftover python virtual env (c:\msys64\tmp\grampspythonenv\bin\python.exe by default) can then be configured in Visual Studio Code and used to debug etc._

stevenyoungs:
> Good change. The last sentence of your description would be a useful addition to the AIO build [README](https://github.com/gramps-project/gramps/blob/master/aio/README.md):
> 
Addressed in the most recent commit. 👏

PR #1876, opened 2025-01-26
Use orjson + encoder

dsblank:
This PR uses orjson with a hand-coded encoder/decoder everywhere, and assumes orjson is installed.

Tests shows that this is 60% faster than master for `db.iter_people()` on a 40k person table.

### DIFF
--- a/.github/workflows/gramps-ci.yml
+++ b/.github/workflows/gramps-ci.yml
@@ -56,6 +56,7 @@ jobs:
         sudo apt-get install python3-libxml2
         sudo apt-get install zlib1g-dev
         sudo apt-get install python3-setuptools
+        python3 -m pip install orjson
     - name: Install addons
       run: |
         mkdir -p ~/.gramps/gramps60/plugins/
--- a/gramps/gen/lib/json_utils.py
+++ b/gramps/gen/lib/json_utils.py
@@ -1,0 +1,219 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2017,2024       Nick Hall
+# Copyright (C) 2024-2025       Doug Blank
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+# ------------------------------------------------------------------------
+#
+# Python modules
+#
+# ------------------------------------------------------------------------
+
+import json
+import orjson
+
+# ------------------------------------------------------------------------
+#
+# Gramps modules
+#
+# ------------------------------------------------------------------------
+
+import gramps.gen.lib as lib
+
+
+NoneType = type(None)
+
+
+class DataDict(dict):
+    """
+    A wrapper around a data dict that also provides an
+    object interface.
+    """
+
+    def __init__(self, data=None):
+        """
+        Wrap a data dict (raw data) or object
+        with an attribute API. If data is an
+        object, we use it to get the attributes.
+        """
+        if isinstance(data, dict):
+            super().__init__(data)
+        else:
+            super().__init__()
+            if data:
+                # Data is actually a Gramps object
+                self.update(object_to_data(data))
+
+    def __str__(self):
+        if "_object" not in self:
+            self["_object"] = data_to_object(self)
+        return str(self["_object"])
+
+    def __getattr__(self, key):
+        if key.startswith("_"):
+            raise AttributeError(
+                "this method cannot be used to access hidden attributes"
+            )
+
+        if "_object" in self:
+            return getattr(self["_object"], key)
+        elif key in self:
+            value = self[key]
+            if isinstance(value, dict):
+                return DataDict(value)
+            elif isinstance(value, list):
+                return DataList(value)
+            else:
+                return value
+        else:
+            self["_object"] = data_to_object(self)
+            return getattr(self["_object"], key)
+
+
+class DataList(list):
+    """
+    A wrapper around a data list.
+    """
+
+    def __getitem__(self, position):
+        value = super().__getitem__(position)
+        if isinstance(value, dict):
+            return DataDict(value)
+        elif isinstance(value, list):
+            return DataList(value)
+        else:
+            return value
+
+
+def remove_object(data):
+    """
+    Remove _object, if any, from dicts.
+    """
+    if isinstance(data, dict):
+        return {
+            key: remove_object(value) for key, value in data.items() if key != "_object"
+        }
+    elif isinstance(data, (list, tuple)):
+        return [remove_object(item) for item in data]
+    else:
+        return data
+
+
+def convert_state_to_object(obj_dict):
+    _class = obj_dict.pop("_class")
+    cls = lib.__dict__[_class]
+    obj = cls.__new__(cls)
+    obj.set_object_state(obj_dict)
+    return obj
+
+
+def convert_object_to_state(obj):
+    return obj.get_object_state()
+
+
+def string_to_data(string: str | bytes) -> DataDict:
+    """
+    Convert a JSON string into its data representation.
+    """
+    return DataDict(orjson.loads(string))
+
+
+def string_to_dict(string: str | bytes) -> dict:
+    """
+    Convert a JSON string into its dict representation.
+    """
+    return orjson.loads(string)
+
+
+def dict_to_string(dict: dict) -> str | bytes:
+    """
+    Convert a dict into its JSON string representation.
+    """
+    return orjson.dumps(dict)
+
+
+def object_to_dict(obj):
+    """
+    Convert any Gramps lib object into a dict representation.
+    """
+    if isinstance(obj, (float, int, str, NoneType)):
+        return obj
+
+    elif isinstance(obj, (list, tuple)):
+        return [object_to_dict(item) for item in obj]
+
+    state = obj.get_object_state()
+    return {k: object_to_dict(v) for k, v in state.items()}
+
+
+def object_to_data(obj):
+    """
+    Convert any Gramps lib object or other value into
+    its dict or list representation or value, respectively.
+    """
+    if isinstance(obj, (float, int, str, NoneType)):
+        return obj
+
+    elif isinstance(obj, (list, tuple)):
+        return DataList([object_to_data(item) for item in obj])
+
+    state = obj.get_object_state()
+    data = {k: object_to_data(v) for k, v in state.items()}
+    # Stash the object in case we need it later:
+    data["_object"] = obj
+    return DataDict(data)
+
+
+def data_to_object(data):
+    """
+    Convert any object dict representation to a Gramps lib object
+    or other value.
+    """
+    if isinstance(data, dict):
+        if "_object" in data:
+            return data["_object"]
+
+        data = {k: data_to_object(v) for k, v in data.items()}
+        return convert_state_to_object(data)
+
+    elif isinstance(data, (list, tuple)):
+        return [data_to_object(item) for item in data]
+
+    return data
+
+
+def object_to_string(obj: object) -> str | bytes:
+    """
+    Convert any Gramps object into a JSON string/bytes.
+    """
+    return orjson.dumps(obj, default=convert_object_to_state)
+
+
+def data_to_string(data: DataDict):
+    """
+    Convert a DataDict into a string.
+    """
+    return orjson.dumps(remove_object(data))
+
+
+def string_to_object(string: str | bytes):
+    """
+    Convert a JSON string/bytes into a Gramps lib object.
+    """
+    return data_to_object(orjson.loads(string))
--- a/gramps/gui/views/treemodels/mediamodel.py
+++ b/gramps/gui/views/treemodels/mediamodel.py
@@ -46,7 +46,7 @@ from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
 from gramps.gen.datehandler import displayer, format_time
 from gramps.gen.lib import Date, Media
-from gramps.gen.lib.serialize import from_dict
+from gramps.gen.lib.json_utils import data_to_object
 from .flatbasemodel import FlatBaseModel
 
 
@@ -135,12 +135,12 @@ class MediaModel(FlatBaseModel):
 
     def column_date(self, data):
         if data.date:
-            date = from_dict(data.date)
+            date = data_to_object(data.date)
             return displayer.display(date)
         return ""
 
     def sort_date(self, data):
-        obj = from_dict(data)
+        obj = data_to_object(data)
         d = obj.get_date_object()
         if d:
             return "%09d" % d.get_sort_value()
--- a/gramps/gui/views/treemodels/repomodel.py
+++ b/gramps/gui/views/treemodels/repomodel.py
@@ -41,7 +41,7 @@ from gi.repository import Gtk
 #
 # -------------------------------------------------------------------------
 from gramps.gen.lib import Address, RepositoryType, Url, UrlType
-from gramps.gen.lib.serialize import from_dict
+from gramps.gen.lib.json_utils import data_to_object
 from gramps.gen.datehandler import format_time
 from .flatbasemodel import FlatBaseModel
 from gramps.gen.const import GRAMPS_LOCALE as glocale
@@ -142,7 +142,7 @@ class RepositoryModel(FlatBaseModel):
     def column_city(self, data):
         try:
             if data.address_list:
-                addr = from_dict(data.address_list[0])
+                addr = data_to_object(data.address_list[0])
                 return addr.get_city()
         except:
             pass
@@ -151,7 +151,7 @@ class RepositoryModel(FlatBaseModel):
     def column_street(self, data):
         try:
             if data.address_list:
-                addr = from_dict(data.address_list[0])
+                addr = data_to_object(data.address_list[0])
                 return addr.get_street()
         except:
             pass
@@ -160,7 +160,7 @@ class RepositoryModel(FlatBaseModel):
     def column_locality(self, data):
         try:
             if data.address_list:
-                addr = from_dict(data.address_list[0])
+                addr = data_to_object(data.address_list[0])
                 return addr.get_locality()
         except:
             pass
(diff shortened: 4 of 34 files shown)


Nick-Hall:
I think that we should make orjson mandatory.  It would simplify the code and there seems no good reason not to do so.  We are already going to include it in the Windows AIO and Mac bundle, and I have checked that all major Linux distributions include it.

dsblank:
> I think that we should make orjson mandatory. 

Yeah, I think you are right. Now that we can use it for all JSON load/dump, and it makes a big difference, it makes sense.

dsblank:
I think we can make it mandatory, but still have a fallback to regular json.loads/dumps. That way if someone runs Gramps on some weird architecture (or orjson fails for some reason), it will still have access to your data. That seems prudent with genealogical data for the future.

Nick-Hall:
By keeping both json and orjson you are adding extra code to maintain and extra tests to run for very little gain.

dsblank:
> By keeping both json and orjson you are adding extra code to maintain and extra tests to run for very little gain.

@Nick-Hall, it's up to you. It was useful during testing to make sure everything is correct, but I can remove it now that all tests are passing. Let me know if you want me to remove the json.loads/dumps and turn the warning into an error.

The next step is to remove to_json, from_json, to_dict, from_dict throughout the code base. Do you want me to do it in this PR, or a follow-up one?

.github/workflows/gramps-ci.yml
```diff
@@ -56,6 +56,7 @@ jobs:
         sudo apt-get install python3-libxml2
         sudo apt-get install zlib1g-dev
         sudo apt-get install python3-setuptools
+        python3 -m pip install orjson
```
Nick-Hall, line 59:
```suggestion
        sudo apt-get install python3-orjson
```

dsblank:
@Nick-Hall : `Unable to locate package python3-orjson` I think that is why I pip installed it.

Nick-Hall:
OK.  It looks like `python3-orjson` is only available on Ubuntu 24.10 and later.

dsblank:
@Nick-Hall, I have a refactor for this PR, and to fix usage across the code base. Do you want me to make a PR on this PR, or add it to this PR?

Nick-Hall:
Please add it to this PR.

Nick-Hall:
@dsblank Thanks.  Now we can review all the changes together as a whole.

Please remember to add new files to either `po/POTFILES.in` or `po/POTFILES.skip`.

dsblank:
WIP

kulath:
I don't think that methods like data_to_object should optionally omit the obj_class parameter. As far as I can see, in most cases the object class is known when the method is called.


```
    def data_to_object(obj_class, data):
    def data_to_object(data, obj_class=None):
        LOG.debug(
            "json, data_to_object: {'_class': %r, ...}",
            data["_class"] if (data and "_class" in data) else data,
        )
        return data_to_object(data)

``` 

(I don't understand how the call In the return in the above method is resolved, presumably to the data_to_object imported from utils, but that is probably just my ignorance)

dsblank:
> I don't think that methods like data_to_object should optionally omit the obj_class parameter. As far as I can see, in most cases the object class is known when the method is called.

None of the other JSON serializer methods require a class, but the Blob `serializer.data_to_object()` does. The class is only required (and added) so that the upgrade code will operate properly on blob data. 

I believe that we can add proper typing to `data` in the future, but we will need to see if it effects the speed of the methods. It *might* look something like:

```python
@singledispatch
def data_to_object(data):
    print("Default implementation for unknown type:", type(data))

@data_to_object.register(PersonData)
def _(data):
    obj: Person = data_to_object(data)
    return obj

@data_to_object.register(FamilyData)
def _(data):
    obj: Family = data_to_object(data)
    return obj
```

but we don't have any of the types created for that yet. At that point, we can even clean up the Blob serializer (and maybe even remove it) basing all operations on type.

Looking forward to @DavidMStraub's #1858 PR on which we can build.

kulath:
> I believe that we can add proper typing to `data` in the future, but we will need to see if it affects the speed of the methods.

I wouldn't like to see the change to JSON going ahead until proper typing has been added (even if it makes a minor decrease in speed).

As I suggest here: https://sourceforge.net/p/gramps/mailman/message/59123008/ I suggest that data_to_object should be modified to become get_foo_from struct_data (e.g. get_person_from_struct_data) so that the typing is explicitly clear in the code. If the data is typed, which would be a great idea for type checking  (thanks) then it should presumably become get_foo_from_foo_struct_data.

gramps/gen/lib/json_utils.py
```diff
@@ -0,0 +1,201 @@
...
+    obj = cls.__new__(cls)
+    obj.set_object_state(obj_dict)
+    return obj
+
+
+# Just call the method directly:
+# def convert_object_to_state(obj):
+#     return obj.get_object_state()
+
+
```
stevenyoungs, line 114:
```suggestion
```
Can this commented out code be deleted?

gramps/gen/lib/json_utils.py
```diff
@@ -0,0 +1,201 @@
...
+        return [data_to_object(item) for item in data]
+
+    return data
+
+
+def object_to_string(obj: object) -> str | bytes:
+    """
+    Convert any Gramps object into a JSON string/bytes.
+    """
+    return orjson.dumps(object_to_dict(obj))
```
stevenyoungs, line 187: I'm thinking out loud here: would having a custom `JSONEncoder` class allow us to skip the `object_to_dict` call?
See the `cls` parameter https://docs.python.org/3/library/json.html#basic-usage

gramps/gui/views/treemodels/mediamodel.py
```diff
@@ -135,12 +135,12 @@ def column_id(self, data):
 
     def column_date(self, data):
         if data["date"]:
-            date = from_dict(data["date"])
+            date = data_to_object(data["date"])
```
stevenyoungs, line 138:
```suggestion
            date = data_to_object(data.date)
```

gramps/gui/views/treemodels/repomodel.py
```diff
@@ -142,7 +142,7 @@ def column_name(self, data):
     def column_city(self, data):
         try:
             if data["address_list"]:
-                addr = from_dict(data["address_list"][0])
+                addr = data_to_object(data["address_list"][0])
```
stevenyoungs, line 145:
```suggestion
                addr = data_to_object(data.address_list[0])
```

gramps/gui/views/treemodels/repomodel.py
```diff
@@ -151,7 +151,7 @@ def column_city(self, data):
     def column_street(self, data):
         try:
             if data["address_list"]:
-                addr = from_dict(data["address_list"][0])
+                addr = data_to_object(data["address_list"][0])
```
stevenyoungs, line 154:
```suggestion
                addr = data_to_object(data.address_list[0])
```

gramps/gui/views/treemodels/repomodel.py
```diff
@@ -160,7 +160,7 @@ def column_street(self, data):
     def column_locality(self, data):
         try:
             if data["address_list"]:
-                addr = from_dict(data["address_list"][0])
+                addr = data_to_object(data["address_list"][0])
```
stevenyoungs, line 163:
```suggestion
                addr = data_to_object(data.address_list[0])
```

gramps/gui/views/treemodels/repomodel.py
```diff
@@ -169,7 +169,7 @@ def column_locality(self, data):
     def column_state(self, data):
         try:
             if data["address_list"]:
-                addr = from_dict(data["address_list"][0])
+                addr = data_to_object(data["address_list"][0])
```
stevenyoungs, line 172:
```suggestion
                addr = data_to_object(data.address_list[0])
```

gramps/gui/views/treemodels/repomodel.py
```diff
@@ -178,7 +178,7 @@ def column_state(self, data):
     def column_country(self, data):
         try:
             if data["address_list"]:
-                addr = from_dict(data["address_list"][0])
+                addr = data_to_object(data["address_list"][0])
```
stevenyoungs, line 181:
```suggestion
                addr = data_to_object(data.address_list[0])
```

gramps/gui/views/treemodels/repomodel.py
```diff
@@ -187,7 +187,7 @@ def column_country(self, data):
     def column_postal_code(self, data):
         try:
             if data["address_list"]:
-                addr = from_dict(data["address_list"][0])
+                addr = data_to_object(data["address_list"][0])
```
stevenyoungs, line 190:
```suggestion
                addr = data_to_object(data.address_list[0])
```

gramps/gui/views/treemodels/repomodel.py
```diff
@@ -196,7 +196,7 @@ def column_postal_code(self, data):
     def column_phone(self, data):
         try:
             if data["address_list"]:
-                addr = from_dict(data["address_list"][0])
+                addr = data_to_object(data["address_list"][0])
```
stevenyoungs, line 199:
```suggestion
                addr = data_to_object(data.address_list.[0])
```

stevenyoungs (commented):
You created `DataDict`, so let's use it!

Nick-Hall:
> You created DataDict, so let's use it!

Yes, that looks neater.

gramps/gen/lib/json_utils.py
dsblank: Not a crazy idea. Which do you think is faster?

1. json.dumps(obj, MyCustomEncoder) - Python converts all dict items to strings
2. orjson.dumps(object_to_dict(obj)) - Python converts objs to dict, orjson does the rest

I am always surprised when I guess, but I bet #2 is faster. Let's find out! Place your bets.

Hey, it looks like [orjson.dumps()](https://github.com/ijl/orjson?tab=readme-ov-file#serialize) accepts `default`. I didn't notice that before. There may be a third option here.

dsblank:
> You created DataDict, so let's use it!

Thanks! I think I converted some of this code before DataDict.

gramps/gen/lib/json_utils.py
dsblank: I left it in because some may wonder why the asymmetry.

dsblank:
> data_to_object should be modified to become get_foo_from struct_data

I think the consensus on the method renaming thread on the dev mailing list was to use singledispatch. I think I'll wait for such changes.

gramps/gen/lib/json_utils.py
stevenyoungs: Nice. I missed the support for `default`.

stevenyoungs: > Which do you think is faster?

well I felt a customencoder should be quicker if it avoids creating the full object hierarchy as a dict, only to then traverse the dict to convert to JSON. There should be fewer memory allocations this way. But the python allocator is quick. I await your test results with interest.....

dsblank:
@kulath said:

> (I don't understand how the call In the return in the above method is resolved, presumably to the (I don't understand how the call In the return in the above method is resolved, presumably to the data_to_object imported from utils, but that is probably just my ignorance) imported from utils, but that is probably just my ignorance)

Yes, you can see at the top of the file that `data_to_object` is imported from json_utils. Otherwise, you would have to use `self.` to call a method in the class.

gramps/gen/lib/json_utils.py
dsblank: I didn't use the class, but the default, which is what we started with (results in seconds, on a 40k person table):

1. `json.dumps(default=convert_object_to_state)`: 6.143402099609375
2. `orjson.dumps(object_to_dict(obj))`: 5.892540454864502
3. `orjson.dumps(person, default=convert_object_to_state)` 4.168360948562622

So, option 3 wins! Another 40% faster! I'll make that change tomorrow.

dsblank: I'll try the class API too, for completeness.

dsblank: 4. `json.dumps(person, cls=MyEncoder)`: 6.621660470962524

I'll go with 3.

Here is the class for inspection:

```python
class MyEncoder(json.JSONEncoder):
    def default(self, obj):
        if obj.__class__.__module__.startswith("gramps"):
            obj = obj.get_object_state()
            return self.default(obj)
        elif isinstance(obj, (list, tuple)):
            return str([self.default(item) for item in obj])
        else:
            return obj
```

gramps/gen/lib/json_utils.py
dsblank: Uncommented it because it is now used.

dsblank:
@stevenyoungs, I tried dealing with `object_to_data(object)` and `DataDict(object)` options a couple of different ways to keep the object, and this method seems the best disruption (it also puts the attributes in the `DataDict`). But it did require `remove_object()` in the `data_to_string()` method which removes "_object" in-place for speed. I didn't want to slow `data_to_string()` too much, but if you feel like making a copy of the dict without "_object" is better, let me know.

stevenyoungs:
Could you do anything with default or a custom json encoder so that _object does not have to be removed from DataDict/Data list, it just gets skipped in the dump?
The two downsides to the current approach are the cost of traversing to find and remove _object's and the potential future cost of adding it back again if the now _object less data continues to be used.
I've no such concerns in the test code.

dsblank:
> Could you do anything with default or a custom json encoder so that _object does not have to be removed from DataDict/Data list, it just gets skipped in the dump?

Yes. Why didn't I think of that?!

gramps/gen/lib/json_utils.py
```diff
@@ -0,0 +1,219 @@
...
+            value = self[key]
+            if isinstance(value, dict):
+                return DataDict(value)
+            elif isinstance(value, list):
+                return DataList(value)
+            else:
+                return value
+        else:
+            self["_object"] = data_to_object(self)
+            return getattr(self["_object"], key)
```
cdhorn, line 85:
```suggestion
        if key in self:
            value = self[key]
            if isinstance(value, dict):
                return DataDict(value)
            if isinstance(value, list):
                return DataList(value)
            return value
        self["_object"] = data_to_object(self)
        return getattr(self["_object"], key)
```
Several unnecessary else after return?

gramps/gen/lib/json_utils.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+
+    def __getitem__(self, position):
+        value = super().__getitem__(position)
+        if isinstance(value, dict):
+            return DataDict(value)
+        elif isinstance(value, list):
+            return DataList(value)
+        else:
+            return value
```
cdhorn, line 100:
```suggestion
        if isinstance(value, list):
            return DataList(value)
        return value
```
Also unneccessary else after return?

gramps/gen/lib/json_utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def object_to_dict(obj):
+    """
+    Convert any Gramps lib object into its DataDict representation.
+    """
+    if isinstance(obj, (float, int, str, NoneType)):
+        return obj
+
+    elif isinstance(obj, (list, tuple)):
```
cdhorn, line 143:
```suggestion
    if isinstance(obj, (list, tuple)):
```
Else after return.

gramps/gen/lib/json_utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+def object_to_data(obj):
+    """
+    Convert any Gramps lib object or other value into
+    its dict representation or value, respectively.
+    """
+    if isinstance(obj, (float, int, str, NoneType)):
+        return obj
+
+    elif isinstance(obj, (list, tuple)):
```
cdhorn, line 158:
```suggestion
    if isinstance(obj, (list, tuple)):
```
Else after return.

gramps/gen/lib/json_utils.py
```diff
@@ -0,0 +1,219 @@
...
+    or other value.
+    """
+    if isinstance(data, dict):
+        if "_object" in data:
+            return data["_object"]
+
+        data = {k: data_to_object(v) for k, v in data.items()}
+        return convert_state_to_object(data)
+
+    elif isinstance(data, (list, tuple)):
```
cdhorn, line 180:
```suggestion
    if isinstance(data, (list, tuple)):
```
Else after return.

gramps/gen/lib/json_utils.py
dsblank: Performance is the same. "Explicit is better than implicit."

gramps/gen/lib/json_utils.py
dsblank: Performance is the same. "Explicit is better than implicit."

gramps/gen/lib/json_utils.py
dsblank: Performance is the same. "Explicit is better than implicit."

gramps/gen/lib/json_utils.py
dsblank: Performance is the same. "Explicit is better than implicit."

gramps/gen/lib/json_utils.py
dsblank: I find it more clear with the else.

dsblank:
Thanks @cdhorn for the review!

gramps/gen/lib/json_utils.py
```diff
@@ -0,0 +1,219 @@
...
+        return [object_to_dict(item) for item in obj]
+
+    state = obj.get_object_state()
+    return {k: object_to_dict(v) for k, v in state.items()}
+
+
+def object_to_data(obj):
+    """
+    Convert any Gramps lib object or other value into
+    its dict representation or value, respectively.
```
stevenyoungs, line 168:
```suggestion
    its dict or list representation or value, respectively.
```
It returns a DataList in some circumstances

Nick-Hall:
@dsblank I'll let you resolve these conflicts.

dsblank:
Ok

dsblank:
@Nick-Hall, conflicts resolved and all tests passing.

Nick-Hall:
@dsblank Thank you.  I will do a final review now, and merge if I don't see any issues.

Nick-Hall:
Added two files to `POTFILES.skip`:

`gramps/gen/proxy/test/__init__.py`
`gramps/gen/proxy/test/proxies_test.py`

dsblank:
Thanks to everyone!

gramps/gen/lib/json_utils.py
```diff
@@ -0,0 +1,219 @@
...
+            super().__init__(data)
+        else:
+            super().__init__()
+            if data:
+                # Data is actually a Gramps object
+                self.update(object_to_data(data))
+
+    def __str__(self):
+        if "_object" not in self:
+            self["_object"] = data_to_object(self)
```
prculley, line 65: I'm not sure it is a good idea to cache the object here...  I was fooling around in my debugger and changed an element of the DataDict; only to be very confused when the debugger refused to show that the value had changed.  After looking at this I realized that the debugger was always showing the original values from the DataDict, not the current values.

Either this should not be cached, or we need to document that values in this dict should not be changed.

dsblank: That's a good point about having a cached version that is out of scope with a changed object. But all of the DataDict would be old too.

PR #1877, opened 2025-01-26
unittest both with and without orjson

stevenyoungs:
orjson is highly recommend, but not mandatory, so test both scenarios

This PR is so that we can discuss if we \ how we want to unit test with orjson
Originated from an observation in #1875. See [this](https://github.com/gramps-project/gramps/pull/1875#issuecomment-2614312272) comment.

### DIFF
--- a/.github/workflows/gramps-ci.yml
+++ b/.github/workflows/gramps-ci.yml
@@ -74,3 +74,5 @@ jobs:
       run: |
         export GDK_BACKEND=-
         python3 setup.py test
+        python3 -m pip install orjson
+        python3 setup.py test


Nick-Hall:
I think that we should make orjson mandatory. It would simplify the code and there seems no good reason not to do so. We are already going to include it in the Windows AIO and Mac bundle, and I have checked that all major Linux distributions include it.

dsblank:
I wrote #1876 so that we can test to make sure the orjson version is the same as the json version.

stevenyoungs:
Closed in favour of #1876

PR #1878, opened 2025-01-26
Add new place to example db

cajturner:
Add a new places with alternative name and place name dates. Added to support testing on this PR: https://github.com/gramps-project/gramps-web-api/pull/608


Nick-Hall:
I often used Saint Petersburg as an example when we were developing the hierarchical place enhancements.  You could also add Petrograd between 1914 and 1924.

cajturner:
@Nick-Hall is it possible for a rc or patch version to be created? Then I can pull this new place data into the web project

Nick-Hall:
I'm planning a beta release around the end of this month.

PR #1881, opened 2025-01-27
Safe mode working with Gramps 5.2 on Windows

hgohel:
Fixes [#13300](https://gramps-project.org/bugs/view.php?id=13300)

Safe mode which was broken with Gramps 5.2.0-5.2.4 on Windows, has been fixed with commit.

Previously code was looking for user configuration in APPDATA, but with the move to XDG-based directory layout (bug [8025](https://gramps-project.org/bugs/view.php?id=8025)), this was no longer correct.

Per Nick Hall's review comment: Use the old directory structure if new structure is not present *and* the old location isn't empty (i.e. upgrade scenario from pre-5.2 Gramps), otherwise, use the new XDG-based directory layout (i.e. new install scenario of Gramps 5.2+).

TESTING
Test to verify that [safe mode](https://gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Command_Line#Safe_mode) functions as described, i.e. starting from a user's customized environment, verify that existing family trees are available in safe mode but settings and configuration are not.

NOTE: This also fixes [#13261](https://gramps-project.org/bugs/view.php?id=13261) which documents another way in which this bug manifests. However, this is bug is closer to the root cause and easier to test.

For previous discussion of this change, see PR #1726

### DIFF
--- a/gramps/gen/const.py
+++ b/gramps/gen/const.py
@@ -100,10 +100,13 @@ if "GRAMPSHOME" in os.environ:
     HOME_DIR = os.path.join(USER_HOME, "gramps")
 elif "USERPROFILE" in os.environ:
     USER_HOME = get_env_var("USERPROFILE")
-    if "APPDATA" in os.environ:
-        HOME_DIR = os.path.join(get_env_var("APPDATA"), "gramps")
-    else:
-        HOME_DIR = os.path.join(USER_HOME, "gramps")
+    HOME_DIR = os.path.join(GLib.get_user_config_dir(), "gramps")
+    if not os.path.exists(HOME_DIR):
+        homedir = get_env_var("APPDATA") if "APPDATA" in os.environ else USER_HOME
+        homedir = os.path.join(homedir, "gramps")
+        dbpath = os.path.join(homedir, "grampsdb")
+        if os.path.isdir(dbpath) and os.listdir(dbpath):
+            HOME_DIR = homedir
 else:
     USER_HOME = get_env_var("HOME")
     HOME_DIR = os.path.join(USER_HOME, ".gramps")


jralls:
This change causes https://gramps-project.org/bugs/view.php?id=13686.
USERDATA [*is* used by macOS](https://github.com/gramps-project/gramps/blob/master/mac/gramps_launcher.py#L41) and [`Glib.get_user_config_dir`](https://docs.gtk.org/glib/func.get_user_config_dir.html) always returns a path so `if not os.path.exists(HOME_DIR):` is always `False`.

hgohel:
@Nick-Hall That's probably the best option right now. I would need to understand what's happening on MacOS and since I don't have a Mac, it would not be quick or easy.

jralls:
@hgohel It seems very likely to me that this patch would break some Windows users too. Unless `$XDG_USER_CONFIG` is set in the environment  `get_user_config_dir` on Windows will return `%USERPROFILE%\AppData\Local`, which is guaranteed to exist. The only situation where `if not os.path.exists(HOME_DIR):` would be true and the following code would execute is the *extremely* unlikely case that `$XDG_USER_CONFIG` is set to a non-existent path. Anybody who has `$APPDATA` pointed somewhere else is screwed, Windows or Mac.

Bugs [#13261](https://gramps-project.org/bugs/view.php?id=13261) and [#13300](https://gramps-project.org/bugs/view.php?id=13300) occur because the first condition in 
`if os.path.exists(HOME_DIR) or "GRAMPSHOME" in os.environ or "SAFEMODE" in os.environ:`
 is false only If `$USERDATA` is set and `$APPDATA` isn't or neither `$USERDATA` or `$HOME` are set. `$HOME` is *always* set on Unix, meaning that the whole XDG implementation applies *only to Windows users* whose environment wasn't completely set up. That on its own is pretty funny because the XDG Base Directory Specification applies only to Gnome Desktop and KDE.

@Nick-Hall Did you remove the `$USERDATA` and/or `$APPDATA` settings from the Windows AIO environment when you applied #1368?

Nick-Hall:
@jralls No changes were made to the Windows AIO environment.  I would expect `$USERDATA` and/or `$APPDATA` to exist.

Note that on Linux the `HOME_DIR` variable is set to `$HOME/.gramps`.  If this directory exists, then we use the old directory layout so that users are able see their existing family trees without having to move them.  I'm not sure if this was a good idea.  For a fresh install, this directory will not exist and the XDG directory structure will be used.

jralls:
@Nick-Hall 
> Note that on Linux the HOME_DIR variable is set to $HOME/.gramps.

There's no OS test, so the logic is:
* if `SAFEMODE` is set, `HOME_DIR` = `$SAFEMODE`
* else If `GRAMPSHOME` is set, `HOME_DIR` = `$GRAMPSHOME/gramps`
* else if `USERPROFILE` is set `HOME_DIR` = `Glib.get_user_config_data()/gramps` (with this commit)
* else if `USERPROFILE` is set, if `APPDATA` is also set `HOME_DIR` = `$APPDATA/gramps`; if not `HOME_DIR` = `$USERPROFILE/gramps`.
* else `HOME_DIR` is set to `get_env_var(HOME)/.gramps`; if `HOME` isn't set `get_env_var` will return `None` and `os.path.join` will raise a TypeError, crashing Gramps.

I just checked `$Env:USERPROFILE` and `$Env:APPDATA` on Windows having no environment customizations. `$Env:USERPROFILE` is set to `C:\Users\John Ralls` and `$Env:APPDATA` is set to `C:\Users\John Ralls\AppData\Roaming`. 

[I said](https://github.com/gramps-project/gramps/pull/1881#issuecomment-2715462288)
> if os.path.exists(HOME_DIR) or "GRAMPSHOME" in os.environ or "SAFEMODE" in os.environ:
is false only If $USERDATA is set and $APPDATA isn't or neither $USERDATA or $HOME are set.

But that's not right. `HOME_DIR` won't exist if this is the first time Gramps has run on the computer; it will exist (before this commit) if `APPDATA` isn't set and wasn't when the previous version ran.

So [#13261](https://gramps-project.org/bugs/view.php?id=13261) happened because Microsoft thinks AppData should be in `$USERPROFILE/AppData/Roaming` and Glib thinks it should be in `$USERPROFILE/AppData/Local` on Windows barring the unlikely case that a Windows user has set `XDG_USER_CONFIG` plus #1368's author not considering the possibility that someone would install 5.2+ and then later try an earlier version.

[#13300](https://gramps-project.org/bugs/view.php?id=13300) is because 
```
if os.path.exists(HOME_DIR) or "GRAMPSHOME" in os.environ or "SAFEMODE" in os.environ:
    USER_DATA = HOME_DIR
    USER_CONFIG = HOME_DIR
    USER_CACHE = HOME_DIR
```
Doesn't differentiate between `USER_DATA` (where grampsdb is supposed to be) and `USER_CONFIG` that is what *should* be pointed at the default configuration.

To fix that part, 
```
- ORIG_HOME_DIR = HOME_DIR
- if "SAFEMODE" in os.environ:
-    if "USERPROFILE" in os.environ:
-         USER_HOME = get_env_var("USERPROFILE")
-     else:
-         USER_HOME = get_env_var("HOME")
-     HOME_DIR = get_env_var("SAFEMODE")
-
-
- if os.path.exists(HOME_DIR) or "GRAMPSHOME" in os.environ or "SAFEMODE" in os.environ:
-    USER_DATA = HOME_DIR
-    USER_CONFIG = HOME_DIR
-    USER_CACHE = HOME_DIR
- else:
-    USER_DATA = os.path.join(GLib.get_user_data_dir(), "gramps")
-    USER_CONFIG = os.path.join(GLib.get_user_config_dir(), "gramps")
-    USER_CACHE = os.path.join(GLib.get_user_cache_dir(), "gramps")
+ if lin() and not os.path.exists(HOME_DIR) and not "GRAMPSHOME" in os.environ:
+   USER_DATA = os.path.join(GLib.get_user_data_dir(), "gramps")
+   USER_CONFIG = os.path.join(GLib.get_user_config_dir(), "gramps")
+ else if "GRAMPSHOME" in os.environ:
+    USER_DATA = HOME_DIR
+    USER_CONFIG = HOME_DIR
+ else:
+    USER_DATA = USER_HOME
+    USER_CONFIG = HOME_DIR
+ USER_CACHE = os.path.join(GLib.get_user_cache_dir(), "gramps")
+ if "SAFEMODE" in os.environ:
+    USER_CONFIG = get_env_var("SAFEMODE")
```

Fixing  [#13261](https://gramps-project.org/bugs/view.php?id=13261) is a little more intricate. We obviously need to look in both `Local` and `Roaming` for configurations. If we go back to Roaming for 6.0 and later then we need check only for `…\Local\gramps\gramps52` and `…\Local\gramps\gramps60` for previous configs that we would import or move respectively. Maybe we could ignore gramps60 as affecting only beta testers.

Nick-Hall:
> Maybe we could ignore gramps60 as affecting only beta testers.

Yes.  I think that we can ask beta testers to move their data.

Nick-Hall:
See PR #2013 for a very much simplified approach.

PR #1882, opened 2025-01-27
Refactor Relationship Calculator and Add a Searchbar

stevenyoungs:
Refactor the relationship calculator tool

- now based on `SelectPerson`
previously the calculator had it's own implementation of a person tree
- `BaseSelector` and `SelectPerson` classes updated to support modal or modeless operation. All existing use remains modal
- Relationship calculator now includes a search bar to aid finding the other person
- supports changes in active person after the tool is opened
- supports changes in families and recalculates the relationship
- handles database change (the current tool crashes)

Related discourse discussion, including user request: https://gramps.discourse.group/t/adding-search-bar-to-select-person-dialog/6845/10?u=stevey

image


stevenyoungs:
Steps to crash the existing relationship calculator tool
1. open the example database
2. open the relationship calculator tool
no need to click anything in the tool
3. close the database
the relationship calculator window is still displayed with all people listed
4. expand a surname in the relationship calculator window

An error similar to the below will be generated

```
36104: ERROR: grampsapp.py: line 188: Unhandled exception
  File "C:\Program Files\GrampsAIO64-5.2.3\gramps\gui\views\treemodels\treebasemodel.py", line 962, in do_get_value
    val = self._get_value(node.handle, col, node.secondary)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\GrampsAIO64-5.2.3\gramps\gui\views\treemodels\treebasemodel.py", line 979, in _get_value
    data = self.map(handle)
           ^^^^^^^^^^^^^^^^
  File "C:\Program Files\GrampsAIO64-5.2.3\gramps\gen\db\generic.py", line 1734, in get_raw_person_data
    return self._get_raw_data(PERSON_KEY, handle)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\GrampsAIO64-5.2.3\gramps\plugins\db\dbapi\dbapi.py", line 1002, in _get_raw_data
    self.dbapi.execute(sql, [handle])
  File "C:\Program Files\GrampsAIO64-5.2.3\gramps\plugins\db\dbapi\sqlite.py", line 136, in execute
    self.__cursor.execute(*args, **kwargs)
sqlite3.ProgrammingError: Cannot operate on a closed database.
```

In Windows AIO at least, there is a never ending stream of errors. The only escape is to kill the gramps process.

stevenyoungs:
rebased

dsblank (changes requested):
## Review: Relationship Calculator Refactor + SelectPerson Search Bar

**Overall:** The core idea is solid — replace the custom `RelCalc` dialog with `SelectPerson` as a base, giving the Relationship Calculator a proper searchable person list. The `_searchbar.py` rename (`search_*` → `filter_*`) is also a clean improvement. However, there are several bugs that need fixing before merge.

---

### Bugs

**1. `map()` is lazy in Python 3 — signals are never disconnected**

`baseselector.py`, `_disconnect_db_signals`:
```python
map(self.dbstate.db.disconnect, self.db_connections)  # does nothing!
self.db_connections.clear()
```
`map()` returns an iterator that is never consumed, so `disconnect()` is never called. This causes signal/memory leaks. Should be:
```python
for key in self.db_connections:
    self.dbstate.db.disconnect(key)
self.db_connections.clear()
```

**2. `self.dbstate.db = db` mutates the shared DbState**

`baseselector.py`, `_db_changed`:
```python
self.dbstate.db = db  # wrong — modifies the global DbState object
```
`dbstate` is a shared application-level object. `BaseSelector` should not be writing to `dbstate.db`. The intent appears to be updating a local reference. Combined with the removal of `self.db = dbstate.db` in `__init__`, subclasses that call `self.db` (e.g. `SelectPerson._connect_db_signals`, `RelCalc._connect_db_signals`) would get `AttributeError` in a standalone `SelectPerson` instance.

**3. `self.db` inconsistency in `_cleanup_on_exit`**

`baseselector.py` still has `self.db = None` in `_cleanup_on_exit`, but `self.db` is no longer initialized in `BaseSelector.__init__` (it was replaced by `self.dbstate`). This creates a confusing attribute that appears only at teardown. The line should be removed or `self.db` properly re-introduced.

**4. `RelCalc._connect_db_signals` doesn't call `super()`**

`relcalc.py`:
```python
def _connect_db_signals(self):
    self.db_connections.append(self.db.connect("family-add", ...))
    # missing: super()._connect_db_signals()
```
`SelectPerson._connect_db_signals` (which connects person-add/delete/update) is never called for `RelCalc`, so the person list won't react to person changes.

**5. `active_changed` parameter is immediately shadowed**

`relcalc.py`:
```python
def active_changed(self, handle) -> None:
    ...
    handle = self.uistate.get_active("Person")  # overwrites the parameter
```
The `handle` argument passed by the `active-changed` signal is silently discarded. If the signal passes the new handle, this should use it directly rather than querying `uistate` again. If querying `uistate` is intentional, the parameter should be `_handle` or `*_`.

**6. Dead code: `self.closebtn` in `relcalc.py`**

```python
self.closebtn = self.glade.get_object("button5")  # "button5" doesn't exist in baseselector.glade
okbutton.connect("clicked", self.close)
```
`button5` was in the deleted `relcalc.glade`, not in `baseselector.glade`. `self.closebtn` is assigned `None` and never used. Can be removed.

---

### Minor Issues

**7. `filter` shadows Python builtin**

Both `baseselector.py` and `selectperson.py` use `filter` as a parameter name, shadowing the built-in. Consider `filter_value`, `initial_filter`, or similar.

**8. License header regression**

All modified files change the modern SPDX URL form:
```
# along with this program; if not, see <https://www.gnu.org/licenses/>.
```
back to the old physical address form. The project has been migrating to the URL style; this PR goes in the opposite direction. Consider keeping the URL style on new/modified files.

---

### Positive aspects

- The core refactoring is elegant: reusing `SelectPerson` as the base for `RelCalc` eliminates ~100 lines of duplicated widget setup and gives the tool the full search/filter functionality for free.
- Deleting `relcalc.glade` and building the UI from `baseselector.glade` simplifies maintenance.
- Adding `_connect_db_signals` / `_disconnect_db_signals` hooks in `BaseSelector` is a good extensibility pattern.
- Type hints on `callback.py`, `selectperson.py`, etc. are welcome.
- Making `RelCalc` modeless (`set_modal(False)`) and updating live as the active person changes is a genuine UX improvement.
- The "only set `WIKI_HELP_PAGE`/`WIKI_HELP_SEC` if not already set" commit is a good fix for subclasses.

stevenyoungs:
> ### Minor Issues
> **7. `filter` shadows Python builtin**
> 
This problem is not introduced by this PR and has already been addressed in the target branch.
The problem was the the `__init__` method of `baseselector.py` and `selectperson.py` both of which has a parameter called `filter`. Now renamed to `search_or_filter`

> **8. License header regression**
> 
This problem is not introduced by this PR and has already been addressed in the target branch.

dsblank (approved):
Looks good; thanks!

stevenyoungs:
Rebased.
Callback changes expanded and moved to #2268.
Problems 1 to 6 above addressed
Corrected the call to `get_relationship_calculator` where `glocale` was incorrectly being passed in as the `reinit` argument. This problem occurs in two other instances within gramps. To be fixed in a separate PR. Update: see #2270 

The use and management of `self.db` and `self.dbstate.db` has been improved. Normally these two are equivalent. But during db change, the `active_changed` callback can be called whilst `self.db` is still referring to the previous db. This is due to the order in which `database-changed` callbacks are invoked. `active_changed` is called from one `database-changed` callback, before `BaseSelector._bd_changed` is called to update `self.db`. This design is quite tricky to get right.

PR #1885, opened 2025-01-28
Use ngettext in probably alive function

Nick-Hall:
Use ngettext in strings which include a substitution for the number of generations.

This was requested by a Weblate translator.

### DIFF
--- a/gramps/gen/utils/alive.py
+++ b/gramps/gen/utils/alive.py
@@ -48,6 +48,7 @@ from ..proxy.proxybase import ProxyDbBase
 LOG = logging.getLogger(".gen.utils.alive")
 
 _ = glocale.translation.sgettext
+ngettext = glocale.translation.ngettext
 
 DEBUGLEVEL = 4  # 4 = everything; 3 much detail; 2= minor detail; 1 = summary
 # -------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


CameronD73:
> @CameronD73 Does this look reasonable to you?

Yes, I had already fixed it in response to a separate comment in a separate branch, but I had not followed your preferred style so please accept this version.
There are a couple of actual errors in the translation code as well - should I raise those as issues here so others do not waste time?

Nick-Hall:
You can add extra translation related fixes to this PR if you like.

CameronD73:
> You can add extra translation related fixes to this PR if you like.

github is not making it easy - I can't edit yours directly, and it did not add your branch to my fork after I resync'ed until _after_ it added my attempted edit as a new branch.
In any case, I was including a change arising from reviewer comments that was unrelated to  translations - so it would be simpler for me to submit a separate PR.
I will also need to  ask on discourse for clarification on possessive.

CameronD73:
@Nick-Hall 
Is there a generally preferred syntax - '%' style or string.format method?

Nick-Hall:
In general, we prefer [str.format](https://docs.python.org/3/library/stdtypes.html#str.format) to the older [printf-style](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting)  formatting.  You could also use [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings), but I don't think that they work well with translations.

PR #1892, opened 2025-01-30
removed unreachable else statement

CameronD73:
there was a bit of useless code left in PR #1790.
The else test is redundant


dsblank (approved):
Thanks!

PR #1893, opened 2025-01-30
Probably alive pr1790 review fixes

CameronD73:
This mainly addresses comments by reviewers that were not attended to before the PR was merged.
1. remove reference to `passnum` as a parameter to spouse check and change to a boolean that selects between  immediate_family_only and a full ancestor/descendant scan
2. Remove bug relating to test for possibly translated "DIRECT" . Also made that string clearer for translators.
3. tried to improve several other strings for translatability
4. reset default debug level to not output so many cryptic messages if debug logging is enabled. Avoids running some code when logging not enabled.

### DIFF
--- a/gramps/gen/utils/alive.py
+++ b/gramps/gen/utils/alive.py
@@ -49,7 +49,9 @@ LOG = logging.getLogger(".gen.utils.alive")
 
 _ = glocale.translation.sgettext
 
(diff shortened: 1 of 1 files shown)


CameronD73:
looks like it did not get squashed after all.   try again!

PR #1894, opened 2025-01-30
Fixes to probably alive code following reviewer comments

CameronD73:
This mainly addresses comments by reviewers that were not attended to before the PR was merged.

1. remove reference to passnum as a parameter to spouse check and change to a boolean that selects between immediate_family_only and a full ancestor/descendant scan
2. Remove bug relating to test for "DIRECT" that would fail in other languages. Also made that string clearer for translators.
3. tried to improve several other strings for translatability
4. reset default debug level to not output so many cryptic messages if debug logging is enabled. Avoids running some code when logging not enabled.

### DIFF
--- a/gramps/gen/utils/alive.py
+++ b/gramps/gen/utils/alive.py
@@ -50,7 +50,9 @@ LOG = logging.getLogger(".gen.utils.alive")
 _ = glocale.translation.sgettext
 ngettext = glocale.translation.ngettext
 
-DEBUGLEVEL = 4  # 4 = everything; 3 much detail; 2= minor detail; 1 = summary
+# DEBUGLEVEL should be 1 for production, as higher levels are rather cryptic unless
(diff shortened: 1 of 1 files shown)


CameronD73:
Groan - I have just noticed that there is one situation where the new text for translation would be incorrect. I needed a bit of extra logic to make the statement always true, but my question is - what is the preferred way for me to proceed? - should I cancel this PR and just provide a new one with all the changes? or leave this in place and create a separare PR?

Nick-Hall:
@CameronD73 Just add more commits to the PR.  I can squash them when I merge.

Nick-Hall:
Let me know when this is ready for review and merge.

CameronD73:
> Let me know when this is ready for review and merge.

It looks like my new commit has appeared here. So it is ready again for review (I think I should stop looking.)

PR #1896, opened 2025-01-30
New indexes for big databases in the narrative web report

SNoiraud:
When we have large databases, the current indexes are too big
We have now 1 page for each letter.

We can limit the size of the page in rows per page in the html tab of the report.
The values are between 10 and 2000. The default is 500.
If we have more than this limit, we create a secondary index for the associated letter.


DaveSch-gramps:
I made the changes to my 5.2.4 basepage f8b6213 for event sorting.

Worked great!

Nick-Hall:
@DaveSch-gramps Thanks for testing.

PR #1897, opened 2025-01-30
Fix "alt_names" typo in the place model

Nick-Hall:
Fixed bug reported in #1872.

### DIFF
--- a/gramps/gui/views/treemodels/placemodel.py
+++ b/gramps/gui/views/treemodels/placemodel.py
@@ -139,7 +139,7 @@ class PlaceBaseModel:
     def search_name(self, data):
         """The search name includes all alt names to enable finding by alt name"""
         return ",".join(
-            [data["name"]["value"]] + [name["value"] for name in data["alt_name"]]
+            [data["name"]["value"]] + [name["value"] for name in data["alt_names"]]
         )
 
     def column_longitude(self, data):


gramps/gui/views/treemodels/placemodel.py
```diff
@@ -139,7 +139,7 @@ def column_name(self, data):
     def search_name(self, data):
         """The search name includes all alt names to enable finding by alt name"""
         return ",".join(
-            [data["name"]["value"]] + [name["value"] for name in data["alt_name"]]
+            [data["name"]["value"]] + [name["value"] for name in data["alt_names"]]
```
stevenyoungs, line 142:
```suggestion
            [data.name.value] + [name["value"] for name in data.alt_names]
```
Do we want to update code use attributes when it is being edited anyway?

Nick-Hall: All of the models seem to use dictionary access, but attributes would be nicer.  Is there any reason why we don't use them?

stevenyoungs: Personally, I'd like to see code upgraded to use attributes when it is being edited.
You were too quick with this PR, creating it and merging before I could get my comment submitted!

Nick-Hall: This PR was only a bug fix.  I made the smallest change possible.

PR #1899, opened 2025-01-31
Add type hints to latitude and longitude methods and functions

stevenyoungs:
In a recent PR I commented on a change to this line
[`if lat and lon:`](https://github.com/gramps-project/gramps/blob/ec2d5d83aef6fc7fd0fb709c48c6caf15a2d890a/gramps/gen/filters/rules/place/_withinarea.py#L119)
in the file [`gramps/gen/filters/rules/place/_withinarea.py`](https://github.com/gramps-project/gramps/blob/master/gramps/gen/filters/rules/place/_withinarea.py)

This PR adds types hints to `lat` and `lon` and methods called to initialise `latit` and `longit`

### DIFF
--- a/gramps/gen/filters/rules/place/_withinarea.py
+++ b/gramps/gen/filters/rules/place/_withinarea.py
@@ -107,25 +107,22 @@ class WithinArea(Rule):
             self.radius = self.radius / 2
 
     def apply(self, dummy_db, place):
-        if self.handle is None:
-            return False
-        if self.latitude is None:
-            return False
-        if self.longitude is None:
+        if not (place and self.handle and self.latitude and self.longitude):
             return False
         if place:
             lat = place.get_latitude()
             lon = place.get_longitude()
         if lat and lon:
             latit, longit = conv_lat_lon(lat, lon, "D.D8")
-            if latit is None or longit is None:
-                return False
-            if (
-                hypot(
-                    float(self.latitude) - float(latit),
-                    float(self.longitude) - float(longit),
+            return (
+                latit
+                and longit
+                and (
+                    hypot(
+                        float(self.latitude) - float(latit),
+                        float(self.longitude) - float(longit),
+                    )
+                    <= self.radius
                 )
-                <= self.radius
-            ):
-                return True
+            )
         return False
--- a/gramps/gen/utils/grampstranslation.py
+++ b/gramps/gen/utils/grampstranslation.py
@@ -29,6 +29,7 @@ for reports so that they can be created in a different language from the UI's.
 # python modules
 #
 # ------------------------------------------------------------------------
+from __future__ import annotations
 import gettext
 import collections
 import inspect
@@ -180,7 +181,7 @@ class GrampsTranslations(gettext.GNUTranslations):
         """
         return self.lang
 
-    def gettext(self, msgid, context=""):
+    def gettext(self, msgid: str, context: str = "") -> str:
         """
         Obtain translation of gettext, return a unicode object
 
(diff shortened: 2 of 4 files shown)


gramps/gen/utils/place.py
```diff
@@ -59,7 +60,7 @@
 West = West.replace("1", " ").strip()
 
 # build dictionary with translation en to local language
-translate_en_loc = {}
+translate_en_loc: dict[str, str] = {}
```
QuLogic, line 63: You need `from __future__ import annotations` here for `dict`, I believe.

gramps/gen/utils/place.py
```diff
@@ -27,6 +27,7 @@
 # Standard python modules
 #
 # -------------------------------------------------------------------------
+from typing import Optional
```
QuLogic, line 30: And these could be `float | None`, etc., with the future import.

gramps/gen/utils/place.py
```diff
@@ -46,11 +47,11 @@
...
-degrees = "1"
-North = _("%(north_latitude)s N") % {"north_latitude": degrees}
-South = _("%(south_latitude)s S") % {"south_latitude": degrees}
-East = _("%(east_longitude)s E") % {"east_longitude": degrees}
-West = _("%(west_longitude)s W") % {"west_longitude": degrees}
+degrees: str = "1"
+North: str = _("%(north_latitude)s N") % {"north_latitude": degrees}
+South: str = _("%(south_latitude)s S") % {"south_latitude": degrees}
+East: str = _("%(east_longitude)s E") % {"east_longitude": degrees}
+West: str = _("%(west_longitude)s W") % {"west_longitude": degrees}
```
QuLogic, line 54: This seems to be induced by the `_` function; would it possible to annotate that instead?

gramps/gen/filters/rules/place/_withinarea.py
```diff
@@ -107,25 +107,22 @@ def prepare(self, db, user):
...
-        if self.latitude is None:
-            return False
-        if self.longitude is None:
+        if not (place and self.handle and self.latitude and self.longitude):
             return False
         if place:
-            lat = place.get_latitude()
-            lon = place.get_longitude()
+            lat: str = place.get_latitude()
+            lon: str = place.get_longitude()
```
QuLogic, line 114: I'm not sure that you need these hints if `get_latitude` has the right return type hinted, which seems to have been done below here? What you probably need is to annotate the method arguments so it knows `place` is a `Place`.

gramps/gen/utils/place.py
stevenyoungs: PR updated. Thanks.

gramps/gen/utils/place.py
stevenyoungs: I've updated the PR, to use this preferred style. Thanks.

gramps/gen/utils/place.py
stevenyoungs: I've added a return type hint to gettext and removed these explicit types. Thanks

gramps/gen/filters/rules/place/_withinarea.py
stevenyoungs: Explicit hint removed in favour of inferred type. Thanks

gramps/gen/utils/grampstranslation.py
```diff
@@ -292,7 +293,7 @@ class GrampsNullTranslations(gettext.NullTranslations):
     neither will be the returned string.
     """
 
-    def gettext(self, msgid, context=""):
+    def gettext(self, msgid, context="") -> str:
```
cdhorn, line 296:
```suggestion
    def gettext(self, msgid: str, context: str = "") -> str:
```
Missed typing the parameters?

stevenyoungs: Thanks. I'd originally tried to types to methods called by gettext but had to step back. I went back a bit too far.

Nick-Hall:
Rebased and squashed commits.

PR #1900, opened 2025-01-31
Report options can memorise options for each database

SNoiraud:
Fixes [#2455](https://gramps-project.org/bugs/view.php?id=2455)

Apply what has been done for web narrative to all reports.

### DIFF
--- a/gramps/gen/plug/report/_options.py
+++ b/gramps/gen/plug/report/_options.py
@@ -948,7 +948,7 @@ class MenuReportOptions(MenuOptions, ReportOptions):
     """
 
     def __init__(self, name, dbase):
-        ReportOptions.__init__(self, name, dbase)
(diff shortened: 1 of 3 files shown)


Nick-Hall:
@SNoiraud I have reopened the feature request and assigned it to you.

There are some unit tests that are failing.

Nick-Hall:
First discussed in the Discourse thread [Thoughts on Narrated Web Report](https://gramps.discourse.group/t/thoughts-on-narrated-web-report/4739/15) in Jan 2024.

Nick-Hall:
This should be OK since we remove [special characters](https://github.com/gramps-project/gramps/blob/master/gramps/gui/dbman.py#L599) from the database name that are not allowed in filenames.

Nick-Hall:
Rebased.

PR #1910, opened 2025-02-02
Update readme to Python 3.9

stevenyoungs:
Based on PR #1859 we are using Python 3.10 on the runners, so update the readme accordingly

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Requirements
 ============
 The following packages **MUST** be installed in order for Gramps to work:
 
-* [**Python**](https://www.python.org/) 3.10 or greater - The programming language used by Gramps.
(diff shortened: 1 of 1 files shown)


Nick-Hall:
We support python 3.9 and above.

stevenyoungs:
> We support python 3.9 and above.

OK. PR updated.
However if the runners are using 3.10 isn't there a risk that support for 3.9 is accidentally broken?

Nick-Hall:
> However if the runners are using 3.10 isn't there a risk that support for 3.9 is accidentally broken?

Possibly, but problems are more likely to occur with new python releases.

Our current policy is to support all versions that have not reached end-of-life at the time of release.

Nick-Hall:
Closed in favour of #1912.

PR #1913, opened 2025-02-02
Optimise FilterProxyDb

stevenyoungs:
Optimise `FilterProxyDb` `get_<object>_from_gramps_id` methods

The first part of each method is to determine if the supplied `gramps_id` exists. For speed, instead of creating the full object, use the raw data.
Add some type hints and rename the `val` parameter to `gramps_id` for consistency

### DIFF
--- a/gramps/gen/proxy/filter.py
+++ b/gramps/gen/proxy/filter.py
@@ -316,9 +316,9 @@ class FilterProxyDb(ProxyDbBase):
         Finds a Person in the database from the passed Gramps ID.
         If no such Person exists, None is returned.
         """
-        person = self.db.get_person_from_gramps_id(val)
-        if person:
-            return self.get_person_from_handle(person.get_handle())
+        data = self.db.get_raw_person_from_id_data(val)
+        if data:
+            return self.get_person_from_handle(data.handle)
         else:
             return None
 
@@ -327,9 +327,9 @@ class FilterProxyDb(ProxyDbBase):
         Finds a Family in the database from the passed Gramps ID.
         If no such Family exists, None is returned.
         """
-        family = self.db.get_family_from_gramps_id(val)
-        if family:
-            return self.get_family_from_handle(family.get_handle())
+        data = self.db.get_raw_family_from_id_data(val)
+        if data:
+            return self.get_family_from_handle(data.handle)
         else:
             return None
 
@@ -338,9 +338,9 @@ class FilterProxyDb(ProxyDbBase):
         Finds an Event in the database from the passed Gramps ID.
         If no such Event exists, None is returned.
         """
-        event = self.db.get_event_from_gramps_id(val)
-        if event:
-            return self.get_event_from_handle(event.get_handle())
+        data = self.db.get_raw_event_from_id_data(val)
+        if data:
+            return self.get_event_from_handle(data.handle)
         else:
             return None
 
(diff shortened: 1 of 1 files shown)


gramps/gen/proxy/filter.py
```diff
@@ -311,102 +317,102 @@ def get_note_from_handle(self, handle):
         else:
             return None
 
-    def get_person_from_gramps_id(self, val):
+    def get_person_from_gramps_id(self, gramps_id: str) -> Person | None:
```
stevenyoungs, line 320: This is a different return type than the equivalent method in `DbGeneric` which always returns a `Person`
Is that going to be a problem? Should they be aligned?

stevenyoungs, line 320: `gramps_id: str`

Is it worth adding a new `GrampsId` type?

cdhorn: I should think so. PersonGrampsId, etc... and an AnyGrampsId to align with the added handle types.

stevenyoungs: I've added GrampsId types. @Nick-Hall would you prefer these to be in a separate PR?

Nick-Hall: Yes, please.  Put the type hints in a separate PR from the performance improvements.

Nick-Hall:
> The first part of each method is to determine if the supplied gramps_id exists. For speed, instead of creating the full object, use the raw data.

Have you considered using the `has_<object>_gramps_id` methods to check for the existence of a `gramps_id`?  It should be possible to optimise these for speed.

stevenyoungs:
> Have you considered using the `has_<object>_gramps_id` methods to check for the existence of a `gramps_id`? It should be possible to optimise these for speed.

I did. We need the handle of the object which has the supplied GrampsID so I think the current implementation will be quicker.
The alternative would, in the positive case, involve two lookups

```
if self.db.has_person_gramps_id(gramps_id): # first lookup
    data = self.db.get_raw_person_from_id_data(val) # second lookup
    return self.get_person_from_handle(data.handle)
return None
```

stevenyoungs:
> Given @dsblank comment about returning None and not checking [#1858 (comment)](https://github.com/gramps-project/gramps/pull/1858#issuecomment-2608114847), would it not be better to check for whether the call was going to return None

If it was going to return None, are you proposing that it raise an exception? That would be a change of behaviour.
Given that the current implementation explicitly checks if the supplied `gramps_id` exists, I was assuming that part of the API contract was that the `gramps_id` did not need to exist. If @Nick-Hall can comment, I've no objection per-se to amending the behaviour, but I'm concerned about the impact of such a change on external code.

Nick-Hall:
Filters can return `None`.  It just means that the object has been excluded by the filter.   This is not an error.

The problem in the past was that the base database layer also returned `None` in the case of broken links. As a result, real errors were missed.  This has long been fixed and the database now raises a `HandleError` in such circumstances.

stevenyoungs:
Review comments implemented and rebased onto master.

gramps/gen/proxy/filter.py
```diff
@@ -24,6 +24,12 @@
 """
 Proxy class for the Gramps databases. Apply filter
 """
+# -------------------------------------------------------------------------
+#
+# Standard python modules
+#
+# -------------------------------------------------------------------------
+from __future__ import annotations
```
dsblank, line 32: I might need to add this to all of the rules that I added typing hints too. Do we support a version of Python that needs this line?

dsblank:
My comment about the possibility of these functions returning `None` was really a question about whether typing could check to see if the code checked for `None` before using it, if we didn't check, raise a typing error.

gramps/gen/proxy/filter.py
Nick-Hall: We support python 3.9 and above.  If you are using a feature such as the vertical bar for unions, you will need the `__future__` import.

dsblank: Yes, there was one: #1918

PR #1916, opened 2025-02-03
Define types for Gramps IDs

stevenyoungs:
Define types for the different primary object Gramps Ids
Use these to add type hints to `DbGeneric`

### DIFF
--- a/gramps/gen/types.py
+++ b/gramps/gen/types.py
@@ -2,6 +2,7 @@
 # Gramps - a GTK+/GNOME based genealogy program
 #
 # Copyright (C) 2025       David Straub
+# Copyright (C) 2025       Steve Youngs
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


stevenyoungs:
This work originated in PR #1913

gramps/gen/types.py
```diff
@@ -1,3 +1,23 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2025       David Straub
```
stevenyoungs, line 4: @DavidMStraub I added the standard license and your copyright to `types.py` Please can you check and confirm?

dstraub-pf: Thank you, confirmed

dstraub-pf:
Looks good to me! :+1:

DavidMStraub:
Looks good to me! 👍

gramps/gen/types.py
DavidMStraub: Thank you, confirmed

Nick-Hall:
As a general comment, identifier is usually abbreviated as "ID" rather than "Id".  So it would be "Gramps IDs" rather than "Gramps Ids" for example.

DavidMStraub:
> As a general comment, identifier is usually abbreviated as "ID" rather than "Id". So it would be "Gramps IDs" rather than "Gramps Ids" for example.

CamelCase and acronyms are a topic of eternal debate... I personally like the non-all-caps acronyms much better because I find them more readable. We also have `DbReadBase` etc., even though database is usually abbreviated as DB. (Then again, we have `DBAPI` rather than `DbApi`...).

Nick-Hall:
Our [coding style](https://gramps-project.org/wiki/index.php/Programming_guidelines#Coding_style) states that we should write PEP8 compatible code. The PEP8 section on [descriptive naming styles](https://peps.python.org/pep-0008/#descriptive-naming-styles) has the following note in the "CapitalizedWords" section:

>Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

You will, however, find old Gramps code that doesn't meet these standards.

stevenyoungs:
> As a general comment, identifier is usually abbreviated as "ID" rather than "Id". So it would be "Gramps IDs" rather than "Gramps Ids" for example.

Thanks. PR updated. Now uses GrampsID.

Nick-Hall:
Rebased and squashed a commit.

cdhorn:
> Our [coding style](https://gramps-project.org/wiki/index.php/Programming_guidelines#Coding_style) states that we should write PEP8 compatible code. The PEP8 section on [descriptive naming styles](https://peps.python.org/pep-0008/#descriptive-naming-styles) has the following note in the "CapitalizedWords" section:
> 
> > Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.
> 
> You will, however, find old Gramps code that doesn't meet these standards.

Note that Id is an abbreviation for identifier, and not an actual acronym. If someone abbreviated Error as Err in a variable name you would not expect them to capitalize it.

https://dictionary.cambridge.org/us/dictionary/english/acronym

PR #1917, opened 2025-02-03
Make the orjson package mandatory

Nick-Hall:
@dsblank Is there a minimum version that we require for orjson?


dsblank:
I don't see anything that sticks out as an obvious release to require: https://github.com/ijl/orjson/releases

PR #1919, opened 2025-02-03
WIP: Add type hints to db classes

stevenyoungs:
Add type hints to the various db classes.


stevenyoungs:
I've created this PR whilst it is still a work in progress to get some input on the impact.
`DbReadBase.get_person_from_handle` has been annotated to return `Person | None`
Why `None`, because `ProxyDbBase.get_person_from_handle` can return `None`.
If `None` is not included, mypy gives the error
```gramps/gen/proxy/proxybase.py:516: error: Return type "Person | None" of "get_person_from_handle" incompatible with return type "Person" in supertype "DbReadBase"  [override]```

Having done that, code such as this
https://github.com/gramps-project/gramps/blob/1280aa45a5f47581badfb655021ce1d430f8a581/gramps/gen/db/base.py#L1959-L1962
now gets (correctly) flagged as an error because it just uses person even if it is `None`

```gramps/gen/db/base.py:2023: error: Item "None" of "Person | None" has no attribute "remove_family_handle"  [union-attr]```

The solution I've taken so far in this PR is to amend the code to handle a return of `None` correctly. In this situation to protect the code with `if person and family:`

Before I go any further, is that the correct approach or is there a better way?

**Comment by kulath** [comment]
(Note, I haven't tried to understand the logic in this particular case, so my comment is just on general principles).

If the code doesn't currently try to deal with a return of `None`, then presumably the only way `None` can occur is through a database corruption (or something of that sort). If corruption has occurred, the it should be flagged as soon as possible. I think this is the point responded to in https://github.com/gramps-project/gramps/pull/1913#issuecomment-2632073505, where AIUI the database does a check and raises `HandleError` , if it doesn't then the earliest point to raise the error would be in the database, which should be done.

I don't think the code should be guarded with `if person and family`, because that isn't really the correct logic if `None` should never occur, (and could conceivably cause problems further downstream).

The problem seems (to me) to be that a Proxy could correctly return `None`, but outside a proxy thay could not happen (except for a database fault of a broken link). So would the correct thing be to have a different method for the call in a proxy (or an additional parameter) to say that `None` was a legitimate return, and then MyPy could do the correct check. - I have no idea whether this would solve the problem or whether it is possible given that the problem is with inheritance, or whether this would involve too much duplication of code!

**Comment by stevenyoungs** [comment]
> Another approach would be to remove the handle from the reference object. It might be possible to pre-process the entire DB, caching the proper references and removing anything pointing to a private object, and replacing sensitive values, like "Living". This would probably be faster and less error prone in the end.

I was trying to restrict this PR to just adding type hints. I've not quite succeeded but I'm reluctant to make larger scale code changes if I can avoid it.

**Comment by stevenyoungs** [comment]
> The problem seems (to me) to be that a Proxy could correctly return `None`, but outside a proxy thay could not happen (except for a database fault of a broken link).

As I understand it a proxy db is substitutable for the real db. So all higher level (business logic if you like) code should be written to correctly handle a None return. If we add an alternative set of methods, which can return None, won't we still have to support a None return type in much of the code?

File: gramps/gen/proxy/proxybase.py [hunk 1]
```diff
@@ -308,83 +348,88 @@ def get_tag_handles(self, sort_handles=False, locale=glocale):
...
         return self.db.get_default_handle()
 
-    def iter_person_handles(self):
+    def iter_person_handles(self) -> Generator[PersonHandle]:
         """
         Return an iterator over database handles, one handle for each Person in
         the database.
         """
-        return filter(self.include_person, self.db.iter_person_handles())
+        return (_ for _ in filter(self.include_person, self.db.iter_person_handles()))
```
stevenyoungs, line 364:
```
        return (_ for _ in filter(self.include_person, self.db.iter_person_handles()))
```
Flagging this change for comment.
Method: `def iter_person_handles(self) -> Generator[PersonHandle]`

This code previously returned a List. The equivalent method in `DbGeneric` returns a `Generator[PersonHandle]`.
I've standardised on returning a generator. Is this the right choice?

gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -78,10 +105,12 @@ class DBAPI(DbGeneric):
     Database backends class for DB-API 2.0 databases
     """
 
+    dbapi: Any  # might be better to have a specific ConnectionBase class
```
stevenyoungs, line 108:
```
    dbapi: Any  # might be better to have a specific ConnectionBase class
```
Flagging this line for comment.
It's not correct to type hint `dbapi` as `Any` but I could not find an appropriate superclass to use.

stevenyoungs:
Cursor functions such as `get_person_cursor` don't have a consistent return type so I'm unsure how to correctly type.
Suggestions welcome

| Class | Return Type |
|-----|-----|
|`DbReadBase`||
|`DummyDb`|`[]`|
|`DbGeneric`|`Cursor`|
|`ProxyDbBase`|`ProxyCursor`|

kulath:
> Having done that, code such as this
> 
> https://github.com/gramps-project/gramps/blob/1280aa45a5f47581badfb655021ce1d430f8a581/gramps/gen/db/base.py#L1959-L1962
> 
> 
> now gets (correctly) flagged as an error because it just uses person even if it is `None`
> `gramps/gen/db/base.py:2023: error: Item "None" of "Person | None" has no attribute "remove_family_handle" [union-attr]`
> 
> The solution I've taken so far in this PR is to amend the code to handle a return of `None` correctly. In this situation to protect the code with `if person and family:`
> 
> Before I go any further, is that the correct approach or is there a better way?

I don't see how protecting the code locally with `if family` etc and then 'carrying on' can be the right thing. Quite apart from the fact that it is a change to the logic (which you are reasonably unwilling to do), you just don't know what the rest of business logic will be doing and whether it will be consistent later on.

When Nick wrote that
> The problem in the past was that the base database layer also returned None in the case of broken links. As a result, real errors were missed. This has long been fixed and the database now raises a HandleError in such circumstances.

I presumed that he meant that anywhere that `None` was returned was already guarded by testing for `None` and doing the right thing. When you said 

> As I understand it a proxy db is substitutable for the real db. So all higher level (business logic if you like) code should be written to correctly handle a `None` return. If we add an alternative set of methods, which can return None, won't we still have to support a `None` return type in much of the code?

From what Nick had said, I imagined that the code is effectively divided into code that can be called with a proxy db, and code that cannot be called with a proxy db. And I imagined that all the code that could be called from a proxy db already dealt with the possibility that `None` might be returned. Hence I would assume that the code you quote simply cannot be called with a proxy db, and hence it can safely assume that `None` is not possible.

So, I think that either you should have a statement like
`if family is None: raise <some exception>` or
`assert family is not None`

dsblank:
> I imagined that the code is effectively divided into code that can be called with a proxy db, and code that cannot be called with a proxy db

This is incorrect. Much code can be called with a limiting proxy (Private, Living) in place. This most likely happens in reports, filters, some utils (like alive.py), exports, and tools.

gramps/gen/proxy/proxybase.py
dsblank: Sounds good to me.

dsblank:
@stevenyoungs I think your approach is sound. We can work on proxies and their ability to return None in another initiative.

gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -1171,13 +1214,13 @@ def _update_secondary_values(self, obj):
...
+            given_name, surname = self._get_person_data(cast(Person, obj))
             sets.append("given_name = ?")
             values.append(given_name)
             sets.append("surname = ?")
             values.append(surname)
         if table == "Place":
-            handle = self._get_place_data(obj)
+            handle = self._get_place_data(cast(Place, obj))
             sets.append("enclosed_by = ?")
             values.append(handle)
```
kulath, line 1226: Wouldn't it be better if this were written as
`if type(obj) is Person`
etc.
(If I am right that this works the same)

I appreciate that this would be changing existing code, but surely no more than adding `cast`. The existing code is not very clear - accessing the type through `table = obj.__class__.__name__` is complex and obscure, the conversion to name is not really necessary, and table is not a good name for the variable.

Testing the type in the if statement would (AIUI) automatically provide the static type checking, and if anyone wanted to add more code in the future, that wouldn't need to be type confirmed by something like a cast.

kulath:
> @stevenyoungs I think your approach is sound. We can work on proxies and their ability to return None in another initiative.

Well, if you are going to work on the ability of proxies to return `None` at another time, don't do the fix of adding the if statement now, do something like an assert for now, mark it with a fixme, and then the consequences of testing for `None` can be properly evaluated at some later time. Adding the if statement now makes it look as though all the consequences (on the calling routines) have been evaluated now. (I'm thinking about other places where something similar occurs, not just this exact piece of code, and the need for a common approach).

dsblank:
As @stevenyoungs mentioned, he is only adding type hints in this PR.

If someone reports a crash because some general code is failing because it isn't properly handling proxies that return None, the immediate fix is to add code that checks the return value. A complete rethink of proxies is going to take some time and energy.

gramps/gen/db/base.py
```diff
@@ -1951,41 +2025,45 @@ def remove_parent_from_family(self, person_handle, family_handle, trans=None):
...
+    def __remove_parent_from_family(
+        self, person_handle: PersonHandle, family_handle: FamilyHandle, trans
+    ) -> str:
         """
         Remove a person as either the father or mother of a family,
         deleting the family if it becomes empty; trans is compulsory.
         """
         person = self.get_person_from_handle(person_handle)
         family = self.get_family_from_handle(family_handle)
+        if person and family:
```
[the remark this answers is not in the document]
stevenyoungs: Ok. The code changes will need to be applied first. The type hints on their own will not pass mypy; the code changes are the minimal set strictly required for mypy to pass.

kulath:
> If someone reports a crash because some general code is failing because it isn't properly handling proxies that return None, the immediate fix is to add code that checks the return value.

Maybe that 'immediate fix' is right, and maybe it isn't.

I'm sorry to point out something that I am sure you already know! I'm sure you don't mean **only just** blindly patch the code where the crash manifests itself.

Not handling proxies that return `None` is a bug, and you don't fix bugs by patching the symptom. You reproduce the fault, and then test whether the patch you propose fixes everything. You need to make sure that not only the place where the crash first occurs is OK, but also any other calling or called code also behaves correctly.

That's why I don't think you should just add an if statement to skip part of the code. That may well turn out to be the necessary and sufficient solution to the problem, but it will probably take some time and energy (and perhaps another initiative)  to determine it.

I'm really excited about the static typing and especially how it is forcing clarification of the code, as mentioned by several of the comments in this PR. However, **_if the code was going to crash before, then it should still crash after the type checking has been added_**. Hence my suggestion of `if family is None: raise <some exception>` or `assert family is not None` (and add a fixme) to still crash, but to satisfy the static type checking.

stevenyoungs:
> Hence my suggestion of `if family is None: raise <some exception>` or `assert family is not None` (and add a fixme) to still crash, but to satisfy the static type checking.

The problem is that the type hinted code does **not** pass type checking _unless_ you modify the code to handle a None return.

Nick-Hall:
I think that allowing `None` as a return value from `get_person_from_handle` is a mistake.  The type checking is highlighting the fact that we have a design issue, which we already know about,  but the solution is to prevent proxies from returning `None`, not fixing code that isn't broken.

A better approach would be to make the proxies return empty objects instead.  Then a check like `if person is None:` would become `if person.is_empty():`.  If the check was omitted by mistake then we would just see an object with all details redacted.

This doesn't actually fix the problem with proxies leaking information, but this could be dealt with at a later date.

stevenyoungs:
> `if person.is_empty():`

Is there a heuristic that can be used to check if a `PrimaryObject` is empty?
no `handle` and no `gramps_id` perhaps?

Nick-Hall:
An empty object would have its handle set to `None`.  An object from the database would always have a handle.

`if person.handle is None:` would work equally as well as `if person.is_empty():` but creating an `is_empty` methid might be neater and easier to understand.

kulath:
> I think that allowing `None` as a return value from `get_person_from_handle` is a mistake. The type checking is highlighting the fact that we have a design issue, which we already know about, 

Agreed!

>but the solution is to prevent proxies from returning `None`, not fixing code that isn't broken.
> 
> A better approach would be to make the proxies return empty objects instead. Then a check like `if person is None:` would become `if person.is_empty():`. If the check was omitted by mistake then we would just see an object with all details redacted.
> 
> This doesn't actually fix the problem with proxies leaking information, but that could be dealt with at a later date.

Doesn't this just move the crash to a different place? If you return an empty object, then isn't it possible that the next piece of code will try to use the non-existent handle and consequently crash? I suppose the question we might ask is "does this handle point to an accessible object?" **_before_** trying to call `get_person_from_handle`.

> The problem is that the type hinted code does not pass type checking unless you modify the code to handle a None return.

If you modify the code to add an assert, will this pass type checking?

My theory is that such a modification would not alter the behaviour of the code (unless debug was switched on), so is the minimal change without extensive modification to fix the underlying problem.

dsblank:
@Nick-Hall said:

> A better approach would be to make the proxies return empty objects instead.

Maybe ([I mentioned above](https://github.com/gramps-project/gramps/pull/1919#issuecomment-2633790420)). An even better approach would be that we would never have a reference handle that didn't point to a an inaccessible object.

@kulath said:

> I suppose the question we might ask is "does this handle point to an accessible object?" before trying to call get_person_from_handle.

I don't think we want to surround every `db.get_object(handle)` with code like that any more than we want to check for the results being `None`.

BUT, we could rethink the proxies to remove all handles to inaccessible objects. Let's consider that before changing every get method.

stevenyoungs:
> If you modify the code to add an assert, will this pass type checking?
> 
> My theory is that such a modification would not alter the behaviour of the code (unless debug was switched on), so is the minimal change without extensive modification to fix the underlying problem.

If I modify the code to return `Person()` instead of `None`, then higher level code does not need modifying to handle `None`. It will always get a `Person` object returned. I believe that can then drop the first commit of this PR - the commit that adds checks for `None` - and still keep mypy happy. We'll see shortly...
At some point, we'll reach a level where this empty `Person` needs to be handled correctly with `if person.is_empty()`

dsblank:
> At some point, we'll reach a level where this empty Person needs to be handled correctly with if person.is_empty()

This approach seems like just kicking the can down the line. Let's explore alternatives.

Nick-Hall:
@dsblank said:
> An even better approach would be that we would never have a reference handle that pointed to a an inaccessible object.

I agree.

stevenyoungs:
Please disregard that I now have code changes in this PR - I'll sort that out when the approach is agreed.

This is a proof of concept of `get_person_from_handle` returning Person() rather than `None`. Methods for other PrimaryObjects can be updated in the same way.

mypy is happy with the code.

Let me have your thoughts....

gramps/gen/db/base.py
```diff
@@ -580,7 +580,7 @@ def get_note_from_handle(self, handle: NoteHandle) -> Note | None:
         """
         raise NotImplementedError
 
-    def get_person_from_handle(self, handle: PersonHandle) -> Person | None:
+    def get_person_from_handle(self, handle: PersonHandle) -> Person:
```
Nick-Hall, line 583: Yes!  This is the change I wanted to see.

stevenyoungs:
This is the raise an error variant. It feels better to me.
Once again, only get_person_from_handle is updated.

I can now rely on `self.db.get_person_from_handle` raising an error, allowing some `if person is None:` code to be removed ❤️ 

https://github.com/gramps-project/gramps/blob/ef5512c4347a3303c31608a3382867cd4994a37d/gramps/gen/proxy/private.py#L87-L89
will now raise an error if passed the handle of a private oject.

gramps/gen/errors.py
```diff
@@ -138,6 +138,18 @@ def __str__(self):
         return self.value
 
 
+class NoneObjectError(HandleError):
```
Nick-Hall, line 141: Would `AccessDeniedError` be more descriptive?

stevenyoungs: Yes, much better. Thanks!

gramps/gen/proxy/proxybase.py
```diff
@@ -163,44 +189,69 @@ def include_something(self, handle, obj=None):
...
+    def include_media(self, handle: MediaHandle) -> bool:
+        raise NotImplementedError
 
-    include_person = include_family = include_event = include_source = (
-        include_citation
-    ) = include_place = include_media = include_repository = include_note = (
-        include_tag
-    ) = None
+    def include_repository(self, handle: RepositoryHandle) -> bool:
+        raise NotImplementedError
```
stevenyoungs, line 214: I was not 100% sure if this change is correct. Any python experts care to comment?
I need to give the predicates the correct type

stevenyoungs: answering my own question, it's not. It alters behaviour in the `gfilter` method which has this line:
`if predicate is not None and obj is not None:`

dsblank:
I'm diving into the proxies this weekend to see what can be done to ensure that handles that point to inaccessible objects are removed from the originating source (like, person.family_list, bookmarks, etc).

dsblank:
I'm already learning things that I never new (or had forgotten): `Proxy._iter_raw_person_data()` does not protect data. There is at one filter that breaks the privacy in the proxy through this loophole.

stevenyoungs:
I think the fact I've got the unit tests to pass with get_person_from_handle never returning None is encouraging (and slightly surprising).
I'll work on bringing the other objects up to the same standard.
What's clear is that if we want to type the db methods, we have to solve this problem

dsblank:
> I think the fact I've got the unit tests to pass with get_person_from_handle never returning None is encouraging (and slightly surprising).

Before last week, there were zero tests for proxies. I added a few here: gramps/gen/proxy/test/proxies_test.py

But they all properly test for None (and in fact test exactly for that). 

The real test would be to do all of the test functions, but with combinations of proxies. Guaranteed fails I'm sure.

kulath:
YES! Changing `get_person_from_handle` to raise an error is so much a better choice than returning `None` or an empty object. It pushes the fault back earlier, which is good and it means you don't have the problem 

> But I also don't think returning an empty object is good. Would it have a handle? Probably, because we used get_from_handle. Would it have a gramps ID? Certainly not. Now we need code to handle primary objects which, unexpectedly, don't have Gramps IDs etc.

Changing the proxies so handles pointing to inaccessible objects are removed is even better.

The only thing that needs to be watched is that any code in the user database check routines that checks for corruption where handles point to nothing will have to work. At the moment this relies on `HandleError` being returned. Perhaps that would be a better exception to return.

gramps/gen/lib/tableobj.py
```diff
@@ -142,6 +143,15 @@ def get_handle(self):
...
         return self.handle
 
+    def is_empty(self) -> bool:
+        """
+        Return true if the object is not associated with a database handle.
+
+        :returns: true if there is no database handle associated with the object, false otherwise
+        :rtype: bool
+        """
+        return self.handle
```
dsblank, line 153: If the return type is bool, should the return be:

```suggestion
        return self.handle is None
```

stevenyoungs: I believe there's an implicit conversion to `bool`

I'll solve anyway by deleting this method. It's no longer required with the raise error approach.

stevenyoungs:
The `get_<object>_from_handle` methods now always return an object or raise an error
`None` is never returned ❤️ 

I am able to build a narrated website with living people excluded.

`ProxyDbBase.gfilter_none` could do with a better name. Suggestions welcome.

Now that `get_<object>_from_handle` never returns `None`, there is plenty of code of this form which can be tidied up, by removing the `if`. I do not want to clutter this PR with such changes so let's do that in a subsequent PR

```
person = self.db.get_person_from_handle(handle)
if person:
```

dsblank:
Good news! Proxies should also never return None from a `db.get_OBJECT(handle)`. (Currently the sanitize methods are using a return value of None as a check, but we can do that in a different way if we want without calling the method.)

In any event, the cases in the past that we were getting `None` was just a bug in the proxies. WIP: #1941

dsblank:
@stevenyoungs, how should we develop this PR and #1941 ? We need the bug fixes so we don't have to have None return values. I could make #1941 be a PR on yours, or vice versa.

stevenyoungs:
> Good news! Proxies should also never return None from a `db.get_OBJECT(handle)`. (Currently the sanitize methods are using a return value of None from the basedb as a check, but we can do that in a different way if we want without calling the method.)

It sounds good, but I'm not sure I follow which methods (in `ProxyDbBase`?) should never return None?
The only ones type hinted to return None are `get_<object>_from_gramps_id`

stevenyoungs:
> @stevenyoungs, how should we develop this PR and #1941 ? 

Ultimately, I want to separate out code changes from type hint changes. This will make it easier to examine in detail the code changes without the noise of type hints.

#1934 is for code changes with no type hints
#1919 [this PR] will require #1934 and then add type hints.

currently #1919 also contains some code changes. I'm keeping these earlier in history and the commit message prefixed with "CODE". I'll move them into #1934 in due course.

Your #1941 looks like it also a pre-requisite of #1919

stevenyoungs:
> Ah, yes, looking up something by gramps_id could be None. (We could change the name of those someday to be something like db.lookup_person_by_gramps_id(gramps_id) as an indicator that they can return None. But I guess type hints will catch that).

❤️ I still think there is value in renaming. I think your proposal get close and I would like to resurrect it.

> I don't think the get_raw data should return None.
> 
> ```python
> def get_raw_person_data(self, handle: PersonHandle) -> Person | None:
> ```
> 
> Why would you add that?

I don't think a `None` return is possible any more. That's left over from an earlier state. Let me check and adjust.

dsblank:
> I don't think a None return is possible any more. That's left over from an earlier state. Let me check and adjust.

If it does return None, then we have a bug in the proxies.

stevenyoungs:
> ```python
> def get_raw_person_data(self, handle: PersonHandle) -> Person | None:
> ```

It doesn't return a `Person` either 😞 So apart from `Person` and `None` being wrong, the change was perfect 😉 
We might want to consider a typed `DataDict`

stevenyoungs:
https://github.com/gramps-project/gramps/blob/ef5512c4347a3303c31608a3382867cd4994a37d/gramps/plugins/db/dbapi/dbapi.py#L1023-L1032

currently `get_raw_person_data` can return `None`. I'll adjust it to raise an error

dsblank:
In the private.py proxy, code like this: 

```python
    def get_repository_from_handle(self, handle):
        repository = self.db.get_repository_from_handle(handle)
        if repository and not repository.get_privacy():
            return sanitize_repository(self.db, repository)
        return None
```

should be changed to:

```python
    def get_repository_from_handle(self, handle):
        repository = self.db.get_repository_from_handle(handle)
        return sanitize_repository(self.db, repository)
```

If internally (say in a report) that causes an issue, then there is a bug. Top-level calls where you are getting a handle from someplace else, then it needs to be checked before calling. Where should that fix go? I don't want to merge-conflict with your typing changes.

(I see that bookmarks don't sanitize their handles, so I'll fix that.)

stevenyoungs:
> should be changed to:
> 
> ```python
>     def get_repository_from_handle(self, handle):
>         repository = self.db.get_repository_from_handle(handle)
>         return sanitize_repository(self.db, repository)
> ```

The current implementation is 

```
    def get_repository_from_handle(self, handle):
        repository = self.db.get_repository_from_handle(handle)
        if not repository.get_privacy():
            return sanitize_repository(self.db, repository)
        raise AccessDeniedError(f"Handle {handle} is private")
 ```

For a transitionary period at least, I think it would be beneficial to get an `AccessDeniedError` raised, with an hint as to why the error was raised. When the dust has settled, then I agree with you. There will be a lot of `if`'s that we can remove throughout the code base.

I'll adjust the implementation to make the future step slightly easier by switching the last two lines:

```
    def get_repository_from_handle(self, handle):
        repository = self.db.get_repository_from_handle(handle)
        if not repository.get_privacy():
            raise AccessDeniedError(f"Handle {handle} is private")
        return sanitize_repository(self.db, repository)
 ```

gramps/plugins/lib/libmixin.py
```diff
@@ -77,8 +83,8 @@ def __find_primary_from_handle(
         obj = class_type()
         handle = str(handle)
         new = True
-        raw = get_raw_obj_data(handle)
-        if raw is not None:
+        if has_obj_handle(handle):
+            raw = get_raw_obj_data(handle)
             obj = data_to_object(raw)
             # references create object with id None before object is really made
             if obj.gramps_id is not None:
```
kulath, line 90: (1) As you are doing for the object_handle the test before getting the obj, could you not collapse get_raw_obj_data and data_to_obj to a single call to get the object form the database?

(2) as you now know that the object exists in the database, is it still possible for gramps_id to be None?

If you don't want to make these changes as you are only trying to make type checking better, can I suggest marking the code with FIXME, so that the optimisations can be investigated in the future?

stevenyoungs: Thanks. I made a similar comment on #1934, where the code changes are being developed. See these two comments
https://github.com/gramps-project/gramps/pull/1934#discussion_r1947916043
https://github.com/gramps-project/gramps/pull/1934#discussion_r1947916066

If this PR gets merged, then I think we need a follow on task to review all calls to `get_<object>_from_handle` and `get_raw<obj>_data` to look for code like this, and checks for a `None` return value (that will not longer be possible)

stevenyoungs:
I think this is ready for review.

This PR is to review the final commit only "**Add type hints to db methods**". It only adds type hints.
Earlier commits, which make code changes, should be reviewed in #1934

stevenyoungs:
The following PRs are pre-requisites: #1934, #1945, #1948, #1949

stevenyoungs:
rebased.
mypy is detecting errors. Fixes for some of these are already merged into the `gramps60` branch.

stevenyoungs:
Also requires PR #2010

gramps/gen/proxy/proxybase.py
bartfeenstra: `collections.Iterable` is the common denominator between `list` and `Generator`.

stevenyoungs:
Converted to draft. I think #2232 is a better starting point but there are still some additional (and useful changes) in this PR which I'll work to separate out

stevenyoungs:
Moved comment fixes into #2234

dsblank:
@stevenyoungs do you want to close this PR? You can still get code out of it even if it is closed.

@Nick-Hall If we can get #2195 merged then we can build proper types all across the code base.

stevenyoungs:
> @stevenyoungs do you want to close this PR? You can still get code out of it even if it is closed.
> 
> @Nick-Hall If we can get #2195 merged then we can build proper types all across the code base.

I intend to use #2232 as the PR to move forwards.
I'll pick through #1919 and #1934 and move whatever makes sense into #2232. The will result in #1919 and #1934 becoming smaller or, most likely, closed.

stevenyoungs:
Relevant changes moved over to #2232

PR #1920, opened 2025-02-04
Fix accessing an attribute in a DataDict

dsblank:
This PR fixes a bug in accessing an attr in a DataDict.

Previously, `person_data.handle` (where person_data is a DataDict) would return the attribute from the "_object" if it existed. This is fine for any non-data values (like strings). But for other items (like a Name), it would return the Object (Name) rather than a DataDict of the Name. This might be not noticed if you continued to access attr, but Object["other_attr"] would crash.

Now, it properly returns a DataDict for all attributes, except those that only exist on the object (such as methods).


dsblank:
We could also make `DataDict["attr"]` an exception to catch any leftover uses and make sure we don't use them. `DataDict` might not be the right name though for a dict that you can't getitem :)

PR #1925, opened 2025-02-05
Fix all_people() simple access method to use new data format

Nick-Hall:
Fixes [#13603](https://gramps-project.org/bugs/view.php?id=13603).

### DIFF
--- a/gramps/gen/simple/_simpleaccess.py
+++ b/gramps/gen/simple/_simpleaccess.py
@@ -812,9 +812,16 @@ class SimpleAccess:
         """
 
         with self.dbase.get_person_cursor() as cursor:
-            # data[3] is primary_name; data[3][5][0][0] is surname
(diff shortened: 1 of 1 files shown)


dsblank:
Nice catch!

dsblank (approved):
Looks good.

dsblank:
Do you want to add all of these bug fixes to 6.0 milestone?

dsblank:
```
git grep -E "data\[[0-9]+\]"
```

Gives 161 matches in 31 files. I'll take a look at these to see if there are any more.

dsblank:
Oh, but that grep doesn't catch what @Nick-Hall is fixing. It should be:

```
git grep -E "data\[[^]+]\]"
```

dsblank:
I didn't find any others beside the simple access methods.

PR #1927, opened 2025-02-05
Fix error when searching by name in the place selector

Nick-Hall:
Fixes [#13605](https://gramps-project.org/bugs/view.php?id=13605).


dsblank (approved):
Looks good! Thank you!

PR #1928, opened 2025-02-05
Added "orjson" and 'all' to setup; updated README

dsblank:
This PR:

1. Adds "orjson" as install_requires
2. Adds a new "all" group with common python packages used in Gramps
3. Updated the README (removed bsddb3, sqlite3 as requirements)


setup.py
```diff
@@ -450,11 +450,20 @@ def run(self):
...
     extras_require={
-        "bsddb": ["bsddb3"],
         "image": ["Pillow"],
         "i18n": ["PyICU"],
         "GUI": ["PyGObject", "pycairo"],
+        "all": [
+            "Pillow",
+            "PyGObject",
+            "PyICU",
+            "PyICU",
```
dsblank, line 462:
```suggestion
```

Nick-Hall:
The `bsddb3` package is still used for database upgrades.  We should move it to the "optional" section rather than removing it completely.

dsblank:
> The bsddb3 package is still used for database upgrades. We should move it to the "optional" section rather than removing it completely.

Moved to optional.

README.md
```diff
@@ -81,6 +79,10 @@ The following packages are **STRONGLY RECOMMENDED** to be installed:
 
 The following packages are optional:
 ------------------------------------
+* [**bsddb3**](https://pypi.python.org/pypi/bsddb3/)
+
+Python bindings for Oracle Berkeley database. Only needed for older Gramps databases.
```
Nick-Hall, line 84:
```suggestion
Python bindings for Oracle Berkeley database. Only needed to upgrade older Gramps databases.
```

Nick-Hall:
Updated the commit message and rebased onto the _gramps60_ branch.

PR #1930, opened 2025-02-06
Update CI to run unit tests with Python 3.9 and 3.13

DavidMStraub:
The current master branch as well as Gramps 6.0 beta fail with syntax errors due to the new typing module on Python 3.9 (sorry, my fault).

To prevent that from happening again, I changed the CI action to test with Python 3.9 (the oldest version we support) and Python 3.13, so we are sure Gramps runs on all supported Python versions.

To make that possible, I installed the Python dependencies in the CI with pip rather than apt.

This already showed one additional benefit: I caught an error due to a function in the `unittest` module that was removed in Python 3.12 (see https://docs.python.org/3/whatsnew/3.12.html#id3), now also fixed.

@Nick-Hall, please apply these fixes to the 6.0 branch as well. Thank you!


stevenyoungs:
Previous relevant discussion in #1910 and in particular https://github.com/gramps-project/gramps/pull/1910#issuecomment-2629357690

Python 3.9 reaches end of life in October 2025. If gramps 6.x is not expected before then, we could target python 3.10 as the minimum on master.

Nick-Hall:
Squashed some commits and rebased onto the _gramps60_ branch.

Nick-Hall:
Why are the builds so much slower?  Previously they were about 4 minutes; now they are 11 minutes and 13 minutes.

QuLogic:
`PyICU` and `PyGObject` are built from source, unlike the others that have wheels.

DavidMStraub:
> `PyICU` and `PyGObject` are built from source, unlike the others that have wheels.

No, that's not the reason. The "install dependencies" step takes 90 seconds instead of 60, a small change.

It seems like the unit test suit actually takes longer:

`Ran 32447 tests in 655.001s`
vs
`Ran 32447 tests in 178.384s`

That's quite suprising :thinking:

DavidMStraub:
I tried changing a couple of things but it didn't change the runtime, not sure what else to try. Still, I think it's a cleaner and more controlled test environment.

Nick-Hall:
Merged the bug fixes separately and rebased.

Nick-Hall:
We should run the unit tests with the `--durations 0` option to find out which ones are slower.

DavidMStraub:
No test longer than 10 s, strange.

Nick-Hall:
Just picking random tests, it looks like most test are about 3 times slower.

I'll investigate further.

Nick-Hall:
Tested locally with a virtual environment.  The unit tests run only very slightly slower compared to the system packages.

Nick-Hall:
With Ubuntu 24.10:
`Ran 32447 tests in 651.465s`  (python 3.9)
`Ran 32447 tests in 523.154s`  (python 3.13)

DavidMStraub:
I suspect it's a peculiarity of the Github runners when using `actions/setup-python` as opposed to Ubuntu's system Python. But the total runtime is still acceptable IMO and it's the only way we can ensure PRs do not break compatibility with individual Python versions.

Nick-Hall:
I'll close this since it has been overwritten by the change in PR #1997.  Both PRs used the same _master_ branch.

PR #1931, opened 2025-02-06
Fix the match method in the GenericFilter class

Nick-Hall:
This was causing problems when using the filter sidebar in the place tree view.  A filtered place was being displayed as "????" if an enclosing place was removed or added back again.

Fixes [#13606](https://gramps-project.org/bugs/view.php?id=13606).

### DIFF
--- a/gramps/gen/filters/_genericfilter.py
+++ b/gramps/gen/filters/_genericfilter.py
@@ -80,7 +80,7 @@ class GenericFilter:
         """
         Return True or False depending on whether the handle matches the filter.
         """
-        obj = self.get_object(handle)
+        obj = self.get_object(db, handle)
         return self.apply_to_one(db, obj)
 
     def is_empty(self):


dsblank (approved):
LGTM

soig:
Humm, I'm still seeing the bug BUT now I'm also seeing this on the console:

 File "/usr/lib/python3.12/site-packages/gramps/gui/views/treemodels/treebasemodel.py", line 792, in add_row_by_handle
    if not self.search or (self.search and self.search.match(handle, self.db)):
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/gramps/gen/filters/_genericfilter.py", line 83, in match
    obj = self.get_object(db, handle)
          ^^^^^^^^^^^^^^^^^^^^^^^
TypeError: GenericPlaceFilter.get_object() missing 1 required positional argument: 'handle'

Nick-Hall:
@soig Thanks for testing. It looks like we also need to swap the order of the parameters in the call in the treebasemodel.

lordemannd:
changing line 83 of gramps/gen/filters/_genericfilter.py
to "obj = self.get_object(db, handle)" (that is: the original change in pr 1931)

works for me on Windows (AIO64-6.0.0-r1-ba2c71f + PR 1927,1931)

soig:
@Nick-Hall as the new exception shows, this is with that that commit…

soig:
Humm I'm confused, I cannot reproduce again.

Nick-Hall:
No. Nothing needed swapping.  The two parameters are in a different order in the different methods.

PR #1, opened 2015-03-10
Remove executable flag of two css files

sebschub:
This is a trivial minor issue:

As far as I can see, the two css files do not have be executable. The rpm package checker of openSUSE complains about this:
````
gramps.noarch: W: script-without-shebang /usr/share/gramps/css/Web_Vertical-Menus.css
gramps.noarch: W: script-without-shebang /usr/share/gramps/css/Web_Horizontal-Menus.css
This text file has executable bits set or is located in a path dedicated for
executables, but lacks a shebang and cannot thus be executed.  If the file is
meant to be an executable script, add the shebang, otherwise remove the
executable bits or move the file elsewhere.
````
This is corrected with this pull request.

The checker also complains about non-executables which include a shebang. Here, I am not sure what to do:
````
gramps.noarch: W: non-executable-script /usr/lib/python2.7/site-packages/gramps/plugins/gramplet/metadataviewer.py 0644L /usr/bin/env
gramps.noarch: W: non-executable-script /usr/lib/python2.7/site-packages/gramps/test/regrtest.py 0644L /usr/bin/env
gramps.noarch: W: non-executable-script /usr/lib/python2.7/site-packages/gramps/gen/utils/place.py 0644L /usr/bin/env
gramps.noarch: W: non-executable-script /usr/lib/python2.7/site-packages/gramps/gen/datehandler/_date_fr.py 0644L /usr/bin/python
gramps.noarch: W: non-executable-script /usr/lib/python2.7/site-packages/gramps/plugins/lib/libhtml.py 0644L /usr/bin/python
gramps.noarch: W: non-executable-script /usr/lib/python2.7/site-packages/gramps/plugins/lib/libmetadata.py 0644L /usr/bin/env
gramps.noarch: W: non-executable-script /usr/lib/python2.7/site-packages/gramps/plugins/quickview/reporef.py 0644L /usr/bin/python
This text file contains a shebang or is located in a path dedicated for
executables, but lacks the executable bits and cannot thus be executed.  If
the file is meant to be an executable script, add the executable bits,
otherwise remove the shebang or move the file elsewhere.
````

PR #1934, opened 2025-02-06
Minimal code changes to support type hinting of db methods

stevenyoungs:
These are the code changes split out from #1919 

Each commit fixes one thing.
The first 7 commits are required for mypy to pass the type hinted code which will be introduced in a later PR #1919 
The last 3 commits fix a spelling mistake and rename parameters for consistency and clarity. These could be moved to a separate PR if required.


gramps/gen/db/dummydb.py
```diff
@@ -1286,151 +1286,170 @@ def iter_citations(self):
         """
         if not self.db_is_open:
             LOG.debug("database is closed")
-        return []
+        return
+        yield
```
DavidMStraub, line 1290: I wouldn't do that. The `yield` is unreachable.

What about just doing `yield from ()` (without `return`)? Same below

gramps/gen/proxy/proxybase.py
```diff
@@ -328,63 +328,68 @@ def iter_family_handles(self):
         Return an iterator over database handles, one handle for each Family in
         the database.
         """
-        return filter(self.include_family, self.db.iter_family_handles())
+        return (_ for _ in filter(self.include_family, self.db.iter_family_handles()))
```
DavidMStraub, line 331: Filter returns a generator. It doesn't make sense to have a generator expression iterating over a filter. Did mypy complain about this?

DavidMStraub:
Some really nice examples in there of bugs that could happen if `None` is not handled properly.

gramps/gen/proxy/proxybase.py
stevenyoungs: Yes, mypy complains:

`gramps/gen/proxy/proxybase.py:371: error: Incompatible return value type (got "filter[Never]", expected "Generator[FamilyHandle, None, None]")  [return-value]`

gramps/gen/db/dummydb.py
stevenyoungs: mypy is happy with `yield from ()`.
I'll update the PR accordingly. Thanks.

DavidMStraub:
> The get_<object>_from_handle methods should never return None in base databases. We should be annotating the return from "get_person_from_handle" as "Person" not "Person | None". Then changes in your commit "Correctly handle None return from get__from_handle" would no longer be required. Returning empty objects instead of None from the proxies would be a better solution.

This is an excellent example how the type hinting can identify poor design choices, such as proxies returning `None`.

But I also don't think returning an empty object is good. Would it have a handle? Probably, because we used `get_from_handle`. Would it have a gramps ID? Certainly not. Now we need code to handle primary objects which, unexpectly, don't have Gramps IDs etc.

In my opinion, the cleanest solution is to raise `HandleError`. Then all `DbReadBase` descendants behave in the same way.

gramps/gen/proxy/proxybase.py
DavidMStraub: Oh! I was wrong, it returns an iterator but not a generator. What about

```suggestion
        yield from filter(self.include_family, self.db.iter_family_handles())
```

stevenyoungs: `yield from filter(self.include_family, self.db.iter_family_handles())` works.
I'll update the PR accordingly. Thanks

stevenyoungs:
As it stands the handle is None and gramps_id is None

The comment in TableObject 
`# If source is None, the handle is assigned as an empty string.`
does not match the current code
`self.handle = None`

Nick-Hall:
We have known for years that proxies returning `None` is a bad idea.  That's why I don't want `get_person_from_handle` annotated as returning "Person | None" for example.

I introduced the `HandleError` back in 2015 as a first step to eliminating the use of `None`.  I have no objection to raising a `HandleError` or perhaps another exception instead of returning `None`.

stevenyoungs:
If it is ultimately considered an error then I'll revise the code to raise an exception of a new type. It that turns out to be too painful, we always have the option to backtrack and return an empty object instead. By using a new exception type, we can identify the locations that were modified.

Is that a reasonable approach?

stevenyoungs:
It could get interesting to start raising an exception instead of returning `None` here:

https://github.com/gramps-project/gramps/blob/ef5512c4347a3303c31608a3382867cd4994a37d/gramps/gen/proxy/private.py#L87-L89

Nick-Hall:
> Is that a reasonable approach?

The best approach is for a proxy never to return a handle for an object that needs to be filtered out.  However, an interim solution is also acceptable, as long as we are heading in the right direction.

Returning fully or partially redacted objects and raising exceptions are alternatives to returning `None`.

kulath:
> In my opinion, the cleanest solution is to raise `HandleError`. Then all `DbReadBase` descendants behave in the same way.

Yes, excellent solution.

`HandleError` instead of any other exception is much better. You could identify each location that is modified with a FIXME to note that a change is needed so "a proxy never to return a handle for an object that needs to be filtered out".

dsblank:
> It could get interesting to start raising an exception instead of returning None here:

Yes, that should not return None. I'll wait to fix that those... I'm just going to fix bugs in #1941

gramps/plugins/importer/importxml.py
```diff
@@ -866,7 +866,7 @@ class object of a primary object.
                     "event": self.db.has_event_handle,
                     "place": self.db.has_place_handle,
                     "source": self.db.has_source_handle,
-                    "citation": self.db.get_raw_citation_data,
+                    "citation": self.db.has_citation_handle,
```
stevenyoungs, line 869: @Nick-Hall this one might be worth worth pushing to 6.0 as well. Let me know and I can make a separate PR

I think it is more of a performance problem, unnecessarily get the citation `DataDict`, just to check if the handle is a valid citation

Nick-Hall: Yes.  Something like this is obviously a mistake and just a single line change.  I think that we can accept it as a bug even if it isn't actually causing a problem.

gramps/plugins/lib/libmixin.py
```diff
@@ -77,8 +83,8 @@ def __find_primary_from_handle(
         obj = class_type()
         handle = str(handle)
         new = True
-        raw = get_raw_obj_data(handle)
-        if raw is not None:
+        if has_obj_handle(handle):
+            raw = get_raw_obj_data(handle)
             obj = data_to_object(raw)
```
stevenyoungs, line 88: I can't see why this doesn't just get the object in one go.

gramps/plugins/lib/libmixin.py
```diff
@@ -103,8 +109,8 @@ def __find_table_from_handle(
         """
         obj = class_type()
         handle = str(handle)
-        raw = get_raw_obj_data(handle)
-        if raw is not None:
+        if has_obj_handle(handle):
+            raw = get_raw_obj_data(handle)
             obj = data_to_object(raw)
```
stevenyoungs, line 114: I can't see why this doesn't just get the object in one go.

gramps/plugins/importer/importxml.py
stevenyoungs: OK. I'll split out later.

stevenyoungs:
The original first commit, 

> Correctly handle None return from &lt;get_object&gt;_from_handle
> when type hints are added, mypy will detect an error without this change

which provoked much discussion is now dropped from this PR 🎆 
Thanks to all commentors for the ideas and encouragement to keep pushing

I think this PR ready for review

gramps/plugins/lib/libmixin.py
stevenyoungs: Addressed in the commit
>__find_primary_from_handle no longer uses get_raw_obj_from_handle to check for handle existence. As a result it can get the object directly rather then convert the raw data into an object

gramps/plugins/lib/libmixin.py
stevenyoungs: Addressed in the commit
>__find_primary_from_handle no longer uses get_raw_obj_from_handle to check for handle existence. As a result it can get the object directly rather then convert the raw data into an object

stevenyoungs:
libmixin.py now has the following code:
```
        handle = str(handle)
        if has_obj_handle(handle):
            obj = get_obj_from_handle(handle)
```

I'm hypothesising that 
1. libmixin clients are not randomly making up `handle` values
any valid handle value must have originated from the DB and therefore exist
2. it is not necessary to call `has_obj_handle`.
Either 
a. `handle` is None or an empty string
=> does not refer to an object
b. the object referred to by `handle` will exist

The only way #1 can be wrong is through object deletion, after the handle has been obtained. In the context of this code, I do not believe that this happens.

Your  thoughts welcome, but I'm not making this change in this PR!

dsblank:
> The only way https://github.com/gramps-project/gramps/pull/1 can be wrong is through object deletion, after the handle has been obtained. In the context of this code, I do not believe that this happens.

I hope that is the case, but I am not 100% sure. If there are left over handles, I think we should treat that as a bug, and that the database is corrupted.

stevenyoungs:
rebased

stevenyoungs:
rebased and re-ordered commits

@dsblank with your proposed proxy db patch I'm not sure the last two commits in this set are still relevant. They introduced a new `AccessDenied` exception which a proxy db would raise if an inaccessible handle was passed to `get_X_from_handle()` instead of the previous behaviour of returning None. What do you think?

dsblank:
> @dsblank with your proposed proxy db patch I'm not sure the last two commits in this set are still relevant. They introduced a new AccessDenied exception which a proxy db would raise if an inaccessible handle was passed to get_X_from_handle() instead of the previous behaviour of returning None. What do you think?

With the new proxy PR, there should never be a case that a handle would be looked up that was not a valid person in the context of that proxy.

That doesn't mean that some random code couldn't call `get_X_from_handle()`. I think we should never return None and that raising an error is the correct thing to do.

stevenyoungs:
closed.
Most commits moved to #2232

PR #1938, opened 2025-02-07
Provide a better description of development versions

Nick-Hall:
Use `git describe` to provide the version text for development versions.

Before this change:
```
> python Gramps.py -v
6.0.0-beta2-149fb610ca
```

After the change:
```
> python Gramps.py -v
v6.0.0-beta1-28-g0abc833f08
```


stevenyoungs:
By default the workflow checkout action only fetches a single commit. See https://github.com/actions/checkout/blob/v4/README.md

Nick-Hall:
The output isn't very useful on the _master_ branch, maybe `git describe` could still be a good way to get the commit hash.

Nick-Hall:
I'm closing this because it doesn't give much benefit, and isn't useful on the _master_ branch.

PR #1941, opened 2025-02-08
Private proxy fix

dsblank:
This PR fixes a missing sanitize method in the `PrivateProxyDb`

### DIFF
--- a/gramps/gen/proxy/private.py
+++ b/gramps/gen/proxy/private.py
@@ -47,6 +47,7 @@ from ..lib import (
     Address,
     EventRef,
     Person,
+    PersonRef,
     Name,
     Source,
     RepoRef,
@@ -165,6 +166,7 @@ class PrivateProxyDb(ProxyDbBase):
         """
         note = self.db.get_note_from_handle(handle)
         if note and not note.get_privacy():
+            # Nothing to sanitize in note object:
             return note
         return None
 
@@ -550,7 +552,8 @@ def copy_associations(db, original_obj, clean_obj):
         if person_ref and not person_ref.get_privacy():
             associated_person = db.get_person_from_handle(person_ref.ref)
             if associated_person and not associated_person.get_privacy():
-                new_person_ref_list.append(person_ref)
+                new_person_ref = sanitize_person_ref(db, person_ref)
+                new_person_ref_list.append(new_person_ref)
     clean_obj.set_person_ref_list(new_person_ref_list)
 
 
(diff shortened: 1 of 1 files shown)


dsblank:
@stevenyoungs I think we should not allow proxies to return None from the `db.get_OBJECT()` (or any method) as that hides bugs. It should be the case that sanitizing the object removes all inaccessible handles/references.

Nick-Hall:
Returning `None` also leaks the information that the object exists, which is another reason not to do so.

dsblank:
The LivingProxyDb is going to require a rewrite. And I think there is an error when one proxy is on top of another.

I'll go ahead and turn off draft, and do a major rewrite when @stevenyoungs is done.

dsblank:
@Nick-Hall, this could go into 6.0 as it is just a regular bug fix.

dsblank:
Originally reported here: https://gramps-project.org/bugs/view.php?id=12886

dsblank:
When we revise the proxies to make sure they work correctly (sanitize all, proper proxy layering, no need to check for None, etc) we also need to look at caching the results. When I exported 2 people, one person was sanitized 6 times, and the other 3 times. Sanitizing is an expensive process and there is no need to do it more than once.

[I would put this note in a github issue if we turned those on. I see issues as being for developers, not a replacement for mantis].

stevenyoungs:
The change looks good to me.

Here are some observations from looking at the code around his change.
The implementation of `copy_citation_ref_list` has the following:

```
            handle = citation.get_reference_handle()
            source = db.get_source_from_handle(handle)
            if source and not source.get_privacy():
                clean_obj.add_citation(citation_handle)
```
1. there is no check on `handle`. 
Is there any way a citation can exist without a `reference_handle`?
I know the standard UI forces a source to be selected when creating a citation.
2. `if source`
A nice example of a check that can be removed if #1934 gets accepted. `get_source_from_handle` would have raised an error. Even today, it will in most scenarios.
3.  if we are prepared to use `source.private` rather than call `source.get_privacy()` it can be optimised to get the raw object
I'm undecided on this.
```
            data = db.get_raw_source_from_handle(handle)
            if not source.private:
                clean_obj.add_citation(citation_handle)
```

dsblank:
I just made this function to work and look exactly like the others. I am re-writing private and living proxies completely so I wouldn't worry about the details of this sanitize function.

Nick-Hall:
Yes.  Keep bug fixes simple.

PR #1943, opened 2025-02-08
Narrative web: Print page does not show the tree.

SNoiraud:
Fixes [#13614](https://gramps-project.org/bugs/view.php?id=13614)


Nick-Hall:
@SNoiraud Do you want this merged into the _gramps60_ branch instead of _master_?

SNoiraud:
Yes. I change the target.

PR #1944, opened 2025-02-09
Refactor proxies

dsblank:
WIP

This PR (targeting 6.1) is a refactor of the proxies to solve the following issues. 

These proxies:

* will never return `None` for an object lookup (eg, `db.get_person_from_handle()`)
* will not re-create an object more than once (if object is in cache)
* work correctly when stacking proxy on proxy, in any order
* requires an initial load (loop through data) but should be much faster

This PR is for discussion. I didn't add any type hints yet, but everything is designed to be type safe. If there are no issues I'll proceed with fleshing out the entire framework.

### DIFF
--- a/gramps/gen/proxy/base.py
+++ b/gramps/gen/proxy/base.py
@@ -101,8 +101,8 @@ class ProxyDbBase:
     def get_number_of_people(self):
         return len(self.person_map)
 
-    def process_person(self, handle):
-        return self.get_person_from_handle(handle)
+    def process_person(self, person):
+        return person
 
     @functools.cache
     def get_person_from_handle(self, handle):
--- a/gramps/gen/proxy/base.py
+++ b/gramps/gen/proxy/base.py
@@ -31,6 +31,7 @@ Proxy class for a Gramps database.
 # -------------------------------------------------------------------------
 import types
 import functools
+import locale
 
 # -------------------------------------------------------------------------
 #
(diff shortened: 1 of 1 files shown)


Nick-Hall:
> This PR (targeting 6.1)

Please rebase onto the _master_ branch.

dsblank:
@stevenyoungs, we're going to have some merge issues in the `gramps.gen.proxy` files. I'm currently keeping all of this new proxy framework in a single new file so we can decide how to move forward. I think we have the following choices:

1. You can revert your changes to the proxy files in #1919, and I can add the type hints to this PR (I'll need to redo them as so much has changed)
2. I can rename the new framework proxy files and delete the old files, but we would lose all of the history.
3. We can wade through a mass of conflicts.

Can you think of another choice? What do you want to do?

dsblank:
(I can't seem to rebase to master cleanly. I'll make a new PR when @stevenyoungs and I have decided how to move forward).

stevenyoungs:
> @stevenyoungs, we're going to have some merge issues in the `gramps.gen.proxy` files. 

I agree that there could be merge problem but I don't see a significant problem. My changes are all pretty simple. You won;t be changing the method signature of the standard db methods so we should be able to accept all your (or my) changes, depending on who goes first. 
My PR has now gone as far as I want to for now. I need to let the review process play our and (hopefully) get some of the pre-requisite PRs merged.

dsblank:
@stevenyoungs, I think we'll be fine. I'll try to work all of your typing hints into a new version of this PR so we can largely just use this one in the merge.

dsblank:
Closing this PR; opening another.

PR #1945, opened 2025-02-09
Correct parameter type hints to accept <object> | None

stevenyoungs:
The type hints on these methods do not currently accept None.
None can be passed in and is handled correctly within the method.
Correct the type hint accordingly.

### DIFF
--- a/gramps/gen/filters/rules/family/_isancestorof.py
+++ b/gramps/gen/filters/rules/family/_isancestorof.py
@@ -71,7 +71,7 @@ class IsAncestorOf(Rule):
     def apply_to_one(self, db: Database, family: Family) -> bool:
         return family.handle in self.selected_handles
 
-    def init_list(self, db: Database, family: Family, first: bool):
+    def init_list(self, db: Database, family: Family | None, first: bool) -> None:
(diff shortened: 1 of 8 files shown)


stevenyoungs:
These problems were identified by mypy on master with PR #1934 and #1919

DavidMStraub:
@Nick-Hall @stevenyoungs some of these files use the pip syntax and are missing `from __future__ import annotations`. This will lead to syntax errors on Python 3.9. @stevenyoungs can you please add them in a follow-up PR?

This would have been caught by the CI in #1930.

stevenyoungs:
Good spot @DavidMStraub 
I think there are some others too

https://github.com/gramps-project/gramps/blob/ef5512c4347a3303c31608a3382867cd4994a37d/gramps/gen/lib/grampstype.py#L111

https://github.com/gramps-project/gramps/blob/ef5512c4347a3303c31608a3382867cd4994a37d/gramps/gen/utils/place.py#L128

and  more. I will fix all that I can find in a follow on PR.

stevenyoungs:
see PR #1956

PR #1949, opened 2025-02-09
mypy: Simplify RelationshipPathBetween.prepare()

stevenyoungs:
mypy complains about the code in its current form:

```
gramps/gen/filters/rules/person/_relationshippathbetween.py:71: error: Item "None" of "Person | None" has no attribute "handle"  [union-attr]
gramps/gen/filters/rules/person/_relationshippathbetween.py:72: error: Item "None" of "Person | None" has no attribute "handle"  [union-attr]
```

Simplify the code and fix the mypy errors

### DIFF
--- a/gramps/gen/filters/rules/person/_relationshippathbetween.py
+++ b/gramps/gen/filters/rules/person/_relationshippathbetween.py
@@ -67,12 +67,10 @@ class RelationshipPathBetween(Rule):
     def prepare(self, db: Database, user):
         self.db = db
         self.selected_handles: Set[PersonHandle] = set()
-        try:
-            root1_handle = db.get_person_from_gramps_id(self.list[0]).handle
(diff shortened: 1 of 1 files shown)


dsblank (approved):
Yes, that previous code could easily crash.

emyoulation:
@stevenyoungs do you think that this feature could be adapted to finding Family objects rather than Person Objects as the RelationshipPathBetween ?  If so, I will submit a feature request to MantisBT. (It seems like Families as Nodes might graph more cleanly than People Nodes.)

stevenyoungs:
My initial thought is that a relationship path exists between two families if there is a relationship path between any pair of people that make up the two families. There must be one person from each of the two families in the pair.
So if family F1 is the people (P1,P2) and family F2 is the people (P3, P4) then you need to test for a relationship path between
P1, P3
P1, P4
P2, P3
P2, P4

PR #1950, opened 2025-02-09
Remove Duplicate Definition of Database

stevenyoungs:
There are currently two definitions of Database
1. [`gb/generic.py`](https://github.com/gramps-project/gramps/blob/309e086e782b8e5afffe9448518acfae501f26ca/gramps/gen/db/generic.py#L2788)
2. [`db/__init__.py`](https://github.com/gramps-project/gramps/blob/309e086e782b8e5afffe9448518acfae501f26ca/gramps/gen/db/__init__.py#L75)

When using Database as a type hint, pylance reports [Variable not allowed in type expression](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportInvalidTypeForm)

The second definition appears to create a variable called `Database` rather than a type. I'm interested in the reasoning if anyone can explain.

Remove the definition from `generic.py` so that Database is a type

### DIFF
--- a/gramps/gen/filters/rules/person/_isdescendantfamilyoffiltermatch.py
+++ b/gramps/gen/filters/rules/person/_isdescendantfamilyoffiltermatch.py
@@ -37,7 +37,7 @@ _ = glocale.translation.gettext
 from ._isdescendantfamilyof import IsDescendantFamilyOf
 from ._matchesfilter import MatchesFilter
 from ....types import PersonHandle
-from ....db.generic import Database
+from ....db import Database
 
 
 # -------------------------------------------------------------------------
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -2783,6 +2783,3 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
             self.serializer = BlobSerializer
         elif serializer_name == "json":
             self.serializer = JSONSerializer
-
-
-Database = DbGeneric


stevenyoungs:
It might be better to move this type definition into `types.py`. That would require a larger change as many filters currently have
`from ....db import Database`
Let me know and I can update this PR as required

Nick-Hall:
Please keep the changes as small as possible if you intend them for the _gramps60_ branch.

gramps/gen/filters/rules/person/_isdescendantfamilyoffiltermatch.py
```diff
@@ -37,7 +37,7 @@
 from ._isdescendantfamilyof import IsDescendantFamilyOf
 from ._matchesfilter import MatchesFilter
 from ....types import PersonHandle
-from ....db.generic import Database
```
dsblank, line 40: What is Database now? It used to be an alias.

dsblank:
I don't mind if you just change `Database` to be `DbGeneric`. I don't think it is as descriptive for this purpose, but I am not going to die on that hill.

Do you know `sed`?

stevenyoungs:
> I don't mind if you just change `Database` to be `DbGeneric`. I don't think it is as descriptive for this purpose, but I am not going to die on that hill.
> 
> Do you know `sed`?

I'd prefer to keep Database as a type alias, if I can understand why it's not working as expected.
I'm hoping it won't come to updating multiple files but if necessary I'll do so. These days I tend to use an IDE instead of cmd line tools like sed. vscode is pretty good - search and replace in files with regex, I can do most git commands too directly from the IDE

Let me do a bit more investigation so I'm confident on the change that is actually required.

stevenyoungs:
Using maintenance/gramps60 and both locations modified as follows:
```
Database = DbGeneric
reveal_type(Database)
reveal_type(DbGeneric)
```

mypy reports
```
gramps/gen/db/generic.py:2789: note: Revealed type is "def (directory: Any =) -> gramps.gen.db.generic.DbGeneric"
gramps/gen/db/generic.py:2790: note: Revealed type is "def (directory: Any =) -> gramps.gen.db.generic.DbGeneric"
gramps/gen/db/__init__.py:76: note: Revealed type is "def (directory: Any =) -> gramps.gen.db.generic.DbGeneric"
gramps/gen/db/__init__.py:77: note: Revealed type is "def (directory: Any =) -> gramps.gen.db.generic.DbGeneric"
```

Modifying HasSourceOfBase.prepare as follows:
```
    def prepare(self, db: Database, user):
        reveal_type(db)
        reveal_type(Database)
```

also gives the expected result:
```
gramps/gen/filters/rules/_hassourceofbase.py:62: note: Revealed type is "gramps.gen.db.generic.DbGeneric"
gramps/gen/filters/rules/_hassourceofbase.py:63: note: Revealed type is "def (directory: Any =) -> gramps.gen.db.generic.DbGeneric"
```

More investigation required. Maybe it's a pylance\pyright problem?

stevenyoungs:
I'm closing this as I'm not convinced there is a problem.
I've unified the definition of the Database type and moved it to types.py as part of #1919.

PR #1952, opened 2025-02-10
Refactor proxies

dsblank:
WIP

This PR (targeting 6.1) is a refactor of the proxies to solve the following issues.

These proxies:

1. will never return None for an object lookup (eg, db.get_person_from_handle())
2. will not re-create an object more than once (if object is in cache)
3. work correctly when stacking proxy on proxy, in any order
4. requires an initial load (loop through data) but should be much faster


gramps/gen/proxy/proxybase.py
```diff
@@ -3,10 +3,11 @@
 #
 # Copyright (C) 2007       Brian G. Matherly
 # Copyright (C) 2011       Tim G L Lyons
+# Copyright (C) 2025       Doug Blank <doug.blank@gmail.com>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
-# the Free Software Foundation; either version 2 of the License, or
+# the Free Software Foundation; either version 2 of the Liceonse, or
```
stevenyoungs, line 10:
```suggestion
# the Free Software Foundation; either version 2 of the License, or
```

dsblank:
> but is there a way to make this work while still having proxy DBs inherit from DbReadBase? 

So these are wrappers, not subclasses. The difference is that these don't need to re-implement methods that are `NotImplemented`, but just call the instance methods, whatever they are.

> If not a common base class, we would need a [protocol](https://docs.python.org/3/library/typing.html#typing.Protocol) I think.

That sounds like a fine PR to propose.

PR #1953, opened 2025-02-10
Fix upgrade bug in 6.0

dsblank:
This PR fixes the issue: https://gramps-project.org/bugs/view.php?id=13622


Nick-Hall:
@dsblank Thanks for fixing this.

Please put "Fixes #13622" on the last line of your commit message.  This will close the bug report automatically when merged and also appear in the ChangeLog and NEWS files.

You can find details in the [Log messages](https://gramps-project.org/wiki/index.php/Committing_policies#Log_messages) section of our committing policies.

PR #1954, opened 2025-02-10
Fixes bsddb upgrade

dsblank:
Thanks @prculley !


Nick-Hall:
This fixes bug [#13627](https://gramps-project.org/bugs/view.php?id=13627).

PR #1955, opened 2025-02-10
Specify the obj_class when calling data_to_object

stevenyoungs:
Make sure the `obj_class` argument is always supplied when calling `data_to_object`

For `JSONSerializer` the `obj_class` parameter defaults to `None`
For `BlobSerializer` there is no default

Found by mypy

### DIFF
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -2012,7 +2012,7 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
         old_data = self._commit_base(person, PERSON_KEY, transaction, change_time)
 
         if old_data:
-            old_person = self.serializer.data_to_object(old_data)
+            old_person = self.serializer.data_to_object(old_data, Person)
             # Update gender statistics if necessary
             if old_person.gender != person.gender or (
                 old_person.primary_name.first_name != person.primary_name.first_name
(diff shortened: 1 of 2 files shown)


stevenyoungs:
@dsblank I would appreciate your review of this in case I missed something.

dsblank:
@stevenyoungs, technically I don't think they are needed here because the class is only needed for reading the the bsddb data on upgrade, but better safe than sorry.

dsblank (approved):
LGTM!

stevenyoungs:
> @stevenyoungs, technically I don't think they are needed here because the class is only needed for reading the the bsddb data on upgrade, but better safe than sorry.

Yeah, I suspect as much too, but mypy was not able to deduce that and gave an error.

A random thought: should the undo\redo code still use BlobSerializer? From memory, undo\redo data does not persist and BlobSerializer is a bit faster.

dsblank:
> should the undo\redo code still use BlobSerializer? 

No, we want to remove all uses of pickle. I'm not sure if we have 100% of them gone yet.

dsblank:
> BlobSerializer is a bit faster.

I'm not sure that is still true now that we use `orjson`. In any event, I don't think we want to maintain two types of "data".

Nick-Hall:
My guess would be that `pickle` is still faster than `orjson`, but not by much.  However, we also need to consider ease of maintenance, so we don't want two data formats.

PR #1958, opened 2025-02-11
Change the progress bar vertical alignment and padding

Nick-Hall:
This will show the progress bar centered with the cancel button and give some space between them.

Fixes [#13623](https://gramps-project.org/bugs/view.php?id=13623).


emyoulation:
Nice revision.
Posted revised image in the Discourse forum.  To make the 3 revised dialogs in the same place and easier to post to the wiki images when it is rolled over to 6.0

https://gramps.discourse.group/t/where-should-gramps-6-0-beta-cycle-bugs-be-reported/6979/2

PR #1960, opened 2025-02-11
Remove BSDDB version and duplicate Gramps version from About dialog

Nick-Hall:
Fixes [#13624](https://gramps-project.org/bugs/view.php?id=13624).


emyoulation:
MantisBT did not want to serve up your revised capture of the dialog.  So here is a capture from my ThinkPad:

image

PR #1963, opened 2025-02-12
Allow filters to be cancelled

dsblank:
This PR does the following:

1. Allows filters to be cancelled
2. Adds User type hints to filters
3. Removed `if user:` because there should always be a user

### DIFF
--- a/gramps/cli/user.py
+++ b/gramps/cli/user.py
@@ -81,7 +81,9 @@ class User(user.UserBase):
                 self._default_callback
             ) = yes
 
-    def begin_progress(self, title, message, steps):
+    def begin_progress(
+        self, title, message, steps, can_cancel=False, cancel_callback=None
+    ):
         """
         Start showing a progress indicator to the user.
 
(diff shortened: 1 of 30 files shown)


dsblank (commented):
## Review: Allow filters to be cancelled

The architecture is sound and the PR achieves its goal cleanly.

### What the PR gets right

- `apply()` now guarantees a `User()` instance (defaulting when `None` is passed), so all downstream code can always call `user.get_cancelled()` safely.
- `cli.User` and `gen.User` implement `get_cancelled()` returning `False` — correct no-op behaviour outside the GUI.
- All the old `if user:` guards are consistently cleaned up.
- `begin_progress` gains `can_cancel=False` as a default — fully backward-compatible.
- Cancellation checks in `_genericfilter.py` correctly reset rules and return `[]` on early exit.

Ready to merge.

dsblank:
Superseded by #2454, reimplemented against current master since this branch predates the filter/rules Optimizer refactor and no longer rebases cleanly. Closing in favor of the new PR.

PR #1964, opened 2025-02-12
Fix bug in filter (swapped lines)

dsblank:
This PR fixes an issue where lines got swapped.


dsblank:
In practice, I don't think this matters (and I'm removing all "if user:" tests in 6.1) but it is confusing to see.

PR #1965, opened 2025-02-12
Remote media cannot be opened in the narrativeweb.

SNoiraud:
Fixes [#13628](https://gramps-project.org/bugs/view.php?id=13628).


Nick-Hall:
This works for me.

PR #1971, opened 2025-02-14
Fix citation filter rules that search the source

Nick-Hall:
Fixes [#13635](https://gramps-project.org/bugs/view.php?id=13635).


dsblank (approved):
Nice catch!

lordemannd:
[6.0.0 beta 2] I can reproduce this on Windows with new empty DB **& this PR**
- add an event
- click to add an EXISTING source citation (empty DB so shouldn't be anything there)
- crash to the Error Reporting Wizard with similar back trace

PR #1972, opened 2025-02-14
Fix crash in source/citation selector with an empty filter

Nick-Hall:
Fixes [#13634](https://gramps-project.org/bugs/view.php?id=13634).

### DIFF
--- a/gramps/gui/selectors/baseselector.py
+++ b/gramps/gui/selectors/baseselector.py
@@ -319,13 +319,9 @@ class BaseSelector(ManagedWindow):
         """
         search_bar_filter = self.search_bar.get_value()
         filter_info = (
-            (
-                0,
-                search_bar_filter,
-                search_bar_filter[0] in self.exact_search(),
-            )
-            if search_bar_filter[1]
-            else None
+            0,
+            search_bar_filter if search_bar_filter[1] else None,
+            search_bar_filter[0] in self.exact_search(),
         )
 
         if self.model:


Nick-Hall:
@stevenyoungs This is related to one of your changes.  I'll send you an invitation to the organisation so that I can add you as a reviewer.

lordemannd:
seems fixed on windows,  tested with beta2-1 + PR 1971,1972

dsblank (approved):
Thanks for the fix!

gramps/gui/selectors/baseselector.py
```diff
@@ -319,13 +319,9 @@ def build_tree(self):
...
         filter_info = (
-            (
-                0,
-                search_bar_filter,
-                search_bar_filter[0] in self.exact_search(),
-            )
-            if search_bar_filter[1]
-            else None
+            0,
+            search_bar_filter,
```
stevenyoungs, line 323:
```suggestion
            search_bar_filter if search_bar_filter[1] else None,
```
I think this is a more performant fix.
The current change works but results in running a filter even if no search text is provided in the search bar. This results in calling `str.find("")` on every item in the model (see `SearchFilter.match()`). `str.find("")` matches matches every string. This is inefficient.

With the suggested change, we only pass in a search if there is some text to search for. Otherwise we pass in `None` which results in all items in the model being added to the view.

stevenyoungs: The above suggestion tested with:
1. Adding a citation to a new Event
2. Adding a Father to a new family. 
The selector correctly only shows males by default
3. Adding a Mother to a new family
The selector correctly only shows females by default
4. Adding a child to a new family

PR #1973, opened 2025-02-14
Remove the default toolbar style option and replace it with "Both"

Nick-Hall:
The "Use default preference" option has been removed because it was confusing users. Using a system desktop default also doesn't make much sense since the Toolbar widget has been removed in Gtk4.

The default of "Both text and icons" is intended to be helpful for new users.

### DIFF
--- a/gramps/gui/configure.py
+++ b/gramps/gui/configure.py
@@ -2121,10 +2121,9 @@ class GrampsPreferences(ConfigureDialog):
         # Toolbar styles:
         obox = Gtk.ComboBoxText()
         formats = [
-            _("Use default preference"),
+            _("Both text and icons"),
             _("Text only"),
             _("Icons only"),
-            _("Both text and icons"),
         ]
         list(map(obox.append_text, formats))
         active = config.get("interface.toolbar-style")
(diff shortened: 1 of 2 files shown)


lordemannd:
works for me, with 1 quibble.
for me, after adding this PR, on the Preferences->General Tab, the initial condition of the field was blank.
can't say if that is confusing for new users,  the blank field probably is understood to mean 'no attention required'

is it possible to initialize the value to one of the 3 values?

Nick-Hall:
@lordemannd Please delete your `gramps.ini` file and try again.

The problem is that you had `toolbar-style` set to 3, which is no longer possible.  The default is 0 (Both text and icons).

For upgrading users this won't be a problem.  For beta testers with the value set to 3, it will display an empty choice in the preferences and the buttons will display only icons instead of both icons and text.  Just choose an options and continue.

emyoulation:
The Zeroth option used to be the defaults INbeta1 and Beta2 

That '0' problem (now that it is not an option) only ever affect beta testers.

So it is an error condition that shouldn't need handling.

PR #1974, opened 2025-02-14
Allow older gramps versions to see (not crash) db schema 21

dsblank:
Older versions of Gramps cannot see the new db format (21) but just crashes.

This PR:

1. adds a `value` column of type `BLOB` to metadata 
2. always writes metadata to both blob and json

Fixes [#13633](https://gramps-project.org/bugs/view.php?id=13633)

### DIFF
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -713,18 +713,14 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
         # run backend-specific code:
         self._initialize(directory, username, password)
 
-        need_to_set_version = False
         if not self._schema_exists():
             self._create_schema()
-            need_to_set_version = True
+            self.set_schema_version(str(self.VERSION[0]))
 
         if self.use_json_data():
             self.set_serializer("json")
         else:
             self.set_serializer("blob")
-
-        if need_to_set_version:
-            self._set_metadata("version", str(self.VERSION[0]))
 
         # Load metadata
         self.name_formats = self._get_metadata("name_formats")
@@ -2773,7 +2769,16 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
 
     def set_schema_version(self, value):
         """set the current schema version"""
+        # First, set anything in the old blob metadata needed
+        # To open properly in gramps 5.2 and earlier:
+        self.set_serializer("blob")
         self._set_metadata("version", str(value))
+
+        # And set the json data, if possible:
+        if self.use_json_data():
+            self.set_serializer("json")
+            self._set_metadata("version", str(value))
+            self._set_metadata("name_formats", [])
 
     def set_serializer(self, serializer_name):
         """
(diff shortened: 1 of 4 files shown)


dsblank:
Actually when #1974 is merged, then we can skip the adding of the metadata value column. (I'll wait until @prculley tests this PR, then remove it).

Nick-Hall:
It looks like you didn't initialise the blob version for new databases.

I am getting the error:  `a bytes-like object is required, not 'NoneType'`

dsblank:
> It looks like you didn't initialise the blob version for new databases.

Are you sure this went through the upgrade? You can force it by removing the json_data in the metadata table.

dsblank:
Oh, I see... it only works on the upgrade. Needs to always work.

dsblank:
@Nick-Hall, ok I believe I have all of the paths correct now. I tested a 6.0 newly create DB, opened in 5.2, and an upgraded 6.0 in 5.2. Both complained about the version 21.

Nick-Hall:
Now I get a problem reading the name formats in the metadata table.  It doesn't give me an error regarding incompatible versions.

Steps to reproduce:
1. Create a new database in _gramps60_
2. Import the example database
3. Open the database in _gramps52_

dsblank:
I'm going to propose some changes to earlier version of Gramps that we check the version number before trying to read the rest of the metadata. Currently gramps5.2 , it checks the version AFTER it reads in a bunch of data. Let's move the check up before that, and release new versions.

Nick-Hall:
Do we really want to release updates to old versions?  How far back would we go?

People probably wouldn't upgrade and would still get the errors.

dsblank:
> Can we add pickled empty lists to the rows still causing a problem?

I can try. I'll look at that this weekend.

Nick-Hall:
@dsblank Don't spend too much time on this one.  It mainly affects developers and doesn't corrupt the database.

dsblank:
@Nick-Hall, ok, I think I have a solution: always write metadata to json_data and blob_data. This was a tricky situation, but always writing blob too solves the issue.

Nick-Hall:
@dsblank Thanks.  I'll test this again tomorrow.

prculley:
This still doesn't work for me when upgrading the 4.1.3 bsddb db.
```Traceback (most recent call last):
  File "D:\Users\PRC\Documents\Gramps\gramps\gui\dbloader.py", line 198, in read_file
    db.load(
  File "D:\Users\PRC\Documents\Gramps\gramps\plugins\db\bsddb\bsddb.py", line 224, in load
    self.dbapi.execute(
  File "D:\Users\PRC\Documents\Gramps\gramps\plugins\db\dbapi\sqlite.py", line 139, in execute
    self.__cursor.execute(*args, **kwargs)
sqlite3.OperationalError: table person has no column named blob_data
```

was it supposed to be built on top of the other PR that created the blob fields?

prculley:
I am working on a test that will take old db directories and load and compare with the correct, more modern XML.  But will need to get this working at least once to check out my test.

Nick-Hall:
Reverted.  This was causing a crash when exiting Gramps.

PR #1977, opened 2025-02-15
New User Defaults

stevenyoungs:
Configure the defaults in gramps so that they are more appropriate for a new user

All suggestions from @emyoulation in the forum post https://gramps.discourse.group/t/6-0-ux-feedback/7002/20

Update: 2025-02-16 the forum post has been deleted.


emyoulation (approved):

gramps/gen/config.py
```diff
@@ -172,7 +172,7 @@ def emit(key):
 register("behavior.startup", 0)
 register("behavior.surname-guessing", 0)
 register("behavior.translator-needed", True)
-register("behavior.use-tips", False)
+register("behavior.use-tips", True)
 register("behavior.welcome", 100)
 register("behavior.web-search-url", "http://google.com/#&q=%(text)s")
```
fxtmtrsine, line 177: While you are at it consider changing  the "http://" to the secure version "https://"  please?  for this and any other url you can find.

emyoulation:
That `https://` probably was not noticed because the websearch feature is only in the Note Editor and doesn't work anyway.

It is a context menu item for selected text in the Notes Editor.
image

For example, select the word "mission" in the note editor and choose "Search selection on the web" from the context menu

It tries to browse to:
`http://www.google.com/#&q=mission` and forwards to the secure 
`https://www.google.com/#&q=mission` 
    which just shows a blank google search form with the failed URL in the addressbar.
But Google expects:
`https://www.google.com/search?q=mission`

gramps/gen/config.py
emyoulation: A search for `http://www` and replace with `https://www`
would get most of the offenders without touching the parsers'

```
if path.startswith(("http://", "https://"))
```

stevenyoungs: I've created #1979 to do this.
Alas, it is not as simple as search and replace http with https.

stevenyoungs:
> That `https://` probably was not noticed because the websearch feature is only in the Note Editor and doesn't work anyway.
> 
> It is a context menu item for selected text in the Notes Editor.

See PR #1980

emyoulation:
Steven,

Thanks for the defaults PR and the search Fix.

Sorry the Discourse thread became so ugly. One of the admins decided that it was unsalvageable and deleted the thread. Probably the right choice although it was unfortunate to lose the code sample for tweaking the Date display format menus

stevenyoungs:
> Steven,
> 
> Thanks for the defaults PR and the search Fix.
> 
> Sorry the Discourse thread became so ugly. One of the admins decided that it was unsalvageable and deleted the thread. Probably the right choice although it was unfortunate to lose the code sample for tweaking the Date display format menus

Ah that's what happened. I just went to add a comment and saw the thread had gone.

here the code to display a date in all available formats:
```
from gramps.gen.datehandler import displayer as dd
from gramps.gen.lib.date import Date

for format_count in range(len(dd.formats)):
    dd.set_format(format_count)
    dd.display_formatted(Date(2025, 2, 17))
```

stevenyoungs:
I'm currently minded to remove the change to `database.backup-on-exit`, leaving the default as `True`
My thinking is as follows:
1. A backup is better than no backup
Let's not get into debate about ideal backup practice.
Having a backup can never be *worse* than not having a backup.
2. If there are other problems associated with performing a backup, let's fix them instead
I think, in the now deleted thread, someone mentioned about the default backup path being wrong?

emyoulation:
I agree. Let's defer this change to the backup on exit to 6.1

The path issue can be touched at the same time. (There's a bug for blank backup path handling that could raise a better solution.)

We've passed the point where more significant changes (than default Preferences switches being tweaked) should be applied. Late beta cycle change tempts fate.

Even the "websearch parameter passing" and https fixes might be too much. They could be pushed to 6.0.1 (which will probably be very soon after 6.0.0 given Paul's discovery of other oversights.)

emyoulation:
I agree. Let's defer this change to the backup on exit to 6.1

The path issue can be touched at the same time. (There's a bug for blank backup path handling that could raise a better solution.)

We've passed the point where more significant changes (than default Preferences switches being tweaked) should be applied. Late beta cycle change tempts fate.

Even the "websearch parameter passing" and https fixes might be too much. They could be pushed to 6.0.1 (which will probably be very soon after 6.0.0 given Paul's discovery of other oversights.)


Linked to your code snippet comment in the Feature Request on MantisBT related to those Date Display Format menus :
https://gramps-project.org/bugs/view.php?id=13396

emyoulation:
a question about the item added to suppress Tag on Import for XML
     gramps/gen/merge/test/merge_ref_test.py

Previous experiments only tagged on import for **GEDCOM** files. The same with adding sources to imports. None of any of the other formats tagged or sourced. 

If that is true, this override may be unnecessary. 

If it is NOT true so that tags/sources are being automatically added to XML files too, then is the added override to suppress the changing the "last modified" timestamp? Or the tag itself?

stevenyoungs:
> a question about the item added to suppress Tag on Import for XML gramps/gen/merge/test/merge_ref_test.py
> 
> Previous experiments only tagged on import for **GEDCOM** files. The same with adding sources to imports. None of any of the other formats tagged or sourced.
> 
> If that is true, this override may be unnecessary.
> 
> If it is NOT true so that tags/sources are being automatically added to XML files too, then is the added override to suppress the changing the "last modified" timestamp? Or the tag itself?

`preferences.tag-on-import` is used by several importers, including XML. These unit tests use an XML string as the input data (tree)  for the test(s) to use. The tests then merges two objects both of which have the import tag. The result of the merge (which also has the tag) is then compared to the original XML input data (which does not have a tag). Since these two strings differ, the test is failed. I've skipped some details but the essentials are correct as I understand it.
By setting `preferences.tag-on-import` to `False` we ensure that no tags are created when importing the XML test input data . The strings then match and the test passes.

emyoulation:
Thanks! That clarification helps!

Another thing that changes due to differing Preferences are Gramps IDs. I wonder if that is ignored in that comparison? (e.g., if the ID preferences is left at the default 4 gigits plus leading zeros, the IDs will likely match)

Nick-Hall:
Why do you want to enable `default-source` and `tag-on-import` by default?

emyoulation:
Drawbacks:
* GEDCOM import is already slow. this slows it further 
* Adding two 2ndry objects per primary makes the database bigger and slower
* adding a tag and a citation means the Timestamp for every object probably changes to today.  So you lose visibility to data that has been stagnating
* removing the tag and/or citation updates the Timestamp again. 
 
(Tagging doesn't improve the data quality but Citing does improve it. So it is unfortunate that the object being tagged has its timestamp updated rather than Tag reference for the tag having a timestamp.)

fxtmtrsine:
> Rationale for Enabling default-source and tag-on-import by Default

Why not when importing from the GUI have a popup that just offers the option maybe as part of the [GEDCOM Encoding dialog](https://gramps-project.org/wiki/index.php/Gramps_5.2_Wiki_Manual_-_Manage_Family_Trees#GEDCOM_Encoding_dialog) to set these options and then remove the options out of preferences altogether?

Nick-Hall:
Limited this PR to the following changes:

"behavior.use-tips" -> True
"interface.grampletbar-close" -> True
"preferences.family-warn" -> False

PR #1979, opened 2025-02-16
Use https instead of http

stevenyoungs:
Update URLs to use https instead of http

I have only updated URLs
1. Which are not part of a copyright
2. Are not a dtd, xmlns etc
My understanding is that these should remain as http. Please let me know if this is wrong.
3. Where the http and https variants both work and deliver the same content


stevenyoungs:
@Nick-Hall I've not fully updated the po files yet. Is updating them the correct approach - can you update weblate from the po files if I do finish updating them? I do not want to create additional work for translators from this change (whether in 6.0 or a late version)

stevenyoungs:
The following gramps URLs no longer work. Is there a working equivalent?
http://library.gramps-project.org
http://developers.gramps-project.org

Nick-Hall:
> I've not fully updated the po files yet. Is updating them the correct approach - can you update weblate from the po files if I do finish updating them?

Don't update the `po` files.  They are controlled by Weblate.

Is there any point in updating the `svg` files?

Ideally the Example database should contain a mix of http:// and https:// references.

Please rebase this on the _master_ branch.  If there are any broken links you can create a PR against the _gramps60_ branch.

Nick-Hall:
> The following gramps URLs no longer work. Is there a working equivalent?
http://library.gramps-project.org
http://developers.gramps-project.org

I'm not sure what the library was, but try:
https://gramps-project.org/wiki/index.php/Portal:Community
https://gramps-project.org/wiki/index.php/Portal:Developers

stevenyoungs:
> Is there any point in updating the `svg` files?

Only for the reason that it makes it easier to find the remaining http:// urls.

jralls:
> The first link is in a test file, so it may even be fictitious. I suggest you leave it - we still need to test the http:// version.

It's definitely fictitious. It's the supposed URL of "Aunt Martha's Attic", a repository that occurs in several test files and the example database.

`http://developers.gramps-project.org` appears at the bottom of several translations of the man pages. https://github.com/gramps-project/gramps/commit/bb6ce5dc641feabb3acb3cab96bd59cba744e9ed changed it to `http://www.gramps\-project.org/wiki/index.php?title=Portal:Developers` in the English man page and https://github.com/gramps-project/gramps/commit/8159aca9b49a8c1604377bd4f2adf2736dcbbd57 updated the scheme to https. The translations should obvs be made to match the English one.

Off topic to this PR: It would be worthwhile to change the man page generation to some mechanism that can use gettext so that the translations are done through Weblate.

stevenyoungs:
> `http://developers.gramps-project.org` appears at the bottom of several translations of the man pages. [bb6ce5d](https://github.com/gramps-project/gramps/commit/bb6ce5dc641feabb3acb3cab96bd59cba744e9ed) changed it to `http://www.gramps\-project.org/wiki/index.php?title=Portal:Developers` in the English man page and [8159aca](https://github.com/gramps-project/gramps/commit/8159aca9b49a8c1604377bd4f2adf2736dcbbd57) updated the scheme to https. The translations should obvs be made to match the English one.

Hopefully I've done that in this PR. I'm not a roff expert (and have no wish to be one) but given that the `-` was not escaped in several of the translated URLs I guess the man pages did not display correctly.

Nick-Hall:
The `gramps.1.in` files are generated using a build script.  I suggest that you run this after updating the `rst` files.

jralls:
> The gramps.1.in files are generated using a build script. I suggest that you run this after updating the rst files.

Something's not working right. All but one of the rst files (pt_BR.rst) has the wiki developer portal URL and all but one of the translated gramps.1.in (fr) has the incorrect `http://developers.gramps-project.org` one. The produced man pages reflect the URL in gramps.1.in, not foo.rst. I don't see any evidence after a build that gramps.1.in is touched, though thanks to jhbuild I'm using the outdated distutils `python setup.py build` formula instead of `build`. I installed gramps-5.2.4 with pip to double check the installed gramps.1.gz files for the URL.

giotodibondone:
Existing issue  [10546 :[Tip of the day] Update obsolete terms in tips.xml (including links from "http" to "https")](https://gramps-project.org/bugs/view.php?id=10546 )  pointed out by @callmedave

stevenyoungs:
> > The gramps.1.in files are generated using a build script. I suggest that you run this after updating the rst files.
> 
> Something's not working right. All but one of the rst files (pt_BR.rst) has the wiki developer portal URL and all but one of the translated gramps.1.in (fr) has the incorrect `http://developers.gramps-project.org` one. The produced man pages reflect the URL in gramps.1.in, not foo.rst. I don't see any evidence after a build that gramps.1.in is touched, though thanks to jhbuild I'm using the outdated distutils `python setup.py build` formula instead of `build`. I installed gramps-5.2.4 with pip to double check the installed gramps.1.gz files for the URL.

I can't see anywhere that `gramps/doc/makefile` is actually run from the pipeline, nor evidence of sphinx being installed.

On Windows (MSYS2) `make man` gives
`make: *** [Makefile:142: man] Segmentation fault`
😞 `make html` also gives a segmentation fault.

is `make` the right way to build the docs? There's also an `update_doc.py` file

Nick-Hall:
With a couple of small tweaks, I can generate the manual pages with:
```
python update_man.py --man
```
They contain a substitution variable for the Gramps version which is populated during the build.

With a few tweaks to `en.rst` I can generate a `gramps.1.in` file to be essentially the same as we are using now.

Experimenting a little I can generate a `pot` file with:
```
sphinx-build -b gettext . output
```
If I create a French `po` file from this, I can generate a French manual page from the `en.rst` file with:
```
sphinx-build -b man -D language=fr . output
```

Nick-Hall:
> There's also an update_doc.py file

To build the API documentation we use:
```
python update_doc.py --build
```

stevenyoungs:
If the `gramps.1.in` files are generated, should they even be in the source tree? Ideally the workflow is then enhanced to build the `gramps.1.in` files. That's not for this PR though.

For this PR is it sufficient for you if I revert the changes to the `gramps.1.in` files?
When running `python update_man.py --man` what do you set SOURCEDIR and DESTDIR to? I get the error 
```
usage: sphinx-build [OPTIONS] SOURCEDIR OUTPUTDIR [FILENAMES...]
sphinx-build: error: the following arguments are required: SOURCE_DIR, OUTPUT_DIR
```
I tried setting two env variables (with both variants of both names) but it did not help.
Cutting a long story short, if I do revert the changes to `gramps.1.in` I'll still need some help to rebuild them.

jralls:
> With a couple of small tweaks, I can generate the manual pages with…

@Nick-Hall OK, so that *could*replace the current build_man in setup.py. But...

> With a few tweaks to en.rst I can generate a gramps.1.in file to be essentially the same as we are using now.
…
>Experimenting a little I can generate a pot file
…
> If I create a French po file from this, I can generate a French manual page from the en.rst file

Must they be a separate potfile and po-files, or can they be incorporated into the existing gramps.pot and po files? Regardless, that sounds like a change for 6.0.x since we're already into string freeze for 6.0. For 6.0 ISTM we're stuck with the existing gramps.1.in and setup.py's `build_man`, so we should merge the URL fixes in this PR.

stevenyoungs:
rebased

PR #1981, opened 2025-02-16
Strip Addon Project URLs

stevenyoungs:
Leading and trailing whitespaces in the URL causes problems as discussed [here](https://gramps.discourse.group/t/leading-trailing-white-space-trimming-in-urls-for-addon-manager/7024)

Call strip on any new URLs enter by the user.
For additional safety, call strip when loading already configured projects.

### DIFF
--- a/gramps/gui/plug/_windows.py
+++ b/gramps/gui/plug/_windows.py
@@ -433,7 +433,7 @@ class AddonManager(ManagedWindow):
         self.projects = config.get("behavior.addons-projects")
         # strip the incoming URLs to avoid any leading or trailing whitespace
         for project in self.projects:
-            project[1] = project[1].strip() 
+            project[1] = project[1].strip()
         self.project_combo = Gtk.ComboBoxText()
         self.project_combo.set_entry_text_column(0)
         self.project_combo.connect("changed", self.__combo_changed)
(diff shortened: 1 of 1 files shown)


gramps/gui/plug/_windows.py
```diff
@@ -431,6 +431,9 @@ def __init__(self, dbstate, uistate, track):
         hbox.pack_start(self.addon_combo, False, False, 0)
 
         self.projects = config.get("behavior.addons-projects")
+        # strip the incoming URLs to avoid any leading or trailing whitespace
+        for project in self.projects:
+            project[1] = project[1].strip()
```
Nick-Hall, line 436: I would prefer to ensure that urls were stripped of whitespace before storing in the preferences.

stevenyoungs: With the current approach all URLs are stripped on the way in to gramps (either from the preference - this code, or when a new URL is entered by the user - the other change in this PR). Therefore I don't think there is any need to strip when storing in the preferences. As a by product the URLs are whitespace free during the lifetime of the addon manager dialog.
Stripping URLs when reading the preference covers the scenario of an upgraded 5.2 preference file already containing whitespace on the URLs.

Nick-Hall:
Closed in favour of PR #1991.

PR #1988, opened 2025-02-18
Fix man page updates

Nick-Hall:
* Fix `update_man.py` script.
* Change man page output from `gramps.1` to `gramps.1.in`.
* Update `en.rst` to generate latest man page.

This will generate an English man page that is very similar to the existing page.


Nick-Hall:
Yes.  The plan will be to add the manual page translations to the _gramps60_ branch at some point so that they can be translated ready for the updated code in the _master_ branch.  We will need to think about the timing of the changes.

PR #1992, opened 2025-02-19
Fix upgrades from bsddb

prculley:
My take on how to do upgrades.

Plan is to make bsddb upgrades go through two steps.
  1) create an sqlite db containing nothing but blobs and then copy data into it.  Which is different than previous PRs.  This makes the resulting db look like gramps v5.x dbs.
  2) this next step does upgrades just like from any more sqlite db with blobs.

This mostly worked, but found a few other issues along the way.

I also wanted to allow older 5.2.x (at least) dbs to be able to open a current db well enough to report a downgrade error.  This required keeping the metadata blobs in the db, and making sure that they were initialized well enough to get the version with the older code.

I've tested with dbs from several older versions of Gramps, an made a PR that can repeat these tests which I will post shortly.  I've also tested less extensively the downgrade, by trying to read a Gramps 6.0 db with 5.2.4.  This also worked, no matter the source of the db, upgrade or new.

As usual there may be cases I missed, so more testing is appreciated.

### DIFF
--- a/gramps/gen/db/upgrade.py
+++ b/gramps/gen/db/upgrade.py
@@ -78,6 +78,20 @@ def gramps_upgrade_21(self):
 
     self._txn_begin()
     self.upgrade_table_for_json_data("metadata")
+    # the *map_keys are not in some older versions of the db so we will initialize
+    # them here; needed in blobs if someone tries to downgrade the db so we don't
+    # crash before the warning is generated.
+    self.set_serializer("blob")
+    self._set_metadata("cmap_index", self.cmap_index, use_txn=False)
+    self._set_metadata("smap_index", self.smap_index, use_txn=False)
+    self._set_metadata("emap_index", self.emap_index, use_txn=False)
+    self._set_metadata("pmap_index", self.pmap_index, use_txn=False)
+    self._set_metadata("fmap_index", self.fmap_index, use_txn=False)
+    self._set_metadata("lmap_index", self.lmap_index, use_txn=False)
+    self._set_metadata("omap_index", self.omap_index, use_txn=False)
+    self._set_metadata("rmap_index", self.rmap_index, use_txn=False)
+    self._set_metadata("nmap_index", self.nmap_index, use_txn=False)
+
     keys = self._get_metadata_keys()
     for key in keys:
         self.set_serializer("blob")
@@ -107,7 +121,10 @@ def gramps_upgrade_21(self):
             # has "handle", "_class", and uses json.dumps()
             self._commit_raw(json_data, key)
             self.update()
+        self._drop_column(table_name, "blob_data")
 
+    self.set_serializer("blob")
+    self._set_metadata("version", 21, use_txn=False)  # keep blob version up to date
     self.set_serializer("json")
     self._set_metadata("version", 21, use_txn=False)
     self._txn_commit()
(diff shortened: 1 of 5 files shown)


gramps/gen/db/upgrade.py
```diff
@@ -107,7 +121,14 @@ def gramps_upgrade_21(self):
             # has "handle", "_class", and uses json.dumps()
             self._commit_raw(json_data, key)
             self.update()
+        # DROP COLUMN is available with Sqlite v 3.35.0, released 2021-03-12
+        summary = self.get_summary()
+        db_ver = summary["Database version"].split(".")
+        if int(db_ver[0]) == 3 and int(db_ver[1]) >= 35:
+            self.dbapi.execute("ALTER TABLE %s DROP COLUMN blob_data;" % table_name)
```
dsblank, line 128: We should turn that into a DB API method (need to create) so that we can have versions for mongodb, postgresql, or whatever.

```suggestion
            self.dbapi.drop_column(table_name, "blob_data")
```

dsblank (commented):
This is looking pretty good! Thanks for taking this on.

gramps/plugins/db/bsddb/bsddb.py
```diff
@@ -181,8 +187,11 @@ def load(
                             if len(fmat) == 3:
                                 fmat = fmat + (True,)
                                 data[format_ix] = fmat
+                        else:
+                            data = []
                     elif key == b"gender_stats":
                         # data is a dict, containing entries (see GenderStats)
+                        self._txn_begin()
                         self.dbapi.execute("DELETE FROM gender_stats")
```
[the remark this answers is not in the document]
Nick-Hall: I don't think we need achange here since the `DbBsddb` upgrade database is a sub-class of `SQLite`.

gramps/gen/db/upgrade.py
Nick-Hall: We can't assume that we are upgrading a DBAPI database here.  The `self.dbapi` variable may not exist.

Perhaps we should keep the `gen/db/upgrade` module for object format upgrades and make backend specific changes elsewhere.

prculley:
I shifted the drop column code, hopefully a better separation of modules.  Although I really have to wonder if anyone will ever try to upgrade older dbs to a non-sqlite db...

I also found and fixed another downgrade issue.  Apparently I missed the brand new db to older version in my testing the first time around.

Nick-Hall:
Rebased.

dsblank:
Thanks @prculley!

@DavidMStraub, another method to add to postgresql if you want: https://github.com/gramps-project/gramps/pull/1992/files#diff-4e69d41ba92f15e611c6ba40ac431f5ca0e134551c8aa163bd2ece432f5c156aR255

Nick-Hall:
@prculley Bug [#13674](https://gramps-project.org/bugs/view.php?id=13674) has been reported in 6.0.0-rc1 related to database upgrades.  There is also a related ["updating schema unexpected errors"](https://gramps.discourse.group/t/6-0rc1-updating-schema-unexpected-errors/7151/2) topic on Discourse.

emyoulation:
@Nick-Hall 
2 of the files updated were plug-ins. Do the related `.gpr.py` registrations need version and maintainer updates too? And are updates needed in the copyright date and credits for both the `.py` and `.gpr.py`?

PR #1993, opened 2025-02-19
Add database upgrade unit tests

prculley:
This adds some tests of the db upgrade process.  I have added several small dbs created with older versions of Gramps.
The imports_test code was updated to allow zips of these dbs as inputs to the test.  I chose to modify imports_test as there was substantial code reuse if I created a new module.

As usual the test compares the extracted and loaded db with the matching Gramps XML file and reports any differences.  This should fail if the upgrade code crashes or misses something important.

I've tested this on top of [PR#1992](https://github.com/gramps-project/gramps/pull/1992) and it seems to work.


emyoulation:
>The imports_test code was updated to allow zips of these dbs as inputs to the test.

Nice! I assume that this accepts the backups automatically generated during a Convert schema?

prculley:
It does not accept zips created during upgrade directly.  You have to massage the names of the zip and the db directory inside the zip to says something like "imp_???????" in order for the test files to be recognized.  And of course, you need a .gramps file with the same name to compare the db with.

Nick-Hall:
Rebased.

Nick-Hall:
@prculley The v4.1.3 unit test fails.

prculley:
I changed the test to report additional debug information.
I got an error (bsddb3.db.DBInvalidArgError: (22, 'Invalid argument -- BDB1009 /home/runner/.gramps/temp/imp_4.1.3/name_group.db: unsupported btree version: 10')

These tests run fine on my system where I am running 
"Berkeley Database library (bsddb3: 6.0.30) (Python-bsddb3 : 6.2.9)"

Do we know what bsddb versions are run during the unit test?

Nick-Hall:
> Do we know what bsddb versions are run during the unit test?

The unit tests are running with bsddb v5.3.28 and python3-bsddb3 v6.2.9 on Unubtu 22.04LTS (Jammy).

prculley:
I have to admit I am confused by this bsd compatibility stuff.  The Gramps 4.1.3 db was created with the AIO for that version, which reports its BSDDB: 6.1.0 (6, 0, 30).  It looks to me like all the AIOs Josip (and I) created used the same 6,0,30 base db.  So I don't understand why some of the test dbs in this PR worked, and some did not.  If they had all failed, I would have attributed it to trying to open a Berkeley db v6.x with a v5.x software, but that doesn't seem to be the case.

I also spotted an early version of Gramps using a DB_BTREE instead of DB_HASH and added code to deal with this, again did not fix the main issue.

Unless someone has another suggestion on how to proceed, I will simply remove the 4.1.3 test db from the test set.

Nick-Hall:
My v4.1.3 database created using BSDDB 6.2.32 (6.2.9) gives exactly the same error in the unit test.  However, it does import into v6.0 on my machine.

prculley:
So I created a vm with ubuntu that has the same bsd versions as the unit test env.  It also fails the same way with the same messages.

I created another test set with the GrampsAIO 4.2.8 to see if the problem was specific to a particular 4.1.3 file and it also fails.
I now believe that the github test env with its bsddb v5.3.28 will simply not open any bsd database created with bsddb v6.x.  From the older bsd, the db is just not recognizable.

I also went back and looked at the gramps 4.x code, and found a test for this situation. https://github.com/tran62/gramps/commit/32b5905aea508d051a4baa466cb76aa500a61fd1#diff-b8c4c99637de2ff0d10955c8c8237608197ed09a565cfa8094dea56c1cf39e5aR353

So now the question is what to do about this?  Opinions wanted...

1. Ignore errors of this type in unit test
2. remove tests created with db6.x
3. try to add tests for older gramps, but made on systems with 5.x or older bsd
4. check for and raise error like we used to for v4.x in upgrade process (better for user experience)
5. get github test env upgraded with db6.x, so all tests will pass

dsblank (approved):
LGTM!

PR #1995, opened 2025-02-23
Fix bug with surname rules in Quick Views

Nick-Hall:
Fixes [#13657](https://gramps-project.org/bugs/view.php?id=13657).


dsblank (approved):
LGTM! Thanks!

PR #1997, opened 2025-02-26
Always return str from to_string methods

DavidMStraub:
This PR changes the JSON serialization methods such that all `to_string` methods actually return strings and not bytes. While it turns out that Python SQLite3 converts the byte strings correctly to strings before storing them in the JSON column, pscopg2 for instance doesn't do that. This breaks compatibility of Gramps 6.0 with PostgreSQL, unless awkward workaraounds are introduced.

Related: https://github.com/gramps-project/addons-source/pull/639#issuecomment-2686053136

### DIFF
--- a/gramps/gen/lib/json_utils.py
+++ b/gramps/gen/lib/json_utils.py
@@ -161,11 +161,11 @@ def string_to_dict(string: str | bytes) -> dict:
     return orjson.loads(string)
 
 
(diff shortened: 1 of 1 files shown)


dsblank (approved):
This seems like a fine change to make. Thanks for this!

Nick-Hall:
Rebased onto the _gramps60_ branch.

PR #1998, opened 2025-02-28
Allow invalid dates in the relationship graph.

SNoiraud:
Fixes #13672

from https://gramps.discourse.group/t/question-mark-in-dates/7121


Nick-Hall:
Rebased.

PR #2001, opened 2025-03-02
New calendar

SNoiraud:
Using the calendar for old dates is very tedious.
Going back to 1625 can take some time.

I added the possibility to select directly the person's birth, the person's death, the family marriage or the event date when you change the active object.

I'll show some screenshot for that.


SNoiraud:
calendar birth 1

calendar death 1

calendar death 2

calendar family 1

calendar family 2

calendar event 1

Nick-Hall:
@SNoiraud You need to upgrade your copy black to version 25.1.0 to avoid the linting errors.

SNoiraud:
Upgrading black make nothing. All my files are correct, but if I look at in the log the problem is in:

 import unittest
 import os
would reformat /home/runner/work/gramps/gramps/po/test/po_test.py

Oh no! 💥 💔 💥
59 files would be reformatted, 1132 files would be left unchanged.

emyoulation:
Serge, 

Is it possible to add a similar expansion in the Date Selection dialog? Or if this PR just has variants of the Calendar gramplet rather than tabs, then your experience in this PR may not be applicable. (Either way such a feature will need a new enhancement request.) I just want you opinion about whether it is a reasonable request.

Dual tabbed Calendars ("Date" and  "Second date") that poll and push dates from the associated fields would be a huge improvement. (I often find newspaper article that refer to "last Thursday" or "next Tuesday" where the curent day of the week was not recorded.)

126e2bd68cea43d1e4902ccaee8d0deb28a4f7c1

SNoiraud:
> I just want you opinion about whether it is a reasonable request.

It seems reasonable.
I have however a little problem with calendars that are not Gregorian (French revolutionary, ...)
The Gtk calendar only shows positives gregorian dates.

I can look at this and will post a snapshot example.

second problem: Glade will not exist in Gtk4. As I don't know what will be the future, I'all add these widget into the table48 grid.

Here is an example of what we could have:
date-editor 1

SNoiraud:
I am not sure Cambalache will be the glade replacement.

PR #2007, opened 2025-03-03
Fix schema upgrade errors

prculley:
Version metadata was prematurely stored, preventing subsequent upgrade steps from working.  Db ended up with blobs instead of json.

This is what happens when I just make sure db upgrades and opens, but never switch to an interesting view.

Fixes [#13674](https://gramps-project.org/bugs/view.php?id=13674).


emyoulation:
Wonderful! The errors during conversion are resolved.  Thank you.

emyoulation:
The 22mb BSDDB database conversion (which had a much more extensive error list with rc1-1 ) is also resolved with this change

dsblank (approved):
Nice catch!

PR #2010, opened 2025-03-08
Fix IsMoreThanNthGenerationDescendantOf.init_list() typing

stevenyoungs:
The `person` argument of `IsMoreThanNthGenerationDescendantOf.init_list()` was typed to expect a `Person`
From  `IsMoreThanNthGenerationDescendantOf.prepare()` a parameter of type `DataDict | None` was passed.
From within `IsMoreThanNthGenerationDescendantOf.init_list` a recursive call is made passing a `Person` object.

Unify so that the parameter always has the type `DataDict | None`

### DIFF
--- a/gramps/gen/filters/rules/person/_ismorethannthgenerationdescendantof.py
+++ b/gramps/gen/filters/rules/person/_ismorethannthgenerationdescendantof.py
@@ -33,7 +33,7 @@ _ = glocale.translation.gettext
 #
 # -------------------------------------------------------------------------
 from .. import Rule
-
+from ....lib.json_utils import DataDict
 
 # -------------------------------------------------------------------------
 #
@@ -77,7 +77,7 @@ class IsMoreThanNthGenerationDescendantOf(Rule):
     def apply_to_one(self, db: Database, person: Person) -> bool:
         return person.handle in self.selected_handles
 
-    def init_list(self, person: Person, gen: int) -> None:
+    def init_list(self, person: DataDict | None, gen: int) -> None:
         if not person:
             return
         if gen >= int(self.list[1]):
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Most of the filter `init_list` methods seem to take a primary object rather than a `DataDict`.  Have look at the [init_list](https://github.com/gramps-project/gramps/blob/maintenance/gramps60/gramps/gen/filters/rules/person/_islessthannthgenerationdescendantof.py#L81) method in the _IsLessThanNthGenerationDescendantOf_ rule.

I think that we should leave this until v6.1 unless there is a real bug associated with it.

stevenyoungs:
> Most of the filter `init_list` methods seem to take a primary object rather than a `DataDict`. Have look at the [init_list](https://github.com/gramps-project/gramps/blob/maintenance/gramps60/gramps/gen/filters/rules/person/_islessthannthgenerationdescendantof.py#L81) method in the _IsLessThanNthGenerationDescendantOf_ rule.

I chose to standardise on `DataDict` as it will be lightly faster. 
I think `init_list` is an internal method so we do not need to be consistent across the different classes. In this class a `DataDict` is sufficient. In other filters, the true object may be required.
There's a general decision to be made here. A `DataDict` and the true object are designed to be interchangeable. As typing is introduced we need guidance on what pattern to follow in situations where a `DataDict` provides sufficient functionality:
* `person: Person`
* `person: DataDict`
* `person: Person | DataDict`

The last option offers the most flexibility, allowing the calling code to provide whichever type is most convenient, at the expense of a little more typing when defining the method.

> I think that we should leave this until v6.1 unless there is a real bug associated with it.

Agreed. So long as we do not make any other code change that exposes the lint error in 6.0, no need to fix in 6.0.

PR #2012, opened 2025-03-11
Fix nested transaction error in the Test Case Generator tool

Nick-Hall:
Fixes [#13680](https://gramps-project.org/bugs/view.php?id=13680).


dsblank (approved):
Nice cleanup!

PR #2013, opened 2025-03-12
Simplify logic for setting data, config and cache directories

Nick-Hall:
This PR is illustrates a simplification of the logic for setting the data, config and cache directories.

* The old `$HOME/.gramps` location for Linux is removed in favour of the newer XDG specification.
* For Windows `$APPDATA` is used with a fallback to `$USERPROFILE/AppData/Roaming`.

The old locations can still be used by setting the `$GRAMPSHOME` variable.

If we decide to adopt this approach then we should consider migrating existing data.

Fixes [#13261](https://gramps-project.org/bugs/view.php?id=13261), [#13300](https://gramps-project.org/bugs/view.php?id=13300), [#13686](https://gramps-project.org/bugs/view.php?id=13686).

### DIFF
--- a/gramps/gen/const.py
+++ b/gramps/gen/const.py
@@ -90,47 +90,33 @@ APP_VCARD = ["text/x-vcard", "text/x-vcalendar"]
 
 # -------------------------------------------------------------------------
 #
-# Determine the home directory. According to Wikipedia, most UNIX like
-# systems use HOME. I'm assuming that this would apply to OS X as well.
-# Windows apparently uses USERPROFILE
+# Determine the user data and user configuration directories.
 #
 # -------------------------------------------------------------------------
 if "GRAMPSHOME" in os.environ:
     USER_HOME = get_env_var("GRAMPSHOME")
-    HOME_DIR = os.path.join(USER_HOME, "gramps")
+    USER_DATA = os.path.join(USER_HOME, "gramps")
+    USER_CONFIG = USER_DATA
 elif "USERPROFILE" in os.environ:
     USER_HOME = get_env_var("USERPROFILE")
-    HOME_DIR = os.path.join(GLib.get_user_config_dir(), "gramps")
-    if not os.path.exists(HOME_DIR):
-        homedir = get_env_var("APPDATA") if "APPDATA" in os.environ else USER_HOME
-        homedir = os.path.join(homedir, "gramps")
-        dbpath = os.path.join(homedir, "grampsdb")
-        if os.path.isdir(dbpath) and os.listdir(dbpath):
-            HOME_DIR = homedir
+    if "APPDATA" in os.environ:
+        USER_DATA = os.path.join(get_env_var("APPDATA"), "gramps")
+    else:
+        USER_DATA = os.path.join(USER_HOME, "AppData", "Roaming", "gramps")
+    USER_CONFIG = USER_DATA
 else:
     USER_HOME = get_env_var("HOME")
-    HOME_DIR = os.path.join(USER_HOME, ".gramps")
-ORIG_HOME_DIR = HOME_DIR
-if "SAFEMODE" in os.environ:
-    if "USERPROFILE" in os.environ:
-        USER_HOME = get_env_var("USERPROFILE")
-    else:
-        USER_HOME = get_env_var("HOME")
-    HOME_DIR = get_env_var("SAFEMODE")
-
-
-if os.path.exists(HOME_DIR) or "GRAMPSHOME" in os.environ or "SAFEMODE" in os.environ:
-    USER_DATA = HOME_DIR
-    USER_CONFIG = HOME_DIR
-    USER_CACHE = HOME_DIR
-else:
     USER_DATA = os.path.join(GLib.get_user_data_dir(), "gramps")
     USER_CONFIG = os.path.join(GLib.get_user_config_dir(), "gramps")
-    USER_CACHE = os.path.join(GLib.get_user_cache_dir(), "gramps")
+
+USER_CACHE = os.path.join(GLib.get_user_cache_dir(), "gramps")
+
+if "SAFEMODE" in os.environ:
+    USER_CONFIG = get_env_var("SAFEMODE")
 
 USER_PICTURES = GLib.get_user_special_dir(GLib.UserDirectory.DIRECTORY_PICTURES)
 if not USER_PICTURES:
-    USER_PICTURES = HOME_DIR
+    USER_PICTURES = USER_DATA
 
 VERSION_DIR_NAME = "gramps%s%s" % (VERSION_TUPLE[0], VERSION_TUPLE[1])
 VERSION_DIR = os.path.join(USER_CONFIG, VERSION_DIR_NAME)
--- a/gramps/gen/const.py
+++ b/gramps/gen/const.py
@@ -34,6 +34,10 @@ import os
 import random
 import sys
 import uuid
+import shutil
+import logging
+
+LOG = logging.getLogger(".")
 
 from gi.repository import GLib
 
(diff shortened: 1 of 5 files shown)


emyoulation:
> - For Windows $APPDATA is used with a fallback to $USERPROFILE/AppData/Roaming.

Please double-check this.  At the same time Gramps switched to XDG, the new installs changed to `$LOCALAPPDATA` (aka `%LocalAppData%` ) 
The path for upgrades remained at `$APPDATA`  (aka %AppData% )

Frankly, the 2 possible locations is making troubleshooting a Windows issue much, much harder. 

Is is reasonable to put in a 6.1 enhancement request to move the `gramps` folder contents in the old location to the new one during 6.1 installation?

Nick-Hall:
> Frankly, the 2 possible locations is making troubleshooting a Windows issue much, much harder.

I agree.  That's why I would like to use a single location for Windows - `AppData/Roaming` or `AppData/Local` - I don't mind which.  It would also be simpler if we removed the old `~/.gramps` location for Linux.

emyoulation:
There were major arguments about the 2 locations when the shift was implemented. I was against changing in the middle of the 5.x family.  It seemed like a 6.x change.  

But everyone else insisted that the `%LocalAppData%` was the only possible option.  So let's not fight that battle again and go with `%LocalAppData%`

I only have an uneducated opinion on Linux file locations.  So I will not offer any Linux preference.

Nick-Hall:
> So let's not fight that battle again and go with %LocalAppData%

I have no objection to using `%LocalAppData%`.  However it seems sensible to use `%AppData%` if set.  I'm not a Windows user though, so I'm looking for feedback.

emyoulation:
I'd be (and was!) for `%AppData%` if this was before the XDG switchover. (It seemed like the benefit of switching to `%LocalAppData%` was not worth the training and support costs.)

But now that the switch is started, aborting and reverting back to `%AppData%` would be saying there was bad planning. I'm distinctly opposed to making that statement when there was so much hard-fought consensus.

jralls:
> But now that the switch is started, aborting and reverting back to %AppData% would be saying there was bad planning. I'm distinctly opposed to making that statement when there was so much hard-fought consensus. (Even when I was on the out-voted side.)

If [this](https://gramps.discourse.group/t/aio-move-of-user-directory-from-appdata-to-localappdata/4156) is the discussion you're referring to it doesn't look to me like the consensus was that it wouldn't make much practical difference between using LocalAppData and Appdata and that grampsdb really belongs in Documents. That discussion happened *after* #1368 was merged so it was more like no planning rather than bad planning.

We also have almost 20 years of using Roaming and only a little over 2 of using Local. A lot more users have there data in Roaming; only those who started using Gramps with 5.2 will have data in Local. 

In order to not freak out those users whose files are in Local we need to look there for their previous config and data. I propose that if we find it we simply move (rename on Windows) it to Roaming. Gramps 5.2 will find it there and be happy should the user go back and forth for comparison.

emyoulation:
You also have the possibility that Gramps exists in both locations. So a simple rename may not always work.

Whatever is decided, so long as it goes back to a single possible path for the Gramps User Directory, I think everyone will be happier. (And the wiki will be simplified. And the Discourse troubleshooting instructions will be less convoluted.)

jralls:
> You also have the possibility that Gramps exists in both locations. So a simple rename may not always work.

That's [#13300](https://gramps-project.org/bugs/view.php?id=13300). Arguably rare enough that we could just bail out and tell the user to sort it out and then restart Gramps.

Or a middle route: If Local\gramps\gramps52 exists and Roaming\gramps\gramps52 doesn't and there aren't any common directory names in the two grampsdb directories then we can go ahead and do the move.

emyoulation:
Has anyone besides me suddenly started getting odd GitHub thread comment notifications? 

Where they have duplicated the content... 1st with a standard email header then with a thread  comment notication header:


This started recently.

Who do we contact at GitHub? It doubles their message sizes and archiving burden.

emyoulation:
Thanks. That means the text/html redundancy is probably another of Yahoo's behind-the-scenes changes to their web mail cleint. 

I still haven't explored to determine when and where Yahoo decides to strip linefeeds/carriage returns on outbound messages through the maillists. So I double them between paragraphs just to be safe. Sometimes they consolidate, sometimes they don't.

gramps/gen/const.py
```diff
@@ -104,10 +105,20 @@
     else:
         USER_DATA = os.path.join(USER_HOME, "AppData", "Roaming", "gramps")
     USER_CONFIG = USER_DATA
+    # Migrate data from AppData\Local to AppData\Roaming on Windows.
+    OLD_HOME = os.path.join(USER_HOME, "AppData", "Local", "gramps")
+    if os.path.exists(OLD_HOME) and not os.path.exists(USER_DATA):
+        shutil.move(OLD_HOME, USER_DATA)
```
jralls, line 111: We should at least log a warning if both exist.

gramps/gen/const.py
```diff
@@ -104,10 +105,20 @@
...
 else:
     USER_HOME = get_env_var("HOME")
     USER_DATA = os.path.join(GLib.get_user_data_dir(), "gramps")
     USER_CONFIG = os.path.join(GLib.get_user_config_dir(), "gramps")
+    # Copy database directory into the XDG data directory.
+    OLD_HOME = os.path.join(USER_HOME, ".gramps")
+    if os.path.exists(OLD_HOME) and not os.path.exists(USER_DATA):
+        shutil.copytree(
+            os.path.join(OLD_HOME, "grampsdb"), os.path.join(USER_DATA, "grampsdb")
+        )
```
jralls, line 121: What about USER_CONFIG?

Nick-Hall: We probably only need the `gramps.ini` file.

I also need to check that the `grampsdb` directory exists before attempting to copy it.

jralls: > We probably only need the gramps.ini file.

I think we just need for [config.py](https://github.com/gramps-project/gramps/blob/master/gramps/gen/config.py#L410) to check the old location.

Nick-Hall: I'll look into it.  Using existing code and the same upgrade path would be a better option.

gramps/gen/const.py
Nick-Hall: OK. I've added a warning.

jralls (approved):

gramps/gen/const.py
Nick-Hall: I've added a check for the old path to the config update code.

Nick-Hall:
@emyoulation Please test that this works on Windows.

emyoulation:
> @emyoulation Please test that this works on Windows.

OK but what tests do you want? <mark>Have you built a new AIOinstaller for me to test? I have not built from source on Windows.</mark>

I am resetting the PC now.  

It seems like a full test would be:
  1) install Gramps 4.28 and import example.gramps into BSDDB in %AppData% 
          (make a backup of the  %AppData%/gramps folder to desktop)
  2) upgrade to Gramps 5.2.4 and import example.gramps into SQLite in %AppData% (copy to backup appdata folder)         
  3) run AIO 6.0 installer candidate  
        see what happened to %AppData% and %LocalAppData%

  4) uninstall all versions of gramps. Strip any `gramps` folder from %AppData% and %LocalAppData%
  5) install 5.2.4 and import example.gramps into SQLite in %LocalAppData% 
          (make a backup of the  %LocalAppData%/gramps folder to desktop)
  6) run AIO 6.0 installer candidate  
        see what happened to %AppData% and %LocalAppData%

  7) uninstall 6.0,
  8) copy the descktop backups of AppData/gramps and LocalAppData/gramps to %AppData% and %LocalAppData% 
  9) run AIO 6.0 installer candidate  
        see what happened to %AppData% and %LocalAppData%

emyoulation:
OK. I've generated fresh sample  `AppData/Roaming/gramps`  and  `AppData/Local/gramps`  directories  with 4.2.8 and 5.2.4 and can test various 6.0 installer scenarios

(For all installs,`enchant` and `gtk-3.0` folders were created in `AppData/Local` )

Nick-Hall:
One option is to merge this and include it in 6.0.0-rc2, but do we wan to test it further first?

emyoulation:
Then I do not understand I would test this. Isn't the `workflows/gramps-ci.yml` either part of the installer executable ? or maybe the build of the installer?  

The other 4 look like they should be easily added to an instance but when are those files called? Does it not need a full installer build to test?

jralls:
@emyoulation It's pure python. Just install the AIO to somewhere that isn't C:\Program Files\ and apply the patch.

emyoulation:
I do not have a development system or full Python environment.  Just what the Gramps AIO installs. 

Doug had questions about what the exe included when we failed to make the AI chat addon functional or pip to install the prerequisites. His comments implied that the undled environment was less than he expected. Or existed in a different configuration.

Since I have little experience with Python outside of the Python Eval and SuperTool gramplets,  maybe "it is pure python" doesn't communicate as much info to me as it does to you.

dsblank:
@emyoulation, ok this might be beyond you capabilities. No problem!

> I do not have a development system or full Python environment. Just what the Gramps AIO installs.

You only need to replace the Python code in this PR. But I don't think it is straightforward for you.

emyoulation:
It seems to be moving folders correctly and complaining appropriately when there is a conflict.

A question.  Are the Enchant dictionaries still supposed be installed in the windows %LocalAppData% even though the gramps folder has reverted to %AppData% (roaming) folder?

PR #2017, opened 2025-03-14
Make `gramps.plugins.db.dbapi.dbapi` typed

bartfeenstra:
A couple of arguments and return values remain untyped, because I'm not familiar enough with this part of Gramps just yet.

Additionally, as a learning note: in the future we should move from typing on `dict` and `list` to `Mapping`, `MutableMapping`, `Sequence`, and `MutableSequence` as they allow other, compatible types to be used. I did not do that now as all these methods are documented to use lists or dicts explicitly.


bartfeenstra:
oops, duplicate of https://github.com/gramps-project/gramps/pull/1919

PR #2019, opened 2025-03-14
Add mypy configuration

bartfeenstra:
This adds centralized mypy configuration so one just has to run `mypy` to get the same results as the CI builds. Instead of blindly ignoring any untyped imports, this explicitly ignores those that are known to be missing, so future errors are not suppressed silently.

This file also serves as a starting point to gradually make the mypy rules stricter, for instance per file or directory (so the codebase can be converted gradually).

As to the `@todo`s: we'd ideally install a few more Python packages (containing typing stubs for third part packages) that are not available through APT. Can we instead manage dependencies through a single, centralized source, and install them with pip?

### DIFF
--- a/.github/workflows/gramps-ci.yml
+++ b/.github/workflows/gramps-ci.yml
@@ -70,7 +70,7 @@ jobs:
     - name: Run mypy static type checker
       run: |
         python3 -m pip install mypy
-        mypy gramps test --ignore-missing-imports --exclude '.*gpr\.py$'
+        mypy
     - name: Run unit test suite
       run: |
         export GDK_BACKEND=-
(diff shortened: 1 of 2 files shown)


DavidMStraub (approved):

bartfeenstra:
That test failure looks like a fluke. Can someone trigger a re-run?

bartfeenstra:
I added two more ignore rules (cairo and orjson) and I removed the to-dos. I think we should tackle the easy installation of dev dependencies in a follow-up, and then leverage that to improve mypy coverage by installing the types packages through pip.

dsblank:
Nice!

PR #2024, opened 2025-03-18
Allow Gramps to continue if an addon rule fails to load

Nick-Hall:


prculley:
Might be helpful to developers if we also spit out a traceback to the console when an addon fails to load.

PR #2027, opened 2025-03-20
Update the INSTALL file

Nick-Hall:
* Change install from `setup.py` to `pip`.
* Update typical installation locations.
* Remove the `--resourcepath` option which no longer exists.

Issue [#13717](https://gramps-project.org/bugs/view.php?id=13717).

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -37,10 +37,10 @@ To install all needed dependencies you can use (in Ubuntu):
 
 To build and install, whether from a tarball or git repo:
    python3 setup.py build
-   sudo python3 setup.py install
+   sudo pip3 install .
 
 You can avoid using sudo for the install step by specifying a prefix to which you have write privilege. The default is /usr/local, which is usually owned by root. You can learn of more options with
-   python3 setup.py --help
+   pip3 install --help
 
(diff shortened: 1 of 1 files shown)


giotodibondone:
Wasn't this meant to be done for Gramps 6.1.x previous releases page mentions Python 3.10+ only while Gramps 6.0.x supports Python 3.9+ only?  

First report:

> [Pierre](https://gramps.discourse.group/u/Pierre)
> [28m](https://gramps.discourse.group/t/distutils-is-deprecated/7477)
> 
> distutils has not been included in the Python distribution since version 3.10. However, setup.py still requires distutils to install version 6.0.1. I’m not a Python installation geek, so it’s difficult for me to work around the problem with distutils. I considered installing in a virtual environment. But this type of installation seems like a cop-out to me. Why not make the Gramps installation method PEP 517/518 compliant?
> 
> In the meantime, I’ll stick with version 6.0.0 as provided by Kubuntu.
> 
>

Nick-Hall:
All of this is relevant to v6.0.  The example installation locations are those used by Ubuntu 22.04.

PR #2028, opened 2025-03-21
Allow Multiple Selection from People, Event and Note Selectors

stevenyoungs:
Continuing on from PR #1836 which added multiple selection of media objects in the gallerytab, allow multiple selection of 
* children when editing a family
* notes in `NotesTab`
* events in `EventEmbedList`

As with the gallery tab, if multiple objects are selected, the references are added without showing the edit ref dialog.

It would be nice to permit multiple selection of citations but currently the multiple selection would permit multiple sources, or sources and citations, to be returned and it is not clear what the user experience should be in this scenario. So for now, I've left it as single selection.

### DIFF
--- a/gramps/gui/editors/displaytabs/citationembedlist.py
+++ b/gramps/gui/editors/displaytabs/citationembedlist.py
@@ -185,33 +185,35 @@ class CitationEmbedList(EmbeddedList, DbGUIElement):
         objct = sel.run()
         LOG.debug("selected object: %s" % objct)
         # the object returned should either be a Source or a Citation
-        if objct:
+        if objct is not None:
             if not isinstance(objct, (Source, Citation)):
                 raise ValueError(
-                    f"selection must be either source or citation, got {type(objct)}"
+                    _("Selection must be either a source or a citation, got %s")
+                    % type(objct)
                 )
-            source = objct if isinstance(objct, Source) else None
-            citation = objct if isinstance(objct, Citation) else Citation()
-            try:
-                from .. import EditCitation
+            if isinstance(objct, Citation):
+                self.add_callback(objct.handle)
+            else:  # objct is a Source
+                try:
+                    from .. import EditCitation
 
-                EditCitation(
-                    self.dbstate,
-                    self.uistate,
-                    self.track,
-                    citation,
-                    source,
-                    callback=self.add_callback,
-                    callertitle=self.callertitle,
-                )
-            except WindowActiveError:
-                from ...dialog import WarningDialog
+                    EditCitation(
+                        self.dbstate,
+                        self.uistate,
+                        self.track,
+                        Citation(),
+                        objct,
+                        callback=self.add_callback,
+                        callertitle=self.callertitle,
+                    )
+                except WindowActiveError:
+                    from ...dialog import WarningDialog
 
-                WarningDialog(
-                    _("Cannot share this reference"),
-                    self.__blocked_text(),
-                    parent=self.uistate.window,
-                )
+                    WarningDialog(
+                        _("Cannot share this reference"),
+                        self.__blocked_text(),
+                        parent=self.uistate.window,
+                    )
 
     def __blocked_text(self):
         """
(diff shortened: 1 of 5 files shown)


dsblank (changes requested):
## Review: Allow Multiple Selection from People, Event and Note Selectors

The PR builds on #1836 (gallery multi-select) and adds multi-select support to children, events, and notes. The approach is generally sound. I have two concrete bugs and a few smaller concerns.

---

### Bug 1 — `editfamily.py`: Scroll-to-cell calculation is wrong in a loop

**Lines 313–318:**
```python
for person in people:
    ref = ChildRef()
    ref.ref = person.get_handle()
    self.family.add_child_ref(ref)
    self.rebuild()
    GLib.idle_add(
        self.tree.scroll_to_cell,
        len(self.family.get_child_ref_list()) - len(people),  # BUG
    )
```

`get_child_ref_list()` grows by 1 each iteration, but `len(people)` is constant. For an empty family adding 3 children the queued scroll targets are `-2`, `-1`, `0`. For a family with 2 children adding 3 more the targets are `0`, `1`, `2` — landing accidentally on the right final position, but scrolling past wrong rows first.

Also, `rebuild()` inside the loop is wasteful — it redraws the tree for every person. The original single-person path (`new_child_added`) adds one ref, rebuilds once, and scrolls once. The equivalent for multiple would be:

```python
if people:
    initial_len = len(self.family.get_child_ref_list())
    refs = []
    for person in people:
        ref = ChildRef()
        ref.ref = person.get_handle()
        self.family.add_child_ref(ref)
        refs.append(ref)
    self.rebuild()
    GLib.idle_add(self.tree.scroll_to_cell, initial_len)
    if len(people) == 1:
        self.call_edit_childref(refs[0])
```

---

### Bug 2 — `citationembedlist.py`: Silent removal of the `ValueError` guard

The refactor drops the `else: raise ValueError("selection must be either source or citation")` branch. The new code handles the unexpected-type case silently:

```python
source = objct if isinstance(objct, Source) else None
citation = objct if isinstance(objct, Citation) else Citation()
# If objct is neither, source=None, citation=Citation() — EditCitation is called with a blank citation
```

If a bug elsewhere caused the wrong type to be returned, the old code would fail loudly at the call site. The new code silently calls `EditCitation(Citation(), None, ...)`, which may produce confusing behaviour or a crash further down the stack with no helpful message. The guard should be preserved:

```python
if not isinstance(objct, (Source, Citation)):
    raise ValueError(f"selection must be either source or citation, got {type(objct)}")
source = objct if isinstance(objct, Source) else None
citation = objct if isinstance(objct, Citation) else Citation()
```

---

### Minor: `eventembedlist.py` — asymmetric branching

```python
if len(events) > 1:
    for event in events:
        ...
else:
    try:
        # single-selection path (open editor)
```

The `else` branch implicitly handles `len(events) == 1`, but an `if len(events) == 1: ... else:` structure would be clearer and mirror the intent expressed in the PR description.

---

### Minor: `SelectEvent` / `SelectNote` — no explicit `__init__`

`SelectPerson` gets an explicit `__init__` that forwards `allow_multiple_selection` to `BaseSelector`. `SelectEvent` and `SelectNote` don't override `__init__`, so they inherit `BaseSelector.__init__` directly — the keyword argument works fine. Not a bug, but it is asymmetric. A comment or note explaining why some selectors have explicit `__init__` overrides and some don't would help future contributors.

---

### No tests

None of the changed behaviour has test coverage. Given that this is UI/GTK-heavy code, unit testing the selector interaction is difficult, but the `ChildEmbedList` logic (ref creation loop, skip-list construction) could benefit from a test to guard the scroll/edit-dialog logic going forward.

---

### Verdict

The `notetab.py` change is clean and correct. The citation refactor reduces duplication but drops a defensive guard. The event multi-select is functional but the branching structure is slightly awkward. The children multi-select has a concrete scroll calculation bug and a rebuild-in-loop inefficiency. Please fix the two bugs before merging.

stevenyoungs:
rebased

PR #2029, opened 2025-03-22
Fix issue with empty metadata fields on new databases in Gramps 6.0

DavidMStraub:
See the discussion in https://github.com/gramps-project/gramps-web-api/issues/631.


DavidMStraub:
Would be great to see this merged into the maintenance branch to be part of an upcoming 6.0.1. The CI of Web API  with Gramps 6 passes with this change.

Nick-Hall:
This bug is on the [roadmap](https://gramps-project.org/bugs/roadmap_page.php) for 6.0.1.

PR #2030, opened 2025-03-22
Fix error when importing a GEDCOM file into an existing tree

Nick-Hall:
Fixes [#13726](https://gramps-project.org/bugs/view.php?id=13726).


dsblank:
Nice!

dsblank (approved):
LGTM!

PR #2032, opened 2025-03-24
Prevent Unhandled Exception in GuiPersonOption

stevenyoungs:
This PR has been heavily revised to restore the previous behaviour, See https://github.com/gramps-project/gramps/pull/2032#issuecomment-2770646907 for further information

**Initial PR description**
This is a quick fix to avoid gramps crashing. As such it simply removes the filter parameter from the call to SelectPerson. The SelectPerson dialog therefore shows all people in the DB. Previously it defaulted to showing the following people
- bookmarked
- home
- active

A longer term solution is still required to restore the previous people shown by default.

Fixes [#13720](https://gramps-project.org/bugs/view.php?id=13720).

### DIFF
--- a/gramps/gui/plug/_guioptions.py
+++ b/gramps/gui/plug/_guioptions.py
@@ -695,6 +695,7 @@ class GuiPersonOption(Gtk.Box):
         """
         # Create a filter for the person selector.
         rfilter = GenericFilter()
+        rfilter.set_name(_("Active, bookmarked or home people"))
         rfilter.set_logical_op("or")
         rfilter.add_rule(rules.person.IsBookmarked([]))
         rfilter.add_rule(rules.person.HasIdOf([self.__option.get_value()]))
@@ -718,7 +719,7 @@ class GuiPersonOption(Gtk.Box):
             self.__uistate,
             self.__track,
             title=_("Select a person for the report"),
-            filter=rfilter,
+            search_or_filter=rfilter,
         )
         person = sel.run()
         self.__update_person(person)
@@ -852,6 +853,8 @@ class GuiFamilyOption(Gtk.Box):
         """
         # Create a filter for the person selector.
         rfilter = GenericFilterFactory("Family")()
+        rfilter.set_name(_("Active or bookmarked families"))
+
         rfilter.set_logical_op("or")
 
         # Add the current family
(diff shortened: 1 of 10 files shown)


stevenyoungs:
@Nick-Hall not sure if you want to merge this as an interim solution pending a longer term fix?

stevenyoungs:
I've reviewed other places in the code that call `SelectorFactory("Person")` and not found any that also use the `filter` parameter when constructing the selector

stevenyoungs:
Extend the `SearchBar` class to support both
1. text search (the current behaviour)
2. gramps filters

Both types of query are available in the query list. Searches are shown first, followed by filter(s). See the image below.
The search text and clear button are only sensitive when a search is selected.

There are two places in the code that use filters with the `SearchBar`:
1. `GuiPersonOption`
2. `GuiFamilyOption`

**Naming**
Previously filter and search were used interchangeably. I've tried to be consistent in using search for a text search and filter for a gramps filter.

image

gramps/gui/plug/_guioptions.py
```diff
@@ -695,6 +695,7 @@ def __get_person_clicked(self, obj):  # IGNORE:W0613 - obj is unused
         """
         # Create a filter for the person selector.
         rfilter = GenericFilter()
+        rfilter.set_name("Active, Bookmarked or Home People")
```
stevenyoungs, line 698: Comments welcome on both the wording and capitalisation

gramps/gui/plug/_guioptions.py
```diff
@@ -852,6 +853,8 @@ def __get_family_clicked(self, obj):  # IGNORE:W0613 - obj is unused
         """
         # Create a filter for the person selector.
         rfilter = GenericFilterFactory("Family")()
+        rfilter.set_name("Active or Bookmarked Families")
```
stevenyoungs, line 856: Comments welcome on both the wording and capitalisation

stevenyoungs:
**Testing**
I've tested the following:
1. Reports > Text Reports > Complete Individual Report...
This uses a default person filter when selecting a person for the report
2. Reports > Text Reports > Family Group Report...
This uses a default family filter when selecting a family for the report
3. Add > Family
Selecting a Father or Mother uses a default text search
4. Add > Family
Selecting Children, Events, Source Citations, Notes, Gallery all use the SearchBar with no default text search or filter

However I'm aware that the selector classes are used in a large number of places.
If you know of other, uncommon, uses of the selector dialogs, please comment.

PR #2034, opened 2025-03-31
MRU Improvements

stevenyoungs:
When adding new objects, also add them to the MRU list.

For now this is a proof of concept and _only_ adds new people created via the `PersonTreeView` to the MRU.

In the screen shot below, of the example database, upon loading, Lewis was the active person. I manually clicked Allison followed by Amy. Amy is the active person at this point. I then added two additional people, Aaron and Aaron Jnr.
image

For now, the active person is shown in the MRU list as well. Whilst I initially did this for debug purposes, there is an argument for keeping it.

Related forum discussion starts here: https://gramps.discourse.group/t/add-a-recent-sub-menu/5760/82

### DIFF
--- a/gramps/gui/displaystate.py
+++ b/gramps/gui/displaystate.py
@@ -122,7 +122,7 @@ class History(Callback):
             if initial_person:
                 self.push(initial_person.get_handle())
 
-    def push(self, handle):
+    def push(self, handle, make_active=True):
         """
         Pushes the handle on the history stack
         """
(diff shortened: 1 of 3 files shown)


gramps/gui/displaystate.py
```diff
@@ -132,8 +132,9 @@ def push(self, handle):
             if handle in self.mru:
                 self.mru.remove(handle)
             self.mru.append(handle)
+            if make_active:
+                self.index += 1
             self.emit("mru-changed", (self.mru,))
-            self.index += 1
```
stevenyoungs, line 136: I've changed the order in which we emit the `mru-changed` signal and increment the index of the active object.
I now need to know the currently active object when updating the MRU menu

Is this likely to cause problems elsewhere?

cdhorn: Possibly for Nick's CombinedView and more likely my experimental CardViews which both use a global navigation history object to try to facilitate navigation across different object types in the same view. We had talked about trying to improve the navigation support in Gramps core to better enable and handle things but nothing has been attempted yet and it has been some time since I reviewed what I did.

PR #2036, opened 2025-04-01
Add HasNoteType filter rules for Media and Repository objects

stevenyoungs:
I can't see any reason not to have these so add for completeness


dsblank (approved):
I love completeness! Nice!

Nick-Hall:
Added the new files to `POTFILES.in`.

PR #2040, opened 2025-04-10
Update Addon Manager wiki Help link from 5.2 to 6.0

emyoulation:
line 500 points to 5.2 URL, update to 6.0

Fixes [#13735](https://gramps-project.org/bugs/view.php?id=13735).


emyoulation:
Great!

You change addresses the hard-coded version Enhancement Request 13762

I found 2 other 6.0 maintenance branch modules that might need the same enhancement from hard-coded wiki version to using a constant. 

The modules & line numbers (and code snippets) are in that request's comments. Is that sufficient?  Or does it require filing additional MantisBT issues?

https://gramps-project.org/bugs/view.php?id=13762#c68595

PR #2041, opened 2025-04-10
Create SECURITY.md

yennhi2906:


yennhi2906:
```html

```

emyoulation:
This file seems to be a lightly modified template from somewhere else.

What was the source?

It adds a version table that would have to be maintained. 

And it does not provide the most important piece of information for a security policy: the preferred method for notifying the GitHub Repository owner of a security violation. (Hopefully, the method chosen will have a way of tracking the progress of resolving the violation. MantisBT and GitHub Issues have public progress feedback. But both require creating an account. And that could be an impediment to being notified.)

QuLogic:
> This file seems to be a lightly modified template from somewhere else.
> 
> What was the source?

This seems to be the default template if you click the "Suggest a policy" button from the Security tab of the repository.

I don't think it makes sense for a random contributor to be writing it though, and not one of the maintainers.

PR #2044, opened 2025-04-11
Notes: display more than 80 characters.

SNoiraud:
The default of 80 characters can be modified in the preferences general tab.
The maximum allowed is 2000.

### DIFF
--- a/gramps/gen/lib/note.py
+++ b/gramps/gen/lib/note.py
@@ -36,6 +36,7 @@ from .primaryobj import BasicPrimaryObject
 from .styledtext import StyledText
 from .styledtexttagtype import StyledTextTagType
 from .tagbase import TagBase
+from gramps.gen.config import config
 
 _ = glocale.translation.gettext
 
@@ -271,6 +272,19 @@ class Note(BasicPrimaryObject):
         :rtype: unicode
         """
         return str(self.text)
+
+    def get_preview(self):
+        """Return the max visible text string associated with the note.
+
+        :returns: either the *clear* text of the note contents or the maximum
+                  visible text.
+        :rtype: unicode
+        """
+        max_vis_len = config.get("interface.note-preview-length")
+        preview = str(self.text).replace("\n", " ")
+        if len(preview) > max_vis_len:
+            return preview[:max_vis_len] + "..."
+        return preview
 
     def set_styledtext(self, text):
         """Set the text associated with the note to the passed string.
--- a/gramps/gui/clipboard.py
+++ b/gramps/gui/clipboard.py
@@ -28,6 +28,7 @@ import pickle
 import os
 from xml.sax.saxutils import escape
 from time import strftime as strftime
+from gramps.gen.config import config
 
 # -------------------------------------------------------------------------
 #
@@ -59,7 +60,6 @@ from .ddtargets import DdTargets
 from .makefilter import make_filter
 from .utils import is_right_click, no_match_primary_mask
 from gramps.gen.const import GRAMPS_LOCALE as glocale
-from gramps.gen.config import config
 from gramps.gui.widgets.multitreeview import MultiTreeView
 
 _ = glocale.translation.sgettext
@@ -399,14 +399,7 @@ class ClipNote(ClipHandleWrapper):
         value = clipdb.get_note_from_handle(self._handle)
         if value:
             self._title = value.get_gramps_id()
-            note = value.get().replace("\n", " ")
-            # String must be unicode for truncation to work for non
-            # ascii characters
-            note = str(note)
-            if len(note) > 80:
-                self._value = note[:80] + "..."
-            else:
-                self._value = note
+            self._value = value.get_preview()
 
 
 class ClipFamilyEvent(ClipObjWrapper):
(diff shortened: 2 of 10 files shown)


gramps/gui/clipboard.py
```diff
@@ -403,8 +404,9 @@ def refresh(self):
             # String must be unicode for truncation to work for non
             # ascii characters
             note = str(note)
```
QuLogic, line 406: On an unrelated note, what does this comment mean? Is this a Python2-ism? If it were `bytes`, it certainly wouldn't do the right thing on Python 3:
```pycon
>>> b = b'abc'
>>> str(b)
"b'abc'"
```
It seems like both the command and `str` cast should be removed (either that or it needs to _conditionally_ `decode`.)

SNoiraud:
> Could you post a screen capture of the revised General tab of Preferences?
note gramps 1

gramps/gui/clipboard.py
SNoiraud: I don't know. it is a very old comment I think.

emyoulation:
How about "Note preview length:" instead of "Maximum visible length of a note.:"? This would more consistent with the phrasing of the option above: "Multiple Surname box height:"

(If you do not care for that suggestion, there is still an extra period at the end of the string.)

Nick-Hall:
We should also move all the duplicated code into a new `Note` method.  Perhaps `note.get_preview()`?

gramps/gui/clipboard.py
SNoiraud: After looking at the source, b is a StyledText, so str is needed to get the text.

gramps/gen/lib/note.py
```diff
@@ -272,6 +273,18 @@ def get(self):
...
+        """Return the max visible text string associated with the note.
+
+        :returns: either the *clear* text of the note contents or the maximum
+                  visible text.
+        :rtype: unicode
+        """
+        max_vis_len = config.get("interface.note-max-vis-length")
+        if len(str(self.text)) > max_vis_len:
+            return str(self.text).replace("\n", " ")[:max_vis_len] + "..."
+        return str(self.text)
```
Nick-Hall, line 286: You will still need to replace the "\n" with " " here.

gramps/gui/clipboard.py
```diff
@@ -28,6 +28,7 @@
 import os
 from xml.sax.saxutils import escape
 from time import strftime as strftime
+from gramps.gen.config import config
```
Nick-Hall, line 31: No longer required.

gramps/gui/editors/displaytabs/backrefmodel.py
```diff
@@ -37,6 +37,7 @@
 # Gramps classes
 #
 # -------------------------------------------------------------------------
+from gramps.gen.config import config
```
Nick-Hall, line 40: No longer required.

gramps/gui/editors/displaytabs/namemodel.py
```diff
@@ -26,6 +26,7 @@
 #
 # -------------------------------------------------------------------------
 from gramps.gen.const import GRAMPS_LOCALE as glocale
+from gramps.gen.config import config
```
Nick-Hall, line 29: No longer required.

gramps/gui/editors/displaytabs/notemodel.py
```diff
@@ -24,6 +24,7 @@
 #
 # -------------------------------------------------------------------------
 from gi.repository import Gtk
+from gramps.gen.config import config
```
Nick-Hall, line 27: No longer required.

gramps/plugins/tool/removeunused.py
```diff
@@ -47,6 +47,7 @@
 from gramps.gui.glade import Glade
 from gramps.gen.filters import GenericFilterFactory, rules
 from gramps.gen.const import GRAMPS_LOCALE as glocale
+from gramps.gen.config import config
```
Nick-Hall, line 50: No longer required.

gramps/gen/lib/note.py
SNoiraud: Yes we need that because get_preview is only used to be displayed in GUI. If we don't do that, we can have multiple lines (i.e in the noteview like below)
multiple-lines

SNoiraud: Were you talking about the second return?

Nick-Hall: Yes.

SNoiraud: Done

gramps/gen/lib/note.py
```diff
@@ -272,6 +273,18 @@ def get(self):
...
 
+    def get_preview(self):
+        """Return the max visible text string associated with the note.
+
+        :returns: either the *clear* text of the note contents or the maximum
+                  visible text.
+        :rtype: unicode
+        """
+        max_vis_len = config.get("interface.note-max-vis-length")
+        if len(str(self.text)) > max_vis_len:
```
Nick-Hall, line 284:
```suggestion
        preview = str(self.text).replace("\n", " ")
        if len(preview) > max_vis_len:
            return preview[:max_vis_len] + "..."
        return preview
```

Nick-Hall: This may be slightly simpler and easier to read.

SNoiraud: Done

Nick-Hall:
I changed the configuration option to "interface.max-preview-length".

PR #2045, opened 2025-04-13
Fix patronymic in name display

ztlxltl:
Fixes a bug in 6.0.0 that causes patronymic surnames to be treated as non-patronymic in the name display. The empty `"string"` value in `raw_surn_data["origintype"]` is compared to the numeric value of `NameOriginType.PATRONYMIC` (and `NameOriginType.MATRONYMIC)`.
With this PR, `raw_surn_data["origintype"]["value"]` is used for these checks.

Fixes bug [13764](https://gramps-project.org/bugs/view.php?id=13764)

### DIFF
--- a/gramps/gen/display/name.py
+++ b/gramps/gen/display/name.py
@@ -166,8 +166,8 @@ def _raw_primary_surname(raw_surn_data_list):
                 not PAT_AS_SURN
                 and nrsur == 1
                 and (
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Merged in commit c2ac928b6ae5924ffcd7b84f21095f15a9712124.

PR #2046, opened 2025-04-14
Provide a fallback for locale module without gettext support

Nick-Hall:


bartfeenstra:
I can confirm that [this change fixes](https://github.com/bartfeenstra/betty/actions/runs/14449075562/job/40516943357?pr=2444) the `GrampsLocale` initialization errors [I was getting before](https://github.com/bartfeenstra/betty/actions/runs/14320625158/job/40137393499?pr=2444#step:14:7683).

jralls:
It has to crash if the GUI is enabled, otherwise we'll get weird crashes when we try to load glade and ui files.

jralls:
@Nick-Hall Thanks.

PR #2047, opened 2025-04-15
Add mimetypes to notes

olivierberten:
Here's a first try at adding mimetypes to notes with a lot of help from Claude AI...

### DIFF
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -421,7 +421,7 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
 
     __callback_map = {}
 
-    VERSION = (21, 0, 0)
+    VERSION = (22, 0, 0)
 
     def __init__(self, directory=None):
         DbReadBase.__init__(self)
@@ -2777,6 +2777,7 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
             gramps_upgrade_19,
             gramps_upgrade_20,
             gramps_upgrade_21,
+            gramps_upgrade_22,
         )
 
         if version < 14:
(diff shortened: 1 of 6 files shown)


Nick-Hall:
> with a lot of help from Claude AI

Our current [contribution policy](https://gramps-project.org/wiki/index.php/Howto:_Contribute_to_Gramps) states that you must "certify that you personally authored the code and did not copy from other implementations or use AI generators".

This was last discussed a couple of years ago, but the community will have to decide if we want to change the policy to allow this contribution.

Nick-Hall:
Please move this to the _master_ branch.  We don't accept enhancements in maintenance branches.

Nick-Hall:
The mime-type field is defined in the [shared note record](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#SHARED_NOTE_RECORD) in the Gedcom 7.0 specification.

olivierberten:
> This was last discussed a couple of years ago, but the community will have to decide if we want to change the policy to allow this contribution.

It seems this has actually been discussed less than a few years ago...
https://gramps.discourse.group/t/gramps-ai-artificial-intelligence-and-the-future/6000
https://gramps.discourse.group/t/guidelines-for-using-ai-when-documenting-gramps/6629
And it seems the current consensus among developers is not to allow this... Too bad for me...

> Please move this to the _master_ branch. We don't accept enhancements in maintenance branches.

I thought so but master is so much behind compared to _maintenance/gramps60_ that I wonder whether that make sense...

> The `MIME` tag is defined in the [shared note record](https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#SHARED_NOTE_RECORD) in the Gedcom 7.0 specification.

As well as
https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#SOURCE_RECORD
https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#NOTE_STRUCTURE
https://gedcom.io/specifications/FamilySearchGEDCOMv7.html#SOURCE_CITATION

That said, GEDCOM7 only allows `text/plain` and `text/html` whereas I'd like Gramps to be more open about it...


> This was last discussed a couple of years ago, but the community will have to decide if we want to change the policy to allow this contribution.

olivierberten:
Closing as I'm working on a pull request against master

PR #2050, opened 2025-04-21
Fix icu/PyICU exception handling in narrated web report

martin-schwenke:
Exception handling is currently incomplete when both icu and PyICU are not installed.

First, fix the error in the most obvious way.  Then remove duplicate exception handling.

Fixes [#13407](https://gramps-project.org/bugs/view.php?id=13407).

### DIFF
--- a/gramps/plugins/webreport/common.py
+++ b/gramps/plugins/webreport/common.py
@@ -64,14 +64,12 @@ except ImportError:
         from PyICU import Locale
 
         HAVE_ICU = True
-        try:
-            from PyICU import AlphabeticIndex as icuAlphabeticIndex
 
-            HAVE_ALPHABETICINDEX = True
-        except ImportError:
-            from gramps.plugins.webreport.alphabeticindex import (
-                AlphabeticIndex as localAlphabeticIndex,
-            )
+        # Failure here is caught by the enclosing exception handling
+        from PyICU import AlphabeticIndex as icuAlphabeticIndex
+
+        HAVE_ALPHABETICINDEX = True
+
     except ImportError:
         from gramps.plugins.webreport.alphabeticindex import (
             AlphabeticIndex as localAlphabeticIndex,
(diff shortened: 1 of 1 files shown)


dsblank (approved):
## Review: Fix icu/PyICU exception handling in narrated web report

**Summary:** This is a small, targeted bug fix for a genuine `NameError` crash (bug #13407) that occurs when neither `icu` nor `PyICU` is installed.

### The Bug

At line 681 in `common.py`, the `else` branch always references `localAlphabeticIndex`:
```python
else:
    AlphabeticIndex = localAlphabeticIndex  # type: ignore
```

In the current code, the outermost `except ImportError: pass` means `localAlphabeticIndex` is **never assigned** when both packages are absent — causing a `NameError` at module load time.

### The Fix

The PR correctly restructures the PyICU block so that when `PyICU` is not installed (or when `AlphabeticIndex` is unavailable in it), the `except ImportError` clause now imports `localAlphabeticIndex` instead of silently passing.

### Correctness Analysis

The fix handles all cases correctly:

| Scenario | Before PR | After PR |
|---|---|---|
| `icu` installed with `AlphabeticIndex` | ✅ works | ✅ unchanged |
| `icu` installed, no `AlphabeticIndex` | ✅ works | ✅ unchanged |
| Only `PyICU` installed with `AlphabeticIndex` | ✅ works | ✅ works |
| Only `PyICU` installed, no `AlphabeticIndex` | ✅ works | ✅ works (`HAVE_ICU` correctly stays `True`) |
| Neither `icu` nor `PyICU` | ❌ `NameError` | ✅ fixed |

### Minor Observation

The PR's comment says "remove duplicate exception handling," but the `icu` block (lines 53–60) still uses the old inner try/except pattern. The PR only simplifies the PyICU block. The `icu` block is functionally correct as-is, but it could be made consistent with the same simplification. This is a cosmetic suggestion, not a blocker.

### Verdict

**Approve.** The fix is correct, minimal, and well-scoped. The code comment (`# Failure here is caught by the enclosing exception handling`) clearly explains the intent. No regressions introduced.

dsblank (approved):
Looks good to me and Claude.

PR #2052, opened 2025-04-25
Refactor Optimizer for correctness

stevenyoungs:
When the filter optimizer combines the result of two or more sub-filters, it is currently using the logical_op of the right hand sub-filter. Fix this so that the optimizer uses the parent filter logical_op

Fixes #13799 https://gramps-project.org/bugs/view.php?id=13799
Related forum discussion: https://gramps.discourse.group/t/strange-filter-behavior-in-6-0-1/7505/16

To reproduce using the example Gramps tree
1. create a new filter called "Ancestors of" with a single rule, "Ancestors of <person>"
set ID to I0044, Inclusive ticked
leave the Options set to "all rules must apply"
2. create a new filter called "Siblings of Ancestors" with a single rule, "Siblings of <filter> match"
set the filter to "Ancestors of"
leave the Options set to "all rules must apply"
3. Create a new filter called "The Family and their Spouces" with two rules
a. rule 1: People matching the <filter> and set the filter to "Ancestors of"
b. rule 2: People matching the <filter> and set the filter to "Siblings of Ancestors"
set the Options to "At least one rule must apply"

Filter the people view using the filter "Siblings of Ancestors". 25 people are expected. 0 are shown

The filter "Ancestors of" returns 7 people
The filter "Siblings of Ancestors" returns 18 people
So the union of these two results must, at a minimum, return 18 people. Therefore a result of 0 people cannot be correct.

### DIFF
--- a/gramps/gen/filters/_genericfilter.py
+++ b/gramps/gen/filters/_genericfilter.py
@@ -4,6 +4,7 @@
 # Copyright (C) 2002-2006  Donald N. Allingham
 # Copyright (C) 2011       Tim G L Lyons
 # Copyright (C) 2012,2024  Doug Blank <doug.blank@gmail.com>
+# Copyright (C) 2025       Steve Youngs <steve@youngs.cc>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -24,6 +25,12 @@
 Package providing filtering framework for Gramps.
 """
 
+# -------------------------------------------------------------------------
+#
+# Standard Python modules
+#
+# -------------------------------------------------------------------------
+from __future__ import annotations
 import logging
 import time
 
@@ -46,6 +53,15 @@ from ..const import GRAMPS_LOCALE as glocale
 from .rules import Rule
 from .optimizer import Optimizer
 
+# -------------------------------------------------------------------------
+#
+# Typing modules
+#
+# -------------------------------------------------------------------------
+from typing import cast, Dict, List, Literal, Set, Tuple
+from ..db import Database
+from ..types import PrimaryObjectHandle
+
 _ = glocale.translation.gettext
 LOG = logging.getLogger(".filter.results")
 
@@ -56,9 +72,10 @@ LOG = logging.getLogger(".filter.results")
 #
 # -------------------------------------------------------------------------
 class GenericFilter:
-    """Filter class that consists of several rules."""
+    """Filter class that consists of zero, one or more rules."""
 
     logical_functions = ["and", "or", "one"]
+    logical_op: Literal["and", "or", "one"]
 
     def __init__(self, source=None):
         if source:
@@ -124,6 +141,9 @@ class GenericFilter:
     def get_rules(self):
         return self.flist
 
+    def get_all_handles(self, db):
+        return db.get_person_handles()
+
     def get_cursor(self, db):
         return db.get_person_cursor()
 
@@ -140,78 +160,44 @@ class GenericFilter:
         return db.get_number_of_people()
 
     def apply_logical_op_to_all(
-        self, db, id_list, apply_logical_op, user=None, tupleind=None, tree=False
+        self,
+        db,
+        possible_handles: Set[PrimaryObjectHandle],
+        apply_logical_op,
+        user=None,
     ):
-        final_list = []
-
-        optimizer = Optimizer(self)
-        handles_in, handles_out = optimizer.get_handles()
-
         LOG.debug(
-            "Optimizer handles_in: %s",
-            len(handles_in) if handles_in is not None else None,
+            "Starting possible_handles: %s",
+            len(possible_handles),
         )
-        LOG.debug("Optimizer handles_out: %s", len(handles_out))
-        if id_list is None:
-            if handles_in is not None:
-                if user:
-                    user.begin_progress(_("Filter"), _("Applying ..."), len(handles_in))
 
-                # Use these rather than going through entire database
-                for handle in handles_in:
-                    if user:
-                        user.step_progress()
+        # use the Optimizer to refine the set of possible_handles
+        optimizer = Optimizer()
+        handles_in, handles_out = optimizer.compute_potential_handles_for_filter(self)
 
-                    if handle is None:
-                        continue
+        # LOG.debug(
+        #    "Optimizer possible_handles: %s",
+        #    len(possible_handles),
+        # )
 
-                    obj = self.get_object(db, handle)
+        # if user:
+        #    user.begin_progress(_("Filter"), _("Applying ..."), len(possible_handles))
 
-                    if apply_logical_op(db, obj, self.flist) != self.invert:
-                        final_list.append(obj.handle)
+        # test each value in possible_handles to compute the final_list
+        final_list = []
+        for handle in possible_handles:
+            if handles_in is not None and handle not in handles_in:
+                continue
+            if handles_out is not None and handle in handles_out:
+                continue
 
-            else:
-                with (
-                    self.get_tree_cursor(db) if tree else self.get_cursor(db)
-                ) as cursor:
-                    if user:
-                        user.begin_progress(
-                            _("Filter"), _("Applying ..."), self.get_number(db)
-                        )
-
-                    for handle, obj in cursor:
-                        if user:
-                            user.step_progress()
-
-                        if handle in handles_out:
-                            continue
-
-                        if apply_logical_op(db, obj, self.flist) != self.invert:
-                            final_list.append(handle)
-
-        else:
-            id_list = list(id_list)
             if user:
-                user.begin_progress(_("Filter"), _("Applying ..."), len(id_list))
-            for handle_data in id_list:
-                if user:
-                    user.step_progress()
+                user.step_progress()
 
-                if tupleind is None:
-                    handle = handle_data
-                else:
-                    handle = handle_data[tupleind]
+            obj = self.get_object(db, handle)
 
-                if handles_in is not None:
-                    if handle not in handles_in:
-                        continue
-                elif handle in handles_out:
-                    continue
-
-                obj = self.get_object(db, handle)
-
-                if apply_logical_op(db, obj, self.flist) != self.invert:
-                    final_list.append(handle_data)
+            if apply_logical_op(db, obj, self.flist) != self.invert:
+                final_list.append(obj.handle)
 
         if user:
             user.end_progress()
@@ -250,15 +236,32 @@ class GenericFilter:
             raise Exception("invalid operator: %r" % self.logical_op)
         return res != self.invert
 
-    def apply(self, db, id_list=None, tupleind=None, user=None, tree=False):
+    def apply(
+        self,
+        db,
+        id_list: List[PrimaryObjectHandle | Tuple] | None = None,
+        tupleind: int | None = None,
+        user=None,
+        tree: bool = False,
+    ) -> List[PrimaryObjectHandle | Tuple]:
         """
         Apply the filter using db.
-        If id_list given, the handles in id_list are used. If not given
-        a database cursor will be used over all entries.
 
-        If tupleind is given, id_list is supposed to consist of a list of
-        tuples, with the handle being index tupleind. So
-        handle_0 = id_list[0][tupleind]
+        If id_list and tupleind given, id_list is a list of tuples. tupleind is the
+        index of the object handle. So handle_0 = id_list[0][tupleind]. The input
+        order in maintained in the output. If multiple tuples have the same handle,
+        it is undefined which tuple is returned.
+
+        If id_list given and tupleind is None, id_list if the list of handles to use
+
+        If id_list is None and tree is False, all handles in the database are searched.
+        The order of handles in the result is not guaranteed
+
+        If id_list is None and tree is True, all handles in the database are searched.
+        The order of handles in the result matches the order of traversal
+        by get_tree_cursor
+
+        id_list takes precendence over tree
 
         user is optional. If present it must be an instance of a User class.
 
@@ -282,9 +285,39 @@ class GenericFilter:
             raise Exception("invalid operator: %r" % self.logical_op)
 
         start_time = time.time()
-        res = self.apply_logical_op_to_all(
-            db, id_list, apply_logical_op, user, tupleind, tree
-        )
+
+        # build the starting set of possible_handles to be filtered
+        possible_handles: Set[PrimaryObjectHandle]
+        if id_list is not None:
+            if tupleind is not None:
+                # construct a dict from handle to corresponding tuple
+                # this is used to efficiently transform final_list from a list of
+                # handles to a list of tuples
+                handle_tuple: Dict[PrimaryObjectHandle, Tuple] = {
+                    data[tupleind]: data for data in cast(List[Tuple], id_list)
+                }
+                possible_handles = set(handle_tuple.keys())
+            else:
+                possible_handles = set(cast(List[PrimaryObjectHandle], id_list))
+        elif tree:
+            tree_handles = [handle for handle, obj in self.get_tree_cursor(db)]
+            possible_handles = set(tree_handles)
+        else:
+            possible_handles = set(self.get_all_handles(db))
+
+        res = self.apply_logical_op_to_all(db, possible_handles, apply_logical_op, user)
+
+        # convert the filtered set of handles to the correct result type
+        if id_list is not None and tupleind is not None:
+            # convert the final_list of handles back to the corresponding final_list of tuples
+            res = sorted(
+                [handle_tuple[handle] for handle in res],
+                key=lambda x: id_list.index(x),
+            )
+        elif tree:
+            # sort final_list into the same order as traversed by get_tree_cursor
+            res = sorted(res, key=lambda x: tree_handles.index(x))
+
         LOG.debug("Apply time: %s seconds", time.time() - start_time)
 
         for rule in self.flist:
@@ -299,6 +332,9 @@ class GenericFilter:
 class GenericFamilyFilter(GenericFilter):
     def __init__(self, source=None):
         GenericFilter.__init__(self, source)
+
+    def get_all_handles(self, db):
+        return db.get_family_handles()
 
     def get_cursor(self, db):
         return db.get_family_cursor()
@@ -320,6 +356,9 @@ class GenericEventFilter(GenericFilter):
     def __init__(self, source=None):
         GenericFilter.__init__(self, source)
 
+    def get_all_handles(self, db):
+        return db.get_event_handles()
+
     def get_cursor(self, db):
         return db.get_event_cursor()
 
@@ -340,6 +379,9 @@ class GenericSourceFilter(GenericFilter):
     def __init__(self, source=None):
         GenericFilter.__init__(self, source)
 
+    def get_all_handles(self, db):
+        return db.get_source_handles()
+
     def get_cursor(self, db):
         return db.get_source_cursor()
 
@@ -359,6 +401,9 @@ class GenericSourceFilter(GenericFilter):
 class GenericCitationFilter(GenericFilter):
     def __init__(self, source=None):
         GenericFilter.__init__(self, source)
+
+    def get_all_handles(self, db):
+        return db.get_citation_handles()
 
     def get_cursor(self, db):
         return db.get_citation_cursor()
@@ -383,6 +428,9 @@ class GenericPlaceFilter(GenericFilter):
     def __init__(self, source=None):
         GenericFilter.__init__(self, source)
 
+    def get_all_handles(self, db):
+        return db.get_place_handles()
+
     def get_cursor(self, db):
         return db.get_place_cursor()
 
@@ -406,6 +454,9 @@ class GenericMediaFilter(GenericFilter):
     def __init__(self, source=None):
         GenericFilter.__init__(self, source)
 
+    def get_all_handles(self, db):
+        return db.get_media_handles()
+
     def get_cursor(self, db):
         return db.get_media_cursor()
 
@@ -426,6 +477,9 @@ class GenericRepoFilter(GenericFilter):
     def __init__(self, source=None):
         GenericFilter.__init__(self, source)
 
+    def get_all_handles(self, db):
+        return db.get_repository_handles()
+
     def get_cursor(self, db):
         return db.get_repository_cursor()
 
@@ -445,6 +499,9 @@ class GenericRepoFilter(GenericFilter):
 class GenericNoteFilter(GenericFilter):
     def __init__(self, source=None):
         GenericFilter.__init__(self, source)
+
+    def get_all_handles(self, db):
+        return db.get_note_handles()
 
     def get_cursor(self, db):
         return db.get_note_cursor()
--- a/gramps/gen/filters/optimizer.py
+++ b/gramps/gen/filters/optimizer.py
@@ -2,6 +2,7 @@
 # Gramps - a GTK+/GNOME based genealogy program
 #
 # Copyright (C) 2024  Doug Blank <doug.blank@gmail.com>
+# Copyright (C) 2025  Steve Youngs <steve@youngs.cc>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -17,121 +18,66 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
-
+# -------------------------------------------------------------------------
+#
+# Standard Python modules
+#
+# -------------------------------------------------------------------------
+from __future__ import annotations
 import logging
+
+# -------------------------------------------------------------------------
+#
+# Typing modules
+#
+# -------------------------------------------------------------------------
+from typing import List, Set, TYPE_CHECKING, Tuple
+from .rules import Rule
+
+if TYPE_CHECKING:
+    from ._genericfilter import GenericFilter
+from ..types import PrimaryObjectHandle
 
 LOG = logging.getLogger(".filter.optimizer")
 
 
-def intersection(sets):
-    if sets:
-        result = sets[0]
-        for s in sets[1:]:
-            result = result.intersection(s)
-        return result
-    else:
-        return set()
-
-
-def union(sets):
-    if sets:
-        result = sets[0]
-        for s in sets[1:]:
-            result = result.union(s)
-        return result
-    else:
-        return set()
-
-
 class Optimizer:
-    """
-    Optimizer to use the filter's pre-selected selected_handles
-    to include or exclude.
-    """
+    def compute_potential_handles_for_filter(
+        self, filter: GenericFilter
+    ) -> Tuple[Set[PrimaryObjectHandle] | None, Set[PrimaryObjectHandle] | None]:
+        if len(filter.flist) == 0:
+            return (None, None)
 
-    def __init__(self, filter):
-        """
-        Initialize the collection of selected_handles in the filter list.
-        """
-        self.all_selected_handles = []
-        self.walk_filters(filter, False, self.all_selected_handles)
-
-    def walk_filters(self, filter, parent_invert, result):
-        """
-        Recursively walk all of the filters/rules and get
-        rules with selected_handles
-        """
-        current_invert = parent_invert if not filter.invert else not parent_invert
-        LOG.debug(
-            "walking, filter: %s, invert=%s, parent_invert=%s",
-            filter,
-            filter.invert,
-            parent_invert,
-        )
-        rules_with_selected_handles = []
-        for item in filter.flist:
-            if hasattr(item, "find_filter"):
-                rule_filter = item.find_filter()
-                if rule_filter is not None:
-                    self.walk_filters(
-                        rule_filter,
-                        current_invert,
-                        result,
-                    )
-            elif hasattr(item, "selected_handles"):
-                rules_with_selected_handles.append(set(item.selected_handles))
-        if rules_with_selected_handles:
-            LOG.debug(
-                "filter %s: parent_invert=%s, invert=%s, op=%s, number of rules with selected_handles=%s",
-                filter,
-                parent_invert,
-                filter.invert,
-                filter.logical_op,
-                len(rules_with_selected_handles),
-            )
-            result.append(
-                (
-                    current_invert,
-                    filter.logical_op,
-                    rules_with_selected_handles,
-                )
-            )
-
-    def get_handles(self):
-        """
-        Returns handles_in, and handles_out.
-
-        `handles_in` is either None, or a set of handles to include.
-        if it is None, then there is no evidence to only include
-        particular handles. If it is a set, then those in the set
-        are a superset of the items that will match.
-
-        `handles_out` is a set. If any handle is in the set, it will
-        not be included in the final results.
-
-        The handles_in are selected if all of the rules are connected with
-        "and" and not inverted. The handles_out are selected if all of the
-        rules are connected with "and" and inverted.
-        """
         handles_in = None
-        handles_out = set()
-        # Get all positive non-inverted selected_handles
-        for inverted, logical_op, selected_handles in self.all_selected_handles:
-            if logical_op == "and" and not inverted:
-                LOG.debug("optimizer positive match!")
-                if handles_in is None:
-                    handles_in = intersection(selected_handles)
-                else:
-                    handles_in = intersection([handles_in] + selected_handles)
+        handles_out = None
+        for rule in filter.flist:
+            if filter.logical_op == "and" or len(filter.flist) == 1:
+                rule_in, rule_out = self.compute_potential_handles_for_rule(rule)
+                if rule_in is not None:
+                    if handles_in is None:
+                        handles_in = rule_in
+                    else:
+                        handles_in = handles_in.intersection(rule_in)
+                if rule_out is not None:
+                    if handles_out is None:
+                        handles_out = rule_out
+                    else:
+                        handles_out = handles_out.union(rule_out)
 
-        # Get all inverted selected_handles:
-        for inverted, logical_op, selected_handles in self.all_selected_handles:
-            if logical_op == "and" and inverted:
-                LOG.debug("optimizer inverted match!")
-                handles_out = union([handles_out] + selected_handles)
+        if filter.invert:
+            handles_in, handles_out = handles_out, handles_in
 
-        if handles_in is not None:
-            handles_in = handles_in - handles_out
-
-        LOG.debug("optimizer handles_in: %s", len(handles_in) if handles_in else 0)
         return handles_in, handles_out
+
+    def compute_potential_handles_for_rule(
+        self,
+        rule: Rule,
+    ) -> Tuple[Set[PrimaryObjectHandle] | None, Set[PrimaryObjectHandle] | None]:
+        """ """
+        if hasattr(rule, "selected_handles"):
+            return (rule.selected_handles, None)
+        if hasattr(rule, "find_filter"):
+            filter = rule.find_filter()
+            if filter:
+                return self.compute_potential_handles_for_filter(filter)
+        return (None, None)
--- a/gramps/gen/filters/test/filter_optimizer_test.py
+++ b/gramps/gen/filters/test/filter_optimizer_test.py
@@ -1,0 +1,427 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2025 Doug Blank <doug.blank@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+import tempfile
+import os
+import unittest
+
+from ...const import DATA_DIR
+from ...db.utils import import_as_dict
+from ...user import User
+from ...filters import reload_custom_filters, FilterList, set_custom_filters
+from ....gen import filters
+
+custom_filters_xml = """<?xml version="1.0" encoding="utf-8"?>
+<filters>
+  <object type="Person">
+    <filter name="Ancestors of" function="and">
+      <rule class="IsAncestorOf" use_regex="False" use_case="False">
+        <arg value="I0044"/>
+        <arg value="1"/>
+      </rule>
+    </filter>
+    <filter name="Family and their Spouses" function="or">
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
+        <arg value="Ancestors of"/>
+      </rule>
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
+        <arg value="Siblings of Ancestors"/>
+      </rule>
+    </filter>
+    <filter name="Siblings of Ancestors" function="and">
+      <rule class="IsSiblingOfFilterMatch" use_regex="False" use_case="False">
+        <arg value="Ancestors of"/>
+      </rule>
+    </filter>
+    <filter name="Everyone" function="or">
+      <rule class="Everyone" use_regex="False" use_case="False">
+      </rule>
+    </filter>
+    <filter name="F1" function="or">
+      <rule class="Everyone" use_regex="False" use_case="False">
+      </rule>
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
+        <arg value="F2"/>
+      </rule>
+    </filter>
+    <filter name="F2" function="and">
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
+        <arg value="F3"/>
+      </rule>
+    </filter>
+    <filter name="F3" function="and">
+      <rule class="IsAncestorOf" use_regex="False" use_case="False">
+        <arg value="I0041"/>
+        <arg value="0"/>
+      </rule>
+    </filter>
+    <filter name="I0001 xor I0002" function="one">
+      <rule class="HasIdOf" use_regex="False" use_case="False">
+        <arg value="I0001"/>
+      </rule>
+      <rule class="HasIdOf" use_regex="False" use_case="False">
+        <arg value="I0002"/>
+      </rule>
+    </filter>
+    <filter name="I0001 or I0002" function="or">
+      <rule class="HasIdOf" use_regex="False" use_case="False">
+        <arg value="I0001"/>
+      </rule>
+      <rule class="HasIdOf" use_regex="False" use_case="False">
+        <arg value="I0002"/>
+      </rule>
+    </filter>
+    <filter name="I0001 and I0002" function="and">
+      <rule class="HasIdOf" use_regex="False" use_case="False">
+        <arg value="I0001"/>
+      </rule>
+      <rule class="HasIdOf" use_regex="False" use_case="False">
+        <arg value="I0002"/>
+      </rule>
+    </filter>
+    <filter name="I0001" function="or">
+      <rule class="HasIdOf" use_regex="False" use_case="False">
+        <arg value="I0001"/>
+      </rule>
+    </filter>
+    <filter name="I0002" function="or">
+      <rule class="HasIdOf" use_regex="False" use_case="False">
+        <arg value="I0002"/>
+      </rule>
+    </filter>
+    <filter name="not I0001" function="or" invert="1">
+      <rule class="HasIdOf" use_regex="False" use_case="False">
+        <arg value="I0001"/>
+      </rule>
+    </filter>
+    <filter name="not I0002" function="or" invert="1">
+      <rule class="HasIdOf" use_regex="False" use_case="False">
+        <arg value="I0002"/>
+      </rule>
+    </filter>
+    <filter name="(not I0001) and (not I0002)" function="and">
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
+        <arg value="not I0001"/>
+      </rule>
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
+        <arg value="not I0002"/>
+      </rule>
+    </filter>
+    <filter name="(not I0001) or (not I0002)" function="or">
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
+        <arg value="not I0001"/>
+      </rule>
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
+        <arg value="not I0002"/>
+      </rule>
+    </filter>
+
+    <filter name="not ((not I0001) and (not I0002))" function="and" invert="1">
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
+        <arg value="not I0001"/>
+      </rule>
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
+        <arg value="not I0002"/>
+      </rule>
+    </filter>
+
+    <filter name="not (I0001 and I0002)" function="and" invert="1">
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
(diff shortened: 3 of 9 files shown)


stevenyoungs:
@dsblank I think this is the correct fix but would appreciate your review

dsblank:
Fantastic! I'll take a look.

dsblank:
Here are the filters in XML:

```xml
<?xml version="1.0" encoding="utf-8"?>
<filters>
  <object type="Person">
    <filter name="Ancestors of" function="and">
      <rule class="IsAncestorOf" use_regex="False" use_case="False">
        <arg value="I0044"/>
        <arg value="1"/>
      </rule>
    </filter>
    <filter name="Family and their Spouses" function="or">
      <rule class="MatchesFilter" use_regex="False" use_case="False">
        <arg value="Ancestors of"/>
      </rule>
      <rule class="MatchesFilter" use_regex="False" use_case="False">
        <arg value="Siblings of Ancestors"/>
      </rule>
    </filter>
    <filter name="Siblings of Ancestors" function="and">
      <rule class="IsSiblingOfFilterMatch" use_regex="False" use_case="False">
        <arg value="Ancestors of"/>
      </rule>
    </filter>
  </object>
</filters>
```

Planning on making this a test in gramps.

Nick-Hall:
We should include a unit test with a fix like this.

dsblank:
@stevenyoungs I couldn't figure out how to make a PR on your PR. So here is a diff for adding a unittest for the above. It passes after your fix!

```diff
diff --git a/gramps/gen/filters/test/__init__.py b/gramps/gen/filters/test/__init__.py
new file mode 100644
index 000000000..e69de29bb
diff --git a/gramps/gen/filters/test/filter_optimizer_test.py b/gramps/gen/filters/test/filter_optimizer_test.py
new file mode 100644
index 000000000..3fc0df3ca
--- /dev/null
+++ b/gramps/gen/filters/test/filter_optimizer_test.py
@@ -0,0 +1,76 @@
+from gramps.gen.filters import FilterList
+import tempfile
+import os
+import unittest
+
+from ...const import DATA_DIR
+from ...db.utils import import_as_dict
+from ...user import User
+from ...filters import reload_custom_filters
+
+reload_custom_filters()
+
+from ...filters import CustomFilters
+
+
+custom_filters_xml = """<?xml version="1.0" encoding="utf-8"?>
+<filters>
+  <object type="Person">
+    <filter name="Ancestors of" function="and">
+      <rule class="IsAncestorOf" use_regex="False" use_case="False">
+        <arg value="I0044"/>
+        <arg value="1"/>
+      </rule>
+    </filter>
+    <filter name="Family and their Spouses" function="or">
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
+        <arg value="Ancestors of"/>
+      </rule>
+      <rule class="MatchesFilter" use_regex="False" use_case="False">
+        <arg value="Siblings of Ancestors"/>
+      </rule>
+    </filter>
+    <filter name="Siblings of Ancestors" function="and">
+      <rule class="IsSiblingOfFilterMatch" use_regex="False" use_case="False">
+        <arg value="Ancestors of"/>
+      </rule>
+    </filter>
+  </object>
+</filters>
+"""
+TEST_DIR = os.path.abspath(os.path.join(DATA_DIR, "tests"))
+EXAMPLE = os.path.join(TEST_DIR, "example.gramps")
+
+class OptimizerTest(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        """
+        Import example database.
+        """
+        cls.db = import_as_dict(EXAMPLE, User())
+
+        with tempfile.NamedTemporaryFile(mode='w+', encoding='utf8') as tmp_file:
+            tmp_file.write(custom_filters_xml)
+            tmp_file.seek(0)
+            fl = FilterList(tmp_file.name)
+            fl.load()
+
+        the_custom_filters = CustomFilters.get_filters_dict("Person")
+        cls.filters = fl.get_filters_dict("Person")
+        for filter_name in cls.filters:
+            the_custom_filters[filter_name] = cls.filters[filter_name]
+
+    def test_ancestors_of(self):
+        filter = self.filters["Ancestors of"]
+        results = filter.apply(self.db)
+        self.assertEqual(len(results), 7)
+
+    def test_siblings_of_ancestors(self):
+        filter = self.filters["Siblings of Ancestors"]
+        results = filter.apply(self.db)
+        self.assertEqual(len(results), 18)
+
+    def test_family_and_their_spouses(self):
+        filter = self.filters["Family and their Spouses"]
+        results = filter.apply(self.db)
+        self.assertEqual(len(results), 25)
```

Let me know if there is way to do this appropriately.

dsblank (approved):
Very nice; thanks!

stevenyoungs:
hhmmmm I'm not sure why other tests are suddenly failing. I'm glad I made you co-author now @dsblank 🤣

dsblank:
I'll take a look :)

dsblank:
It is probably because I altered the global CustomFilters. Let me see if I can cleanup after the new tests.

dsblank:
Definitely caused by my test. Attempting to fix...

dsblank:
Okay, I think I have made the pytest and Gramps gods happy.

Nick-Hall:
Please add entries in the `po/POTFILES.skip` file for the new files.

stevenyoungs:
> Please add entries in the `po/POTFILES.skip` file for the new files.

Done.

kkujansuu:
Sorry to intrude but as far as I can see filtering still does not work in Gramps 6 even with this optimizer fix. See for example the filters below.

The filter F1 obviously should yield all individuals in the database (since it is an "or" filter with the "Everyone" rule).
But it only yields the same results as filter F3:

```
  <object type="Person">
    <filter name="F1" function="or">
      <rule class="Everyone" use_regex="False" use_case="False">
      </rule>
      <rule class="MatchesFilter" use_regex="False" use_case="False">
        <arg value="F2"/>
      </rule>
    </filter>
    <filter name="F2" function="and">
      <rule class="MatchesFilter" use_regex="False" use_case="False">
        <arg value="F3"/>
      </rule>
    </filter>
    <filter name="F3" function="and">
      <rule class="IsAncestorOf" use_regex="False" use_case="False">
        <arg value="I0041"/>
        <arg value="0"/>
      </rule>
    </filter>
  </object>

```
test

Or have I misunderstood something?

dsblank:
This looks like another test to add to the test suite. Thanks!

dsblank:
Added @kkujansuu 's additional issue as another set of unittests. Now we have one failure to fix in the optimizer.

stevenyoungs:
Here's a tabular version of the nice diagram @kkujansuu created. It includes additional information on the rules which have a `selected_handles` attribute.

|name|rule|logical_op|optimizer|
|--|--|--|--|
|F3|Ancestors of <person> "I0041"|and|selected_handles|
|F2|People matching the <filter> "F3"|and||
|F1|Everyone<br>People matching the <filter> "F2"|or||

The filter optimizer walks the tree of filters, looking for rules where
1. there is a `selected_handles` attribute
2. the logical_op is "and"

Where both conditions are true, the `selected_handles` are used to reduce the search space for any other, non-optimized rules. It reduces the search space by limiting handles which are checked in subsequent rules to those in `selected_handles` (`handles_in`) or for inverted rules, testing all handles in the db, excluding those handles in `selected_handles` (`handles_out`).

Earlier in this PR, I modified test 2 (`logical_op == 'and'`) to check the `logical_op` of the parent filter. The parent filter is F2 in this scenario. As can be seen, this also has a `logical_op` of `'and'`.

From the table above, we can see that the rule in filter F3 meets both conditions (`selected_handles` and `logical_op=='and'` [or `parent_logical_op=='and'`]). The optimizer therefore *only* considers the 16 handles resulting from filter F2 when evaluating the "Everyone" rule. Whilst this is quicker, it is also, unfortunately, wrong.

Alas, I'm still not sure what the correct fix is....

dsblank:
@stevenyoungs, I made a change [cc8bc77](https://github.com/gramps-project/gramps/pull/2052/commits/cc8bc775777df1f37a4891ca9d54d68a14a9d968) that I believe fixes the issues, but not 100% sure yet. Need further testing.

dsblank:
I don't think the failing tests have anything to do with these changes.

stevenyoungs:
> @stevenyoungs, I made a change [cc8bc77](https://github.com/gramps-project/gramps/pull/2052/commits/cc8bc775777df1f37a4891ca9d54d68a14a9d968) that I believe fixes the issues, but not 100% sure yet. 

Like you, I'm not entirely confident, but it feels wrong. `parent_logical_op` is now always going to be the `logical_op` of the root (top-most) filter. i.e. constant for a given execution of Optimizer. So every entry appended to `result` will have the same logical_op. 
`get_handles`, which tests the value of `logical_op` for each value in `result` (now called `all_selected_handles`) will consider everything as "and" or as "or" (because `parent_logical_op` is always the same value). so if one of the lower filters has a logical_op of "or", I believe it would get treated by `get_handles` as "and".

My assumption is that this works for @kkujansuu's example because the root filter is "or". If there was an F0 filter added on top, which used "and", my expectation is that the wrong result will be generated.

>  I think we can make up a tree of AND's and OR's that still fails.

I believe so too. I think we need a multi-level filter, where the filter on each level has multiple rules. We then alternate these filters to use "and" or "or". Then repeat with each filter using "or" and repeat again with each filter using "and". Then for completeness, repeat with negation. @kkujansuu's example leads us in this direction.

dsblank:
> Could something like this work?

Yes, I think so. The thing I missed in my first version was the ability to short-circuit any further processing when encountering an "or" at any level.

stevenyoungs:
I've taken @kkujansuu's idea and implemented. Initial results look promising.
His example filters from https://github.com/gramps-project/gramps/pull/2052#issuecomment-2834278763 now produce the correct results.

It does not yet handle inverted filters, therefore tests are expected to fail.

stevenyoungs:
Implemented support for inverted filters.

I'm not sure that "one" is handled correctly. I think it should use `symmetric_difference()` rather than `union`. `union` will give us results where more than one rule match

stevenyoungs:
> I'm not sure that "one" is handled correctly. I think it should use `symmetric_difference()` rather than `union`. `union` will give us results where more than one rule match

whilst the result of Optimizer is not what I expected, the end result is correct due to logic in `GenericFilter `
```
                    if apply_logical_op(db, obj, self.flist) != self.invert:
                        final_list.append(obj.handle)
```

I need to think about that...

stevenyoungs:
Something in `OptimizerTest.setUpClass()` and/or `OptimizerTest.tearDownClass()` appears to be causing 9 tests from `gramps.plugins.test.reports_test` to fail.

The errors are all of the form:
```
ERROR: test_WebCal (gramps.plugins.test.reports_test.TestDynamic)
  File "/home/runner/work/gramps/gramps/gramps/plugins/test/reports_test.py", line 93, in test_method
    self.assertTrue(test_function(out, err, report_name, **options))
  File "/home/runner/work/gramps/gramps/gramps/plugins/test/reports_test.py", line 171, in test_output_file
    raise Exception("can't find '%s' in order to delete it" % filename)
Exception: can't find '/tmp/WEBCAL' in order to delete it
```

I've modified `OptimizerTest` to avoid using a temporary file. I can't see anything else that could be the cause.

stevenyoungs:
It appears to be one or both calls to `filters.set_custom_filters` that cause the other tests to fail.
Once the problem is identified the 3 most recent commits can be discarded.

dsblank:
I'll take a look at this now.

dsblank:
Thank @kkujansuu for the suggestion, and @stevenyoungs for the implementation! I think this is much better not only that it is correct, but because it is easier to understand and debug.

Is this an accurate summary of the changes:

1. Optimizer now starts with the set of all handles, and removes non-matching ones. (This might have a slight impact on memory, but even 100k handles are not that big)
2. Now adds a new optimization for "only one of" types of filters where only one rule should apply
3. Collects the matching handles via "two functions that call each other recursively"

As we move forward, we can probably expand the tests (which all kinds of variations) to prove the correctness.

kkujansuu:
I think there are still a couple of special cases that don't work:

1. Empty filter with invert.

```
    <filter name="invert" function="and" invert="1">
    </filter>

```

2. The symmetric difference for "one" filter does not always work, e.g.

```
    <filter name="I0001 or I0002" function="one">
      <rule class="HasIdOf" use_regex="False" use_case="False">
        <arg value="I0001"/>
      </rule>
      <rule class="HasIdOf" use_regex="False" use_case="False">
        <arg value="I0002"/>
      </rule>
    </filter>
```

And I think it is not necessary to generate the "all_handles" sets. Instead one could use a special class instead of the sets. Something like

```
INCLUDE = 1
EXCLUDE = 2

@dataclass
class Handles:
    rectype: int  # INCLUDE or EXCLUDE
    handles: set[str]
    
```
Then 

    Handles(INCLUDE, set1)
    
would represent the set "set1"

    Handles(EXCLUDE, set1)

would represent the inverse of "set1", i.e. all handles not in set1.

    Handles(INCLUDE, set())

would be the empty set and

    Handles(EXCLUDE, set())
    
would represent "all handles".

The functions compute_xxx would return a Handles object instead of a set of handles.

One would also need to define the set operations for the Handles objects, e.g.

so that e.g.

    Handles(INCLUDE, set1) & Handles(INCLUDE, set2) == Handles(INCLUDE, set1 & set2)
    Handles(INCLUDE, set1) & Handles(EXCLUDE, set2) == Handles(INCLUDE, set1 - set2)

Does that make sense?

stevenyoungs:
> Thank @kkujansuu for the suggestion, and @stevenyoungs for the implementation! I think this is much better not only that it is correct, but because it is easier to understand and debug.
> 
> Is this an accurate summary of the changes:
> 
> 1. Optimizer now starts with the set of all handles, and removes non-matching ones. (This might have a slight impact on memory, but even 100k handles are not that big)
> 2. Now adds a new optimization for "only one of" types of filters where only one rule should apply
> 3. Collects the matching handles via "two functions that call each other recursively"
> 

Optimizer attempts to reduce the search space of the filter tree. The handles in this, possibly reduced, search space, are then evaluated against the filters (in `GenericFilter`)
Optimizer:
1. recurses the tree of filters in a depth first manner. 
2. When a leaf rule is reached, if returns one of
a. the selected_handles of the rule, if available
or
b. all_handles
3. Once all rules in a filter have been evaluated, (stored in `handlesets`) Optimizer uses the logical_op of the filter to combine the 0, 1 or more sets of results in `handlesets`, using intersection, union or symmetric_difference as appropriate

There some further optimisations that we can make 
1. for "union" if one of the sets in `handlesets` has the same size as `all_handles` we can avoid computing the union and simply return `all_handles`
2. for "interesction" if one of the sets in `handlesets` is empty, we can avoid computing the intersection and return an empty set
We can go further and avoid even calling `compute_potential_handles_for_rule`

> As we move forward, we can probably expand the tests (which all kinds of variations) to prove the correctness.

Yes, I think we can dynamically generate some tests using the example tree and the Everyone and Home Person rules, for example using "and", "or, and "one" rules. Additionally tests with an empty filter and more complex trees of filter using mixes of "and", "or, and "one".

stevenyoungs:
> 1. Empty filter with invert.

In Gramps 5.2, a filter with no rules
|logical_op|invert|result|
|---|---|---|
|and|False|all handles|
|and|True|zero handles|
|or|False|zero handles|
|or|True|all handles|
|one|False|zero handles|
|one|True|all handles|

`logical_op` behaves logically, inverting the result in all cases.
It's not entirely obvious why the non-inverted result is as it is. How does a filter with no rules actually match *anything*? Why does changing the `logical_op` vary the result - there is nothing to and\or\one of with.

stevenyoungs:
> 2\. The symmetric difference for "one" filter does not always work, e.g.

When a rule does not offer the `selected_handles` optimisation, all_handles is returned.
The "HasIdOf" rules does not offer `selected_handles` (currently) so both rules return all_handles. The symmetric_difference of two equal sets is an empty set. This is obviously wrong. Now we know why, I'll look at a fix.

gramps/gen/filters/optimizer.py
```diff
@@ -43,95 +44,73 @@ def union(sets):
         return set()
 
 
+def symmetric_difference(sets):
```
Nick-Hall, line 47: You could write this function as:
```
def symmetric_difference(sets):
    return reduce(lambda x, y: x^y, sets, set())
```

This will need `from functools import reduce` added to the imports.

The same approach could be used for _union_ and _intersection_.

gramps/gen/filters/optimizer.py
```diff
@@ -43,95 +44,73 @@ def union(sets):
...
-                    current_invert,
-                    filter.logical_op,
-                    rules_with_selected_handles,
-                )
-            )
+        self.all_handles = all_handles
+        self.handles_in = self.compute_potential_handles_for_filter(filter)
+        self.handles_out = set()
+
+    def compute_potential_handles_for_filter(self, filter):
```
Nick-Hall, line 71: You should add a docstring for this method.

gramps/gen/filters/optimizer.py
```diff
@@ -43,95 +44,73 @@ def union(sets):
...
+            handles = union(handlesets)
+        elif filter.logical_op == "one":
+            handles = symmetric_difference(handlesets)
+
+        if filter.invert:
+            handles = self.all_handles.difference(handles)
+
+        return handles
+
+    def compute_potential_handles_for_rule(self, rule):
```
Nick-Hall, line 90: You should add a docstring for this method.

Nick-Hall (changes requested):
Please add the docstrings.  I'll leave the code change decision up to you.

gramps/gen/filters/optimizer.py
stevenyoungs: Done

gramps/gen/filters/optimizer.py
stevenyoungs: Done

gramps/gen/filters/_genericfilter.py
```diff
@@ -144,51 +147,29 @@ def apply_logical_op_to_all(
...
+        optimizer = Optimizer(set(self.get_all_handles(db)), self)
+        handles_in = optimizer.get_handles()
 
         LOG.debug(
             "Optimizer handles_in: %s",
-            len(handles_in) if handles_in is not None else None,
+            len(handles_in),
         )
-        LOG.debug("Optimizer handles_out: %s", len(handles_out))
         if id_list is None:
```
stevenyoungs, line 157: I think there is a further optimization: when `id_list` is not None, we can precompute the set of handles and then pass this set into Optimizer to use as `all_handles`.
I don't want to pursue this until we have a working implementation.

gramps/gen/filters/optimizer.py
```diff
@@ -18,29 +19,18 @@
...
 
 def intersection(sets):
-    if sets:
-        result = sets[0]
-        for s in sets[1:]:
-            result = result.intersection(s)
-        return result
-    else:
-        return set()
+    return reduce(lambda x, y: x & y, sets, sets[0] if sets else set())
```
stevenyoungs, line 29: if any set in handlesets is empty, we can optimize by immediately returning set()

gramps/gen/filters/optimizer.py
stevenyoungs: Nice suggestion @Nick-Hall!
I've implemented this. For intersection I've avoided computing x&x when we only have a single set.

stevenyoungs:
> 2. The symmetric difference for "one" filter does not always work, e.g.
> 
> ```
>     <filter name="I0001 or I0002" function="one">
>       <rule class="HasIdOf" use_regex="False" use_case="False">
>         <arg value="I0001"/>
>       </rule>
>       <rule class="HasIdOf" use_regex="False" use_case="False">
>         <arg value="I0002"/>
>       </rule>
>     </filter>
> ```

@kkujansuu, this is now working and I've added tests for _one_ (xor), _or_ and _and_ variants of this filter.
Thanks for testing!

stevenyoungs:
> 1. Empty filter with invert.
> 
> ```
>     <filter name="invert" function="and" invert="1">
>     </filter>
> ```

Observations of the behaviour in Gramps v5.2 and v6.0 with this PR applied

logical_op | invert | v5.2|v6.0|
-- | -- | --| --|
|and|False|all handles|all handles|
|and|True|zero handles|zero handles|
|or|False|zero handles|zero handles|
|or|True|all handles|all handles|
|one|False|zero handles|zero handles|
|one|True|all handles|all handles|

I've added 6 new unit tests to capture the above.

gramps/gen/filters/_genericfilter.py
dsblank: Now that everything is working I can work on this, and the line 29 optimization if any set is empty. (I'll do some speed tests to see if that is worthwhile)

gramps/gen/filters/optimizer.py
dsblank: I generalized your basic idea: if you put the shorter sets first, then it will reduce faster.

stevenyoungs: Nice!
Given that you now have a sorted list of sets, for one if statement, you could test if sets[0] has length zero. Whether it is worth it is a different matter!

dsblank:
Although the optimization of `intersection` *only* reduced the time from milliseconds to nanoseconds, it was easy and could speedup could accumulate. I tested with large sets of handles, with an empty set at the end. Test code:

```python
import random
from functools import reduce
from gramps.gen.utils.id import create_id

def make_sets(n):
    sets = []
    for i in range(n):
        sets.append(set([create_id() for i in range(random.randint(20, 999_999))]))
    sets.append(set([]))
    return sets

sets = make_sets(10)
```

Without optimization:
```ipython
In [19]: %%time
    ...: functools.reduce(lambda x, y: x & y, sets, sets[0] if sets else set())
    ...: 
    ...: 
CPU times: user 46 ms, sys: 8.99 ms, total: 54.9 ms
Wall time: 54.9 ms
```

With optimization:
```ipython
In [20]: %%time
    ...: functools.reduce(lambda x, y: x & y, sorted(sets, key=lambda s: len(s)), sets[0] if sets else set())
    ...: 
    ...: 
CPU times: user 6 μs, sys: 1 μs, total: 7 μs
Wall time: 8.11 μs
```

The optimization of union could not make such large increases in speed, but I sped it quite a bit (4x) by using a trick to do more processing in C:

Without optimization:
```ipython
In [22]: %%time
    ...: x = reduce(lambda x, y: x | y, sets, set())
    ...: 
    ...: 
CPU times: user 1.79 s, sys: 653 ms, total: 2.45 s
Wall time: 2.45 s
```

With optimization:
```ipython
In [21]: %%time
    ...: x = set.union(*sets)
    ...: 
    ...: 
CPU times: user 289 ms, sys: 144 ms, total: 433 ms
Wall time: 433 ms
```

The optimization when you have an id_list was easy:

```python
        if id_list:
            optimizer = Optimizer(set(id_list), self)
        else:
            optimizer = Optimizer(set(self.get_all_handles(db)), self)
        possible_handles = optimizer.get_possible_handles()
```

gramps/gen/filters/_genericfilter.py
```diff
@@ -144,51 +147,34 @@ def apply_logical_op_to_all(
     ):
         final_list = []
 
-        optimizer = Optimizer(self)
-        handles_in, handles_out = optimizer.get_handles()
+        if id_list:
+            optimizer = Optimizer(set(id_list), self)
+        else:
+            optimizer = Optimizer(set(self.get_all_handles(db)), self)
```
stevenyoungs, line 153:
```suggestion
        optimizer = Optimizer(set(id_list if id_list else self.get_all_handles(db)), self) 
```
This might be cleaner

stevenyoungs: There's more to it than this. I'll add a commit

dsblank: Nice catch!

stevenyoungs:
sorry @dsblank, I hadn't intended to get in your way!

gramps/gen/filters/optimizer.py
dsblank: True, but it may be more expensive to do the test :) Anyway, I think this is clean, elegant, and not prone to errors.

gramps/gen/filters/_genericfilter.py
dsblank: Thanks for fixing the id_list speedup, and handling the tuple interface.

gramps/gen/filters/_genericfilter.py
```diff
@@ -142,76 +145,44 @@ def get_number(self, db):
     def apply_logical_op_to_all(
         self, db, id_list, apply_logical_op, user=None, tupleind=None, tree=False
```
stevenyoungs, line 146: The tree parameter does not appear to be used. Should it be?
It is a parameter of `GenericFilter.apply` from where it is passed straight through

dsblank:
@stevenyoungs I think if you are ready, this can be merged. Ended up with a couple of more optimizations along with the major bug fix. This is worth an immediate 6.0.2

CC: @Nick-Hall 

Nice work everyone!

gramps/gen/filters/_genericfilter.py
dsblank: I didn't alter the original API. We can add this to a clean up in 6.1 since we are in bug-fix mode on 6.0.

stevenyoungs:
> @stevenyoungs I think if you are ready, this can be merged. Ended up with a couple of more optimizations along with the major bug fix. This is worth an immediate 6.0.2
> 
See my comment above about the `tree` parameter to `apply` and `apply_logical_op_to_all`

The docstring of the return value from `apply` does not match the code. See line 236 of _genericfilter.py
```
        :Returns: if id_list given, it is returned with the items that
                do not match the filter, filtered out.
                if id_list not given, all items in the database that
                match the filter are returned as a list of handles
```

the code always returns the list of handles returned by `apply_logical_op_to_all`.

gramps/gen/filters/_genericfilter.py
stevenyoungs: in 5.2, `tree` is [used](https://github.com/gramps-project/gramps/blob/0f8d4ecd429431b4df64910962f8764af9ff1766/gramps/gen/filters/_genericfilter.py#L175C13-L175C86)
```
    def check_and(self, db, id_list, user=None, tupleind=None, tree=False):
        final_list = []
        flist = self.flist
        if user:
            user.begin_progress(_("Filter"), _("Applying ..."), self.get_number(db))
        if id_list is None:
            with self.get_tree_cursor(db) if tree else self.get_cursor(db) as cursor:
 ```

and then in `GenericPlaceFilter` this gives different behaviour

```
    def get_cursor(self, db):
        return db.get_place_cursor()

    def get_tree_cursor(self, db):
        return db.get_place_tree_cursor()
 ```
 
 I've changed the behaviour this when I introduced `self.get_all_handles(db)` - although clearly we do not have any tests for this.

dsblank:
https://github.com/gramps-project/gramps/pull/2052/files#diff-9970918f74dbce48e9687b440d5d9352c5f7909e6f42785ec4ba1ce27f2ff02dL170

gramps/gen/filters/_genericfilter.py
stevenyoungs: So far I cannot see any code that passes in the `tree` argument to `apply`. It can only be a filter of which the class or super class is `GenericPlaceFilter`.  `GenericPlaceFilter` is the only type where `get_cursor` and `get_tree_cursor` give different results.

dsblank: I noted the missing code below.

dsblank:
The only object where get_tree_cursor() is different from get_cursor() is in the `place` filters.

dsblank:
And if id_list is used, it needs to return objects (or data?), not handles.

gramps/gen/filters/_genericfilter.py
stevenyoungs: Can you find anywhere where the tree parameter is set to True?

dsblank:
I can take a look at what is missing, and add tests.

stevenyoungs:
> I can take a look at what is missing, and add tests.

I've added support for the `tree` parameter and in the process discovered that the "Every place" yields a result of nothing :-(

stevenyoungs:
1. How on earth did the tests pass when I broke `intersection` when I converted it to use `reduce`?
This reinforces my concern that coverage is not yet good enough
2. I'm concerned that placemodel is relying on the order of results.
So, now that the "Every place" filter returns the correct number of handles, the display is not as expected. I believe that this is down to [`PlaceModel.add_row`](https://github.com/gramps-project/gramps/blob/96f2b6bc04fa248dca85e15763efdb513c9db0ed/gramps/gui/views/treemodels/placemodel.py#L365) expecting parent places to be added to the model _first_.
In the latest version of this PR, place P1354 is the first handle passed to add_row. It does have a parent place. The parent place is not already in the model. So place P1354 is added as a root node. This is incorrect. I would expect one of places [P0441, P0442, P0957] to be the first ones added, these being the true root node countries Greece, Russia, USA. We now store the `possible_handles` in a `set`, which may have a different order than the order in which the handles were retrieved via  `self.get_tree_cursor`.

Nick-Hall:
> I'm concerned that placemodel is relying on the order of results.

Yes. That is why we use tree cursors.

stevenyoungs:
I see two options to solve this
1. when `tree==True`, do not use `Optimizer` and process handles in the order returned by `get_tree_cursor`
filter performance for Places will be broadly the same as gramps 5.2
2. continue to use Optimizer and then at the end of `apply`, sort the result back into the expected order
filter performance for Places should be better than Gramps 5.2, so long as the gains from `Optimizer` outweigh the final sorting of the result

@dsblank any thoughts?

stevenyoungs:
I've coded the second option

> continue to use Optimizer and then at the end of apply_logical_op_to_all, sort the result back into the expected order

The code is shorter than the comment.
Some unit tests of Place filters are required that check not only the expected number of handles are returned but that those handles are also in the expected order.

gramps/gen/filters/_genericfilter.py
stevenyoungs: For the record, the Places view sets tree=True

stevenyoungs:
> The docstring of the return value from `apply` does not match the code. See line 236 of _genericfilter.py
> 
> ```
>         :Returns: if id_list given, it is returned with the items that
>                 do not match the filter, filtered out.
>                 if id_list not given, all items in the database that
>                 match the filter are returned as a list of handles
> ```
> 
> the code always returns the list of handles returned by `apply_logical_op_to_all`.

I've added code to return tuples when the input id_list is tuples. We need to check the performance.  A simple test is to go to the Events view and in the search bar set Type=Birth and click Find

> @stevenyoungs I think if you are ready, this can be merged. Ended up with a couple of more optimizations along with the major bug fix. This is worth an immediate 6.0.2

All problems that I'm aware of are now addressed 🤞
This PR has grown considerably, and taken a number of wrong turns. As such it needs careful review and testing.

dsblank:
I'm free for the next few hours to look over the changes and add more tests. I can look at the typing conflicts too.

stevenyoungs:
> I'm free for the next few hours to look over the changes and add more tests. I can look at the typing conflicts too.

Thanks. I'll try and get the typing fixed but will leave you to review and look at unit tests.

gramps/gen/filters/_genericfilter.py
```diff
@@ -300,6 +324,9 @@ class GenericFamilyFilter(GenericFilter):
     def __init__(self, source=None):
         GenericFilter.__init__(self, source)
 
+    def get_all_handles(self, db):
```
dsblank, line 327: You can actually avoid having to add another method to all of the filter types.

gramps/gen/filters/_genericfilter.py
```diff
@@ -282,9 +278,37 @@ def apply(self, db, id_list=None, tupleind=None, user=None, tree=False):
...
+                    data[tupleind]: data for data in cast(List[Tuple], id_list)
+                }
+                possible_handles = set(handle_tuple.keys())
+            else:
+                possible_handles = set(cast(List[PrimaryObjectHandle], id_list))
+        elif tree:
+            tree_handles = [handle for handle, obj in self.get_tree_cursor(db)]
+            possible_handles = set(tree_handles)
+        else:
+            possible_handles = set(self.get_all_handles(db))
```
dsblank, line 299: We already have a method specific to the filter type. Let's use it, and make it parallel with the tree case:

```suggestion
            all_handles = [handle for handle, obj in self.get_cursor(db)]
            possible_handles = set(all_handles)
```

dsblank (changes requested):
You can remove all of the get_all_handles() methods.

stevenyoungs:
`IsDescendantFamilyOf` was adding `None` to it's list of `selected_handles`.  Preventing this allows GenericFilter to also avoid remove `None` from `possible_handles`. Conceptually this is the right change. It remains to be seen if other files add `None` to their `selected_handles`

dsblank:
This is a clever and efficient method to not lose the order of the id_list, considering it can only shrink (be filtered) in the result.

```python
res = sorted(res, key=lambda x: tree_handles.index(x))
```

gramps/gen/filters/_genericfilter.py
stevenyoungs: Speed was my concern. `get_all_handles` is O(1) which is better the O(n) of using a `cursor`. The `cursor` also constructs the object which we do not need at this stage.

dsblank:
A possibility occurs to me (for Gramps 6.1):  if all of the filter.prepare() methods were lazy (eg used set comprehension rather than sets) then we needn't compute the actual members until needed. But that would remove the optimization based on set length (the sort). But could be worth it overall.

stevenyoungs:
> This is a clever and efficient method to not lose the order of the id_list, considering it can only shrink (be filtered) in the result.
> 
> ```python
> res = sorted(res, key=lambda x: tree_handles.index(x))
> ```

My only doubt is performance, but we can optimise by building a Dict[PrimaryObjectHandle, index] if so.

stevenyoungs:
> A possibility occurs to me (for Gramps 6.1): if all of the filter.prepare() methods were lazy (eg used set comprehension rather than sets) then we needn't compute the actual members until needed. But that would remove the optimization based on set length (the sort). But could be worth it overall.

And some other ideas:
1. For some rules, rule.selected_handles is the exact result. Currently we ignore that and still call rule.apply_to_one in GenericFilter
2. prepare currently works across the entire db. If it knew the starting possible_handles, it may be able to reduce the scope of it's work.

Let's get 6.0 working first!

gramps/gen/filters/_genericfilter.py
stevenyoungs: Unless you feel strongly, I will leave the new `get_all_handles` methods

Nick-Hall:
> Let's get 6.0 working first!

Yes.  You should keep the scope of this PR to the bug fixes.  I'm not too worried about performance.

> This PR has grown considerably, and taken a number of wrong turns. As such it needs careful review and testing.

I agree.  This is a much bigger change than we would normally accept in a maintenance branch.

We need to test every rule.  Are there any rules without unit tests?

Then we need to test combinations of rules in filters.

After that we need to do more testing, and then even more testing.

dsblank:
I just added tests for id_list, tupleind, and tree.

gramps/gen/filters/_genericfilter.py
dsblank: Yes, you have a point. Actually I think they are both O(n), but one is in SQL and the other is in Python. Let's leave it as you have it.

dsblank:
> We need to test every rule. Are there any rules without unit tests?

Every rule has a test.

> After that we need to do more testing, and then even more testing.

I believe that we have tests for all of the options. Of course we can continue to add tests (there are a large combination of rules x options). We also now have tests for things that have NEVER been tested before. 

But we also have to balance that fact with the issue that we currently have a broken release.

I would suggest that we test and us this for a couple of more days and get a release out sooner than later. This is a serious breakage.

gramps/gen/filters/test/filter_optimizer_test.py
```diff
@@ -183,9 +182,12 @@ def test_everyone_with_id_list_tupleind_3(self):
         filter = self.filters["Everyone"]
         default_person = self.db.get_default_person()
         id_list = [(default_person, default_person.handle)]
-        with pytest.raises(IndexError) as excinfo:
```
Nick-Hall: You can use [self.assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) here.

dsblank (approved):
LGTM!

gramps/gen/filters/optimizer.py
```diff
@@ -18,120 +19,86 @@
...
-        for s in sets[1:]:
-            result = result.intersection(s)
-        return result
-    else:
-        return set()
+    # sort the sets by length, shortest first.
+    # with interrsection, the shortest of the starting sets determines the
+    # maximum size of the result
+    sorted_sets = sorted(sets, key=lambda s: len(s))
+    return reduce(lambda x, y: x & y, sorted_sets[1:], sorted_sets[0])
```
dsblank, line 33: This is equally fast and perhaps a little easier to read:

```suggestion
    return set.intersection(*sorted_sets)
```

And then we also don't have to import/use functools.

stevenyoungs:
Another bug: https://gramps-project.org/bugs/view.php?id=13799#c68867

This is calling apply with id_list, tupleind == 1, tree == False
I'm guessing that the problem is the order of results in the scenario.

gramps/gen/filters/_genericfilter.py
```diff
@@ -282,9 +278,37 @@ def apply(self, db, id_list=None, tupleind=None, user=None, tree=False):
...
 
         start_time = time.time()
-        res = self.apply_logical_op_to_all(
-            db, id_list, apply_logical_op, user, tupleind, tree
-        )
+
+        # build the starting set of possible_handles to be filtered
+        possible_handles: Set[PrimaryObjectHandle]
+        if id_list is not None:
+            if tupleind is not None:
```
stevenyoungs, line 285:
```suggestion
        if id_list:
            if tupleind is not None:
```
Good catch on tupleind to avoid problems when it is zero. I'm not sure we need to use `is not None` for id_list though

stevenyoungs:
> Another bug: https://gramps-project.org/bugs/view.php?id=13799#c68867
> 
> This is calling apply with id_list, tupleind == 1, tree == False I'm guessing that the problem is the order of results is different to the input order in this scenario.

This is now working correctly.

stevenyoungs:
Hopefully we now have the holy grail of performance and correctness.
I've no further changes planned and this is ready for review from my perspective.

dsblank:
I won't have further time to add more tests for awhile, and this looks very solid to me.

stevenyoungs:
> Found another anomaly.
> 
> On the example database, LAGvZ is the only person with the tag "ToDo".
> 
> Build the custom filter: People with the Tag="ToDo" and set the options to **Return values that do not match the filter rules**
> 
> Running the filter returns zero (0) matches.

I can reproduce this.
With **Return values that do not match the filter rules** _unticked_ the correct result of 1 person (LAGvZ) is returned.
With **Return values that do not match the filter rules** _ticked_, no people are returned.

The error originates in Optimizer method `compute_potential_handles_for_filter`
```        
        if filter.invert:
            handles = self.all_handles.difference(handles)
```

`handles` and `self.all_handles` are the same so the result of calling `difference` is empty.

Working backwards, `compute_potential_handles_for_rule` returns `self.all_handles`, because this rule is not currently optimized.
I'm not sure it is safe to carry out the "invert" calculation where we currently do. There are two scenarios where it will give the wrong result
1. a rule is not optimized
this is the current scenario. we end up discarding all handles
2. a rule is optimized but the `selected_handles` is an over estimate
we will discard those handles from `selected_handles` which do not match the rule(s) [and are discarded early because of the invert]

In both cases, we will end up discarding handles too early

dsblank:
@stevenyoungs do you want me to add the tests that fail, or do you have them coded already?

stevenyoungs:
> @stevenyoungs do you want me to add the tests that fail, or do you have them coded already?

That would be great. I've not done any more than the investigation in my comment above so far.

stevenyoungs:
> > @stevenyoungs do you want me to add the tests that fail, or do you have them coded already?
> 
> That would be great. I've not done any more than the investigation in my comment above so far.

@dsblank I've added a couple of additional tests, one of which should fail at the moment
I then have a potential fix to push

stevenyoungs:
> Found another anomaly.
> 
> On the example database, LAGvZ is the only person with the tag "ToDo".
> 
> Build the custom filter: People with the Tag="ToDo" and set the options to **Return values that do not match the filter rules**
> 
> Running the filter returns zero (0) matches.

Good news first
* the above problem is now fixed.

Now the bad news
* I'm not particularly happy with the fix. 
Something feels wrong and then....
* the code is still vulnerable if a rule is optimised and provides a superset of the potentially matching handles.
For example if I optimize the Tag rule we used in this problem such that it always sets `selected_handles` to every handle in the DB. This matches the specification; `selected_handles` is a superset of the actual rule result. But when we invert inside optimizer, we will discard all handles and essentially recreate @DaveSch-gramps problem.
* If we keep the contract that a rules `selected_handles` is a superset of the handles matching the rule, then I'm not sure that we can get invert working correctly within Optimizer.

dsblank:
I have time today to look at this

DaveSch-gramps:
The last fix works the main rule is the filter to return results that are NOT the rule.

Create a filter to find positive results. Then Create a second filter that takes the first filter to find the negative results. No results are returned for the second filter.

These custom filters were created and tested on the example database.

<filters>
  <object type="Person">
    <filter name="Family Name" function="and">
      <rule class="HasNameOf" use_regex="False" use_case="False">
        <arg value=""/>
        <arg value=""/>
        <arg value=""/>
        <arg value=""/>
        <arg value=""/>
        <arg value=""/>
        <arg value=""/>
        <arg value="Garner"/>
        <arg value=""/>
        <arg value=""/>
        <arg value=""/>
      </rule>
    </filter>
    <filter name="NOT Family Name" function="and" invert="1">
      <rule class="MatchesFilter" use_regex="False" use_case="False">
        <arg value="Family Name"/>
      </rule>
    </filter>
  </object>
</filters>

DaveSch-gramps:
I had a hard time triaging the previous post. I have a large database (239K people) and running even some of the simplest filters were taking 30+ minutes to resolve/display the list of people. Gramps was displaying "Not Responding". Actually running the filter was not the issue. Running the same filters on 5.2.4 results were returned 2-3 seconds after the final final "applying filter" progress bar was completed.

dsblank:
> I had a hard time triaging the previous post. I have a large database (239K people) and running even some of the simplest filters were taking 30+ minutes to resolve/display the list of people. Gramps was displaying "Not Responding". Actually running the filter was not the issue. Running the same filters on 5.2.4 results were returned 2-3 seconds after the final "applying filter" progress bar was completed.

Thanks @DaveSch-gramps for the info. The optimizer should never in any case take longer, so there must be a bug. I'm adding your tests and that will help track the issue down.

dsblank:
I think this version properly uses the new recursive calls. I'll have to think about what we can do with "one" and "or". I tested also to make sure that all of the items that were previously optimized still are.

emyoulation:
As one of the "others", let me express my gratitude.

gramps/gen/filters/optimizer.py
```diff
@@ -17,121 +18,93 @@
...
-        particular handles. If it is a set, then those in the set
-        are a superset of the items that will match.
 
-        `handles_out` is a set. If any handle is in the set, it will
-        not be included in the final results.
+class Optimizer:
+    def compute_potential_handles_for_filter(
+        self, filter: GenericFilter
+    ) -> Tuple[
+        Union[Set[PrimaryObjectHandle], None], Union[Set[PrimaryObjectHandle], None]
```
stevenyoungs, line 60:
```suggestion
        Set[PrimaryObjectHandle] | None, Set[PrimaryObjectHandle] | None
```

In other places, `|` is used in preference to the more verbose `Union[]`

gramps/gen/filters/optimizer.py
```diff
@@ -17,121 +18,93 @@
...
-            handles_in = handles_in - handles_out
-
-        LOG.debug("optimizer handles_in: %s", len(handles_in) if handles_in else 0)
         return handles_in, handles_out
+
+    def compute_potential_handles_for_rule(
+        self,
+        rule: Rule,
+    ) -> Tuple[
+        Union[Set[PrimaryObjectHandle], None], Union[Set[PrimaryObjectHandle], None]
```
stevenyoungs, line 101:
```suggestion
        Set[PrimaryObjectHandle] |None, Set[PrimaryObjectHandle] | None
```

gramps/gen/filters/optimizer.py
```diff
@@ -17,121 +18,93 @@
...
-        return result
-    else:
-        return set()
-
+# -------------------------------------------------------------------------
+#
+# Typing modules
+#
+# -------------------------------------------------------------------------
+from typing import List, Set, TYPE_CHECKING, Tuple, Union
```
stevenyoungs, line 34:
```suggestion
from typing import List, Set, TYPE_CHECKING, Tuple
```

gramps/gen/filters/optimizer.py
```diff
@@ -54,79 +54,53 @@ def union(sets: List[Set[PrimaryObjectHandle]]) -> Set[PrimaryObjectHandle]:
...
+                if rule_in is not None:
+                    if handles_in is None:
+                        handles_in = rule_in
+                    else:
+                        handles_in.intersection(rule_in)
+                if rule_out is not None:
+                    if handles_out is None:
+                        handles_out = rule_out
+                    else:
+                        handles_out.intersection(rule_out)
```
stevenyoungs, line 79: Should this be union?
Given a filter with two rules and logical_op "and"
rule 1 says handle 'a' is out. So the path `handles_out = rule_out` is executed
rule 2 says handle 'b' is out. So the path `handles_out.intersection(rule_out)` is executed. This expands to 'a'.intersection('b') which is obviously an empty set

Logically, neither of handles 'a' or 'b' can be in the result so I would expect `handles_out` for the filter to be the set ('a', 'b')

stevenyoungs: I think it will produce the correct result as it is, but `GenericFilter` will call `apply_logical_op` for handles 'a' and 'b' when it doesn't really need to

dsblank: > Should this be union?

I think intersection is correct, because the set can only get smaller.

dsblank: > I think it will produce the correct result as it is, but GenericFilter will call apply_logical_op for handles 'a' and 'b' when it doesn't really need to

Not sure I follow this. You want to make a suggestion?

dsblank: > This expands to 'a'.intersection('b') which is obviously an empty set

I think it is the same set. I think intersection is correct.

```ipython
In [1]: a = {1, 2, 3}
In [2]: b = {1, 2, 3}
In [3]: a.intersection(b)
Out[3]: {1, 2, 3}
```

stevenyoungs:
> Thanks @DaveSch-gramps! With the last fix, and more tests, I think we finally have all cases covered.

Brave words given the history of this bug and PR - but I'm also feeling optimistic now.

stevenyoungs:
`test_not_i0001_and_i0002` uses the filter "not (I0001 and I0002)" which in turn requires two sub-filters "I0001" and "I0002"
These sub-filters did not exist, yet the unit tests were still passing. Messages were generated in the build output
```
Can't find filter I0001 in the defined custom filters
Can't find filter I0002 in the defined custom filters
```

I've fixed the immediate problem by defining the filters "I0001" and "I0002". In my view the unit test should also have failed.

dsblank:
> I've fixed the immediate problem by defining the filters "I0001" and "I0002". In my view the unit test should also have failed.

Good catch. Looking into how the tests could have passed...

dsblank:
> Looking into how the tests could have passed...

I see... it returns all handles, which is correct. I'll add more tests where "all" doesn't necessarily give a good answer.

stevenyoungs:
> @stevenyoungs if you test with either #2071 or #2062 you'll see that 12 tests fail. I like the idea, but we should probably wait for 6.1 to add features to this PR.

Good find. I've added a test that catches this scenario. The build has just failed, due to this test, so it is working.
Now I'll reorder the commits so we have the test first.

gramps/gen/filters/optimizer.py
```diff
@@ -17,121 +18,97 @@
...
 
-def union(sets):
-    if sets:
-        result = sets[0]
-        for s in sets[1:]:
-            result = result.union(s)
-        return result
-    else:
-        return set()
+LOG = logging.getLogger(".filter.optimizer")
```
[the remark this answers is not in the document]
dsblank: I don't think there is.

PR #2055, opened 2025-04-27
rename Rule.apply to apply_to_one

stevenyoungs:
In gramps 6.0, the `apply` method of a `Rule` was renamed to `apply_to_one`. The method in `Rule` itself appears to have been missed. Fix that.


emyoulation:
The rules that fail because of this renaming are too well hidden.

If the old method name is no longer valid, could that name be re-purposed to be a Warning generator instead of a confusing crash?

If the warning identified the the calling Rule and Custom Filter, users will have a better chance of finding and fixing.

stevenyoungs:
The current behaviour (before this change) is, I agree, bad. The default implementation of `apply` returns True and so all objects match the rule. This silently breaks rules that have not been updated for gramps 6.0. The rule appears to the user to have worked correctly, but the result is likely to be wrong. See for example the `IsRelatedWithFilterMatch` rule from the FilterRules addon.
With this change, the `IsRelatedWithFilterMatch` rule now fails with an error 

```
    if person and self.filt.apply(db, person):
                  ^^^^^^^^^^^^^^^
AttributeError: 'MatchesFilter' object has no attribute 'apply'
```
 which in my view both (a) means the user does not get the wrong result and (b) makes it easier to diagnose the underlying problem.

I agree that the `apply` function could be repurposed to generate a warning. The warning may be slightly more user friendly than the above error. We would still need to generate an error to avoid the rule returning an incorrect result and it adds another message for translation. I am unsure of the benefit of doing this. The core rules have all been updated. I have a [PR](https://github.com/gramps-project/addons-source/pull/715) pending to update the FilterRules addon. The only other addon requiring checking \ updating is FilterRules2. We probably don't have any data on who else has written custom rules outside of the gramps github org.

dsblank (approved):
Thank you! LGTM!

dsblank:
@Nick-Hall this should be ready to merge for 6.0 bug fixes.

PR #2057, opened 2025-04-30
Respect user choice of CSS files for existing narrated web site

hgohel:
Fixes [#13792](https://gramps-project.org/bugs/view.php?id=13792)

When a Narrated Web Site is generated to an existing folder and the CSS for the web pages is changed, the screen CSS would not be overwritten. This has been changed so that the user's choice of CSS is correctly copied to the output directory

### DIFF
--- a/gramps/plugins/webreport/narrativeweb.py
+++ b/gramps/plugins/webreport/narrativeweb.py
@@ -1269,7 +1269,7 @@ class NavWebReport(Report):
         # copy screen style sheet
         if CSS[self.css]["filename"]:
             fname = CSS[self.css]["filename"]
-            self.copy_file(fname, _NARRATIVESCREEN, "css")
+            self.copy_file(fname, _NARRATIVESCREEN, "css", True)
 
         # copy printer style sheet
         fname = CSS["Print-Default"]["filename"]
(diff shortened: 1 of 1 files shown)


hgohel:
@SNoiraud This address defect 13792 and I would like your review on it. There are other files being copied to the output folder which are not overwritten either; should all of them be overwritten in case of changes, or stick with just this one file?

SNoiraud:
The origin of this:
When you copy files (images, ...), if these files are already copied because they are shared, we do not copy them to save time. If you use archives, we save memory.

hgohel:
@SNoiraud Thanks for confirming the rationale for current behavior. Reading in-between the lines, I take you would prefer to make minimal changes and overwrite only this file and not the others. Removed draft status of PR; it is ready for further review and merging.

@Call-Me-Dave Open to your comments on this PR in the context of the bug report if you would verbalize them.

hgohel:
The fix was tested and validated by reporter on Linux, and validated by me on Windows as well. Ready to merge.

Call-Me-Dave:
> @Call-Me-Dave Open to your comments on this PR in the context of the bug report if you would verbalize them.

It was meant to be a thumbs up emoticon to represent that I like this 👍

PR #2059, opened 2025-05-08
Fixes #0013789 Incorrect addons project after upgrade from Gramps 5.2

prculley:
Upgrades from older Gramps (Gramps v5.2.x) copy the gramps.ini from the older version, including the "behaviour.addons-projects" values, which then point to the older github addons.  This fixes that.

### DIFF
--- a/gramps/gen/config.py
+++ b/gramps/gen/config.py
@@ -41,7 +41,7 @@ import logging
 # Gramps imports
 #
 # ---------------------------------------------------------------
-from .const import USER_CONFIG, USER_DATA, USER_HOME, VERSION_DIR
+from .const import USER_CONFIG, USER_DATA, USER_HOME, VERSION_DIR, VERSION_DIR_NAME
 from .utils.configmanager import ConfigManager
 from .const import GRAMPS_LOCALE as glocale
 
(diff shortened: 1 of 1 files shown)


emyoulation:
Addresses https://gramps-project.org/bugs/view.php?id=13789

emyoulation:
Instead of resetting to defaults, could this just deselect all but the 1st (default) "Project" instead of flushing the row?

This would allow users to Edit the URL to a current version path without having to find the URL or remember which Projects that they had added.  

Also, if the Addon Manager becomes smarter about not trying to install `gramps_target_version` addons that are mismatches, then the JSON might have mixed sets of `gramps_target_version` rows. ( I do not know how the "major_version" hack in Isotammi .gpr.py will work/fail with smarter `gramps_target_version` attribute handling.) 

Also, if a URL Validator enhancement (as suggested in [13069](https://gramps-project.org/bugs/view.php?id=13069)) is added, then this routine could 'invalidate' as well as deselect project rows.

There are a series of Addon Manager feature requests and bugs (with some duplication) linked to: 
https://gramps-project.org/bugs/view.php?id=13069

PR #2062, opened 2025-05-11
Refactor "Has Id Of" rules for filter Optimizer

stevenyoungs:
Refactor the set of "Has Id Of" rules to allow the filter Optimizer to take the fast path

### DIFF
--- a/gramps/gen/filters/rules/citation/_hasidof.py
+++ b/gramps/gen/filters/rules/citation/_hasidof.py
@@ -34,6 +34,7 @@ _ = glocale.translation.gettext
 #
 # -------------------------------------------------------------------------
 from .. import HasGrampsId
+from ....db import Database
 
 
 # -------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


gramps/gen/filters/rules/citation/_hasidof.py
```diff
@@ -46,3 +47,4 @@ class HasIdOf(HasGrampsId):
 
     name = _("Citation with <Id>")
     description = _("Matches a citation with a specified Gramps ID")
+    _get_raw_object_from_id_data = staticmethod(Database._get_raw_citation_from_id_data)
```
Nick-Hall, line 50: The `_get_raw_citation_from_id_data` method is protected and should only be used in sub-classes.  It does not form part of the public API.

stevenyoungs: Agreed. I also think it should be part of the public API. It makes no sense to construct the full object here when only the handle is required.

dsblank: Looking over the later messages, perhaps there is some agreement over most of the devs.

PR #2063, opened 2025-05-12
Dev version correctly reports the SHA of the local HEAD commit

hgohel:
Local builds report the Gramps version in the form of Major.Minor.Patch-SHA. However, the commit SHA reported did not correspond to the repo root, but that of the gramps/gramps folder which was unexpected. This change fixes that by removing the path argument to get_git_revision so the git log command does not limit commits, and the dev version is reported as expected.

A brief history: the path argument was a carry-over from SVN, and wasn't initially used (see commit 57daf4f), but a subsequent change attempted to use it (commit 6836dbd), and finally the path argument was appended to the git log command in commit f92ee32.

To test this change, run `python Gramps.py -v` from the root of the source directory, or other subfolders (adjusting the path to Gramps.py as required) and verify that the version reported matches the HEAD commit of your source.

Testing performed/needed
- [X] MSYS environment on Windows
- [X] Ubuntu (in a WSL environment)
- [ ] Apple MacOS
- [x] Linux

### DIFF
--- a/gramps/gen/const.py
+++ b/gramps/gen/const.py
@@ -182,9 +182,7 @@ ROOT_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir))
 
 sys.path.insert(0, ROOT_DIR)
 if DEV_VERSION:
-    git_revision = get_git_revision(ROOT_DIR).replace("\n", "")
-    if sys.platform == "win32" and git_revision == "":
-        git_revision = get_git_revision(os.path.split(ROOT_DIR)[1])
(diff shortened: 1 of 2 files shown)


Nick-Hall:
This works for me using Arch Linux.

hgohel:
@Nick-Hall CI check seems to be hung on 3 of 4 tasks since yesterday. Can it be restarted? How?

Nick-Hall:
The CI checks are passing.  The 3 of 4 tasks refer to your check boxes.

PR #2065, opened 2025-05-13
Avoid multiple copies of Rules after Plugin manager reload

prculley:
If the Plugin Manager "Reload" button is used (to reload current plugins), then the filter rules list shown in the edit filter dialog add rule will get multiple copies of addon type rules.  This doesn't affect the standard plugin rules.

Fixes [#13844](https://gramps-project.org/bugs/view.php?id=13844).

### DIFF
--- a/gramps/gen/plug/_manager.py
+++ b/gramps/gen/plug/_manager.py
@@ -227,7 +227,8 @@ class BasePluginManager:
             # make the new rule findable via import statements
             setattr(obj_rules, plugin.ruleclass, r_class)
             # and add it to the correct fiter editor list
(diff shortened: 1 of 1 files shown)


gramps/gen/plug/_manager.py
```diff
@@ -227,7 +227,8 @@ def reg_plugins(
             # make the new rule findable via import statements
             setattr(obj_rules, plugin.ruleclass, r_class)
             # and add it to the correct fiter editor list
```
stevenyoungs, line 229:
```suggestion
            # and add it to the correct filter editor list
```
Fix spelling

PR #2066, opened 2025-05-13
Fix crash in Citation View

stevenyoungs:
Fix a crash in Citation View will when sorting by "Last Changed" date.

To reproduce:

1. Start with Gramps60 branch with https://github.com/gramps-project/gramps/pull/2060 in place.
2. Preconditions: in Citation View, Sidebar turned off (Menu/View/Sidebar). Configure the view so "Last Changed" is visible.
3. Open a tree with some citations.
4. In Citation View, click on "Last Changed" column to resort it.

Fixes bug [0013796](https://gramps-project.org/bugs/view.php?id=13796)

### DIFF
--- a/gramps/gui/views/listview.py
+++ b/gramps/gui/views/listview.py
@@ -787,8 +787,13 @@ class ListView(NavigationView):
         if not self.search_bar.is_visible():
             filter_info = (True, self.generic_filter, False)
         else:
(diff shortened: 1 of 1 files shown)


stevenyoungs:
Let me know if you want me to split this PR. I kept the commits separate for this eventuality. Only the first one is required for 6.0

PR #2067, opened 2025-05-15
Fix broken compound dates with bce year in XML import

prculley:
If date spans have a negative year, the import of the date fails and converts to a text string.

A similar issue was fixed for normal dates and I copied the solution.

The issue goes back a ways, I found it in gramps52...  Might want to look into back porting this.

Fixes [#13631](https://gramps-project.org/bugs/view.php?id=13631).


hgohel:
@prculley CI build failure may have been a transient issue. Since I don't know how to kick off a new CI build, I'm thinking closing the PR and re-opening it will trigger it, if you want to try it.

PR #2068, opened 2025-05-19
Connect the Help button in the repository reference editor

stevenyoungs:
Previously the button was displayed but not connected to anything

Fixes [0013352](https://gramps-project.org/bugs/view.php?id=13352)

### DIFF
--- a/gramps/gui/editors/editreporef.py
+++ b/gramps/gui/editors/editreporef.py
@@ -36,11 +36,21 @@ from gramps.gen.const import GRAMPS_LOCALE as glocale
 _ = glocale.translation.gettext
 from gramps.gen.lib import NoteType
 from gramps.gen.db import DbTxn
+from gramps.gen.const import URL_MANUAL_SECT2
 
 from .displaytabs import NoteTab, AddrEmbedList, WebEmbedList, SourceBackRefList
 from ..widgets import MonitoredEntry, PrivacyButton, MonitoredDataType
 from .editreference import RefTab, EditReference
 from ..glade import Glade
+
+# -------------------------------------------------------------------------
+#
+# Constants
+#
+# -------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


gramps/gui/editors/editreporef.py
```diff
@@ -36,12 +36,22 @@
...
 from ..glade import Glade
 
+# -------------------------------------------------------------------------
+#
+# Constants
+#
+# -------------------------------------------------------------------------
+
+WIKI_HELP_PAGE = URL_MANUAL_SECT2
+WIKI_HELP_SEC = _("Repository_Reference_Editor", "manual")
```
Nick-Hall, line 53: References to manual pages should no longer be translated.  The translation should be handled by page re-direction in the wiki instead.

Nick-Hall (dismissed):
References to manual pages should no longer be translated. The translation should be handled by page re-direction and anchors in the wiki instead.

gramps/gui/editors/editreporef.py
```diff
@@ -50,7 +50,7 @@
 # -------------------------------------------------------------------------
 
 WIKI_HELP_PAGE = URL_MANUAL_SECT2
-WIKI_HELP_SEC = _("Repository_Reference_Editor", "manual")
+WIKI_HELP_SEC = ("Repository_Reference_Editor", "manual")
```
Nick-Hall: This looks wrong.  The string "manual" is a translation context.

Nick-Hall:
> Is any change required on the wiki side?

Changes to the wiki will only be required when changing a reference to an existing wiki entry.

gramps/gui/editors/editreporef.py
stevenyoungs: I've revised the code and tested in English. The correct page is displayed.

PR #2070, opened 2025-05-19
Bug 13702: EditNote Callback: Pass object not handle

stevenyoungs:
Fix EditNote so that it passes the object when calling the callback function, rather than the object handle. 
Fixes the error described in 13702 and gives consistency with other Edit<Obj> classes.

To reproduce:
1. create a new note- type "a"
2. select "a" and click the Link button
3. Change the Link Type to Note and click New
4. type "a" and click ok

Fixes [#13702](https://gramps-project.org/bugs/view.php?id=13702) and possibly [#12260](https://gramps-project.org/bugs/view.php?id=12260)

Reviewed and adjusted other places that create an EditNote instance with callback

### DIFF
--- a/gramps/gui/editors/displaytabs/notetab.py
+++ b/gramps/gui/editors/displaytabs/notetab.py
@@ -166,13 +166,13 @@ class NoteTab(EmbeddedList, DbGUIElement):
         except WindowActiveError:
             pass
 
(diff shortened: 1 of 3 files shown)


hgohel:
Looks good to me!

PR #2071, opened 2025-05-28
Add optimization to has-id rules

dsblank:
This is an alternative implementation to #2062 

CC: @stevenyoungs, @Nick-Hall


stevenyoungs:
> This is an alternative implementation to #2062

I'm struggling to spot substantive differences.... The best I can spot so far is staticmethod vs a prepare method for each object type but the actual behaviour looks like it will be the same

dsblank:
Agreed.

PR #2074, opened 2025-05-29
Filter optimizer logical op "or" and "one"

stevenyoungs:
Enable the filter optimizer to optimize filters the use logical operations of "one" or "or"

Note: this branch needs to be rebased once #2052 is merged into maintenance/gramps60 and onward merged into master

### DIFF
--- a/gramps/gen/filters/__init__.py
+++ b/gramps/gen/filters/__init__.py
@@ -41,6 +41,10 @@ def reload_custom_filters():
     CustomFilters = FilterList(CUSTOM_FILTERS)
     CustomFilters.load()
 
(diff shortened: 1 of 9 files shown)


dsblank:
The only way we can optimize "or" would be if all rules (at that level) have selected_handles.

stevenyoungs:
> The only way we can optimize "or" would be if all rules (at that level) have selected_handles.

Yes I reached that conclusion as well and gave up for a bit. It may just be inverted unoptimized rules that are a problem when the logical op is "or". Let's see if my latest attempt passes the tests.....

PR #2076, opened 2025-06-04
Default to SOURCE_DATE_EPOCH for timestamps recording in gzip metadata

Antiz96:
Default to [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) (if set) as timestamps with gzip when compressing man pages to avoid recording non-deterministic information in gzip metadata across builds.

This is required for [reproducible builds](https://reproducible-builds.org/).

### DIFF
--- a/gramps/gen/const.py
+++ b/gramps/gen/const.py
@@ -182,9 +182,7 @@ ROOT_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir))
 
 sys.path.insert(0, ROOT_DIR)
 if DEV_VERSION:
-    git_revision = get_git_revision(ROOT_DIR).replace("\n", "")
-    if sys.platform == "win32" and git_revision == "":
-        git_revision = get_git_revision(os.path.split(ROOT_DIR)[1])
+    git_revision = get_git_revision().replace("\n", "")
     VERSION += git_revision
 # VERSION += "-1"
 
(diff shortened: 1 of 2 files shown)


QuLogic:
I think it might make sense to default this to [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) instead of 0.

Antiz96:
@QuLogic MR updated to use `SOURCE_DATE_EPOCH` (if set), instead of enforcing "0" :)

setup.py
```diff
@@ -164,8 +165,9 @@ def build_man(build_cmd):
 
                 if filename:
                     # Binary io, so open is OK
-                    with open(newfile, "rb") as f_in, gzip.open(
-                        man_file_gz, "wb"
+                    build_date = int(os.environ.get("SOURCE_DATE_EPOCH", time.time()))
```
QuLogic: Since this is in a loop, you can probably pull this out to the top to prevent the fallback `time.time` from changing, and at least keep things internally consistent.

Antiz96: Done: https://github.com/gramps-project/gramps/compare/71fdbbe3c253e4d16958b40681f666d386906841..485a8d19d873bb9ad187545aed7b8acb64b1592f

Antiz96:
Sorry, I unexpectedly closed this merge request when rebasing my branch against master and accidentally discarding my commit in the process :sweat_smile:

I opened a new one at https://github.com/gramps-project/gramps/pull/2078, targeting `scripts/build_hook.py` from a rebased branch.

PR #2078, opened 2025-06-11
Default to SOURCE_DATE_EPOCH for timestamps recording in gzip metadata

Antiz96:
Default to [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) (if set) as timestamps with gzip when compressing man pages to avoid recording non-deterministic information in gzip metadata across builds.

This is required for [reproducible builds](https://reproducible-builds.org/).


Antiz96:
I guess this PR can be closed now (superseded by https://github.com/gramps-project/gramps/pull/2081)

PR #2081, opened 2025-06-19
Normalize gzip mtime for reproducible builds

bmwiedemann:
Normalize gzip mtime for reproducible builds.

This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -164,9 +164,10 @@ def build_man(build_cmd):
 
                 if filename:
                     # Binary io, so open is OK
-                    with open(newfile, "rb") as f_in, gzip.open(
-                        man_file_gz, "wb"
-                    ) as f_out:
+                    with (
+                        open(newfile, "rb") as f_in,
+                        gzip.GzipFile(man_file_gz, "wb", mtime=0) as f_out,
+                    ):
                         f_out.writelines(f_in)
                         _LOG.info("Compiling %s >> %s", filename, man_file_gz)
 


Nick-Hall:
This looks like a duplicate of #2078.

bmwiedemann:
Yes, that serves the same purpose.
However my variant is simpler, because nobody needs that mtime value in the `.gz` header anyway, so we can just zero it, similar to `gzip -n` or `echo foo | gzip`

Antiz96:
Hi,

This is exactly the change I initially proposed in #2076 (later re-opened as #2078). After a short discussion there, we ended up agreeing that making use of `SOURCE_DATE_EPOCH` rather than "enforcing" a null / 0 `mtime` may eventually be a cleaner / "more elegant" solution, but I don't know if the reproducible builds standard actually suggests one over the other to be honest.

No strong feeling on my side.

bmwiedemann:
Both variants would be reproducible. From the experience of my previous 2000 patches for reproducibility I just think the simpler version would be preferable here because it is unlikely that anything uses that timestamp.

In the end, it is the maintainer's call.
I'm fine with either as long as it is merged this year.

Nick-Hall:
I don't mind either solution.  We could even backport this to the _gramps60_ branch.

Nick-Hall:
Backported to the _gramps60_ branch.

PR #2083, opened 2025-06-21
Pass an object rather than a handle to the note editor callback #2

stevenyoungs:
Fix `EditNote` so that it passes the object when calling the callback function, rather than the object handle.
Fixes the error described in 13702 and gives consistency with other Edit classes.

To reproduce:

1. create a new note- type "a"
2. select "a" and click the Link button
3. Change the Link Type to Note and click New
4. type "a" and click ok

Related bugs:
- [#13884](https://gramps-project.org/bugs/view.php?id=13884)
- [#13702](https://gramps-project.org/bugs/view.php?id=13702)
- possibly [#12260](https://gramps-project.org/bugs/view.php?id=12260)

Reviewed and adjusted other places that create an `EditNote` instance with callback
Reviewed and adjusted other places that call `NoteTab.add_callback`

This was originally PR #2070 merged into 6.0.2
Rolled back in 6.0.3 due to [#13884](https://gramps-project.org/bugs/view.php?id=13884)

### DIFF
--- a/gramps/gui/editors/displaytabs/notetab.py
+++ b/gramps/gui/editors/displaytabs/notetab.py
@@ -19,6 +19,13 @@
 
 # -------------------------------------------------------------------------
 #
+# Standard Python modules
+#
+# -------------------------------------------------------------------------
+from __future__ import annotations
+
+# -------------------------------------------------------------------------
+#
 # Python classes
 #
 # -------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


hgohel:
Given the regression in 6.0.3, for safety let's be more explicit about testing. For example the steps from #13884

Select a person. Open it.
Select the "Notes" tab.
Try to pick an existing note to attach to that person
(should not result in an unhandled exception)

Nick-Hall:
Changing the pubic API is also likely to cause third-party addons to crash.

dsblank:
As this changes the public API, targeting Gramps 6.1

dsblank:
@stevenyoungs Do you want to change the target branch to master?

Nick-Hall:
Rebased.

PR #2084, opened 2025-06-28
Internet indexing for the Narrative Web.

SNoiraud:
This was requested from https://gramps.discourse.group/t/can-you-make-the-noindex-html-meta-tags-optional-for-search-engines/5986/21


Nick-Hall:
Rebased.

PR #2085, opened 2025-07-06
Fix regression in relationship path between people filter

DavidMStraub:
See https://github.com/gramps-project/gramps-web-api/issues/664#issuecomment-3041901672


dsblank:
Can you add a test for this?

Could probably work it in here:

gramps/gen/filters/test/filter_optimizer_test.py

or a version of that file.

Nick-Hall:
Rebased.

Nick-Hall:
Fixes bug [#13949](https://gramps-project.org/bugs/view.php?id=13949).

PR #2088, opened 2025-07-09
add sort by date button on individual events and citations

jmichault:
This pull request adds a sort button to the citation tabs and the person's event tab.
image

Clicking the button sorts events or citations in ascending order by date.


Nick-Hall:
Enhancements should be made in the _master_ branch.

jmichault:
> The interface on this seems... not quite right.
> 
> Users can already sort (up or down) by clicking the header of any given column.
> 
> If the button is in the toolbar, then isn't sort criteria ambiguous?
> 
> However, there is a need for a way to lock-in the order after sorting. (I have been using a workaround after sorting. That is selecting any row and using the manual Up button then the Down button. That locks in the order.)
> 
> So, although the function could be useful, this interface seems wrong.

It seems I forgot to mention that the sorting is done in the database, not just in the display, so it is locked.

PR #2090, opened 2025-07-13
Don't allow the optimizer to cross namespace boundaries

tomhughes:
Currently if we have a source filter that matches a source against a repository filter and that repository filter can be optimized the repository handle is propagated back to the source filter and of course we don't find any sources which match a repository handle.

This change stops the optimizer chasing filters into a different namespace as any handles returned from such a filter will not be useful.

I've offered this against `gramps60` rather than `master` just because the code has been changed there.

### DIFF
--- a/gramps/gen/filters/optimizer.py
+++ b/gramps/gen/filters/optimizer.py
@@ -52,7 +52,9 @@ class Optimizer:
         handles_out = None
         for rule in filter.flist:
             if filter.logical_op == "and" or len(filter.flist) == 1:
-                rule_in, rule_out = self.compute_potential_handles_for_rule(rule)
+                rule_in, rule_out = self.compute_potential_handles_for_rule(
+                    filter, rule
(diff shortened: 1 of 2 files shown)


dsblank:
@tomhughes is it possible to add a test to test this change? It could be added to `gramps/gen/filters/rules/test/`   like those in `person_rules_test.py`

tomhughes:
Yes I know there's a real failure but I'm afraid my python isn't up to knowing how to fix it... It missing the type declaration but trying to import the necessary module to get the type creates a cyclic dependency.

dsblank:
I have a fix for this here: https://github.com/gramps-project/gramps/pull/2099

dsblank:
Thanks @tomhughes ! With you suggestion I was able to find a solution fairly quickly.

Nick-Hall:
Closed in favour of PR #2099.

PR #2093, opened 2025-07-20
Added new fieldtype 48 to support Pro-Gen version 3.5

Anthon-S:
The Pro-Gen importer was originally written for Pro-Gen Version 3.0
With the change from Pro-Gen 3.0 to 3.5, the record length of the person table changed from 309 to 332/336 bytes (PG35-1D/PG35-2D) and a new field type 48 was introduced. The latter is crucial, as parsing for the 3.5 record format of the person table (recfmt) fails.
Added the new fieldtype 48 to support Pro-Gen version 3.5. Pro-Gen data definition files PG35-1D/PG35-2D

### DIFF
--- a/gramps/plugins/lib/libprogen.py
+++ b/gramps/plugins/lib/libprogen.py
@@ -386,7 +386,7 @@ class PG30DefTable(object):
                 fmt += "i"
             elif fldtyp == 31:
                 pass
-            elif fldtyp == 32 or fldtyp == 44 or fldtyp == 45:
+            elif fldtyp == 32 or fldtyp == 44 or fldtyp == 45 or fldtyp == 48:
                 fmt += "%ds" % fld.size
             elif fldtyp == 41:
                 fmt += "h"


gramps/plugins/lib/libprogen.py
```diff
@@ -386,7 +386,7 @@ def get_recfmt(self):
                 fmt += "i"
             elif fldtyp == 31:
                 pass
-            elif fldtyp == 32 or fldtyp == 44 or fldtyp == 45:
+            elif fldtyp == 32 or fldtyp == 44 or fldtyp == 45 or fldtyp == 48:
```
dsblank, line 389: Consider checking to see if they are in a list of values:

```suggestion
            elif fldtyp in [32, 44, 45, 48]:
```

Anthon-S: Sure, that makes sense, but then we should do it for the entire for... block:
```
  for fld in flds:
            fldtyp = fld.type_
            if fldtyp == 2 or fldtyp == 3 or fldtyp == 22 or fldtyp == 23:
                fmt += "i"
            elif fldtyp == 31:
                pass
            elif fldtyp == 32 or fldtyp == 44 or fldtyp == 45:
                fmt += "%ds" % fld.size
            elif fldtyp == 41:
                fmt += "h"
            elif fldtyp == 42 or fldtyp == 43 or fldtyp == 46 or fldtyp == 47:
```

Nick-Hall:
Rebased onto the _gramps60_ branch.

PR #2095, opened 2025-07-26
Maintenance/gramps60 MantisBT issue #0013909 failure while searching for filter match

lordemannd:
issue #0013909 can be traced to an error in a format descriptor in:
gramps\gen\filters\rules\event\_matchespersonfilter.py

approximately line 86 has:
person = db.method("get_%_from_handle", classname)(handle)

should be:
person = db.method("get_%s_from_handle", classname)(handle)

That is: there should be an "s" after the "%".


lordemannd:
closing, i made too many mistakes, it seems easier to create a new pull request

PR #2097, opened 2025-07-31
Disambiguate name in glade strings

izikeros:
This PR adds context to the 'Name' strings in .glade files to enable disambiguation and using different translations depending on the context. The problem is described in bug [12527](https://gramps-project.org/bugs/view.php?id=12527). 

Here is problem example.

Currently there is context missing and in .po file there is e.g. for Polish localization:

#: ../gramps/gen/filters/rules/person/_hassoundexname.py:57
#: ../gramps/gui/glade/editfamily.glade:116
#: ../gramps/gui/glade/editfamily.glade:379
#: ../gramps/gui/glade/editplaceformat.glade:201
#: ../gramps/gui/glade/editplacename.glade:184
#: ../gramps/gui/glade/editplaceref.glade:234
#: ../gramps/gui/glade/mergeperson.glade:218
#: ../gramps/gui/glade/mergeperson.glade:233
#: ../gramps/gui/glade/mergeplace.glade:426
#: ../gramps/gui/glade/mergeplace.glade:443
#: ../gramps/plugins/gramplet/soundgen.py:68
msgid "Name:"
msgstr "Imię i nazwisko:"

The msgstr is correct in `person` context but is totally wrong in `place` context and they can't be translated separately dur to context missing.

This PR enables this distinction.

### DIFF
--- a/gramps/gui/glade/editfamily.glade
+++ b/gramps/gui/glade/editfamily.glade
@@ -113,7 +113,7 @@
                                 <property name="can-focus">False</property>
                                 <property name="halign">start</property>
                                 <property name="margin-start">6</property>
-                                <property name="label" translatable="yes">Name:</property>
+                                <property name="label" translatable="yes" context="family">Name:</property>
                               </object>
                               <packing>
                                 <property name="left-attach">0</property>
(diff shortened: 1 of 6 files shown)


Nick-Hall:
This PR appears to contain more commits than intended.

PR #2098, opened 2025-08-02
SQLite Optimizations

dsblank:
This PR was entirely written by Cursor.ai. This is largely provided to give ideas and suggestions for improving SQLite performance. It also includes ideas for bulk operations and JSON query functions at the DBI level.

The prompts were:

> Can you look at how SQLite is used and suggest performance improvements. Here is an example SQLite database: /home/dsblank/.gramps/grampsdb/67ca17c8/sqlite.db If you want to change the database, please use a copy.

and:

> Nice! Can you change the code base to use the new DBI methods? 

The summary results are in:

1. DATABASE_OPTIMIZATION_USAGE.md
2. SQLITE_PERFORMANCE_GUIDE.md

### DIFF
--- a/DATABASE_OPTIMIZATION_USAGE.md
+++ b/DATABASE_OPTIMIZATION_USAGE.md
@@ -1,0 +1,249 @@
+# Database Optimization Usage Guide
+
+This guide explains how to use the new database optimization features in Gramps.
+
+## Overview
+
+The database optimization features provide significant performance improvements for Gramps SQLite databases, especially for larger datasets. These optimizations are automatically applied when connecting to databases and can be further enhanced through manual optimization tools.
+
+## Automatic Optimizations
+
+### SQLite Configuration
+The following optimizations are automatically applied when connecting to a database:
+
+- **WAL Mode**: Write-Ahead Logging for better concurrency
+- **NORMAL Synchronous**: Faster than FULL mode, still safe
+- **64MB Cache**: 32x larger than default
+- **Memory Mapping**: 256MB for faster data access
+- **Incremental Vacuum**: Automatic space reclamation
+
+### Performance Indexes
+Additional indexes are automatically created for common query patterns:
+
+- Composite indexes for person name searches
+- Family relationship optimizations
+- Citation and source optimizations
+- Event and place optimizations
+- Partial indexes for filtered queries
+
+## New Database Methods
+
+### Optimized Query Methods
+
+```python
+# Get persons by name
+handles = db.get_person_by_name("Smith", "John")
+
+# Get families by parent
+handles = db.get_families_by_parent(parent_handle)
+
+# Get events by type
+handles = db.get_events_by_type("Birth")
+
+# Search people with full-text search
+handles = db.search_people_full_text("John Smith")
+
+# Get places by location
+handles = db.get_places_by_location("New York")
+
+# Get sources by author
+handles = db.get_sources_by_author("Smith")
+
+# Get citations by source
+handles = db.get_citations_by_source(source_handle)
+
+# Get media by type
+handles = db.get_media_by_type("image/jpeg")
+
+# Search notes by content
+handles = db.get_notes_by_content("birth certificate")
+
+# Get repositories by name
+handles = db.get_repositories_by_name("Library")
+```
+
+### Bulk Operations
+
+```python
+# Bulk retrieve persons
+persons = db.bulk_get_persons(person_handles)
+
+# Bulk retrieve families
+families = db.bulk_get_families(family_handles)
+
+# Bulk insert data
+data_list = [(handle1, json_data1), (handle2, json_data2)]
+db.bulk_insert("person", data_list)
+
+# Bulk update data
+update_data = [(handle1, new_json1), (handle2, new_json2)]
+db.bulk_update("person", update_data, "handle")
+```
+
+### JSON Query Optimizations
+
+```python
+# Direct JSON path queries
+results = db.json_query("person", "$.name.first", "John")
+
+# Full-text search in JSON
+results = db.json_search("person", "search_term", [
+    "$.name.first", "$.name.surname", "$.name.suffix"
+])
+
+# Complex JSON queries
+results = db.json_query("person", "$.birth.date", "1900", ">=")
+```
+
+### Database Management
+
+```python
+# Optimize database
+db.optimize_database()
+
+# Get database statistics
+stats = db.get_database_stats()
+print(f"Database size: {stats['database_size_bytes']} bytes")
+print(f"Person count: {stats['person_count']}")
+
+# Monitor query performance
+result, execution_time = db.monitor_query_performance(
+    "SELECT * FROM person WHERE surname = ?", ["Smith"]
+)
+
+# Get optimization suggestions
+suggestions = db.suggest_indexes()
+```
+
+## Performance Monitoring
+
+### Using the Performance Monitor
+
+```python
+from gramps.gen.db.performance import DatabasePerformanceMonitor
+
+# Create monitor
+monitor = DatabasePerformanceMonitor(db)
+
+# Monitor specific queries
+@monitor.monitor_query
+def get_people_by_surname(surname):
+    return db.get_person_by_name(surname)
+
+# Run monitored query
+results = get_people_by_surname("Smith")
+
+# Get performance statistics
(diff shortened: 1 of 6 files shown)


DATABASE_OPTIMIZATION_USAGE.md
```diff
@@ -0,0 +1,249 @@
...
+handles = db.get_places_by_location("New York")
+
+# Get sources by author
+handles = db.get_sources_by_author("Smith")
+
+# Get citations by source
+handles = db.get_citations_by_source(source_handle)
+
+# Get media by type
+handles = db.get_media_by_type("image/jpeg")
```
kulath, line 56: Plurals. Should be get_people_by_name (you already have analogous search_people_full_text) and get_medias_by_name (I know this is terribly ugly and media is already a plural, but it gets more than one media handle.

glamberson:
Hi Doug,

Sorry I didn't look at this earlier - I've been focused on getting the PostgreSQL addon stable and didn't want to context-switch. Now that I've had a chance to review your SQLite optimizations, they look like solid improvements. The performance gains from WAL mode and the cache settings should help a lot of users.

Since you asked about DBAPI recommendations, I've been thinking about the architecture quite a bit while working on the PostgreSQL implementation. The SQLite-specific optimizations you've implemented make sense where you've put them - in the SQLite backend. The PRAGMA settings and tuning belong there.

For the DBAPI base class, I noticed we're both addressing some of the same performance bottlenecks (prepared statements, batch operations) from different angles. Perhaps we could consider putting backend-agnostic versions in the base DBAPI class so all backends benefit. Here's what I'm thinking:

### 1. Prepared Statements (Move to Base DBAPI)
Your implementation is good, but we could make it backend-agnostic:

```python
# In base DBAPI class (works for all backends):
def prepare(self, name, query):
    """Cache prepared statements for any backend that supports them."""
    if name not in self._prepared_statements:
        if hasattr(self.dbapi, 'prepare'):
            # PostgreSQL, MySQL, etc.
            self._prepared_statements[name] = self.dbapi.prepare(query)
        else:
            # SQLite and others - just cache the query
            self._prepared_statements[name] = query
    return self._prepared_statements[name]
```

This way SQLite caches query strings, PostgreSQL uses real prepared statements, and future backends get the best their driver offers.

### 2. Batch Operations (Move to Base DBAPI)
Similar idea - your `executemany()` usage could benefit everyone:

```python
# In base DBAPI class:
def batch_commit_persons(self, persons, trans):
    """Commit multiple persons efficiently."""
    if hasattr(self.dbapi, 'executemany'):
        data = [(p.handle, p.serialize()) for p in persons]
        self.dbapi.executemany(
            "INSERT OR REPLACE INTO person VALUES (?, ?)", data)
        # Still fire signals for GUI
        for person in persons:
            self.emit('person-add', ([person.handle],))
    else:
        # Fallback for backends without executemany
        for person in persons:
            self.commit_person(person, trans)
```

### 3. Additional Quick Wins (New)
While we're improving DBAPI, here are two more easy wins:

**Real Cursors** (5 lines, huge memory savings):
```python
def get_person_cursor(self, limit=None):
    """Return iterator instead of list - saves memory."""
    if self.dbapi and hasattr(self.dbapi, 'cursor'):
        cursor = self.dbapi.cursor()
        cursor.execute("SELECT handle FROM person")
        return cursor  # Real cursor for PostgreSQL/MySQL
    else:
        handles = self.get_person_handles()
        return iter(handles)  # Iterator for SQLite/BSDDB
```

**Lazy Loading** (10 lines, optional):
```python
def get_person_from_handle(self, handle, lazy=False):
    """Optional lazy loading to reduce memory usage."""
    if lazy:
        return LazyPerson(handle, self)  # Load on access
    return super().get_person_from_handle(handle)
```

### How This Would Work

1. Your PR: Keep all SQLite-specific optimizations in `sqlite.py`
2. Base DBAPI: Move backend-agnostic improvements there
3. Result: SQLite gets your optimizations + base improvements, PostgreSQL gets base improvements + can add its own

### Collaboration Proposal

Would you like to:
A) Keep this PR as-is and I'll submit a follow-up PR with base DBAPI improvements
B) Split this into two PRs - SQLite-specific and base DBAPI improvements  
C) Expand this PR to include base DBAPI improvements

Whatever approach you prefer works for me. Your SQLite optimizations are valuable either way.

The immediate changes would be about 50 lines in base DBAPI:
- Real cursors (5 lines)
- Prepared statements (20 lines)  
- Lazy loading (10 lines)
- Batch operations (15 lines)

These are all non-breaking, backward-compatible changes with graceful fallbacks.

### Longer-term Vision

Looking ahead, I've been thinking about how we might evolve DBAPI to better support different backends while maintaining backward compatibility. The idea would be to gradually build a more modern architecture underneath while keeping the existing interface stable - essentially a compatibility layer that lets us improve the internals without breaking existing code.

This could let us progressively adopt features like proper transaction isolation, connection pooling, and even multi-user support for backends that can handle it, while SQLite continues to work exactly as it does today. But that's a longer conversation and certainly not something that needs to block your improvements here.

What are your thoughts on the immediate changes, and does this general direction make sense to you?

DavidMStraub:
Thanks for this interesting experiment. I am using AI gents at work extensively myself, and recently have noticed a significant increase in code contributions using them. In general, I feel the more promising apporach is to identify a problem in the code base and giving as detailed a prompt as possible, rather than letting the AI come up with suggestions based on a very generic prompt. I am not saying what you/they came up with is not potentially valuable (I haven't reviewed the code), but in general I made bad experiences with "let the AI come up with something and then let's try to understand it".

This is particulary true given that the Gramps code base contains lots of things that are nowadays quite non standard, if you know what I mean...

My general suggestion would be

- let's make sure the entire code base is cleaned up and type hinted, so it's more understandable to both humans and AI, and static analysis can catch more refactoring bugs humans or AI
- let's identify the actual features or optimizations we need based on user reports and performance analysis

Again, these are just general remarks, not specifically about the code in this PR.

glamberson:
> > This way your PR stays focused on SQLite performance improvements, and I'll handle the base DBAPI enhancements
> > separately. Should have the PR to your branch ready shortly.
> 
> Yes. Please keep each PR small and focused.

I preferred to separate them, but I dind't realize Doug had already compined the sqlite and DBAPI changes here. Thus I deferred to his chosen way. I'm all for splitting them.

dsblank:
> I preferred to separate them, but I dind't realize Doug had already compined the sqlite and DBAPI changes here. Thus I deferred to his chosen way. I'm all for splitting them.

Either way is fine with me. I find that writing both together is useful, not only because one depends on the other, but doing something in the low-level often leads to ideas in the high-level.

However the team feels is best is fine with  me.

And I should add that I am so happy that we have generated such collaboration and ideas already!

Nick-Hall:
> I am hoping that just because this PR was entirely written by AI, that doesn't mean we can't refine it and make it so that it could be include in the core Gramps code, right?

The consensus following our previous discussions seemed to be that we should allow AI generated code in contributions.

If we allow some code in a contribution to be AI generated, then I don't see why we shouldn't allow all of it to be AI generated.  I also don't see any point in trying to define how a development tool should be used.

It is ultimately the developer's responsibility to submit well written code that conforms to our guidelines.

dsblank:
Here is a proposal:

1. Create a Gramps Enhancement Proposal (GEP) for adding methods to DPI-DB, and implementations for PostgreSQL and SQLite. [I can do this]
2. Create a PR for DB-API changes off of master (planning for Gramps 6.1)
3. Create a PR for PostegreSQL off of (3)
4. Create a PR for SQLite off of (3)

We'll have to figure out a way to measure the effectiveness of the implementations to compare with the baseline (current code base).

@glamberson if you would like to create 2, 3, 4, that is fine with me. We can close this PR, and move the relevant parts to 2 and 4.

Does this sound viable?

We can also create a GEP for adding type hints to the code base. I assume that the new code above will have type hints. @DavidMStraub do you want to create the GEP?

Nick-Hall:
> We can also create a GEP for adding type hints to the code base.

You don't need to GEP to add type hints.  However, you should create a GEP if you intend to extend the public database API.

dsblank:
> So don't be too surprised when I start suggesting some far more drastic architectural ideas in the near future.

I can't wait to hear! But we have a deep history of code here. It will be interesting to see if we can span that gap.

PR #2099, opened 2025-08-04
Fixed mixed filter types

dsblank:
This PR:

1. Fixes the problem of mixed filter types
2. Had to add the fix from #2096 to test
3. Clean up unused code in the testing system

### DIFF
--- a/gramps/gen/filters/optimizer.py
+++ b/gramps/gen/filters/optimizer.py
@@ -42,6 +42,9 @@ LOG = logging.getLogger(".filter.optimizer")
 
 
 class Optimizer:
+    def __init__(self, top_level_filter):
+        self.top_level_filter = top_level_filter
+
     def compute_potential_handles_for_filter(
         self, filter: GenericFilter
     ) -> Tuple[Set[PrimaryObjectHandle] | None, Set[PrimaryObjectHandle] | None]:
@@ -73,11 +76,29 @@ class Optimizer:
         self,
         rule: Rule,
     ) -> Tuple[Set[PrimaryObjectHandle] | None, Set[PrimaryObjectHandle] | None]:
-        """ """
+        """
+        Find the the handles for a particular rule.
+        """
+        # Has to be of the appropriate type
         if hasattr(rule, "selected_handles"):
             return (rule.selected_handles, None)
         if hasattr(rule, "find_filter"):
             filter = rule.find_filter()
-            if filter:
+            if filter and self.is_same_namespace(filter):
                 return self.compute_potential_handles_for_filter(filter)
         return (None, None)
+
+    def is_same_namespace(self, filter):
+        """
+        Determine if the given filter is in the 'same namespace' as the top-level filter.
+
+        In this context, 'same namespace' means that both filters are instances of the same class,
+        as determined by comparing their class names.
+
+        Parameters:
+            filter: The filter object to compare against the top-level filter.
+
+        Returns:
+            True if both filters have the object type, False otherwise.
+        """
+        return type(self.top_level_filter.make_obj()) == type(filter.make_obj())
--- a/gramps/gen/filters/test/filter_optimizer_test.py
+++ b/gramps/gen/filters/test/filter_optimizer_test.py
@@ -207,6 +207,23 @@ custom_filters_xml = """<?xml version="1.0" encoding="utf-8"?>
         <arg value="Person I1041"/>
       </rule>
     </filter>
+    <filter name="Home Person" function="and">
+      <rule class="IsDefaultPerson" use_regex="False" use_case="False">
+      </rule>
+    </filter>
+    <filter name="Person of Events of Home Person" function="and">
+      <rule class="MatchesEventFilter" use_regex="False" use_case="False">
+        <arg value="Events of Home Person"/>
+      </rule>
+    </filter>    
+  </object>
+  <object type="Event">
+    <filter name="Events of Home Person" function="and">
+      <rule class="MatchesPersonFilter" use_regex="False" use_case="False">
+        <arg value="Home Person"/>
+        <arg value="0"/>
+      </rule>
+    </filter>
   </object>
 </filters>
 """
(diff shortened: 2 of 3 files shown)


gramps/gen/filters/optimizer.py
```diff
@@ -73,11 +76,17 @@ def compute_potential_handles_for_rule(
...
             return (rule.selected_handles, None)
         if hasattr(rule, "find_filter"):
             filter = rule.find_filter()
-            if filter:
+            if filter and self.is_same_namespace(filter):
                 return self.compute_potential_handles_for_filter(filter)
         return (None, None)
+
+    def is_same_namespace(self, filter):
+        return self.top_level_filter.__class__.__name__ == filter.__class__.__name__
```
dsblank: I'm not 100% sure this is the right test, but the tests pass. Could probably use some more mixed-namespace filter tests.

dsblank:
> Does this replace https://github.com/gramps-project/gramps/pull/2090? If so, I think that we are ready for another maintenance release.

Yes, and yes! I've reviewed (and marked bug) on some other fixes.

gramps/gen/filters/optimizer.py
dsblank: @stevenyoungs if you are around, do you have an opinion on how to make sure filters are in the right "namespace"?

stevenyoungs: @dsblank I saw you comment a few days ago and will take a look...

gramps/gen/filters/optimizer.py
```diff
@@ -73,11 +76,29 @@ def compute_potential_handles_for_rule(
...
+        In this context, 'same namespace' means that both filters are instances of the same class,
+        as determined by comparing their class names.
+
+        Parameters:
+            filter: The filter object to compare against the top-level filter.
+
+        Returns:
+            True if both filters have the same class name (i.e., are of the same type), False otherwise.
+        """
+        return type(self.top_level_filter) == type(filter)
```
stevenyoungs: I think this is safe, but could be too cautious. As an example, `DeferredFamilyFilter` subclasses `GenericFamilyFilter`. Therefore their types are different and `is_same_namespace` will return false. [this is from reading the code, I've not constructed a test case to see if it is actually possible to contruct a rule using a filter of each of these types]

We really want to check the type of the objects returned by the filter, perhaps a property can be added to `GenericFilter`, initialised by the the `Generic*Filter` classes. This property can then be used in `is_same_namespace`

stevenyoungs:
I also tested a slightly more complex scenario layering another filter on top that returns people. This also worked with the revised code

|Number|Filter|Type|
|--|--|--|
|3|Persons with events matching the <event filter>|GenericPersonFilter|
|2|Events of persons matching the <person filter>|GenericEventFilter|
|1|Home Person|GenericPersonFilter|

Because #1 and #3 have the same type, `is_same_namespace` returns true. I don't think that this is problematic, but I need to do some more investigation to be sure.

stevenyoungs:
Whilst this approach seems pragmatic for gramps 6.0, for gramps 6.1,  I wonder if we could modify the code such that optimization is still possible in mixed type scenarios like this? Perhaps construct a new `Optimizer` instance each time the type changes?

Nick-Hall:
We should keep the fix simple in the maintenance branch.

gramps/gen/filters/optimizer.py
dsblank: You are right @stevenyoungs . I changed the code a bit to use:

```python
    return type(self.top_level_filter.make_obj()) == type(filter.make_obj())
```

gramps/gen/filters/optimizer.py
```diff
@@ -73,11 +76,29 @@ def compute_potential_handles_for_rule(
...
+        Determine if the given filter is in the 'same namespace' as the top-level filter.
+
+        In this context, 'same namespace' means that both filters are instances of the same class,
+        as determined by comparing their class names.
+
+        Parameters:
+            filter: The filter object to compare against the top-level filter.
+
+        Returns:
+            True if both filters have the same class name (i.e., are of the same type), False otherwise.
```
dsblank:
```suggestion
            True if both filters have the object type, False otherwise.
```

gramps/gen/filters/optimizer.py
stevenyoungs: Nice idea. Much simpler than adding a new method!

gramps/gen/filters/test/filter_optimizer_test.py
```diff
@@ -207,6 +207,18 @@
         <arg value="Person I1041"/>
       </rule>
     </filter>
+    <filter name="Home Person" function="and">
+      <rule class="IsDefaultPerson" use_regex="False" use_case="False">
+      </rule>
+    </filter>
```
stevenyoungs, line 213:
```suggestion
    </filter>
    <filter name="Person of Events of Home Person" function="and">
      <rule class="MatchesEventFilter" use_regex="False" use_case="False">
        <arg value="Events of Home Person"/>
      </rule>
    </filter>    
```

Add an additional filter where we go from Person -> Event -> Person

gramps/gen/filters/test/filter_optimizer_test.py
```diff
@@ -425,3 +437,14 @@ def test_not_parents_of_person_I1041(self):
...
+    def test_home_person(self):
+        filter = self.filters["Home Person"]
+        results = filter.apply(self.db)
+        self.assertEqual(len(results), 1)
+        self.assertEqual(results[0], "GNUJQCL9MD64AM56OH")
+
+    def test_events_of_home_person(self):
+        filter = self.event_filters["Events of Home Person"]
+        results = filter.apply(self.db)
+        self.assertEqual(len(results), 3)
```
stevenyoungs, line 455:
```suggestion
        self.assertEqual(len(results), 3)

    def test_person_of_events_of_home_person(self):
        filter = self.event_filters["Person of Events of Home Person"]
        results = filter.apply(self.db)
        self.assertEqual(results[0], "GNUJQCL9MD64AM56OH")
```
Add a test for the "Person of Events of Home Person" filter

stevenyoungs:
> I also tested a slightly more complex scenario layering another filter on top that returns people. This also worked with the revised code
I've suggested adding this as an additional unit test case. Feel free to discard if you think this is overkill.

gramps/gen/filters/test/filter_optimizer_test.py
```diff
@@ -425,3 +442,19 @@ def test_not_parents_of_person_I1041(self):
...
+
+    def test_person_of_events_of_home_person(self):
+        filter = self.event_filters["Person of Events of Home Person"]
```
stevenyoungs:
```suggestion
        filter = self.filters["Person of Events of Home Person"]
```
Fix my copy-and-paste error

dsblank: Yep, just fixed that :) Thanks!

Nick-Hall:
Rebased and squashed commits.

PR #2100, opened 2025-08-07
NarrativeWeb: Add non-sharable notes to media

SNoiraud:
Fixes [#13922](https://gramps-project.org/bugs/view.php?id=13922).

### DIFF
--- a/gramps/plugins/webreport/media.py
+++ b/gramps/plugins/webreport/media.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-#!/usr/bin/env python
 #
 # Gramps - a GTK+/GNOME based genealogy program
 #
@@ -53,8 +52,11 @@ import logging
 # Gramps module
 # ------------------------------------------------
 from gramps.gen.const import GRAMPS_LOCALE as glocale
-from gramps.gen.lib import Date, Media
+from gramps.gen.lib import Date, Media, Note, NoteType, Name
+from gramps.gen.display.name import displayer as _nd
+from gramps.gen.display.place import displayer as _pd
 from gramps.gen.plug.report import Bibliography
+from gramps.gen.utils.db import get_referents
 from gramps.gen.utils.file import media_path_full
 from gramps.gen.utils.thumbnails import run_thumbnailer
 from gramps.gen.utils.image import image_size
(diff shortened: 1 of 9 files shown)


gramps/plugins/webreport/media.py
```diff
@@ -677,6 +679,147 @@ def mediapage(self, report, the_lang, the_title, media_handle, info):
...
+                    with Html("div", class_="subsection narrative") as section:
+                        section += Html("h5", header, inline=True)
+                        with Html("ol") as listn:
+                            section += listn
+                            for notehandle in note_list:
+                                note = self.r_db.get_note_from_handle(notehandle)
+                                note_text = self.get_note_format(note, True)
+                                note_text = note.text
+                                listn += Html("li", note_text)
+                return section
```
dsblank, line 696: **Bug: `NameError` if object has no media references**

If `mref.get_media_list()` returns an empty list, the `for` loop body never executes, so `section` is never defined. `return section` on this line then raises `NameError` and crashes the report.

Fix: initialize `section = None` before the loop and guard the return.

gramps/plugins/webreport/media.py
```diff
@@ -677,6 +679,147 @@ def mediapage(self, report, the_lang, the_title, media_handle, info):
                         )
                         table += trow
 
+            def show_referent(mref, header):
+                for mlist in mref.get_media_list():
```
dsblank, line 684: **Bug: iterates all media refs, not just the one for the current media**

`mref.get_media_list()` returns every media reference attached to the object (Person, Family, etc.), not just the reference that points to the media page being rendered. Notes from *other* media references on the same object would incorrectly appear here.

The loop should filter to only the reference pointing at `media.handle`:
```python
for mlist in mref.get_media_list():
    if mlist.get_reference_handle() != media.handle:
        continue
    ...
```

gramps/plugins/webreport/media.py
```diff
@@ -677,6 +679,147 @@ def mediapage(self, report, the_lang, the_title, media_handle, info):
...
+                    note_list = mlist.get_note_list()
+                    with Html("div", class_="subsection narrative") as section:
+                        section += Html("h5", header, inline=True)
+                        with Html("ol") as listn:
+                            section += listn
+                            for notehandle in note_list:
+                                note = self.r_db.get_note_from_handle(notehandle)
+                                note_text = self.get_note_format(note, True)
+                                note_text = note.text
```
dsblank: **Bug: `get_note_format` result is silently discarded**

Line 691 calls `self.get_note_format(note, True)` and assigns the result to `note_text`, but line 692 immediately overwrites it with `note.text`. The formatted HTML is thrown away.

`note.text` is a `StyledText` object, not a plain string. Passing it directly to `Html("li", ...)` will likely not render the note content correctly.

The fix is to use the formatted result:
```python
note_text = self.get_note_format(note, True)
listn += Html("li", note_text)
```
(remove line 692)

gramps/plugins/webreport/media.py
```diff
@@ -677,6 +679,147 @@ def mediapage(self, report, the_lang, the_title, media_handle, info):
...
+                                listn += Html("li", note_text)
+                return section
+
+            # get media reference notes
+            note = Note()
+            note.set_type(NoteType.MEDIAREF)
+            mediaref = self._(note.get_type().xml_str())
+
+            # add section title
+            mediadetail += Html("h4", mediaref, inline=True)
```
dsblank: **Visual bug: section heading always rendered even when there are no notes**

The `<h4>` "Media Reference" heading is unconditionally added to `mediadetail` here, before any referents are checked. If no object referencing this media has any notes on its media reference, the page will show an orphaned section heading with nothing under it.

The heading should only be added if at least one referent note is found.

gramps/plugins/webreport/media.py
```diff
@@ -677,6 +679,147 @@ def mediapage(self, report, the_lang, the_title, media_handle, info):
...
+
+            idx = 0
+            for referents in "Person", "Family", "Event", "Place", "Source", "Citation":
```
dsblank, line 707: Minor: `get_referents` is called with a one-element tuple `(referents,)`, so `refs` is always a one-element tuple. The double-loop `for refl in refs: for ref in refl:` works but is confusing. A cleaner equivalent: `for ref in refs[0]:` (or just unpack: `(ref_list,) = refs`).

gramps/plugins/webreport/media.py
```diff
@@ -53,8 +52,11 @@
 # Gramps module
 # ------------------------------------------------
```
dsblank, line 53: `Name` is imported but only used to copy a `primary_name` for display formatting. This could use the name displayer directly on the person object instead:
```python
surname = _nd.display(mref)
```
That is equivalent and does not require importing `Name` or constructing a copy.

gramps/plugins/webreport/media.py
```diff
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
```
dsblank, line 1: `black` reports this file would be reformatted. Please run:
```bash
git diff --name-only --diff-filter=ACMR origin/master...HEAD | grep '\.py$' | xargs --no-run-if-empty black
```
The CI black check will fail otherwise.

gramps/plugins/webreport/media.py
SNoiraud: good catch

gramps/plugins/webreport/media.py
SNoiraud: We can't use None. We need to use section = Html()

gramps/plugins/webreport/media.py
SNoiraud: This is because we need to set a different name format than the defaut format for gramps.

PR #2102, opened 2025-08-09
Fix regex matches in filter match method

dsblank:
This PR fixes https://gramps-project.org/bugs/view.php?id=13926

The issue was that regex matches require `rule.requestprepare()` to be called, and apparently this was not properly called when `filter.match()` was called. 

This was not an issue of the filter optimizer, but appears to be a side effect of the refactoring. It only appeared in those places where `filter.match()` was used. I believe that the fanchart widget was the only place in core gramps.


Nick-Hall:
Rebased and squashed commits.

PR #2116, opened 2025-08-31
Localized plugin for name guessing

ChristinaRau:
Hello, 

the context to this pull request is this discussion: https://gramps.discourse.group/t/new-localized-options-for-surname-guessing/8204

I added a new NameGuesser plugin with localizations (for DE, ES, PT, IS).
The plugin is used when adding a new person to a family (in the classes EditFamily and RelationshipView).

Default locale (EN):

- Surname guessing is like the default before: When adding a child, it gets the father's surname; when adding a father, he gets the child's surname.

Locale DE:

- When adding a child, it gets the father's surname, unless the family is unmarried, then it gets the mother's surname.
- If unmarried, the mother gets the child's surname when adding the mother.
- If not unmarried, the father gets the child's surname when adding the father.

Locale ES:

- Surname guessing is like the Latin American surname guessing option that has existed before. The child's first surname is the father's first surname, the child's second surname is the mother's first surname.

Locale PT:
- Like ES, but reversed: The child's first surname is the mother's first surname, the child's second surname is the father's first surname.

Locale IS:
- When a child is added and a father exists in the family, the child's surname will be the father's given name + "sson" or "sdóttir"
- If a child has a matronymic surname and a mother is added to the family, the mother's given name will be the child's surname minus "sson" or "sdóttir"
- If a child has a non-matronymic surname and a father is added, the father's given name will be the child's surname minus "sson" or "sdóttir"

### DIFF
--- a/gramps/gen/nameguesser.py
+++ b/gramps/gen/nameguesser.py
@@ -51,7 +51,7 @@ class NameGuesser:
     The name guesser guesses the names of a person based on their family members' names.
     """
 
-    def fathers_surname_from_child(self, db, family):
+    def fathers_name_from_child(self, db, family):
         """
         Find a child and return their name for the father.
         """
@@ -68,7 +68,7 @@ class NameGuesser:
 
         return name
 
-    def mothers_surname_from_child(self, db, family):
+    def mothers_name_from_child(self, db, family):
         """
         Mother's surname cannot be guessed in the default locale, return empty name.
         """
@@ -79,7 +79,7 @@ class NameGuesser:
 
         return name
 
-    def childs_surname(self, db, family):
+    def childs_name(self, db, family):
         """
         Child inherits name from father.
         """
--- a/gramps/gen/nameguesser.py
+++ b/gramps/gen/nameguesser.py
@@ -1,0 +1,147 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2025 Christina Rauscher
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+# -------------------------------------------------------------------------
+#
+# Python modules
+#
+# -------------------------------------------------------------------------
+import logging
+
+# -------------------------------------------------------------------------
+#
+# Gramps modules
+#
+# -------------------------------------------------------------------------
+from .lib import Name, Surname
+from .utils.db import preset_name
+from .plug import PluginRegister, BasePluginManager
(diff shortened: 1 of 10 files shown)


dsblank (changes requested):
## Review: Localized plugin for name guessing

The architectural idea is solid — moving locale-specific name guessing into plugins is a clean extension point. The code is well-structured, well-tested, and clearly motivated by the linked Discourse discussion. There are a few issues that should be addressed before merging.

---

### 🔴 Critical Bug: `_pluginreg.py` attribute name mismatch

In `PluginData.__init__()`:
```python
# NAMEGUESSER attr
self._nameguesser = None   # ← initialized as _nameguesser
```

But the property getter/setter reference `self._nameguesserclass`:
```python
@property
def nameguesserclass(self):
    return self._nameguesserclass   # ← AttributeError! Never initialized
```

The setter initializes `self._nameguesserclass` on first use, so it works when the setter is called (as in `nameplugins.gpr.py`). But accessing the getter on a plugin that never called the setter will raise `AttributeError`. The `__init__` should initialize `self._nameguesserclass = None` instead of `self._nameguesser = None`.

---

### 🔴 Regression: `behavior.surname-guessing` config is orphaned

The PR removes the code that reads `config.get("behavior.surname-guessing")` in `editfamily.py` and `relview.py`, but:
- `config.py:172` still registers `behavior.surname-guessing`
- `configure.py:1678–1681` still shows the UI for it (4 options: Father's surname, None, Latin American, Icelandic)

Users who had set option 1 ("None") or option 3 ("Icelandic") now have their preference silently ignored. For EN locale users who chose Latin American style, the new EN `NameGuesser` always defaults to father's surname. The config UI now does nothing. This should either be removed from the Preferences UI or the locale-based guesser should respect the user's explicit override.

---

### 🟡 Global state / caching issue in `get_nameguesser`

After the first call, subsequent calls with different `clocale` values are silently ignored — the cached class is returned. The `clocale` parameter is documented as allowing "selection of the language" but effectively only works on the very first call (or when `reinit=True`). There is also no mechanism for a runtime locale change to trigger a reinit. In practice this may be fine since `clocale=glocale` is always passed, but the API is misleading.

---

### 🟡 Unconditional `StreamHandler` added to logger

```python
LOG = logging.getLogger("gen.relationship")
LOG.addHandler(logging.StreamHandler())   # ← adds a handler on every import
```

This unconditionally attaches a `StreamHandler` on import, which can cause duplicate log output. Other Gramps modules don't do this — the line should be removed.

---

### 🟡 Unused import: `FamilyRelType` in `nameguesser_es.py` and `nameguesser_pt.py`

Both files import `FamilyRelType` but never reference it. These should be removed.

---

### 🟡 Missing `__init__.py` in test directory

`gramps/plugins/nameguesser/test/` has no `__init__.py`, which may prevent test discovery depending on the test runner configuration.

---

### 🟢 Minor: confusing comments in ES/PT name guessers

In `nameguesser_es.py`'s `mothers_name_from_child`:
```python
# set surname to patrilineal because she probably gave her patrilineal surname to her child
surname.origintype = NameOriginType(NameOriginType.PATRILINEAL)
```
Applying `PATRILINEAL` to the *mother's* returned name is semantically confusing (even if technically the surname she passes down was patrilineally received). Consider using `MATRILINEAL` since it represents the mother's line from the child's perspective, or at least improve the comment.

Similarly, the PT `fathers_name_from_child` comment says "he probably gave his **matrilineal** surname to his child" — a father giving a "matrilineal" surname is contradictory wording.

---

### 🟢 Minor: use `removesuffix()` in `nameguesser_is.py`

```python
father_given_name = surname.replace("sson", "")
```

Using `str.replace()` could mishandle edge cases (e.g. a name that contains "sson" not at the end). Python 3.9+'s `str.removesuffix("sson")` is safer and more explicit. Same applies to the other suffix stripping in that file.

PR #2118, opened 2025-09-02
Fixes MatchesPersonFilter

dsblank:
This PR fixes: https://gramps-project.org/bugs/view.php?id=13947

For some reason, the logic was corrupted in Gramps 6.0.


Nick-Hall:
Thanks.  I could release 6.0.5 over the weekend.

PR #2127, opened 2025-09-27
Preserve order in find_children

DavidMStraub:
A user in the German Gramps forum [pointed out](https://discourse.genealogy.net/t/kinder-in-falscher-reihenfolge-in-ahnentafel-unter-diagramme/836721/7) that the order of children in the pedigree view (when clicking the left arrow) is seemingly random. I traced that back to the function `find_children`, which iterates over families and children and adds the handles to a set, eventually returning a list.

However, since sets are not ordered, the resulting order is random.

I fixed that by simply using a list instread of the set and removing duplicates in the end (using the [famous Stackoverflow trick](https://stackoverflow.com/questions/7961363/removing-duplicates-in-lists)).

### DIFF
--- a/gramps/gen/utils/db.py
+++ b/gramps/gen/utils/db.py
@@ -374,13 +374,14 @@ def find_children(db, person):
     """
     Return the list of all children's IDs for a person.
     """
-    children = set()
+    children = []
     for family_handle in person.get_family_handle_list():
         family = db.get_family_from_handle(family_handle)
         if family:
             for child_ref in family.get_child_ref_list():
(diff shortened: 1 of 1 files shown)


DavidMStraub:
I think the CI failure is random, can we just rerun it please?

DavidMStraub:
Sorry, but this is out of scope for this function. It is returning the children in the order they are in the database. It has never resepected any user-chosen sort order. Implementing that would be a **new feature**. Here, I'm fixing a **bug** that led to a random order.

PR #2128, opened 2025-09-27
Protocols for typing of Gramps objects

DavidMStraub:
Typing of functions/methods involving Gramps objects has become significantly more complicated in Gramps 6.0 due to the introduction of `DataDict`.

As was discussed in various places, e.g. [here](https://github.com/gramps-project/gramps/pull/2010#issuecomment-2719256251), protocols seem to be the right way to address this.

This PR contains a proof of concept, adding protocols for ~~all primary~~ the `Person` primary type and some secondary Gramps types. The code at the bottom checks that the instances are compatible with the protocols (at list w.r.t. existence of methods).

Related: #2018, #1934, #1919, #2010.

CC @dsblank, @stevenyoungs, @bartfeenstra

### DIFF
--- a/gramps/gen/types.py
+++ b/gramps/gen/types.py
@@ -25,6 +25,7 @@ from typing import (
     TYPE_CHECKING,
     NewType,
     Protocol,
+    Sequence,
     Type,
     TypeVar,
     Union,
@@ -50,16 +51,33 @@ if TYPE_CHECKING:
     from .lib.address import Address
     from .lib.attribute import Attribute
     from .lib.attrtype import AttributeType
+    from .lib.childref import ChildRef
+    from .lib.childreftype import ChildRefType
     from .lib.date import Date
     from .lib.eventref import EventRef
     from .lib.eventroletype import EventRoleType
+    from .lib.eventtype import EventType
+    from .lib.familyreltype import FamilyRelType
     from .lib.grampstype import GrampsType
     from .lib.ldsord import LdsOrd
+    from .lib.location import Location
     from .lib.mediaref import MediaRef
     from .lib.name import Name
     from .lib.nameorigintype import NameOriginType
     from .lib.nametype import NameType
+    from .lib.notetype import NoteType
     from .lib.personref import PersonRef
+    from .lib.placename import PlaceName
+    from .lib.placeref import PlaceRef
+    from .lib.placetype import PlaceType
+    from .lib.reporef import RepoRef
+    from .lib.repotype import RepositoryType
+    from .lib.srcattribute import SrcAttribute
+    from .lib.srcattrtype import SrcAttributeType
+    from .lib.srcmediatype import SourceMediaType
+    from .lib.styledtext import StyledText
+    from .lib.styledtexttag import StyledTextTag
+    from .lib.styledtexttagtype import StyledTextTagType
     from .lib.surname import Surname
     from .lib.url import Url
     from .lib.urltype import UrlType
@@ -149,11 +167,11 @@ class NameLike(Protocol):
     """Protocol for Name-like objects."""
 
     private: bool
-    citation_list: list[str]
-    note_list: list[str]
+    citation_list: Sequence[str]
+    note_list: Sequence[str]
     date: DateLike
     first_name: str
-    surname_list: list[SurnameLike]
+    surname_list: Sequence[SurnameLike]
     suffix: str
     title: str
     type: GrampsTypeLike  # NameType
@@ -170,9 +188,9 @@ class EventRefLike(Protocol):
     """Protocol for EventRef-like objects."""
 
     private: bool
-    citation_list: list[str]
-    note_list: list[str]
-    attribute_list: list[AttributeLike]
+    citation_list: Sequence[str]
+    note_list: Sequence[str]
+    attribute_list: Sequence[AttributeLike]
     ref: str  # Event handle
     role: GrampsTypeLike  # EventRoleType
 
@@ -182,11 +200,11 @@ class MediaRefLike(Protocol):
     """Protocol for MediaRef-like objects."""
 
     private: bool
-    citation_list: list[str]
-    note_list: list[str]
-    attribute_list: list[AttributeLike]
+    citation_list: Sequence[str]
+    note_list: Sequence[str]
+    attribute_list: Sequence[AttributeLike]
     ref: str  # Media handle
-    rect: list[int] | None  # Region coordinates [x, y, width, height]
+    rect: Sequence[int] | None  # Region coordinates [x, y, width, height]
 
 
 @runtime_checkable
@@ -194,8 +212,8 @@ class AddressLike(Protocol):
     """Protocol for Address-like objects."""
 
     private: bool
-    citation_list: list[str]
-    note_list: list[str]
+    citation_list: Sequence[str]
+    note_list: Sequence[str]
     date: DateLike
     street: str
     locality: str
@@ -212,8 +230,8 @@ class AttributeLike(Protocol):
     """Protocol for Attribute-like objects."""
 
     private: bool
-    citation_list: list[str]
-    note_list: list[str]
+    citation_list: Sequence[str]
+    note_list: Sequence[str]
     type: GrampsTypeLike  # AttributeType
     value: str
 
@@ -232,8 +250,8 @@ class UrlLike(Protocol):
 class LdsOrdLike(Protocol):
     """Protocol for LdsOrd-like objects."""
 
-    citation_list: list[str]
-    note_list: list[str]
+    citation_list: Sequence[str]
+    note_list: Sequence[str]
     date: DateLike
     type: int  # LDS ordinance type
     place: str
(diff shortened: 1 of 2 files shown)


gramps/gen/types.py
```diff
@@ -84,3 +110,194 @@
     NoteGrampsID,
 ]
 AnyGrampsID = PrimaryObjectGrampsID  # No Tag IDs
+
+
+@runtime_checkable
+class GrampsTypeLike(Protocol):
+    """Protocol for GrampsType-like objects (NameType, AttributeType, etc.)."""
```
bartfeenstra, line 98: If protocols are used, they should be the 'primary' types. In other words, concrete types should refer to the protocols, and not the other way around, so a `GrampsType` is defined as a `GrampsTypeLike`, but a `GrampsTypeLike` is not defined as a `GrampsType`-like.

DavidMStraub: I don't agree actually - unless I'm misunderstanding what you mean. The whole point of protocols is that they can be used without inheritance. For instance, `DataDict` cannot inherit from `PersonLike` because not every `DataDict` has a `surname_list` (it might be a family or an event etc.).

It's like with database proxies in Gramps that sometimes don't inherit from `DbGeneric` but just share a subset of its methods.

bartfeenstra: If protocols are used like interfaces (as in the abstract programming concepts), such interfaces should not reference concrete types. For example, a 'person-like' interface is not a 'person' (concrete type). Instead, it's the other way around: a 'person' concrete type is a 'person-like' interface. That way other code can type on the interface (the protocol, in Python terminology), instead of any concrete type.

DavidMStraub: I agree. Anything in the code or in what I said that goes against that?

bartfeenstra: Just the docstrings, I believe. I wasn't quite clear about that

In essence, with my limited knowledge of Gramps internals, this looks like a good improvement. Are there plans to introduce additional concrete implementations in the future?

DavidMStraub: I'd be happy to complete it for all the other primary types so we can make use of it in other PRs if the feedback is positive.

Nick-Hall:
This is an interesting proof of concept, but I'm not sure that it is good solution.  In particular, I don't like the way that the internals of objects are exposed.

I think that sub-typing `DataDict` should be considered.  In that way a `PersonLike` could be defined as `Person | PersonDataDict`.

DavidMStraub:
I was about about to remove the "PoC" and mark this as ready.

> I think that sub-typing DataDict should be considered. In that way a PersonLike could be defined as Person | PersonDataDict.

The problem (actually, one of several problems) is that `DataDict` uses `__getattr__`, which is not visible to static type checkers. So what I did is exactly like sub-typing `DataDict`. In fact, I originally had

`PersonLike = PersonDataDict | Person`

and `PersonDataDict` was the `Protocol`. The neat thing is, this is actually not necessary because `Person` turns out to pass an `isinstance` check of the protocol at runtime – this is perfect duck typing at work, both `Person` and a `DataDict` derived from a person walk like the protocol, talk like it, and quack like it.

DavidMStraub:
> Can we generate the protocols so that we don't have to maintain them manually? I really don't want to go back to having two places to update.

I don't think so, not without major runtime-code refactoring.

In fact, it would be three places in the future as we already have two places: the properties themselves and the schema in `get_schema`.

In a clean sheet design, one would probably nowadays implement all primary objects as ~~dataclasess~~ Pydantic models, then the schema, type hints *and* dict-like behaviour would come for free. But I don't think we can realistically go there. 

I understand that duplication of schema changes is annoying and I don't disagree, but as long as unit tests (and static type checks!) monitor internal consistency, for me it would be an acceptable nuisance (or a necessary evil).

dsblank (changes requested):
Except for the issue noted in https://github.com/gramps-project/gramps/pull/2184 it looks good!

DavidMStraub:
Linter errors should be fixed by #2184.

dsblank (approved):
LGTM!

gramps/gen/types.py
```diff
@@ -84,3 +91,414 @@
...
+    postal: str
+    phone: str
+    parish: str
+
+
+@runtime_checkable
+class PersonLike(Protocol):
+    """Protocol for Person-like objects."""
+
+    handle: str
```
stevenyoungs:
```suggestion
    handle: PersonHandle
```
Can this use PersonHandle rather than str.
Similar for the `handle` attribute of the other `*Like` classes

gramps/gen/types.py
```diff
@@ -84,3 +91,414 @@
...
+    gramps_id: str
```
stevenyoungs:
```suggestion
    gramps_id: PersonGrampsID
```
Similar for the `gramps_id` attribute of the other `*Like` classes

gramps/gen/types.py
```diff
@@ -84,3 +91,414 @@
...
+    private: bool
+    person_ref_list: list[PersonRefLike]
+
+
+@runtime_checkable
+class FamilyLike(Protocol):
+    """Protocol for Family-like objects."""
+
+    handle: str
+    gramps_id: str
```
stevenyoungs:
```suggestion
    gramps_id: FamilyGrampsID
```

gramps/gen/types.py
```diff
@@ -84,3 +91,414 @@
...
+    tag_list: list[str]
+    private: bool
+    person_ref_list: list[PersonRefLike]
+
+
+@runtime_checkable
+class FamilyLike(Protocol):
+    """Protocol for Family-like objects."""
+
+    handle: str
```
stevenyoungs:
```suggestion
    handle: FamilyHandle
```

gramps/gen/types.py
```diff
@@ -84,3 +91,414 @@
...
+    tag_list: list[str]
+    private: bool
+
+
+@runtime_checkable
+class EventLike(Protocol):
+    """Protocol for Event-like objects."""
+
+    handle: str
+    gramps_id: str
```
stevenyoungs:
```suggestion
    handle: EventHandle
    gramps_id: EventGrampsID
```

gramps/gen/types.py
```diff
@@ -84,3 +91,414 @@
...
+    tag_list: list[str]
+    private: bool
+
+
+@runtime_checkable
+class MediaLike(Protocol):
+    """Protocol for Media-like objects."""
+
+    handle: str
+    gramps_id: str
```
stevenyoungs:
```suggestion
    handle: MediaHandle
    gramps_id: MediaGrampsID
```

gramps/gen/types.py
```diff
@@ -84,3 +91,414 @@
...
+    tag_list: list[str]
+    private: bool
+
+
+@runtime_checkable
+class PlaceLike(Protocol):
+    """Protocol for Place-like objects."""
+
+    handle: str
+    gramps_id: str
```
stevenyoungs:
```suggestion
    handle: PlaceHandle
    gramps_id: PlaceGrampsID
```

gramps/gen/types.py
```diff
@@ -84,3 +91,414 @@
...
+    tag_list: list[str]
+    private: bool
+
+
+@runtime_checkable
+class SourceLike(Protocol):
+    """Protocol for Source-like objects."""
+
+    handle: str
+    gramps_id: str
```
stevenyoungs:
```suggestion
    handle: SourceHandle
    gramps_id: SourceGrampsID
```

gramps/gen/types.py
```diff
@@ -84,3 +91,414 @@
...
+    tag_list: list[str]
+    private: bool
+
+
+@runtime_checkable
+class CitationLike(Protocol):
+    """Protocol for Citation-like objects."""
+
+    handle: str
+    gramps_id: str
```
stevenyoungs:
```suggestion
    handle: CitationHandle
    gramps_id: CitationGrampsID
```

gramps/gen/types.py
```diff
@@ -84,3 +91,414 @@
...
+    tag_list: list[str]
+    private: bool
+
+
+@runtime_checkable
+class RepositoryLike(Protocol):
+    """Protocol for Repository-like objects."""
+
+    handle: str
+    gramps_id: str
```
stevenyoungs:
```suggestion
    handle: RepositoryHandle
    gramps_id: RepositoryGrampsID
```

gramps/gen/types.py
```diff
@@ -84,3 +91,414 @@
...
+    tag_list: list[str]
+    private: bool
+
+
+@runtime_checkable
+class NoteLike(Protocol):
+    """Protocol for Note-like objects."""
+
+    handle: str
+    gramps_id: str
```
stevenyoungs:
```suggestion
    handle: NoteHandle
    gramps_id: NoteGrampsID
```

gramps/gen/types.py
```diff
@@ -84,3 +91,414 @@
...
+    change: int
+    tag_list: list[str]
+    private: bool
+
+
+@runtime_checkable
+class TagLike(Protocol):
+    """Protocol for Tag-like objects."""
+
+    handle: str
```
stevenyoungs:
```suggestion
    handle: TagHandle
```

stevenyoungs (changes requested):

DavidMStraub:
Thanks, suggestions applied and rebased.

It's really time to merge this so we can start using it. IMO it doesn't make sense to keep this open for so long, because work to use it in the code base STARTS after it's merged.

Nick-Hall:
I'm still not happy with this yet.   I'll review it after v6.1 has been released.

DavidMStraub:
Ok. I still think it would be better to add type hints and then refactor and not the other way around, but I accept it of course. Will wait with adding more type hints until after 6.1.

PR #2129, opened 2025-09-27
Decouple the core library code from GObject introspection

Nick-Hall:
Replaces #2016 by an approach suggested by @jralls.

### DIFF
--- a/gramps/gen/const.py
+++ b/gramps/gen/const.py
@@ -38,18 +38,22 @@ import logging
 
 LOG = logging.getLogger(".")
 
-from gi.repository import GLib
-
 # -------------------------------------------------------------------------
 #
 # Gramps modules
 #
 # -------------------------------------------------------------------------
 from .git_revision import get_git_revision
-from .constfunc import get_env_var
+from .constfunc import (
+    get_env_var,
+    get_user_cache_dir,
+    get_user_config_dir,
+    get_user_data_dir,
+)
 from ..version import VERSION, VERSION_TUPLE, major_version, DEV_VERSION
 from .utils.resourcepath import ResourcePath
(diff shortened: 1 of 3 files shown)


bartfeenstra:
Which Gramps version is this targeted at?

Nick-Hall:
It targets the _master_ branch which will probably become the v6.1 feature release.

jralls:
LGTM.

dsblank (approved):
This looks like a fine solution to the issue.

gramps/gen/const.py
```diff
@@ -128,12 +132,12 @@
             if os.path.exists(db_dir):
                 shutil.copytree(db_dir, os.path.join(USER_DATA, "grampsdb"))
 
-USER_CACHE = os.path.join(GLib.get_user_cache_dir(), "gramps")
+USER_CACHE = os.path.join(get_user_cache_dir(), "gramps")
```
stevenyoungs, line 135: @Nick-Hall this change is throwing an exception on Windows (for me at least).
Inside `get_user_cache_dir()`, the is no env variable called `XDG_CACHE_HOME`. It then it assumes that there will be an env variable called `HOME`. On Windows this does not exist. the subsequent call to `join` then throws an exception when passed None as the first arg
`os.path.join(get_env_var("HOME"), ".cache")`

We've already defined `USER_HOME`. Could it use that?

Nick-Hall: Yes, Using `USER_HOME` would solve the problem, but importing it may cause an import cycle.

stevenyoungs: See PR #2201 for my proposed fix

PR #2141, opened 2025-11-18
Coalesce event role types

stevenyoungs:
PR #1715 introduced new built-in event role types
If a database, created with an earlier version of gramps, defined custom role types of the same name, duplicates would exist.
There is no easy way for the user to disambiguate these two role types
This PR forces an update to the db schema number and coalesces the event role types during upgrade


stevenyoungs:
> Please move this to the _master_ branch.

Rebased to master. I've not yet been able to test as master is currently broken on Windows.

Code revised to only commit the person or family if at least one event_ref is upgraded

> A solution for v6.0 would be to add a fix to the Check & Repair tool.

I'll look into this and commit a separate PR if I get something working.

stevenyoungs:
[PR #2145](https://github.com/gramps-project/gramps/pull/2145) has, in my view, a slightly better algorithm. If it passes review, I will make the corresponding updates to this PR.

dsblank (changes requested):
## Summary

The PR adds a database upgrade (v21→v22) to fix a real problem introduced in PR #1715: databases created before those new built-in `EventRoleType` values (FATHER, MOTHER, PARENT, CHILD, etc.) may have had user-defined *custom* types with the same name, causing duplicates.

The approach is sound, but there are a few issues worth raising:

---

## Issues

### 1. Critical (correctness): Hidden logic in a no-op-looking line

In `upgrade_event_ref_role_22` (`upgrade.py:98`):
```python
event_role_type.string = event_role_type.string
```
This looks like a no-op, but it's actually the core of the whole upgrade. `string` is a property whose **setter** (`__set_str`) does a lookup in `_S2IMAP` (the localized-string → integer map). So assigning `"Father"` to `.string` actually changes `.value` from 0 (CUSTOM) to 11 (FATHER) — if the string matches a built-in localized name.

This is completely non-obvious. A future reader will almost certainly delete it as dead code. It should be rewritten explicitly, e.g.:
```python
# Attempt to map the custom string to a known built-in type:
event_role_type.set(event_role_type.string)
```
Or at minimum add a comment explaining the mechanism. `set(str)` dispatches to `__set_str` via `singledispatchmethod`, which is equivalent and more readable.

### 2. Moderate (robustness): Locale dependency

The lookup uses `_S2IMAP`, which is indexed by the **localized** string. This means:
- If the user created a custom role type that matched a built-in **by its English name** (e.g., "Father") but the upgrade runs with a non-English locale (where the built-in maps to "Vater" / "Père" / etc.), the coalescing will silently fail.
- Conversely, if the user's locale changed between creating the custom type and running the upgrade, the lookup can also fail.

A safer approach would be to also try `set_from_xml_str()`, which uses `_E2IMAP` (English XML names), as a fallback. That said, the localized lookup is the *right* first attempt since user-entered strings would normally be in the display locale.

### 3. Minor: `new_event_ref_list` is a redundant copy

The function mutates `event_ref["role"]` in-place, then appends the same `event_ref` DataDict to `new_event_ref_list`. The returned list is functionally identical to the input list — both contain the same (mutated) objects. The whole `new_event_ref_list` construction could be removed:
```python
def upgrade_event_ref_role_22(event_ref_list):
    upgraded = False
    for event_ref in event_ref_list:
        if event_ref.role.value == EventRoleType.CUSTOM:
            event_role_type = data_to_object(event_ref.role)
            event_role_type.set(event_role_type.string)  # attempt built-in lookup
            if event_role_type.value != EventRoleType.CUSTOM:
                upgraded = True
                event_ref["role"] = object_to_data(event_role_type)
    return upgraded, event_ref_list
```

### 4. Minor: Copyright year

Line 7 of `upgrade.py` says `# Copyright (C) 2025` but it's 2026.

### 5. Minor: No tests

Other upgrade functions don't always have tests either, but given the subtle locale-dependent mechanism here a unit test would be valuable to guard against regression (especially the no-op-looking line getting deleted).

---

## What's Good

- The version bump (21→22) and `_gramps_upgrade` dispatch wiring look correct and consistent with existing patterns.
- Both `Person` and `Family` event_ref lists are handled (the only two object types with `event_ref_list`).
- `self.update()` is called in both loops for progress reporting.
- The `_txn_begin/commit` and `_set_metadata("version", 22, use_txn=False)` pattern matches `gramps_upgrade_21`.
- Import additions (`EventRoleType`, `data_to_object`, `object_to_data`) are correct.

dsblank (commented):
## Review of Updated Changes

The author has made substantial revisions addressing most of the issues raised in the previous review. Here's how each concern was handled:

---

### Previous Issues — Status

**Issue 1 (Critical): Hidden no-op-looking line** ✅ Fixed
The cryptic `event_role_type.string = event_role_type.string` is gone. The logic now uses proper Gramps ORM objects: `event_ref.get_role()` / `event_ref.set_role(event_role.string)`. This is clear and correct.

**Issue 2 (Moderate): Locale dependency** ✅ Acknowledged with a comment
The author added a clear comment explaining the locale-sensitivity limitation. The approach is pragmatic — matching localized strings is correct for the common case. This is an acceptable trade-off for a schema upgrade.

**Issue 3 (Minor): Redundant `new_event_ref_list` copy** ✅ Fixed
The redundant list construction is gone. The function now mutates objects in-place and returns a `modified` boolean. Clean.

**Issue 4 (Minor): Copyright year** ✅ Fixed
Copyright now reads `2025-2026`.

**Issue 5 (Minor): No tests** ⚠️ Partially addressed
Test data files were added (`imp_5_2_4_Duplicate_EventRef_Role_Names.gramps` and `.zip`) but there is no Python test file actually exercising the upgrade path. The test data exists but nothing runs it automatically.

---

### New Issues Found

**1. Dead import: `data_to_object`, `object_to_data`** — Minor
`upgrade.py:43` imports these but neither is used anywhere in the file. Should be removed:
```python
# Remove this line:
from gramps.gen.lib.json_utils import data_to_object, object_to_data
```

**2. `event_role_names.discard()` bypasses the DB** — Moderate
The custom role deletion at the end mutates an in-memory set but doesn't call any persistence method to write the updated role names back to the database. This could mean the duplicate custom roles reappear after a restart. Worth verifying that this in-memory mutation is sufficient for all backends, or comparing with how other upgrade functions flush custom type lists.

**3. `object` shadows the builtin** — Minor style
The inner function `def update_event_ref_list(object, duplicate_roles)` uses `object` as a parameter name, shadowing Python's built-in. Rename to `obj` or something more specific like `person_or_family`.

**4. `if len(duplicate_roles):` — Minor style**
Prefer `if duplicate_roles:` (standard Python idiom for checking non-empty collections).

**5. No Python unit test exercising the upgrade** — Moderate
The `.gramps` test fixture is useful for manual testing but isn't wired into the test suite. Given how subtle the upgrade logic is, a proper automated test that imports the fixture and verifies role types were coalesced would be valuable.

---

### Summary

The PR is significantly improved — the core logic is now correct and readable. The two items most worth addressing before merge are: (1) the unused import of `data_to_object`/`object_to_data`, and (2) verifying that `event_role_names.discard()` actually persists the removal to the database backend across restarts.

stevenyoungs:
> **2. `event_role_names.discard()` bypasses the DB** — Moderate

@dsblank I could not work out how to force a sync to the backend. Any ideas?

stevenyoungs:
> **5. No Python unit test exercising the upgrade** — Moderate
> The `.gramps` test fixture is useful for manual testing but isn't wired into the test suite. Given how subtle the upgrade logic is, a proper automated test that imports the fixture and verifies role types were coalesced would be valuable.

I believe claude is wrong here 😟 . Can you ask it to examine `imports_test.py`. This code dynamically creates a test called `gramps.plugins.test.imports_test.TestImports.test_imp_5_2_4_Duplicate_EventRef_Role_Names_zip`

I can manually run the test as 
`LANGUAGE= LANG=en_US.utf-8 GRAMPS_RESOURCES=. python -m unittest gramps.plugins.test.imports_test.TestImports.test_imp_5_2_4_Duplicate_EventRef_Role_Names_zip`

stevenyoungs:
> ### New Issues Found
> 1. Dead import: `data_to_object`, `object_to_data`
> 
> 3. `object` shadows the builtin
> 
> 4. `if len(duplicate_roles):`

The above points are addressed in the latest commits

dsblank:
@stevenyoungs there are some merge conflicts between this and FamilySearch. I was going to resolve them, but is going to require a bit more thought (one uses Person object, one uses Person data).

stevenyoungs:
> @stevenyoungs there are some merge conflicts between this and FamilySearch. I was going to resolve them, but is going to require a bit more thought (one uses Person object, one uses Person data).

@dsblank rebased and updated to schema 23 for gramps 6.2

PR #2142, opened 2025-11-20
subdirs and comments in narweb css files

SNoiraud:
fixes [#14013](https://gramps-project.org/bugs/view.php?id=14013)


SNoiraud:
Problem with build:
gramps/plugins/webstuff/webstuff.py:26: error: Skipping analyzing "pyparsing": module is installed, but missing library stubs or py.typed marker  [import-untyped]
gramps/plugins/webstuff/webstuff.py:26: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports

dsblank (changes requested):
## Review: subdirs and comments in narweb css files

The fix addresses a real bug and the approach is reasonable, but there are a few issues to address before merge.

### 1. New undeclared dependency (blocker)

`pyparsing` is imported unconditionally at module level (`webstuff.py:26`) but is **not listed in `pyproject.toml` dependencies** (only `orjson` is listed there). This will cause an `ImportError` for users without pyparsing installed. It should be added to `dependencies` in `pyproject.toml`:

```toml
dependencies = ["orjson", "pyparsing"]
```

### 2. Deprecated `pyparsing` API

`pyparsing.nestedExpr` has been deprecated since pyparsing 3.x in favour of `pyparsing.nested_expr`:

```python
# Should be:
comments = pyparsing.nested_expr("/*", "*/").suppress()
```

### 3. Subdirectory path parsing is fragile

In `narrativeweb.py`, the splitting logic only handles one level of nesting. For a path like `.../css/subdir1/subdir2/file.css`:
- `cssdir = subdir[1].split("/")` → `["subdir1", "subdir2", "file.css"]`
- `cssdir[1]` = `"subdir2"` (wrong — should be the filename)
- `cssdir[0]` = `"subdir1"` (misses `subdir2`)

A more robust approach using `rsplit`:

```python
css_idx = from_path.find("css/")
if css_idx != -1:
    rel = from_path[css_idx + 4:]  # portion after "css/"
    if "/" in rel:
        subdir, fname = rel.rsplit("/", 1)  # handles any depth
        newdir = os.path.join("css", subdir)
        self.copy_file(from_path, fname, newdir)
        continue
dummy_fdir, fname = os.path.split(from_path)
self.copy_file(from_path, fname, "images")
```

### 4. Ad-hoc string replacements in URL cleanup

The chained `.replace("/css/", "")` call in `webstuff.py` will strip any occurrence of `/css/` anywhere in a URL, not just as a leading path prefix. Using `strip("'\"")` for quote removal and a more targeted path cleanup would be safer.

kulath:
Is there an alternative to pyparsing, because it is undesirable to add another module to the build, and to the contents of the various AIO builds

PR #2149, opened 2025-12-11
fix: add resilience to set_birth_death_index() for storage backend failures

mahula:
### Context

This fix was discovered while investigating a bug in Gramps Web where birth/death dates would disappear after editing a person. The investigation revealed two separate issues:

1. **gramps-web-api**: Unnecessary recalculation of indices on every Person update (https://github.com/gramps-project/gramps-web-api/pull/744)
2. **gramps core** (this PR): `set_birth_death_index()` not resilient to storage backend read failures

### Problem

The `set_birth_death_index()` method reads Event objects from the database to determine which event references correspond to birth and death events. On certain storage backends, these reads can fail or return `None`:

- **NAS devices with Btrfs filesystem** (Copy-on-Write can cause read inconsistencies during write transactions)
- **Network-mounted storage** with latency or locking issues
- **Any storage** where `get_event_from_handle()` might fail transiently

When an event read fails, the current implementation simply doesn't find a matching birth/death event and resets the index to `-1`, even if a valid index existed before.

### Environment Where Issue Was Discovered

- NAS with Btrfs filesystem
- Docker containers with host-mounted volumes  
- SQLite database in WAL mode
- Gramps Web API calling `set_birth_death_index()` during Person updates

### Solution

Add defensive handling to preserve existing indices when event reads fail:

1. **Store original indices** (`birth_ref_index`, `death_ref_index`) before recalculation
2. **Handle `None` return** from `get_event_from_handle()` gracefully
3. **Wrap event reads in try/except** to catch storage-related exceptions
4. **Preserve existing indices** when reads fail, if they point to the current event reference

### Changes

- `gramps/gen/db/base.py`: Enhanced `set_birth_death_index()` with fallback logic

### Testing

- Tested on ext4 filesystem - no regression
- Tested on Synology NAS with Btrfs filesystem - birth/death dates preserved correctly

### Breaking Changes

None. This is a defensive improvement that maintains existing behavior on well-functioning storage backends.

### DIFF
--- a/gramps/gen/db/base.py
+++ b/gramps/gen/db/base.py
@@ -2025,24 +2025,58 @@ class DbWriteBase(DbReadBase):
     def set_birth_death_index(self, person):
         """
         Set the birth and death indices for a person.
+
+        This method iterates through the person's event references and
+        identifies which events are birth and death events based on their
+        type and the role of the person in the event.
+
+        Note: On some storage backends (e.g., NAS with host-mounted volumes),
+        reading events during a write transaction may fail. In such cases,
+        we preserve the existing indices if they appear valid.
         """
         birth_ref_index = -1
         death_ref_index = -1
         event_ref_list = person.get_event_ref_list()
+
+        # Store original indices to use as fallback
+        original_birth_index = person.birth_ref_index
+        original_death_index = person.death_ref_index
+
         for index, ref in enumerate(event_ref_list):
-            event = self.get_event_from_handle(ref.ref)
-            if (
-                event.type.is_birth()
-                and ref.role.is_primary()
-                and (birth_ref_index == -1)
-            ):
-                birth_ref_index = index
-            elif (
-                event.type.is_death()
-                and ref.role.is_primary()
-                and (death_ref_index == -1)
-            ):
-                death_ref_index = index
+            try:
+                event = self.get_event_from_handle(ref.ref)
+                # If event cannot be loaded (can happen with file system issues),
+                # preserve existing indices if they point to this index
+                if event is None:
+                    if original_birth_index == index:
+                        birth_ref_index = index
+                    if original_death_index == index:
+                        death_ref_index = index
+                    continue
+
+                # Check if this is a birth event with primary role
+                if (
+                    event.type.is_birth()
+                    and ref.role.is_primary()
+                    and (birth_ref_index == -1)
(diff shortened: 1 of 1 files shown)


DavidMStraub:
I don't think this is the right approach *or* I simply don't understand it.

If `self.get_event_from_handle(ref.ref)` raises a `HandleError`, you'll catch it with your (too broad) `except Exception` and leave the indices unchanged - even if the event disappeared?

And even if it doesn't raise a `HandleError` but returns `None` - it could simply be because the event was made private and we're in a `PrivateProxyDb` (which doesn't raise `HandleError`). Then again, we shouldn't just pretend the indices don't change.

If we *really* have a filesystem error that prevents us from working with the database, I don't see why we shouldn't simply let the exception happen here rather than swallow it, and have it handled by higher level code.

dsblank (changes requested):
Thank you for investigating this and providing a clear description of the problem. The concern is valid — the current code would crash with `AttributeError` if `get_event_from_handle()` returns `None`, and silent index resets to `-1` on transient failures are a real problem.

That said, I have a few concerns with the proposed solution before this can be merged:

### 1. No logging on failure

When the fallback kicks in, there is no log message. Storage failures should be logged at `WARNING` level so users and administrators can detect when this defensive behavior is triggered. Silent fallback makes diagnosing issues much harder. Something like:

```python
except Exception as e:
    LOG.warning("Failed to read event %s for person %s: %s", ref.ref, person.handle, e)
```

### 2. Bare `except Exception` hides bugs

Catching all exceptions without logging swallows programming errors (e.g., `AttributeError`, `TypeError`) that should surface during development. At minimum, log the exception.

### 3. Correctness of the fallback logic

The fallback preserves `birth_ref_index = index` when `original_birth_index == index` — but if the read failed, we don't actually know that the event at that index is still a birth event. The event may have changed type, been deleted, or the handle may be dangling. Preserving the old index in this case could silently point to the wrong event. It may be better to log a warning and leave the index at `-1` rather than preserve a potentially stale value.

### 4. Minor: implementation detail in docstring

The new docstring block describes internal implementation details (fallback logic, storage backend caveats) that are better as inline comments — which you've also added. Consider keeping the docstring at a higher level.

Please address at least the logging concern before merging — visibility into fallback behavior is important for production deployments.

PR #2153, opened 2025-12-17
Use filter.apply() rather than apply_to_one()

dsblank:
This PR replaces this construct when using a filter inside a filter:

```python
for person in db.iter_people():
    if self.filt.apply_to_one(db, person):
        ...
```

with the better:

```python
for handle in self.filt.find_filter().apply(db):
    person = db.get_raw_person_data(handle)
```

This has possible speed increases in two ways:

1. The filter.apply() can use the filter optimizer
2. The selected_handles are then used rather than the whole database 

This could be consider a bug in Gramps 6.0, but I think it makes sense as part of a set of optimizations for the next version of Gramps.

### DIFF
--- a/gramps/gen/filters/rules/person/_deeprelationshippathbetween.py
+++ b/gramps/gen/filters/rules/person/_deeprelationshippathbetween.py
@@ -24,7 +24,6 @@
 # Standard Python modules
 #
 # -------------------------------------------------------------------------
-from __future__ import annotations
 from collections import deque
 
 # -------------------------------------------------------------------------
@@ -41,7 +40,6 @@ from ....const import GRAMPS_LOCALE as glocale
 # Typing modules
 #
 # -------------------------------------------------------------------------
-from typing import Union, List, Set, Dict
 from ....lib import Person
 from ....db import Database
 from ....types import FamilyHandle, PersonHandle
@@ -57,7 +55,7 @@ _ = glocale.translation.gettext
 def get_family_handle_people(
     db: Database, exclude_handle: str, family_handle: FamilyHandle
 ):
-    people: Set[str] = set()
+    people: set[str] = set()
 
     family = db.get_family_from_handle(family_handle)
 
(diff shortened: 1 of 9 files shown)


gramps/gen/filters/rules/person/_deeprelationshippathbetween.py
```diff
@@ -166,28 +166,18 @@ def prepare(self, db: Database, user):
...
-            if user:
-                user.step_progress()
-        if user:
-            user.end_progress()
-            user.begin_progress(
-                _("Finding relationship paths"),
-                _("Evaluating people"),
-                db.get_number_of_people(),
-            )
+        target_people = self.filt.find_filter().apply(db)
```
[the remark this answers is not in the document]
dsblank: The return list supports mutation.

gramps/gen/filters/rules/person/_deeprelationshippathbetween.py
```diff
@@ -166,28 +166,18 @@ def prepare(self, db: Database, user):
...
+
+        user.begin_progress(
```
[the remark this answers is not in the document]
dsblank: Good point.

gramps/gen/filters/rules/person/_hascommonancestorwithfiltermatch.py
```diff
@@ -66,23 +66,13 @@ def prepare(self, db, user):
...
-            if person and self.filt.apply_to_one(db, person):
-                # store all people in the filter so as to compare later
-                self.with_people.append(person.handle)
-                # fill list of ancestor of person if not present yet
-                if person.handle not in self.ancestor_cache:
-                    self.add_ancs(db, person)
-        if user:
-            user.end_progress()
+
+        for handle in self.filt.find_filter().apply(db):
```
stevenyoungs: Is this loop now fast enough that user progress updates are no longer justified\required?

stevenyoungs: This comment also applies to the other loops where `user.step_progress()` is no longer called

dsblank: It isn't that the loop is fast enough to no longer need to update the user, but the expensive part has been moved to "apply()". If we need to update the user, it should happen there (and I think it already does).

dsblank:
Changed the target base.

PR #2154, opened 2025-12-21
Improved Manage Family Trees behavior when deleting trees: preserve sort column and better next item selection

hgohel:
Manage Family Trees list unexpectedly (1) reverts the sort column selection to Name when a tree is deleted from the list and (2) always selects the most recently accessed tree as the current selection after a tree is deleted, rather than the next/previous item.

There are two commits in this PR for easy review:
1. 68bb59880369addb4e4117004090844d79742c19 will preserve the sort order when the list is rebuilt after a tree is deleted
2. 0057cd0a4ad873b89ed392598443e29bc8dc0370 will remember the item that was next after an item that has been deleted (or previous item if there's no next), and selects that after the list is rebuilt.

Testing performed:
With sort order set to name, last accessed etc., repeat each step below and verify that the selection after delete is reasonable, i.e. next item, previous item, or none if there are no more items in family tree list.

- delete item at top of list
- delete item at bottom of list
- delete item in middle of list

Fixes [#12187](https://gramps-project.org/bugs/view.php?id=12187)

To execute three new unit tests:

`python -m unittest gramps.gui.test.dbman_selection_test.TestDbManSelection`

AI Usage: KwaiKAT's KAT-Coder-Pro V1 used within Cline host to fix the bugs, with manual tweaking before commit. Further changes and unit tests by GitHub Copilot.

### DIFF
--- a/gramps/gui/dbman.py
+++ b/gramps/gui/dbman.py
@@ -446,6 +446,11 @@ class DbManager(CLIDbManager, ManagedWindow):
         """
         Builds the display model.
         """
+        # Store current sort order before rebuilding model
+        current_sort_column, current_sort_order = None, None
+        if self.model is not None:
+            current_sort_column, current_sort_order = self.model.get_sort_column_id()
+        
         self.model = Gtk.TreeStore(str, str, str, str, int, bool, str, str)
 
         # use current names to set up the model
@@ -476,7 +481,12 @@ class DbManager(CLIDbManager, ManagedWindow):
                 self.model.append(node, data)
         if self._current_node is None:
             self._current_node = last_accessed_node
-        self.model.set_sort_column_id(NAME_COL, Gtk.SortType.ASCENDING)
+        
+        # Restore the previous sort order, or default to NAME_COL ascending if no previous sort
+        if current_sort_column is not None and current_sort_order is not None:
+            self.model.set_sort_column_id(current_sort_column, current_sort_order)
+        else:
+            self.model.set_sort_column_id(NAME_COL, Gtk.SortType.ASCENDING)
         self.dblist.set_model(self.model)
 
     def existing_name(self, name, skippath=None):
--- a/gramps/gui/dbman.py
+++ b/gramps/gui/dbman.py
@@ -237,13 +237,13 @@ class DbManager(CLIDbManager, ManagedWindow):
     def _select_next_after_deletion(self, next_item_name):
         """
         Select the item that would have been next after deletion.
-        
+
         Args:
             next_item_name: The name of the item that should be selected
         """
         if self.model is None or not next_item_name:
             return
-            
+
         # Find the item with the matching name in the rebuilt model
         iter = self.model.get_iter_first()
         while iter:
@@ -471,7 +471,7 @@ class DbManager(CLIDbManager, ManagedWindow):
         current_sort_column, current_sort_order = None, None
         if self.model is not None:
             current_sort_column, current_sort_order = self.model.get_sort_column_id()
-        
+
         self.model = Gtk.TreeStore(str, str, str, str, int, bool, str, str)
 
         # use current names to set up the model
@@ -502,7 +502,7 @@ class DbManager(CLIDbManager, ManagedWindow):
                 self.model.append(node, data)
         if self._current_node is None:
             self._current_node = last_accessed_node
-        
+
         # Restore the previous sort order, or default to NAME_COL ascending if no previous sort
         if current_sort_column is not None and current_sort_order is not None:
             self.model.set_sort_column_id(current_sort_column, current_sort_order)
@@ -808,7 +808,7 @@ class DbManager(CLIDbManager, ManagedWindow):
         path = store.get_path(node)
         node = self.model.get_iter(path)
         filename = self.model.get_value(node, FILE_COL)
-        
+
         # Remember what should be selected next (the item after the deleted one)
         next_iter = self.model.iter_next(node)
         next_item_name = None
@@ -819,7 +819,7 @@ class DbManager(CLIDbManager, ManagedWindow):
             prev_iter = store.iter_previous(node)
             if prev_iter is not None:
                 next_item_name = self.model.get_value(prev_iter, NAME_COL)
-        
+
         try:
             with open(filename, "r", encoding="utf-8") as name_file:
                 file_name_to_delete = name_file.read()
--- a/gramps/gui/dbman.py
+++ b/gramps/gui/dbman.py
@@ -234,6 +234,27 @@ class DbManager(CLIDbManager, ManagedWindow):
             self.selection.select_path(tree_path)
             self.dblist.scroll_to_cell(tree_path, None, 1, 0.5, 0)
 
+    def _select_next_after_deletion(self, next_item_name):
+        """
+        Select the item that would have been next after deletion.
+        
+        Args:
+            next_item_name: The name of the item that should be selected
+        """
+        if self.model is None or not next_item_name:
+            return
+            
+        # Find the item with the matching name in the rebuilt model
+        iter = self.model.get_iter_first()
+        while iter:
+            name = self.model.get_value(iter, NAME_COL)
+            if name == next_item_name:
+                path = self.model.get_path(iter)
+                self.selection.select_path(path)
+                self.dblist.scroll_to_cell(path, None, 1, 0.5, 0)
+                break
+            iter = self.model.iter_next(iter)
+
     def __connect_signals(self):
         """
         Connects the signals to the buttons on the interface.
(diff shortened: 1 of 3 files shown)


gramps/gui/dbman.py
```diff
@@ -790,7 +833,7 @@ def __really_delete_db(self):
             ErrorDialog(_("Could not delete Family Tree"), str(msg), parent=self.top)
         # rebuild the display
         self.__populate()
-        self._select_default()
+        self._select_next_after_deletion(next_item_name)
 
     def __really_delete_version(self):
```
hgohel, line 838: Regarding `__really_delete_version()` I am not sure whether specific version deletion (RCS) is still being used in Gramps. If yes, deleting a version scenario will also remember sort order. However, I did not add the code here to choose an appropriate next item because I am not sure how versions are displayed and selected. Did not want to jam in code without understanding how it is supposed to work.

dsblank (changes requested):
## Review of PR #2154 — Manage Family Trees: preserve sort order and better selection after delete

The core logic is sound and fixes both reported bugs. A few issues to address:

### Fix 1: Preserve sort order in `_populate_model` — ✅ Good

Correct approach: capture sort column/order before rebuilding the model, restore it afterward.

Minor note: `Gtk.TreeSortable.get_sort_column_id()` can return sentinel values `GTK_TREE_SORTABLE_DEFAULT_SORT_COLUMN_ID` (-1) or `GTK_TREE_SORTABLE_UNSORTED_SORT_COLUMN_ID` (-2). These pass the `is not None` check and get passed back to `set_sort_column_id`. In practice GTK handles these constants, but worth being aware of.

---

### Fix 2: Select next/previous item after deletion

**Bug: Mixed `store` vs `self.model` references** in `__really_delete_db`:

```python
store, node = self.selection.get_selected()
path = store.get_path(node)
node = self.model.get_iter(path)           # node re-fetched from self.model

next_iter = self.model.iter_next(node)     # uses self.model ✓
...
prev_iter = store.iter_previous(node)      # uses store ✗ (inconsistent)
```

`store` and `self.model` should be the same object here so it works, but the inconsistency is fragile. Please change `store.iter_previous(node)` to `self.model.iter_previous(node)`.

**Minor: `iter` shadows Python built-in** in `_select_next_after_deletion`:
```python
iter = self.model.get_iter_first()
```
Consider renaming to `tree_iter` (same pattern already exists in `existing_name()`, but worth fixing in new code).

**Minor: docstring style inconsistency** — `_select_next_after_deletion` uses Google-style `Args:` section while the rest of the file uses plain single-paragraph docstrings.

---

### Summary

| | |
|---|---|
| Sort order fix | ✅ Correct |
| Next-item selection logic | ✅ Correct behavior |
| `store` vs `self.model` inconsistency | Please fix |
| `iter` shadowing built-in | Minor (optional) |
| Docstring style | Minor (optional) |

Please fix the `store.iter_previous` → `self.model.iter_previous` inconsistency; the other items are optional suggestions.

hgohel:
@dsblank Made the consistency fix requested, rebased onto master to obtain latest changes and re-tested locally.

dsblank (changes requested):
## Review by Claude

First, apologies — the project has a new **AGENTS.md** (agent/contributor guidelines) that spells out several coding conventions, but it hasn't been merged yet, so you had no way of knowing about it. The issues below come from those guidelines; none of them reflect poorly on the work itself, which is correct and well-structured.

---

### AGENTS.md Compliance

#### ❌ Type hints missing

All new functions and methods must have type hints using Python 3.10+ syntax. The new method currently has none:

```python
# Current
def _select_next_after_deletion(self, next_item_name):

# Required
def _select_next_after_deletion(self, next_item_name: str | None) -> None:
```

#### ❌ Docstring format — should be Sphinx style

All docstrings must use Sphinx format with `:param:`, `:type:`, `:rtype:` tags. The new method uses Google-style `Args:`:

```python
# Current (Google style — non-compliant)
"""
Select the item that would have been next after deletion.

Args:
    next_item_name: The name of the item that should be selected
"""

# Required (Sphinx style)
"""
Select the item that would have been next after deletion.

:param next_item_name: The name of the item that should be selected.
:type next_item_name: str | None
"""
```

#### ❌ No unit tests

The guidelines require tests for each distinct logic flow (happy path, error cases, edge cases). The new `_select_next_after_deletion` logic has several flows that should be covered:

- Deleted item has a next item → next is selected
- Deleted item is last in list → previous is selected
- Deleted item was the only item → nothing selected (`next_item_name` is `None`)
- `self.model is None` guard

Tests should use `unittest` and live in `gramps/gui/test/` with a `_test.py` suffix.

---

### Previously noted (still optional)

- `iter` as a variable name shadows the Python built-in — `tree_iter` would be more idiomatic (same pattern used elsewhere in the file)

---

### Summary

| Rule | Status |
|---|---|
| Black formatting | ✅ Passes |
| Commit messages | ✅ (commits will be squashed) |
| Type hints | ❌ Add `str \| None` / `-> None` |
| Sphinx docstrings | ❌ Replace `Args:` with `:param:`/`:type:` |
| Unit tests | ❌ Add tests for new logic flows |

The logic itself is correct and the required fix from the previous review was addressed — thank you for that. Once these three items are resolved the PR should be good to go.

dsblank:
@Nick-Hall please merge #2236; it will make coding (and reviewing) much easier on the developers. AGENTS.md will probably change often and we need to have it available for devs to make PRs.

hgohel:
Added unit tests to complete requested changes - ready for review and approval:

`python -m unittest gramps.gui.test.test_dbman_selection.TestDbManSelection`

hgohel:
@Nick-Hall I've re-targeted this to gramps61 since this is a bug fix.

dsblank (commented):
Two blocking items must be addressed before this PR can merge.

**1. Test file name (see inline comment)**

The file must be renamed to `dbman_selection_test.py`. CI discovers tests with:

```
python3 -m unittest discover -p "*_test.py"
```

The current name `test_dbman_selection.py` matches `test_*.py` (pytest convention) but **not** `*_test.py`, so these tests will never be executed by CI. All other test files in this directory follow the correct pattern:
- `gramps/gui/test/display_test.py`
- `gramps/gui/test/user_test.py`
- `gramps/gui/test/utils_test.py`

After the rename, the test command in the PR description will also need updating.

**2. `po/POTFILES.skip`**

`gramps/gui/test/dbman_selection_test.py` (after rename) must be added to `po/POTFILES.skip`. Per the contributor guidelines, every new `.py` file must appear in either `POTFILES.in` or `POTFILES.skip`. The two existing test files in this directory both appear there; this one is missing.

**3. Copyright (see inline comment)**

AI tools cannot hold copyright. The `GitHub Copilot` copyright line should be removed; `Himanshu Gohel` is the correct copyright holder.

---

Everything else from the previous review has been addressed correctly — type hints, Sphinx docstring, `tree_iter` rename, and unit tests. Thank you.

gramps/gui/test/dbman_selection_test.py
```diff
@@ -0,0 +1,105 @@
+#
```
dsblank, line 1: This file must be renamed to `dbman_selection_test.py`. CI discovers tests with:

```
python3 -m unittest discover -p "*_test.py"
```

The current name `test_dbman_selection.py` matches `test_*.py` (pytest convention) but **not** `*_test.py`, so these tests will never be executed by CI. All other test files in this directory follow the correct pattern:

```
gramps/gui/test/display_test.py
gramps/gui/test/user_test.py
gramps/gui/test/utils_test.py
```

gramps/gui/test/test_dbman_selection.py
```diff
@@ -0,0 +1,105 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  GitHub Copilot
```
dsblank: AI tools cannot hold copyright. Please remove this line — `Himanshu Gohel` on the next line is the correct copyright holder.

hgohel:
@dsblank @Nick-Hall 
The issue with the failing units is that because the unit tests run in headless mode on the CI system by setting `GDK_BACKEND=-`, at some point a low-level function which expects a screen fails, and logs critical messages, followed by a segfault loading DbManager. I kept trying to mock more and more things but could not resolve it.

Ending up using AI help, and after a lot of back and forth, it was clear that just mocking wasn't enough; the resolution was to change core code. It added a `has_gtk_display()` method and several places were modified such that GTK objects are created at runtime, not at import, so importing the module is safe in headless environments. I don't know whether you are open to such a change, but now the UT runs cleanly. Modified files include:

clipboard.py
dbman.py
pluginmanager.py
utils.py
dbman_selection_test.py
__init__.py
persistenttreeview.py

Let me know if you want me to push up these changes for review, or if you have other suggestions.

dsblank:
> the resolution was to change core code. It added a has_gtk_display() method and several places were modified such that GTK objects are created at runtime, not at import, so importing the module is safe in headless environments. I don't know whether you are open to such a change, but now the UT runs cleanly. 

Yes, this has been a goal to make the separation for this kind of use. Please make a PR and we can further evaluate.

Nick-Hall:
> Let me know if you want me to push up these changes for review, or if you have other suggestions.

We should really be using something like `xvfb` to test GUI code.

You shouldn't have to create a `has_gtk_display()` method.  Any code in `gramps/gui` will expect a display and code in `gramps/gen` should run without one.

hgohel:
> We should really be using something like `xvfb` to test GUI code.

@Nick-Hall That was one of the suggestions by AI. I ran `xvfb-run` on Ubuntu and it seemed to work, but xvfb isn't available for MSYS2, so I don't know how we would run those tests in that environment.

> Any code in `gramps/gui` will expect a display and code in `gramps/gen` should run without one.

Makes sense, but tests run with `GDK_BACKEND=-` which seems to oppose that idea. When I ran the entire test suite with the changes I referenced in the last post, a few tests in `gramps/gui`, and several others started failing so I won't push the changes to this PR, but I'll upload it to a different PR for a review of the direction it was headed in.

Nick-Hall:
Please rebase on the _master_ branch.

dsblank:
Closing this one, in favor of one based on headless fixes.

PR #2160, opened 2025-12-30
Optimizer: skip loop when all rules are optimized

dsblank:
Before:

We needed to loop over all optimized handles, in case there was a non-optimized rule (a rule that didn't have `selected_handles`)

After:

We now detected if there are non-optimized rules, and only loop in that case. Otherwise we can return the optimized handles directly.

### DIFF
--- a/gramps/gen/filters/_genericfilter.py
+++ b/gramps/gen/filters/_genericfilter.py
@@ -180,24 +180,37 @@ class GenericFilter:
 
         # use the Optimizer to refine the set of possible_handles
         optimizer = Optimizer(self)
-        handles_in, handles_out = optimizer.compute_potential_handles_for_filter(self)
+        handles_in, handles_out, all_optimized = (
+            optimizer.compute_potential_handles_for_filter(self)
+        )
 
-        # LOG.debug(
-        #    "Optimizer possible_handles: %s",
-        #    len(possible_handles),
-        # )
+        # Intersect optimizer results with possible_handles to ensure we only
+        # consider handles in the subset being filtered (important when id_list
+        # is provided and rules computed selected_handles from all handles)
+        if handles_in is not None:
+            possible_handles = possible_handles & handles_in
 
-        # if user:
-        #    user.begin_progress(_("Filter"), _("Applying ..."), len(possible_handles))
+        if handles_out is not None:
+            possible_handles = possible_handles - handles_out
 
-        # test each value in possible_handles to compute the final_list
+        LOG.debug(
+            "After optimization possible_handles: %s",
+            len(possible_handles),
+        )
+
+        # If all rules are optimized, we can skip the loop and return the optimized set directly
+        if all_optimized:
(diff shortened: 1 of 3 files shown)


gramps/gen/filters/_genericfilter.py
```diff
@@ -180,24 +180,37 @@ def apply_logical_op_to_all(
 
         # use the Optimizer to refine the set of possible_handles
         optimizer = Optimizer(self)
-        handles_in, handles_out = optimizer.compute_potential_handles_for_filter(self)
+        handles_in, handles_out, all_optimized = (
```
stevenyoungs, line 183: @dsblank is it a requirement of the filter optimizer that all filters shall return the _exact_ set of handles matching (handles_in) the filter or not matching the filter (handles_out)?

stevenyoungs: Just to clarify: previously if `handles_in` was a superset of the exact `handles_in` and `handles_out` was a subset of the exact `handles_out`, the loop would ensure the correct result was still computed (at the computational expense of looping over some additional handles)
Are there any scenarios where a filter may want to exploit this e.g. the filter can cheaply produce the _approximate_ `handles_in` or `handles_out` values but it is much more expensive to produce the _exact_ `handles_in` or `handles_out`.

stevenyoungs:
Looks good to me, subject to confirmation that `handles_in` and `handles_out` shall always be the exact sets for all optimized filters. See my earlier [comment](https://github.com/gramps-project/gramps/pull/2160#discussion_r2655529757)

dsblank:
> is it a requirement of the filter optimizer that all filters shall return the exact set of handles matching (handles_in) the filter or not matching the filter (handles_out)?

This is indeed a tricky question. The answer is "the the handles_in and handles_out are the exact 'in' and 'out' only if all of the rules have selected_handles".

> Are there any scenarios where a filter may want to exploit this e.g. the filter can cheaply produce the approximate handles_in or handles_out values but it is much more expensive to produce the exact handles_in or handles_out.

Good point. We must require that selected_handles not be used for any other purpose other than exactly in, or exactly out. It cannot be used for other purposes.

gramps/gen/filters/_genericfilter.py
dsblank: The final handles_in is the exact set of items (minus handles_out) that match if all_optimized is True. If all_optimized is not true, then it is a superset, and other rules may make the final set smaller. 

A rule cannot approximate either handles_in or handles_out: those are considered the true set, in case we further optimize.

dsblank:
@Nick-Hall this could be considered a bug. Let me know if so.

Nick-Hall:
This looks more like a performance enhancement to me.  It's also a large enough change to risk introducing a bug.

dsblank:
Closing in favor of #2455 — I no longer have write access to push to this branch, so I rebased and reopened from my fork to resolve conflicts with recent master changes.

PR #2161, opened 2025-12-31
Add order change UI for rules in filter editor

dsblank:
This PR adds the ability to change the order of rules in the filter editor by dragging/dropping, up/down arrow buttons, and keyboard controls.

Visual info is given by a textual hint below the rule list (requires translation), and via changing the mouse-over cursor.

image
 
Thanks to @emyoulation for bringing this issue up.

Note: This PR was written with help from Cursor (various models) and Copilot.

### DIFF
--- a/gramps/gui/glade/rule.glade
+++ b/gramps/gui/glade/rule.glade
@@ -502,6 +502,21 @@
                   </packing>
                 </child>
                 <child>
+                  <object class="GtkLabel" id="reorder_hint">
+                    <property name="visible">True</property>
+                    <property name="can-focus">False</property>
+                    <property name="halign">center</property>
+                    <property name="margin-top">4</property>
+                    <property name="use-markup">True</property>
+                    <property name="label" translatable="yes">&lt;span style=&quot;italic&quot; size=&quot;small&quot;&gt;Drag and drop rules to change order&lt;/span&gt;</property>
+                  </object>
+                  <packing>
+                    <property name="left-attach">0</property>
+                    <property name="top-attach">5</property>
+                    <property name="width">2</property>
+                  </packing>
+                </child>
+                <child>
                   <object class="GtkLabel" id="label16">
                     <property name="visible">True</property>
                     <property name="can-focus">False</property>
@@ -513,7 +528,7 @@
                   </object>
                   <packing>
                     <property name="left-attach">0</property>
-                    <property name="top-attach">5</property>
+                    <property name="top-attach">6</property>
                   </packing>
                 </child>
                 <child>
@@ -599,7 +614,7 @@
                   </object>
                   <packing>
                     <property name="left-attach">0</property>
-                    <property name="top-attach">7</property>
+                    <property name="top-attach">8</property>
                     <property name="width">2</property>
                   </packing>
                 </child>
@@ -618,7 +633,7 @@
                   </object>
                   <packing>
                     <property name="left-attach">0</property>
-                    <property name="top-attach">6</property>
+                    <property name="top-attach">7</property>
                     <property name="width">2</property>
                   </packing>
                 </child>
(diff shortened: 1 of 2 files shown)


dsblank:
Note: the ordering of rules can't be done automatically, as the system doesn't know how expensive a rule is. (Optimized rules already cut down on the selected handles).

stevenyoungs:
> Visual info is given by a textual hint below the rule list (requires translation), and via changing the mouse-over cursor.

Would image buttons act as a visual prompt, instead of the current text, and as a benefit provide an additional way to adjust the order? [Image taken from the children list of the edit family dialog]

Is there keyboard support for re-ordering lists in gramps?

stevenyoungs:
> Note: the ordering of rules can't be done automatically, as the system doesn't know how expensive a rule is. (Optimized rules already cut down on the selected handles).

If the optimizer could query rules to see if they were optimized, it could then process optimized rules first where the condition was "and". I've not spent long thinking about it, but I don't think there is any benefit if the condition is "or" or "one" as you need to run the slower rules on all handles anyway

dsblank:
> If the optimizer could query rules to see if they were optimized, it could then process optimized rules first where the condition was "and". I've not spent long thinking about it, but I don't think there is any benefit if the condition is "or" or "one" as you need to run the slower rules on all handles anyway

I tried that (sorted on rules with shorter `selected_handles` first) but it doesn't make a significant difference:

1. The handle set has already been narrowed down (given #2160 and other fixes)
2. The test for optimized rules is very fast "handle in self.selected_handles"

dsblank:
> Would image buttons act as a visual prompt, instead of the current text, and as a benefit provide an additional way to adjust the order? [Image taken from the children list of the edit family dialog]

Added the up/down arrows (on right side of list).

image

> Is there keyboard support for re-ordering lists in gramps?

There is now: Control + Arrow keys.

Thanks for the review!

gramps/gui/glade/rule.glade
```diff
@@ -493,14 +493,90 @@
...
                   </packing>
                 </child>
+                <child>
+                  <object class="GtkLabel" id="reorder_hint">
+                    <property name="visible">True</property>
+                    <property name="can-focus">False</property>
+                    <property name="halign">center</property>
+                    <property name="margin-top">4</property>
+                    <property name="use-markup">True</property>
+                    <property name="label" translatable="yes">&lt;span style=&quot;italic&quot; size=&quot;small&quot;&gt;To change rule order, use drag and drop, up/down buttons, or Ctrl+Arrow keys&lt;/span&gt;</property>
```
hgohel, line 572: Assuming that Cmd is the appropriate key, is there a way to code the tips so that "Ctrl" gets converted to "Cmd" when building for Mac?

dsblank: I have no idea. And this string is complicated for translators with all of the HTML entities. We could just leave out the "Ctrl + Arrows" (and maybe the italics?)

@Nick-Hall we could use your advice here.

dsblank: Good catch. There were actually two related issues here:

1. **The hint text** — the label in `rule.glade` had the string hardcoded as `Ctrl+Arrow keys`. Since `glade.py`'s existing `is_quartz()` substitution only replaces GDK mask constants (for accelerators), not arbitrary text, we can't fix this purely in the glade file. Instead the label text is now set programmatically in `filtereditor.py` after load, using `mac()` from `constfunc` to pick `"Cmd"` or `"Ctrl"` at runtime.

2. **The shortcut itself** — on Mac/Quartz, the Command key maps to `META_MASK` in GTK, not `CONTROL_MASK`. The `on_key_press` handler was only checking `CONTROL_MASK`, so the shortcut wouldn't have worked on Mac at all. That's also fixed: the handler now checks `META_MASK` on Mac and `CONTROL_MASK` elsewhere.

Committed in 3cdca2872e.

dsblank:
Closing in favour of #2305, which is opened from a fork branch as required. All commits and review history carried over.

PR #2164, opened 2026-01-04
Better default Tag colors

dsblank:
This PR uses better default colors for Tags. Previously the color was always black.

With this PR:

1. Create a palette of colors that work well with light or dark themes
2. Uses the assigned colors to find an unused color, if possible

image

Resolves #0011974

This PR was written by me and Cursor, with the following prompt:

> Black is not a very good default color to show that the record is tagged. Can you come up with a set of colors to use for new Tags that minimizes collisions with existing tags, and that would work with either a dark or light palette.

### DIFF
--- a/gramps/gen/lib/tag.py
+++ b/gramps/gen/lib/tag.py
@@ -35,6 +35,77 @@ _ = glocale.translation.gettext
 
 # -------------------------------------------------------------------------
 #
+# Default Tag Color Palette
+#
+# -------------------------------------------------------------------------
+# Color palette designed to work well on both light and dark GTK themes.
+# These are medium-saturation colors that provide good contrast and are
+# visually distinct from each other.
+DEFAULT_TAG_COLORS = [
+    "#3584e4",  # Blue - GTK Adwaita blue
+    "#c64600",  # Orange/Red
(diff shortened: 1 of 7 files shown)


gramps/gen/lib/tag.py
```diff
@@ -33,6 +33,76 @@
...
+    "#1a5fb4",  # Dark Blue
+    "#c2185b",  # Pink/Magenta
+    "#0d52bf",  # Deep Blue
+    "#ff7800",  # Bright Orange
+    "#33d17a",  # Light Green
+    "#9141ac",  # Purple
+]
+
+
+def get_default_tag_color(db=None):
```
Nick-Hall, line 63: This should really be a utility function in `gramps/gen/utils`.

gramps/gen/lib/tag.py
```diff
@@ -44,12 +114,14 @@ class Tag(TableObject):
     attached to a primary object.
     """
 
-    def __init__(self, source=None):
+    def __init__(self, source=None, db=None):
```
Nick-Hall, line 117: Instead of passing a database instance to the constructor like this, it would be more consistent with the rest of Gramps to use a utility function.

PR #2165, opened 2026-01-04
Save filter dates in locale-independent format

dsblank:
# Fix Filter Date Locale Compatibility

Fixes https://gramps-project.org/bugs/view.php?id=13992

## Problem

When entering a date in a filter (like "Events with &lt;data&gt;"), the date is saved in the locale language format. When that filter is used in a different locale, it doesn't parse correctly because the date parser expects dates in the current locale's format.

For example:
- A user in French locale enters "avant 15/01/2024" (before January 15, 2024)
- The filter is saved with this French text
- When opened in English locale, the parser can't understand "avant" and the date becomes text-only
- The filter fails to match dates correctly

## Solution

This PR implements a two-part solution:

1. **Normalize dates to English format when saving filters**: Dates are converted to locale-independent English format (e.g., "bef 2024-01-15") before being saved to the filter XML file.

2. **Add English parser fallback when loading**: When loading dates in the filter editor, if the locale parser fails to parse a date string, the code falls back to the English parser. This ensures that saved English-format dates can be loaded and displayed correctly in any locale.

### `gramps/gui/editors/filtereditor.py`

- Added import for `Date` class
- Modified `rule_ok()` method to detect `DateEntry` widgets and normalize dates to English format before saving
- Uses `Date.__str__()` which produces English keywords ("bef", "aft", "abt", "from", "to") plus ISO date format (YYYY-MM-DD)
- Handles edge cases: empty dates, text-only dates, and non-DateEntry widgets

### `gramps/gui/widgets/monitoredwidgets.py`

- Added import for `LANG_TO_PARSER` to access the English parser
- Modified `set_date()` method to add English parser fallback when locale parser fails
- Updates display text to locale format after successful parsing, ensuring users see dates in their language

### Saving a Filter
1. User enters date in their locale (e.g., "avant 15/01/2024" in French)
2. Date is parsed and stored in the `DateEntry.date` object
3. When saving, `Date.__str__()` converts it to English format: "bef 2024-01-15"
4. English format is saved to the filter XML file

### Loading a Filter
1. English format date string is loaded from XML (e.g., "bef 2024-01-15")
2. Date string is set to the `DateEntry` widget
3. Locale parser attempts to parse it
4. If locale parser fails (text-only), English parser is tried as fallback
5. After successful parsing, display text is updated to locale format (e.g., "avant 15/01/2024" in French)
6. User sees the date in their locale language

## Supported Date Formats

This solution handles all Gramps date formats:
- **Modifiers**: before, after, about, from, to
- **Quality indicators**: estimated, calculated
- **Calendar types**: Gregorian, Julian, Hebrew, French Republican, Persian, Islamic, Swedish
- **Date ranges and spans**: "2024-01-15 - 2024-12-31"
- **Partial dates**: "2024-01-00", "2024-00-00"
- **New year variants**: Mar1, Mar25, Sep1

All of these are preserved in the English format and can be parsed and displayed correctly in any locale.

## Backwards Compatibility

✅ **Fully backwards compatible**

- **Old filters continue to work**: Existing filters with locale-specific dates continue to work in their original locale
- **Old filters can be edited**: When editing old filters, the English parser fallback ensures they can be parsed in any locale
- **Auto-upgrade**: Old filters are automatically upgraded to English format when saved
- **No breaking changes**: Filter execution behavior is unchanged; same parsing logic and error handling

The only difference a user would see is that the string representation in the rule list uses the default format of a date:

Screenshot from 2026-01-04 11-33-53

Disclaimer: this PR was written by me and Cursor.

Initial prompt:

> When entering a Date in a filter (like "Events with &lt;data&gt;") the date is saved in the locale language. When that filter is used in a different locale, it doesn't parse. What can be done to allow filter dates to be used across locales? 

After suggesting that it change the Filter editor (rather than filters) and additional dialog, I reviewed the suggestions and made changes.

### DIFF
--- a/gramps/gui/editors/filtereditor.py
+++ b/gramps/gui/editors/filtereditor.py
@@ -71,6 +71,7 @@ from gramps.gen.lib import (
     NoteType,
     PlaceType,
 )
+from gramps.gen.lib.date import Date
 from gramps.gen.filters import rules
 from ..autocomp import StandardCustomSelector, fill_entry
 from ..selectors import SelectorFactory
@@ -839,7 +840,24 @@ class EditRule(ManagedWindow):
         try:
             page = self.notebook.get_current_page()
             class_obj, vallist, tlist, use_regex, use_case = self.page[page]
-            value_list = [str(sclass.get_text()) for sclass in tlist]
+
+            # Normalize dates to locale-independent format before saving
+            value_list = []
+            for sclass in tlist:
+                if isinstance(sclass, DateEntry):
+                    # For DateEntry widgets, get the Date object and convert to English format
+                    # This ensures dates are stored in locale-independent format
+                    date_obj = sclass.date
+                    if date_obj and date_obj.get_modifier() != Date.MOD_TEXTONLY:
+                        # Use Date.__str__() which produces English keywords + ISO format
+                        normalized_date = str(date_obj)
+                        value_list.append(normalized_date)
+                    else:
+                        # If date parsing failed or is text-only, use the text as-is
+                        value_list.append(str(sclass.get_text()))
+                else:
+                    # For other widgets, use get_text() as before
+                    value_list.append(str(sclass.get_text()))
             if class_obj.allow_regex:
                 new_rule = class_obj(
                     value_list, use_regex.get_active(), use_case.get_active()
--- a/gramps/gui/widgets/monitoredwidgets.py
+++ b/gramps/gui/widgets/monitoredwidgets.py
@@ -63,6 +63,8 @@ _ = glocale.translation.gettext
 from ..autocomp import StandardCustomSelector, fill_entry
 from gramps.gen.datehandler import displayer, parser, LANG_TO_PARSER
 from gramps.gen.lib.date import Date, NextYear
+
+_english_parser = LANG_TO_PARSER["C"]() if "C" in LANG_TO_PARSER else None
 from gramps.gen.errors import ValidationError
 
 # -------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


dsblank (commented):
**Code Review (by Claude)**

This PR fixes a real bug: filter rules with dates were saved in the user's locale format, causing them to fail when loaded in a different locale. The two-part fix (save normalized, load with fallback) is the right approach.

**`filtereditor.py` — `rule_ok()`**

Instead of blindly calling `get_text()` on all widgets, it now detects `DateEntry` widgets and serializes them via `str(date_obj)`. `Date.__str__()` is locale-independent (hardcoded English prefixes like `bef`, `aft`, `abt` and ISO `YYYY-MM-DD` format), so this correctly normalizes the stored value. The `MOD_TEXTONLY` fallback to `get_text()` is also correct — if the date couldn't be parsed, there's nothing better to save.

**`monitoredwidgets.py` — `MonitoredDate.set_date()`**

Adds an English fallback parser so that when the locale parser returns `MOD_TEXTONLY` (parse failure), it retries with the `"C"` locale parser. This handles loading previously-saved English-format dates in any locale. The display is then updated to the user's locale format.

Two issues were identified and fixed in a follow-up commit:
- The English parser was being instantiated on every keystroke; it is now cached as a module-level constant (`_english_parser`).
- Calling `set_text()` inside `set_date()` emits `content-changed` synchronously, causing a redundant recursive call. A `_setting_date` re-entrancy guard was added to short-circuit this.

**`DateEntry.date` freshness** — confirmed not an issue. `content-changed` fires on every keystroke, so `date_obj` is always current when `rule_ok()` reads it.

Overall the fix is correct and well-scoped. ✓

PR #2178, opened 2026-03-01
Add JSON field indexes on json_data handles with dialect support

dsblank:
## Summary

- Add \`dialect\` class attribute to \`DBAPI\` base class (raises if not overridden) and set to \`"sqlite"\` in the \`SQLite\` subclass
- Add \`_format_json_extract_for_index()\` to emit the correct dialect-specific JSON-extract expression for SQLite and PostgreSQL
- Move \`_create_secondary_columns()\` call to after standard indices
- During \`reindex()\`, drop and recreate per-table \`<table>_handle_json\` indices on the JSON handle field, enabling efficient JOIN lookups
- Add \`_quote_column(col)\` no-op hook to the \`DBAPI\` base class; dialect subclasses override it to quote column names that are reserved words in that database (e.g. PostgreSQL needs \`"desc"\` quoted). Used at the \`CREATE INDEX\` and \`ORDER BY\` sites for the media \`desc\` column. This eliminates the need for post-hoc string replacement in database adapters.

## Test plan

- [ ] Open an existing database and verify \`reindex()\` completes without error
- [ ] Confirm \`<table>_handle_json\` indices exist in the SQLite database after reindex
- [ ] Verify queries that JOIN on the JSON handle field use the new index (EXPLAIN QUERY PLAN)
- [ ] Test that a subclass without \`dialect\` set raises an appropriate error
- [ ] Confirm \`get_media_handles(sort_handles=True)\` works correctly (exercises \`_quote_column\` at \`ORDER BY desc\`)

### DIFF
--- a/gramps/plugins/db/dbapi/dbapi.py
+++ b/gramps/plugins/db/dbapi/dbapi.py
@@ -281,22 +281,23 @@ class DBAPI(DbGeneric):
         self.dbapi.execute("CREATE INDEX note_gramps_id ON note(gramps_id)")
         self.dbapi.execute("CREATE INDEX reference_obj_handle ON reference(obj_handle)")
 
-        # JSON handle indexes for faster JOINs on json_data
-        for table in [
-            "person",
-            "source",
-            "citation",
-            "media",
-            "place",
-            "family",
-            "event",
-            "repository",
-            "note",
-        ]:
-            self.dbapi.execute(
-                f"CREATE INDEX {table}_handle_json "
-                f"ON {table}({self._format_json_extract_for_index()})"
-            )
+        # JSON handle indexes for faster JOINs on json_data (only when using JSON storage)
+        if json_data:
+            for table in [
+                "person",
+                "source",
+                "citation",
+                "media",
+                "place",
+                "family",
+                "event",
+                "repository",
+                "note",
+            ]:
+                self.dbapi.execute(
+                    f"CREATE INDEX {table}_handle_json "
+                    f"ON {table}({self._format_json_extract_for_index()})"
+                )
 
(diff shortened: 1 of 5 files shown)


gramps/plugins/db/dbapi/dbapi.py
```diff
@@ -251,6 +255,29 @@ def _create_schema(self, json_data):
...
+            "json_extract(json_data, '$.handle')" for SQLite or
+            "JSON_EXTRACT_PATH(json_data, 'handle')" for PostgreSQL)
+        """
+        if self.dialect is None:
+            raise Exception("Please add 'dialect' property to your subclass")
+
+        if self.dialect == "sqlite":
+            return "json_extract(json_data, '$.handle')"
+        elif self.dialect == "postgres":
+            return "JSON_EXTRACT_PATH(json_data, 'handle')"
```
stevenyoungs, line 325: Is it better to push this db driver specific logic into the actual driver file (sqlite.py etc)
i.e. in sqlite.py add
```
def _do_format_json_extract_for_index(self):
    return "json_extract(json_data, '$.handle')"
````
and call `return self._do_format_json_extract_for_index()` here

You may not need `dialect` then either (check for the absence of `_do_format_json_extract_for_index` to raise an exception)

dsblank: We'll need dialect for other things that are coming :)

dsblank: I'm going to move all of the SQL-dialect formatting to a module after we get some of these PRs merged. Will also clean up the hacky postgresql formatting too.

PR #2180, opened 2026-03-03
Live filter UI updates, responsiveness, and O(n²) fix

dsblank:
## Summary

- **Live sidebar timing display**: The filter sidebar now shows `Prepare time:`, `Apply time:`, and `Elapsed time:` as they happen, in a consistent order, rather than only after the filter completes. `Gdk.flush()` ensures each update is committed to the display immediately.

- **UI responsiveness on flat views**: `step_progress()` in `gui/user.py` now pumps the GTK main loop at most 10×/second when no `ProgressMeter` is active. This prevents the OS "not responding" freeze on flat views (People, Citations, etc.) which previously had no event pumping during filter execution.

- **Pulsing status-bar progress indicator**: `flatbasemodel._rebuild_filter()` now shows and pulses `uistate.progress` during `search.apply()`, giving the same visual feedback that tree views already had via `ProgressMonitor`.

- **O(n²) → O(n) result ordering fix**: The post-filter result reordering in `_genericfilter.apply()` used `list.index()` as a sort key — O(n) per item, O(n²) overall. On a 100k-person database this caused a ~50 second freeze after every filter run. Fixed with a single O(n) pass:
  - `tupleind` path: filter `id_list` by a result set instead of sorting by index
  - `tree` path: precomputed position dict replaces repeated `index()` calls

## Files changed

| File | Change |
|---|---|
| `gramps/gen/user.py` | Add `User.print()` and `_print_func` hook |
| `gramps/cli/user.py` | Wire `_print_func` for CLI user |
| `gramps/gen/filters/_genericfilter.py` | Timing messages + O(n²) fix |
| `gramps/gui/user.py` | GTK event pumping + pulse progress in `step_progress()` |
| `gramps/gui/filters/sidebar/_sidebarfilter.py` | Live label updates with `render`/`live_print`/`live_step` |
| `gramps/gui/views/treemodels/flatbasemodel.py` | Pulsing progress bar in `_rebuild_filter()` |

## Test plan

- [ ] Open People view, apply a sidebar filter rule on a large database
- [ ] Verify `msg_label` shows `Searching...\nPrepare time: X.XXXs` while apply phase runs
- [ ] Verify `Apply time:` appears in label immediately after apply completes (not only at the end)
- [ ] Verify final label shows `Prepare time` / `Apply time` / `Elapsed time` in order
- [ ] Verify status-bar progress bar pulses during filter execution and disappears on completion
- [ ] Verify no "not responding" OS dialog appears during filtering
- [ ] Verify filter on a tree view (e.g. Places) shows no regression
- [ ] Verify the ~50 second post-filter freeze is gone on large databases

### DIFF
--- a/gramps/cli/user.py
+++ b/gramps/cli/user.py
@@ -70,6 +70,7 @@ class User(user.UserBase):
         self.steps = 0
         self.current_step = 0
         self._input = input
+        self._print_func = lambda msg: self._fileout.write(msg + "\n")
 
         def yes(*args, **kwargs):
             return True
--- a/gramps/gen/filters/_genericfilter.py
+++ b/gramps/gen/filters/_genericfilter.py
@@ -277,10 +277,13 @@ class GenericFilter:
                 if id_list not given, all items in the database that
                 match the filter are returned as a list of handles
         """
-        start_time = time.time()
+        t0 = time.perf_counter()
         for rule in self.flist:
             rule.requestprepare(db, user)
-        LOG.debug("Prepare time: %s seconds", time.time() - start_time)
+        t1 = time.perf_counter()
+        LOG.debug("Prepare time: %s seconds", t1 - t0)
+        if user:
+            user.notify("Prepare time: %.3fs" % (t1 - t0))
 
         if self.logical_op == "and":
             apply_logical_op = self.and_test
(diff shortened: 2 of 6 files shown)


Nick-Hall:
Please rebase on the _master_ branch.

dsblank:
@Nick-Hall this has the critical bug fix that reduces filters from O(n**2) to O(n). That will immediately make filters work up to 10x faster.

It also updates the output so you never see "Wait or Force Quit" dialog and shows the timing in the different processes of filtering. I can remove that part but we should absolutely include the O(n**2) issue.

dsblank:
This is the second issue that occurs with filters that take more than about 5 seconds:

image

Nick-Hall (approved):

PR #2192, opened 2026-03-20
Add PyInstaller-based Windows AIO build

dsblank:
## Summary

Adds `aio-pyinstaller/` as a **parallel** Windows AIO build system using PyInstaller, leaving the existing `aio/` (cx_Freeze) system completely intact.

### New files in `aio-pyinstaller/`

| File | Purpose |
|------|---------|
| `gramps.spec` | PyInstaller spec: 3 Analysis objects → 3 EXEs → single flat COLLECT |
| `grampsaiow.py` | GUI entry point (`grampsw.exe`) |
| `grampsaioc.py` | Console entry point (`gramps.exe`) |
| `grampsaiocd.py` | Debug console entry point (`grampsd.exe`) |
| `grampsaio64.nsi.template` | NSIS installer adapted for PyInstaller flat layout |
| `build.sh` | Build script using `pyinstaller` + `pyinstaller-hooks-contrib` |

### New workflow

`.github/workflows/windows-aio-pyinstaller.yml` — manual-trigger only (`workflow_dispatch`), mirrors `windows-aio.yml`.

### Key technical differences from cx_Freeze

- **Flat layout**: `contents_directory="."` on each `EXE()` object puts all DLLs/extensions alongside the executables, matching the cx_Freeze 6.0.6 layout
- **orjson**: Built from source (no pre-built wheel for Python 3.14); manually copied into the bundle as a fallback since `collect_all()` skips single-file extension modules
- **SSL fix**: `ssl.create_default_context` is patched at startup to use the bundled CA cert, rather than relying on `SSL_CERT_FILE` which is silently ignored in Python 3.13+
- **GI typelibs**: Handled automatically by the `pyinstaller-hooks-contrib` gi community hook
- **GExiv2**: Not available in MSYS2 for the mingw64 environment (only ucrt64/clang64); a non-fatal warning dialog appears on launch and can be suppressed

### Notable NSIS changes vs cx_Freeze template

- Hunspell dictionary, locale, and voikko `File` directives use `/nonfatal` since not all optional files may be present
- `fc-cache` post-install step removed (hangs indefinitely under Wine; GTK handles font scanning at first launch)
- en-US dictionary pre-selected by default alongside mandatory en-GB

## Test plan

- [x] Trigger the `Windows AIO (PyInstaller)` workflow and verify installer is produced
- [x] Verify Gramps launches successfully from the installer
- [x] Verify flat install layout matches 6.0.6 (no `_internal/` subfolder)
- [ ] Verify on real Windows (tested under Wine 11.0 on Linux)
- [ ] Verify the existing `Windows AIO` (cx_Freeze) workflow is unaffected


.github/workflows/windows-aio-pyinstaller.yml
```diff
@@ -0,0 +1,56 @@
...
+    - uses: actions/checkout@v6
+    - uses: msys2/setup-msys2@v2
+      with:
+        msystem: MINGW64
+        update: true
+    - name: Build
+      run: |
+        cd aio-pyinstaller
+        ./build.sh ${{ inputs.cleanup }} ${{ inputs.build-number }}
+    - uses: actions/upload-artifact@v4
```
Nick-Hall: The version for this action should probably be updated to v6.

dsblank: Do you know how to use github's suggestion button? 

image

hgohel: Cool! Had not noticed the suggestion button in the comment toolbar. Thanks for the tip.
```suggestion
    - uses: actions/upload-artifact@v6
```

hgohel (approved):
Since this is a parallel build process, LGTM.

aio-pyinstaller/build.sh
```diff
@@ -0,0 +1,228 @@
...
+else
+    echo "WARNING: could not copy orjson to bundle (ORJSON_DIR='$ORJSON_DIR', BUNDLE_INTERNAL dir exists: $([ -d "$BUNDLE_INTERNAL" ] && echo yes || echo no))"
+fi
+
+# stage NSIS support files alongside the built executables
+mkdir -p dist/grampsaio/src
+cat grampsaio64.nsi.template | sed "s/yourVersion/$appversion/;s/yourBuild/$appbuild/" >dist/grampsaio/src/grampsaio64.nsi
+cp ../aio/gramps.ico  dist/grampsaio/src/
+cp ../aio/grampsc.ico dist/grampsaio/src/
+cp ../aio/grampsd.ico dist/grampsaio/src/
```
stevenyoungs, line 210: Rather than reach into the `../aio` directory, maybe it is better to restructure these directories. Something like

```
aio
aio\cx_freeze
aio\pyinstaller
```
common files could then go in aio

A subsequent PR could rename aio to windows for consistency with `mac` (though `debian` is inconsistent)

aio-pyinstaller/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+4. Build AIO :
+
+```
+cd gramps/aio-pyinstaller
+./build.sh
+```
+To capture the full output of the build, use `./build.sh >& build_log.txt`
+
+The resulting AIO installer is in `gramps/aio-pyinstaller/mingw64/src/GrampsAIO-[appversion]-[appbuild]-[hash]_win64.exe`
```
stevenyoungs, line 30:
```suggestion
The resulting AIO installer is in `gramps/aio-pyinstaller/dist/grampsaio/src/GrampsAIO-[appversion]-[appbuild]-[hash]_win64.exe`
```

dsblank:
@stevenyoungs you should commit any of your suggestions that you want, and feel free to make any needed changes. I know nothing except what Claude tells be re: Windows.

PR #2211, opened 2026-03-29
Add GUI feedback to Addon Manager

emyoulation:
### Interactive Lozenge Buttons 
[0013099](https://gramps-project.org/bugs/view.php?id=13099): [6.1 Addon Manager] Make filter indicator lozenges provide filter toggle functionality

Enhancement Specification • Addon Manager • gramps/gui/plug/_windows.py
Brian McCullough (emyoulation) • March 2026 
code generated by Claude (Anthropic), claude.ai, March 2026

#### Overview
The existing (Gramps 5.2 - 6.0.7)  lozenge widgets on each addon row are passive labels. This enhancement converts the four filterable lozenges into two-state interactive buttons that both reflect and drive the filter combo boxes.

image

image

#### Lozenge → Filter Mapping
Four of the six lozenges correspond to filter combos. Two do not and remain passive labels.

Lozenge 				label | addon 				field | Filter 				combo | "All" 				sentinel value
-- | -- | -- | --
Project | addon["_p"] | project_combo | _("All 				projects")  (active index 0)
Type | PTYPE_STR[addon["t"]] | type_combo | model 				value −1 at active index 0
Audience | AUDIENCETEXT[addon["a"]] | audience_combo | model 				value −1 at active index 0
Status | STATUSTEXT[addon["s"]] | status_combo | model 				value −1 at active index 0
Version | addon["v"] | — (no 				filter) | always 				blue, not clickable
Installed 				version | addon["_v"] | — (no 				filter) | always 				blue, not clickable

#### Edge Cases &amp; Constraints
* If no combo entry matches the lozenge text (e.g. the project list has changed since the addon list was loaded), treat the lozenge as non-interactive: disable the button and keep it blue.
* The addon_combo (All addons / Not installed / Installed / Updates) has NO corresponding lozenge — it is not affected by this change.
* The search entry has no lozenge — unchanged.
* Multiple lozenges on the same row may be orange simultaneously if multiple filters are active.
* Clicking an orange lozenge resets only its own filter; other active filters are unaffected.
* The clear-all button (icon) already resets all combos; lozenge colors update automatically on the next filter pass.

Document prepared with assistance from Claude (Anthropic), claude.ai, March 2026.

### DIFF
--- a/data/gramps.css
+++ b/data/gramps.css
@@ -1,10 +1,31 @@
-.lozenge {
+/*
+Gramps genealogy software 
+https://github.com/gramps-project/gramps/edit/maintenance/gramps60/data/gramps.css
+   Addon Manager; Addons tab  
+*/
+.lozenge { 
   font-size: small;
   color: #ffffff;
   background-color: #0d6efd;
   padding: 3px;
   border-radius: 6px;
-  box-shadow: 0px 0px 6px black;
+  box-shadow: 1px 3px 2px rgba(10, 10, 50, 0.33); /* Default dark blue shadow */
+  transition: background-color 0.3s ease, box-shadow 0.3s ease;
+}
+
+.lozenge-active { /* 0013099: Make filter indicator lozenges provide filter toggle functionality  */
+  font-size: small;
+  color: #ffffff;
+  background-color: #E67E22;
+  padding: 3px;
+  border-radius: 6px;
+  box-shadow: 1px 3px 2px rgba(10, 10, 50, 0.33); /* Default dark blue shadow */
+  transition: background-color 0.3s ease, box-shadow 0.3s ease;
+}
+
+.lozenge-button {
+  padding: 2px;
+  min-height: 0;
 }
 
 .addon-row {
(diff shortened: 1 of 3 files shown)


dsblank:
@emyoulation please copy the file in https://github.com/gramps-project/gramps/pull/2199 to the root of gramps repo folder. This will provide instructions for Claude.

@Nick-Hall feel free to merge #2199 sooner than later... it will save us all time as it will make agents do the right thing.

dsblank:
@emyoulation no problem! I'm going to put this into "draft" mode until it is ready.

emyoulation:
I think there is a flaw in the in the validation for a local URI that is typed instead of browsed.  

The final version Gramps MS Windows users will see is the format: `file:///C:/User/`... but what they will paste/transcribe from a file explorer will have the format: `C:\User\`...
And it does not gracefully validate both and transform them to the expected format. 

The multiple OS format (plus Gramps) format validation is more than I feel confident building. 

(Which is another reason why that code is a separate `gramps/gui/plug/_projectdialog.py` module. Done right, the validation code would be useful for any Gramps feature with a filepath combobox. But I fear this code is not that mature.)

PR #2215, opened 2026-04-02
Add support for multiple named dashboards

dsblank:
Users can now create, rename, and delete named dashboards from a dropdown selector bar above the gramplet pane. Each dashboard has its own gramplet layout stored in a separate .ini file (keyed by SHA-256 digest to stay filesystem-safe). On first run the legacy single-dashboard config is copied forward so existing layouts are preserved.

### DIFF
--- a/gramps/gui/widgets/statusbar.py
+++ b/gramps/gui/widgets/statusbar.py
@@ -115,3 +115,28 @@ class Statusbar(Gtk.Box):
     def clear_filter(self):
         """Clear the filter status text."""
         self.__filter.set_text("")
+
+    def set_view_widget(self, widget):
+        """
+        Replace the status message area with a view-specific widget.
+        The normal status text is hidden until clear_view_widget() is called.
+        """
+        self.__view_widget = widget
+        # Force the view widget to the same height as the status text widget
+        # it replaces. __status is always realized so its preferred height is
+        # reliable; _selector_bar may not be realized on the first visit.
+        _, status_natural = self.__status.get_preferred_height()
+        self.__view_widget.set_size_request(-1, status_natural)
+        self.__status.hide()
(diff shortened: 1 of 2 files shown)


emyoulation:
Please add your 2026 "copyright line" to both changed files.

I tried swapping the "Files Changed" into an instance of "the build that we are using for the FamilySearch Integration testing" (GrampsAIO64-6.0.0-aio-pr2308-c35097) to test this feature. 
That caused an error in the FS statusbar hijack:

```
WARNING .: Failed to connect FamilySearch status button
  File "C:\Program Files\GrampsAIO64-6.1.0\gramps\gui\viewmanager.py", line 347, in __build_main_window
    fs_btn = self.statusbar.get_fs_status_button()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Statusbar' object has no attribute 'get_fs_status_button'
```


Have to admit that, so far, I do not care for the idea of either the FS or "named Dashboards" using the Status bar. 

(For FS, I feel that the FS icon is like "branding" Gramps as a FamilySearch product.  But that objection is irrelevent to comments for this 'named dashboards' PR.)

For this Dashboards feature, there are existing View mode navigation features in the Navigator sidebar and the Toolbar. Adding another location to discover view Mode now requires scanning 3/4 of a circle around the perimeter of the Gramps application window. 

The other Toolbar customizations add Actions (e.g., Previous Next Home Print Export) into a section between the "'Manage Databases' ▾" section and the Configure icon. You could co-opt that for Dashboards actions: Add, Rename, Remove. (It could use the same icons as the People view modes add for similar Add, Edit, Remove person actions.)
image

emyoulation:
Thought a screen capture might encourage more experimenters:
Dashboards

PR #2226, opened 2026-04-08
Fix xgettext warning message

Nick-Hall:
The warning "'msgid' format string with unnamed arguments cannot be properly localized" was reported when extracting strings.


Nick-Hall:
@stevenyoungs Does this look OK to you?  I'll regenerate the `gramps.pot` file as part of the release.

PR #2229, opened 2026-04-09
Reduce number of db queries on open/close

DavidMStraub:
## Summary

- Opening a database issued one SELECT per metadata key (~28 queries); closing issued one BEGIN/INSERT-or-UPDATE/COMMIT per key (~38 transactions, ~114 round-trips). Both costs are paid on every request in web apps that reopen the database per request, and the transaction overhead is especially painful over network databases.
- On open: DBAPI now fetches all metadata rows in a single SELECT into a short-lived cache that serves the per-key `_get_metadata()` calls with no further queries. The cache is discarded before `load()` returns.
- On close: DBAPI overrides `_set_all_metadata()` to wrap `super()._set_all_metadata(use_txn=False)` in a single `BEGIN`/`COMMIT`, reducing ~38 transactions to one. A `use_txn` parameter is added to `_set_all_metadata()` so non-DBAPI backends retain their existing per-key transaction behaviour unchanged. Gender stats saving is extracted from `_set_all_metadata()` into its callers so it manages its own transaction independently.

## Context

This might be a small optimization for Gramps running on SQLite, but for Gramps Web it's an irreducible overhead that dominates the request duration for small/simple requests. It becomes particularly bad when using PostgreSQL, since a network lag is paid for each and every request.

### DIFF
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -744,6 +744,10 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
         if need_to_set_metadata:
             self._set_metadata("version", str(self.VERSION[0]))
 
+        # Load all metadata in one batch query where possible, then
+        # populate individual attributes via the (possibly cached) getter.
+        self._prime_metadata_cache()
+
         # Load metadata
         self.name_formats = self._get_metadata("name_formats")
         self.owner = self._get_metadata("researcher", default=Researcher())
@@ -805,12 +809,16 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
         self.rmap_index = self._get_metadata("rmap_index", 0)
         self.nmap_index = self._get_metadata("nmap_index", 0)
 
+        # Discard the batch cache; subsequent _get_metadata() calls
+        # (e.g. for default-person-handle) must always hit the database.
+        self._clear_metadata_cache()
+
         if need_to_set_metadata:
             # for new db we always need blob metadata to allow prior Gramps versions
             # to make it to the downgrade version check
             # Note: downgrade check only works from v5.1.2 and later on sqlite
             self.set_serializer("blob")
-            self.has_changed = 1  # to make sure genderstats gets saved
+            self.save_gender_stats(self.genderStats)
             self._set_all_metadata()
             if self.use_json_data():
                 self.set_serializer("json")
@@ -858,6 +866,8 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
                 # the database for gramps.cli.clidbman:
                 filename = os.path.join(self._directory, "meta_data.db")
                 Path(filename).touch()
+                if self.has_changed:
+                    self.save_gender_stats(self.genderStats)
                 self._set_all_metadata()
 
             self._close()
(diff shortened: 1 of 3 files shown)


dsblank:
This looks generally good. The one question I have is if it works in the case of upgrading from earlier versions of Gramps (without the JSON field) to Gramps 6. Need to check if we have a test for that.

DavidMStraub:
Good point. I think the answer is: yes it works, but there is no test for it. `gramps_upgrade_21` has its own transaction and metadata handling and it runs after `load` (which is what is modified in this PR) is called, so the cache is gone again.

dsblank (changes requested):
## Review (by Claude)

### v21 Upgrade Path

**Short answer: The PR does not break the v21 upgrade path. But there is one latent bug worth flagging.**

---

### The upgrade flow with this PR

For any database upgrading from v20 → v21:

1. `use_json_data()` returns `False` (no `json_data` column yet)
2. `set_serializer("blob")` → `BlobSerializer.metadata_field = "value"`
3. `_prime_metadata_cache()` executes: `SELECT setting, value FROM metadata` ← correct column name, works fine
4. Metadata loaded from cache into attributes
5. `_clear_metadata_cache()` → cache is `None`
6. `dbversion = 20 < 21` → `_gramps_upgrade(20, ...)` called

Inside `gramps_upgrade_21()` (cache is already `None`):
- Line 98: `_get_metadata(key, "not-found")` with `serializer="blob"` → hits DB directly ✓
- Line 80: `upgrade_table_for_json_data("metadata")` adds `json_data` column ✓
- Reads blob values, writes json values — all via live DB, unaffected by PR ✓

The cache is fully out of the picture before the upgrade runs. **No regression here.**

---

### Latent bug introduced by the PR

In `_get_metadata()` with an active cache:

```python
if self._metadata_cache is not None:
    raw = self._metadata_cache.get(key)
    if raw is not None:          # ← treats None as "missing"
        return self.serializer.metadata_to_object(raw)
    if default == "_":
        return []
    return default
```

Without the cache, if a key exists in the DB with a `NULL` value, `fetchone()` returns `(None,)` — a truthy tuple — so `metadata_to_object(None)` would be called. With the cache, `dict.get(key)` returns `None` for a NULL-valued key, and the code falls through to return the default instead.

**In practice this cannot happen today** — Gramps always stores real values, never NULL, in the metadata table. But the sentinel check should use `key not in self._metadata_cache` rather than `raw is not None` to be semantically correct:

```python
if self._metadata_cache is not None:
    if key not in self._metadata_cache:
        return [] if default == "_" else default
    raw = self._metadata_cache[key]
    return self.serializer.metadata_to_object(raw)
```

This would also correctly handle a hypothetical NULL metadata value (passing `None` to the deserializer, same as the non-cached path).

---

### Summary

| Concern | Result |
|---|---|
| `_prime_metadata_cache()` column name on pre-v21 DB | Safe — `BlobSerializer.metadata_field = "value"`, matches actual schema |
| Cache active during `gramps_upgrade_21` | No — cache is cleared before version check and upgrade |
| `_set_all_metadata()` called from within upgrade transaction | Never happens — upgrade functions use `_set_metadata()` directly |
| Gender stats moved out of `_set_all_metadata()` | Equivalent behavior, just a different call site |
| NULL metadata values treated differently when cached | Latent bug, not triggered in practice today |

DavidMStraub:
Good catch! This should be fixed now: I now use `key not in self._metadata_cache` instead of `raw is not None`.

dsblank (approved):

DavidMStraub:
Merge conflict resolved.

PR #2236, opened 2026-04-16
Fix AGENTS.md import header style to match codebase conventions

dsblank:
## Summary

- Corrects the import section header example from the short 24-character \`# ---\` form to the 73-character \`# -------------------------------------------------------------------------\` form with blank \`#\` lines around the section name, matching the style used throughout the codebase
- Adds a GTK/Gnome modules section to the example (a common third import group in GUI files)
- Clarifies that not all sections are required and lists the most common section names
- Notes that inline \`# pylint: disable=...\` suppressions are acceptable when the author has a deliberate reason
- Clarifies docstring verbosity: simple functions need only a concise one-liner; full Sphinx \`:param:\`/\`:type:\`/\`:returns:\`/\`:rtype:\` fields are reserved for non-obvious cases not already expressed by type hints
- Adds a Pull Requests section requiring all PRs to be submitted from a personal fork, not from a branch pushed directly to the upstream remote

## Test plan

- [ ] Verify import header style against existing files (e.g. \`gramps/gen/db/base.py\`, \`gramps/gui/views/listview.py\`)
- [ ] Verify docstring style against existing files (e.g. \`gramps/gen/db/base.py\`)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -77,7 +77,6 @@ Code should be PEP 8 compatible, except where that conflicts with Black formatti
 
 Run pylint on new code. Ideally new code should score 9 or higher and changes should not reduce the overall pylint score. This is not strictly enforced and should never come at the expense of code clarity, Black formatting, or any other rule in this guide.
 
-Inline suppression comments such as `# pylint: disable=import-outside-toplevel` are acceptable when the author has a deliberate reason (e.g. a debug-only import, a circular-import workaround).
 
 ### Class Headers
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,25 +40,34 @@ def get_person_from_gramps_id(self, gramps_id: PersonGrampsID) -> Person | None:
 
 ### Import Grouping
 
-Imports must be organized into three sections, each separated by a blank line and preceded by a comment header:
+Imports must be organized into sections, each preceded by a comment header of this form:
 
 ```python
-# ------------------------
-# Python modules
-# ------------------------
+# -------------------------------------------------------------------------
+#
+# Standard Python modules
+#
+# -------------------------------------------------------------------------
 import os
 import logging
 
-# ------------------------
-# Gramps modules
-# ------------------------
-from gramps.gen.db.base import DbReadBase
+# -------------------------------------------------------------------------
+#
+# GTK/Gnome modules
+#
+# -------------------------------------------------------------------------
+from gi.repository import Gtk
 
-# ------------------------
-# Gramps specific
-# ------------------------
+# -------------------------------------------------------------------------
+#
+# Gramps modules
+#
+# -------------------------------------------------------------------------
+from gramps.gen.db.base import DbReadBase
 from .mymodule import MyClass
 ```
+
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -40,26 +40,35 @@ def get_person_from_gramps_id(self, gramps_id: PersonGrampsID) -> Person | None:
 
 ### Import Grouping
 
-Imports must be organized into three sections, each separated by a blank line and preceded by a comment header:
+Imports must be organized into sections, each preceded by a comment header of this form:
```
stevenyoungs, line 43: Do we have any preference on the order of imports. I tend to list them in alphabetical order but that might be just me

kulath: Yes, but I don’t think it is necessary to specify anything about that, it over specifically things, and people may have reasonable reasons to do something different (e.g. to group imports in some particular way - perhaps to group basic object imports, and list them in the order in the user interface categories).

Nick-Hall: I don't have any strong preference here.  I tend to group related items together though.   For example, when importing a list of objects I will often group `Repository`, `Source` and `Citation`.

AGENTS.md
```diff
@@ -68,6 +77,8 @@ Code should be PEP 8 compatible, except where that conflicts with Black formatti
 
 Run pylint on new code. Ideally new code should score 9 or higher and changes should not reduce the overall pylint score. This is not strictly enforced and should never come at the expense of code clarity, Black formatting, or any other rule in this guide.
 
+Inline suppression comments such as `# pylint: disable=import-outside-toplevel` are acceptable when the author has a deliberate reason (e.g. a debug-only import, a circular-import workaround).
```
Nick-Hall: The current policy is not to allow disabling of pylint warnings.  The reason is that it is still useful to see them sometimes.

dsblank: I can't tell what the action item is here. Please make a "suggestion" if you would like this changed.

Nick-Hall:
```suggestion
```

dsblank:
@Nick-Hall , @stevenyoungs : added points about:

1. Use concise docstrings
2. Use a PR on your own fork for PRs

Please merge to guide all current and future contributions. Authors should rebase their master.

AGENTS.md
kulath: @dsblank @Nick-Hall I don't understand why it would be helpful to see pylint warnings. If there is a good reason why a warning should be disabled, then the presence of the '# pylint: disable' allows you to see that there was a pylint violation. If they are not disabled, then you get two warnings, one for pylint and one for AI code review. If you disable, then you get one warning, the disable comment.

It may be helpful to leave a comment in the code as to why the pylint check was disabled, but this should not be part of the Agents advice, and should probably only be done occasionally, otherwise a simple pylint comment becomes a long essay, for no particularly good reason. So a reason comment should only be a discretionary thing.

dsblank:
Closing this PR because it was submitted directly from the upstream repository branch rather than from a personal fork. All pull requests should come from a branch in your own fork of the repository — this keeps the upstream branch list clean and follows the project contribution guidelines. A replacement PR will be opened from the fork shortly.

PR #2237, opened 2026-04-17
Fix Resource Leaks

stevenyoungs:
There's a long standing problem in gramps AIO that manifests as gramps crashing

I discussed this in 2023 but made little progress on fixing it. See https://gramps.discourse.group/t/possible-gdi-object-leak-aio-5-2-0-r1-54cc357/4182. That conversation suggested that there were several existing bug reports of similar sounding problems

The attached set of fixes may fix some or all of these problems. 

This PR entirely authored by Github Copilot. I'm not an expert on GTK object lifetimes so would appreciate a detailed review by someone with more knowledge

### DIFF
--- a/gramps/gui/editors/displaytabs/backreflist.py
+++ b/gramps/gui/editors/displaytabs/backreflist.py
@@ -156,12 +156,3 @@ class BackRefList(EmbeddedList):
     def edit_button_clicked(self, obj):
         reftype, ref = self.find_node()
         edit_object(self.dbstate, self.uistate, reftype, ref)
-
-    def clean_up(self):
-        """
-        Clean up GTK objects
-
-        """
-
-        # Call parent class cleanup
-        EmbeddedList.clean_up(self)
--- a/gramps/gui/editors/displaytabs/backreflist.py
+++ b/gramps/gui/editors/displaytabs/backreflist.py
@@ -156,3 +156,12 @@ class BackRefList(EmbeddedList):
     def edit_button_clicked(self, obj):
         reftype, ref = self.find_node()
         edit_object(self.dbstate, self.uistate, reftype, ref)
+
+    def clean_up(self):
+        """
+        Clean up GTK objects
+
+        """
+
+        # Call parent class cleanup
+        EmbeddedList.clean_up(self)
--- a/gramps/gui/editors/displaytabs/buttontab.py
+++ b/gramps/gui/editors/displaytabs/buttontab.py
@@ -338,10 +338,3 @@ class ButtonTab(GrampsTab):
             # if self.up_btn :
             #    self.up_btn.set_sensitive(False)
             #    self.down_btn.set_sensitive(False)
-
-    def clean_up(self):
-        """
-        Clean up GTK objects. Called via ManagedWindow.clean_up() chain.
-        """
-        # Call parent class cleanup
-        GrampsTab.clean_up(self)
--- a/gramps/gui/editors/displaytabs/buttontab.py
+++ b/gramps/gui/editors/displaytabs/buttontab.py
@@ -338,3 +338,10 @@ class ButtonTab(GrampsTab):
             # if self.up_btn :
             #    self.up_btn.set_sensitive(False)
             #    self.down_btn.set_sensitive(False)
+
+    def clean_up(self):
+        """
+        Clean up GTK objects. Called via ManagedWindow.clean_up() chain.
+        """
+        # Call parent class cleanup
+        GrampsTab.clean_up(self)
--- a/gramps/gui/editors/displaytabs/embeddedlist.py
+++ b/gramps/gui/editors/displaytabs/embeddedlist.py
@@ -93,6 +93,7 @@ class EmbeddedList(ButtonTab):
         Create a new list, using the passed build_model to populate the list.
         """
         self.config_key = config_key
+        self.column_header_images = []
 
         ButtonTab.__init__(
             self,
@@ -121,6 +122,7 @@ class EmbeddedList(ButtonTab):
         self.col_icons = {}
         self.columns = []
         self.build_columns()
+        self.track_ref_for_deletion("column_header_images")
 
         if self._DND_TYPE:
             self._set_dnd()
@@ -489,6 +491,8 @@ class EmbeddedList(ButtonTab):
 
         list(map(self.tree.remove_column, self.columns))
         self.columns = []
+        # Clear previously created column header images
+        self.column_header_images = []
         self.setup_editable_col()
 
         # loop through the values returned by column_order
@@ -547,6 +551,8 @@ class EmbeddedList(ButtonTab):
                 image.show()
                 column.set_widget(image)
                 column.set_resizable(False)
+                # Track image objects for proper cleanup to prevent GDI leaks
+                self.column_header_images.append(image)
             else:
                 # insert the colum into the tree
                 column.set_resizable(True)
@@ -649,7 +655,9 @@ class EmbeddedList(ButtonTab):
 
         This method is called automatically when the tab/window closes via
         the ManagedWindow.clean_up() mechanism. It ensures that all GTK
-        objects are properly destroyed.
+        objects are properly destroyed:
+        - Clears and destroys the model to release rendering cache
+        - Clears column header image references
         """
         # Clear and destroy the model to release GDI resources
         if hasattr(self, "model") and self.model and hasattr(self.model, "destroy"):
@@ -657,5 +665,8 @@ class EmbeddedList(ButtonTab):
             self.model.destroy()
             self.model = None
 
-        # Call parent class cleanup
+        # Clear column header images to allow garbage collection
+        self.column_header_images = []
+
+        # Call parent class cleanup which will handle tracked references
         ButtonTab.clean_up(self)
--- a/gramps/gui/editors/displaytabs/embeddedlist.py
+++ b/gramps/gui/editors/displaytabs/embeddedlist.py
@@ -604,7 +604,7 @@ class EmbeddedList(ButtonTab):
         selectedpath = None
         if node:
             selectedpath = model.get_path(node)
-        if self.model and hasattr(self.model, "destroy"):
+        if self.model is not None and hasattr(self.model, "destroy"):
             self.tree.set_model(None)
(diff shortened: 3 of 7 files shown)


dsblank (changes requested):
## Review

The core approach is correct and the GDI resource leak fixes are real. The `gallerytab.py` bug fix (where the original `super(ButtonTab, self).clean_up()` skipped `ButtonTab.clean_up()`) is particularly important now that `ButtonTab` has its own `clean_up`. A few issues to address:

### Required changes

**1. Truthiness check on GTK model (`embeddedlist.py:663`)**

An empty `Gtk.ListStore`/`TreeStore` may evaluate to `False` in boolean context (row count = 0), so this guard can silently skip cleanup on an empty model:

```python
# current
if hasattr(self, "model") and self.model and hasattr(self.model, "destroy"):

# should be
if hasattr(self, "model") and self.model is not None and hasattr(self.model, "destroy"):
```

The same pre-existing pattern appears in `rebuild()` at line 607 — worth fixing there too for consistency.

**2. `GalleryTab` doesn't call `_cleanup_callbacks()` (`gallerytab.py`)**

`GalleryTab` inherits from `DbGUIElement` and registers media handles via `callman`, but its `clean_up` never disconnects those DB signal handlers. `EventEmbedList` correctly calls `self._cleanup_callbacks()` — `GalleryTab` should do the same before calling `ButtonTab.clean_up(self)`.

### Minor / optional

**3. Pass-through `clean_up` methods**

`BackRefList.clean_up` and `GroupEmbeddedList.clean_up` do nothing except delegate to their parent — identical to the inherited behavior without the override. Unless there's a plan to add logic later, they're unnecessary noise. `ButtonTab.clean_up` is similarly a pure pass-through, though at least it makes the chain explicit.

**4. Old-style super calls**

`EmbeddedList.clean_up(self)`, `ButtonTab.clean_up(self)`, etc. work here because `DbGUIElement` has no `clean_up`, but `super().clean_up()` is idiomatic Python 3 and less fragile if the hierarchy changes.

**5. Docstring length**

The docstrings are considerably longer than the rest of this file's style. Brief one-liners would be more consistent.

stevenyoungs:
@dsblank thanks for the review. I've addressed all five points.
Good to know that one AI agrees with another on the core problem.

dsblank (commented):
## Review of Latest Changes

The latest commit addresses all five points from the previous review. Here's the status:

### Previously Required Changes

**1. Truthiness check on `self.model` — `embeddedlist.py`**
- `clean_up()` (new): ✅ Uses `self.model is not None` — correct.
- `rebuild()` (line 604): ✅ Also fixed to `self.model is not None` — good catch on the pre-existing bug too.

**2. `GalleryTab` missing `_cleanup_callbacks()` — `gallerytab.py`**
- ✅ Fixed. `clean_up()` now correctly calls `self._cleanup_callbacks()` before `super().clean_up()`.
- The original `super(ButtonTab, self).clean_up()` bug (which skipped `ButtonTab.clean_up`) is also fixed — now just `super().clean_up()`.

### Previously Minor/Optional Changes

**3. Pass-through `clean_up` methods in `BackRefList` / `GroupEmbeddedList`**
- ✅ Removed entirely — those pass-throughs are gone.
- `ButtonTab.clean_up` is still a pure pass-through to `GrampsTab.clean_up`. Still noise, but low risk.

**4. Old-style super calls**
- `embeddedlist.py`, `eventembedlist.py`, `gallerytab.py`: ✅ All now use `super().clean_up()`.
- `surnametab.py` (line 462): ❌ Still uses `EmbeddedList.clean_up(self)` — old-style. Should be `super().clean_up()` for consistency.

**5. Docstring length / style**
- `buttontab.py`: ✅ Brief one-liner — good.
- `eventembedlist.py` and `surnametab.py`: ❌ Still have overly long multi-paragraph docstrings inconsistent with file style.

### Remaining Issues (Minor)

1. **`surnametab.py:462` — old-style super call** (only remaining one):
   ```python
   # current
   EmbeddedList.clean_up(self)
   # should be
   super().clean_up()
   ```

2. **Long docstrings in `eventembedlist.py` and `surnametab.py`** — still verbose vs. the rest of the codebase. One-liners would suffice.

3. **`ButtonTab.clean_up` pass-through** — still present, adds no value. Could be deleted since `GrampsTab.clean_up` is the parent and will be called via the MRO naturally. Low risk either way.

### Summary

All required changes are addressed and most optional ones are too. Only two minor issues remain: the old-style `super()` call in `surnametab.py` and some overly verbose docstrings. The core resource leak fixes look correct. This is close to merge-ready — the `surnametab.py` super call is the one thing worth fixing before approval.

stevenyoungs:
> 2\. Long docstrings

@dsblank Is this something you can attempt to tame in `AGENTS.md`?

stevenyoungs:
I've addressed the three remaining issues

dsblank (changes requested):
## PR #2237 Review — Fix Resource Leaks

**Overall:** The core changes are correct and address a real problem. The `gallerytab.py` bug (missing `_cleanup_callbacks()` + wrong `super()` target) is the most impactful fix. The `embeddedlist.py` `clean_up()` addition and the `model is not None` correction are solid improvements. The main ask before merging is trimming the Copilot-generated comments to match project style.

---

### Analysis by File

#### `embeddedlist.py`

**Good:**
- `if self.model and ...` → `if self.model is not None and ...` is a genuine correctness fix. Some GTK model objects can be falsy without being `None`.
- Clearing `column_header_images` in `build_columns()` before rebuilding prevents stale refs accumulating across rebuilds.

**Concerns:**
- `self.column_header_images` is registered with `track_ref_for_deletion`, but `EmbeddedList.clean_up()` already explicitly clears it before calling `super().clean_up()`. The tracked deletion will then also `delattr` it — harmless but redundant.
- `clean_up()` calls `self.tree.set_model(None)` while `tree` is also tracked for deletion. The ordering is correct (explicit cleanup runs before `super().clean_up()` processes tracked refs), but worth noting.

#### `gallerytab.py` — Most impactful fix

**Good:**
- The old `clean_up()` called `ButtonTab.clean_up(self)` instead of `super().clean_up()`, and — critically — never called `self._cleanup_callbacks()` at all. DB signal handlers were never disconnected on editor close. Both are now fixed.
- Promoting `text_renderer` from local variable to `self.text_renderer` is correct — Python's GC couldn't see the reference while `IconView` held it.
- Initializing `self.text_renderer = None` and `self.menu = None` before `ButtonTab.__init__` is safe and defensive.

**Concern:**
- After `clean_up()` calls `iconmodel.clear()`, `super().clean_up()` (`GrampsTab`) will `delattr(self, 'iconmodel')`. A hypothetical second `clean_up()` call would then raise `AttributeError` on `if self.iconmodel is not None`. Not a real-world issue since `clean_up()` is called once, but the pattern is inconsistent with the `hasattr` guard used in `embeddedlist.py`.

#### `eventembedlist.py`

No concerns. `EventEmbedList(DbGUIElement, GroupEmbeddedList)` needs `_cleanup_callbacks()` on close, and `super().clean_up()` correctly chains through to the new `EmbeddedList.clean_up()` via MRO.

#### `surnametab.py`

No concerns. Initializing `self.cmborig = None` before `EmbeddedList.__init__()` is important — if `clean_up()` is ever called before `build_columns()` (e.g. on init failure), the `is not None` guard in `clean_up()` will work correctly.

---

### Changes Requested

1. **Remove the inline comments.** Every `super().clean_up()` call has a `# Call parent class cleanup which will handle tracked references` comment; `self._cleanup_callbacks()` has `# Disconnect all database callbacks`; etc. These describe what the code already says and don't match Gramps project style. Please strip them before merging.

2. **Trim the `clean_up()` docstrings.** The parent class documents the protocol. The override docstrings restate it with GDI-specific detail that belongs in the PR description, not the source. A brief one-liner (or no docstring, deferring to the parent) is appropriate.

3. **Clarify the framing.** These are GTK object lifecycle fixes that benefit all platforms — the GDI leak is most visible on Windows AIO, but the fixes are not Windows-specific. Commit messages and any remaining comments should reflect that.

4. **Manual test confirmation.** Was the AIO Windows package tested with this fix? If so, please note it in the PR description (reproduction steps / before-after crash frequency). If not, it would be good to verify before merge.

dsblank:
> 2. Long docstrings

I'll see if I can formulate a description of docstrings that matches our standard.

stevenyoungs:
rebased.

> * **Remove the inline comments.** 

Done.

> * **Trim the `clean_up()` docstrings.** 

These were trimmed before your review earlier today.

> * **Clarify the framing.** 

Commit messages and code comments adjusted

> * **Manual test confirmation.** 

I manually tested opening and closing the EditPerson dialog 150 times and was not able to reproduce the problem

stevenyoungs:
> * After `clean_up()` calls `iconmodel.clear()`, `super().clean_up()` (`GrampsTab`) will `delattr(self, 'iconmodel')`. A hypothetical second `clean_up()` call would then raise `AttributeError` on `if self.iconmodel is not None`. Not a real-world issue since `clean_up()` is called once, but the pattern is inconsistent with the `hasattr` guard used in `embeddedlist.py`.

`hasattr(self, "iconmodel")` guard added

dsblank (commented):
A few items to address before merging:

**1. Missing guard in `EmbeddedList.clean_up()`**

`self.tree` is accessed without checking if it exists:
```python
if hasattr(self, "model") and self.model is not None and hasattr(self.model, "destroy"):
    self.tree.set_model(None)  # can raise AttributeError if tree is gone
```
Should add `hasattr(self, "tree")` to the condition.

**2. Redundant `_cleanup_callbacks()` in `EventEmbedList.clean_up()`**

`editprimary.py` already calls `tab._cleanup_callbacks()` for every tab during `_cleanup_db_connects()`. The extra call here is harmless but redundant — either remove it or add a comment explaining why it's needed a second time.

**3. Comment style**

Several comments describe *what* the code does rather than *why*:
```python
# Clear and destroy the model to release GDI resources
# Call parent class cleanup which will handle tracked references
```
The project convention is to only comment non-obvious intent. These can be removed.

dsblank:
Paraphrasing @Nick-Hall:

"If a developer targets the current maintenance branch with a bugfix then it is considered for fairly quick release"

PR #2239, opened 2026-04-17
Explicitly convert label parameter to string in Gtk.Label

hgohel:
The argument `label` to Gtk.Label must be of type string. In recent versions of Python and/or PyGObject, this check has become stricter and as a result is throwing an exception instead of silently converting to a string. So now we explicitly convert to string.

Note: I have not been able to reproduce the crash as it is data and view configuration dependent, so it is important to validate the changes.

Fixes #[14183](https://gramps-project.org/bugs/view.php?id=14183) which was reported for Place details. However, similar code exists in both Person details and Repository details so I've made a defensive change in both those locations as well.

### DIFF
--- a/gramps/plugins/gramplet/persondetails.py
+++ b/gramps/plugins/gramplet/persondetails.py
@@ -82,7 +82,7 @@ class PersonDetails(Gramplet):
         )
         label.set_selectable(True)
         label.show()
-        value = Gtk.Label(label=value, halign=Gtk.Align.START)
+        value = Gtk.Label(label=str(value), halign=Gtk.Align.START)
         value.set_selectable(True)
         value.show()
         self.grid.add(label)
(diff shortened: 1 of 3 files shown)


Nick-Hall:
Both #2238 and this PR fix the same bug.

Nick-Hall (changes requested):
There is only one value that is not already passed in as a string.  This bug can be fixed by replacing `place.get_type()` with `str(place.get_type())`.

stevenyoungs:
@hgohel looks like we both found the same problem via different routes - I stumbled across it by accident, testing something else.
I'll take your explanation and add it to #2238, with attribution. Between us, I think we then have a good explanation and steps to reproduce the problem ❤️

Nick-Hall:
Closed in favour of PR #2238.

PR #2240, opened 2026-04-17
Fix Windows AIO addon installation failing on missing stdlib modules

eduralph:
## Summary

Fixes Windows AIO addon installation, which crashed with `ImportError` before pip could do useful work whenever an addon declared pypi dependencies. Reported as [#13921](https://gramps-project.org/bugs/view.php?id=13921) for the S3 Media Uploader addon (missing `ctypes.wintypes`), but the class of bug is broader: pip and its vendored libraries (truststore, rich, urllib3, distlib) import various stdlib submodules conditionally or lazily, and cx_Freeze's static scan misses them, so any addon with Python dependencies can trip a fresh `ModuleNotFoundError`.

## Approach

- **`aio/setup.py`** — bundle the affected stdlib packages (`ctypes`, `encodings`, `email`, `http`, `urllib`, `importlib`) **in full** via cx_Freeze's `packages` list so every submodule is included, regardless of whether the static scan picks it up. Add single-file `getpass` to `includes`. This replaces the whack-a-mole pattern of enumerating individual submodules.

- **`aio/build.sh`** — smoke test the frozen `pip.exe` right after `python setup.py build_exe` and before NSIS. If pip emits `ImportError`, `ModuleNotFoundError`, or any traceback, the build aborts before the installer is produced. Future coverage gaps now surface on the build host rather than in user bug reports.

- **`aio/build.sh`** (separate commit) — sanitize `/` in branch names when computing `appbuild`, so builds on branches like `fix/issue-NNNN-xxx` don't break the sed substitution into the NSIS template and the resulting installer filename. This isn't part of the bug itself, but it's what let me build the fix off this branch in the first place, so I've kept it here for review.

## Test plan

Tested on a Windows VM with a clean MSYS2/MINGW64 environment, building from this branch:

- [x] Build completes; smoke test passes.
- [x] Installer produced and installs cleanly.
- [x] **S3 Media Uploader** addon installs and runs. (Biggest dependency graph of the set: boto3 → botocore + urllib3 + jmespath + s3transfer.)
- [x] **TMG Project Backup** addon installs and runs. (Pulls `dbf`.)
- [x] **DNA Segment Map** addon installs and runs. (Pulls `pandas`/native wheels.)

Diff footprint is 40 lines across two files; no changes outside `aio/`.

Fixes #13921.

### DIFF
--- a/aio/build.sh
+++ b/aio/build.sh
@@ -182,6 +182,33 @@ cd aio
 cat grampsaio64.nsi.template | sed "s/yourVersion/$appversion/;s/yourBuild/$appbuild/" >grampsaio64.nsi
 # build cx_freeze executables
 python setup.py build_exe
+
+# Smoke test the frozen pip before packaging so stdlib-coverage gaps are
+# caught here instead of by end users running the Addon Manager.
+# $MSYSTEM is MINGW64/UCRT64/etc.; the matching build tree is the
+# lowercase equivalent, so the command below works across environments
+# without needing to be edited when the rest of the script is modernized.
+echo "Smoke-testing bundled pip..."
+msys_dir="${MSYSTEM,,}"
+smoke_fail=0
+for cmd in "./${msys_dir}/pip.exe --version" "./${msys_dir}/pip.exe install --dry-run --ignore-installed certifi"; do
+    echo "--- $cmd ---"
+    smoke_out=$($cmd 2>&1) || true
+    echo "$smoke_out"
+    if echo "$smoke_out" | grep -qiE "ImportError|ModuleNotFoundError|Traceback"; then
+        echo "ERROR: Frozen pip emitted an error or traceback running: $cmd"
+        smoke_fail=1
+    fi
+done
+if [ "$smoke_fail" -ne 0 ]; then
+    echo "ERROR: pip smoke test failed; aborting build before NSIS."
+    echo "  Likely cause: a stdlib submodule pip (or one of its vendored"
+    echo "  libraries) imports lazily was not picked up by cx_Freeze's"
+    echo "  static scan. Add the missing module to INCLUDES or its parent"
+    echo "  package to PACKAGES in aio/setup.py and rebuild."
+    exit 1
+fi
+
 # build installer
 cd mingw64/src
 makensis grampsaio64.nsi
--- a/aio/setup.py
+++ b/aio/setup.py
@@ -66,7 +66,19 @@ else:
 
 INCLUDE_DLL_PATH = os.path.join(sys.base_exec_prefix, "bin")
 INCLUDE_FILES = []
-INCLUDES = ["gi", "colorsys", "site"]
+# Simple (single-file) stdlib modules pip's vendored libraries import
+# lazily and therefore cx_Freeze's static scan misses.
+INCLUDES = [
+    "gi",
+    "colorsys",
+    "site",
+    "getpass",
+]
+# Bundle these stdlib packages in full. pip and its vendored libraries
+# (truststore, rich, urllib3, distlib) import submodules conditionally
+# or lazily, and cx_Freeze's static scan misses them. Listing the parent
+# package here pulls in every submodule, so addon installation no longer
+# crashes on the first un-frozen import.
 PACKAGES = [
     "gi",
     "cairo",
@@ -90,6 +102,12 @@ PACKAGES = [
     "pydot",
     "orjson",
     "selenium",
+    "ctypes",
+    "encodings",
+    "email",
+    "http",
+    "urllib",
+    "importlib",
 ]
 EXCLUDES = [
     "tkinter",


dsblank:
Welcome @eduralph ! We very much appreciate these contributions! What brought you to gramps? Hope to create an engaging and productive developer environment here.

eduralph:
Thanks @dsblank - I started doing some work for the TMG Importer in the add-on section and ran into an issue with windows not installing correctly - the issue I fixed. I've been working on that corner of the system, focusing on resolving errors and suggesting some improvements to make the plugin process more resilient.

stevenyoungs:
@Nick-Hall this PR will need adjusting as well if #2198 is merged.
I suggest you make the call on #2198 and then @eduralph and I can fix our various AIO related PRs to target UCRT64

aio/build.sh
```diff
@@ -168,7 +168,7 @@ rm -rf dist aio/dist
 python setup.py bdist_wheel
 if `grep -q '^DEV_VERSION\s*=\s*True' gramps/version.py`; then
     # <branch_name>-<short_commit_id>
-    appbuild="$(git rev-parse --abbrev-ref HEAD)-$(git rev-parse --short HEAD)"
+    appbuild="$(git rev-parse --abbrev-ref HEAD | tr '/' '-')-$(git rev-parse --short HEAD)"
```
stevenyoungs: I provided a different approach in #2200 which escapes characters in `appversion` and `appbuild` and replaces all characters which are invalid in an NTFS filename
Do you want to drop this part of your change in deference to #2200?

aio/setup.py
```diff
@@ -66,7 +66,15 @@ def normalize(version):
...
 INCLUDE_DLL_PATH = os.path.join(sys.base_exec_prefix, "bin")
 INCLUDE_FILES = []
-INCLUDES = ["gi", "colorsys", "site"]
+INCLUDES = [
+    "gi",
+    "colorsys",
+    "site",
+    # Simple (single-file) stdlib modules pip's vendored rich imports lazily
+    # and cx_Freeze's static scan misses. Package-form stdlib with the same
+    # problem is handled in PACKAGES below.
```
stevenyoungs:
```suggestion
    # Simple (single-file) stdlib modules pip's vendored libraries import lazily
    # and therefore cx_Freeze's static scan misses. 
    # Package-form stdlib packages with the same problem are handled in 
    # PACKAGES below.
```
I was struggling to parse the original comment - which I accept could be a me problem 😄 
Is my proposed wording still technically accurate?

Is it correct to use "stdlib modules" here, and "stdlib packages" on line 101?

aio/build.sh
```diff
@@ -182,6 +182,25 @@ cd aio
 cat grampsaio64.nsi.template | sed "s/yourVersion/$appversion/;s/yourBuild/$appbuild/" >grampsaio64.nsi
 # build cx_freeze executables
 python setup.py build_exe
+
+# Smoke test the frozen pip before packaging so stdlib-coverage gaps are
+# caught here instead of by end users running the Addon Manager.
+echo "Smoke-testing bundled pip..."
+smoke_fail=0
+for cmd in "./mingw64/pip.exe --version" "./mingw64/pip.exe install --dry-run --ignore-installed certifi"; do
```
stevenyoungs: If #2198 is merged first, this will need revising. Probably:

```suggestion
for cmd in "./ucrt64/pip.exe --version" "./ucrt64/pip.exe install --dry-run --ignore-installed certifi"; do
```

stevenyoungs (changes requested):

aio/build.sh
eduralph: Yes, you are infinitly more familiar with the code base and what makes more sense. will do that.

aio/build.sh
stevenyoungs: #2198 now merged. I've made a local build and the changed proposed above for ucrt64 is correct.

aio/build.sh
eduralph: Done — dropped the slash-sanitization commit; deferring to #2200 for the broader NTFS-character handling.

aio/setup.py
eduralph: Applied your wording almost verbatim (slight reflow to keep line lengths consistent

aio/build.sh
eduralph: Didn't do anything on this one - switch to ucrt64 is on gramps61. Let me know if there is any follow-up work

stevenyoungs: I assume Nick will merge gramps61 into master before long

hgohel: To eliminate the question of which branch this code will be committed to, use `$MSYSTEM` instead of an explicit environment like MINGW64 or UCRT64. At some point making a pass through the build scripts to do this everywhere would be good.

aio/build.sh
```diff
@@ -182,6 +182,25 @@ cd aio
...
+    echo "--- $cmd ---"
+    smoke_out=$($cmd 2>&1) || true
+    echo "$smoke_out"
+    if echo "$smoke_out" | grep -qiE "ImportError|ModuleNotFoundError|Traceback"; then
+        echo "ERROR: Frozen pip emitted an error or traceback running: $cmd"
+        smoke_fail=1
+    fi
+done
+if [ "$smoke_fail" -ne 0 ]; then
+    echo "ERROR: pip smoke test failed; aborting build before NSIS."
```
hgohel, line 204: It would be helpful to suggest a course of action to the person reading the build log. Would it be something like "make sure to include dependencies in setup.py"?

aio/setup.py
```diff
@@ -66,7 +66,15 @@ def normalize(version):
...
 INCLUDE_DLL_PATH = os.path.join(sys.base_exec_prefix, "bin")
 INCLUDE_FILES = []
-INCLUDES = ["gi", "colorsys", "site"]
+INCLUDES = [
+    "gi",
+    "colorsys",
+    "site",
+    # Simple (single-file) stdlib modules pip's vendored libraries import
+    # lazily and therefore cx_Freeze's static scan misses. Package-form
+    # stdlib packages with the same problem are handled in PACKAGES below.
```
hgohel: Suggest moving the comment outside the list of INCLUDES, and second, remove the part about "Package-form stdlib packages..." because that is covered separately in the PACKAGES section.

aio/setup.py
```diff
@@ -90,6 +98,17 @@ def normalize(version):
     "pydot",
     "orjson",
     "selenium",
+    # Bundle these stdlib packages in full. pip and its vendored libraries
+    # (truststore, rich, urllib3, distlib) import submodules conditionally
+    # or lazily, and cx_Freeze's static scan misses them. Listing the parent
+    # package here pulls in every submodule, so addon installation no longer
+    # crashes on the first un-frozen import.
```
hgohel: Same as above, move the comment outside the list.

hgohel (commented):
Good change, thanks for bringing your expertise to Gramps.

eduralph:
Thanks for the review @hgohel, @stevenyoungs — addressed in [282535e4](https://github.com/gramps-project/gramps/commit/282535e4):

- **`aio/build.sh` smoke-test path** — switched from hard-coded `mingw64` to `${MSYSTEM,,}` (lowercased `MSYSTEM`). No edit needed when this branch later picks up the UCRT64 switch from #2198, and it works unchanged today on MINGW64. Covers @hgohel's `$MSYSTEM` suggestion and @stevenyoungs' UCRT64 follow-up in one change.
- **`aio/build.sh` smoke-test failure message** — now prints an actionable hint pointing at `INCLUDES` / `PACKAGES` in `aio/setup.py`.
- **`aio/setup.py`** — moved both list-describing comments out of the list bodies; dropped the `INCLUDES` cross-reference to `PACKAGES` so each list documents only itself.

stevenyoungs (approved):
great change. Thanks @eduralph

hgohel (approved):
Looks good to me.

PR #2243, opened 2026-04-18
Log plugin registration and load failures instead of printing

eduralph:
## Summary
**User impact:** When an add-on fails to load or register, the error is printed to the terminal — where it is easy to miss and hard to capture in a bug report — instead of going through Gramps' normal message channel. In the worst case a stray failure in one add-on could swallow an unrelated interruption (like Ctrl-C).

This routes every plugin scan/load failure through Gramps' logging so the diagnostics land in one place users and packagers can control and attach to reports.

Origin: robustness / code-quality change to the plugin loader — no Mantis tracker ticket.

## What to look at
The change swaps bare `except:` clauses for typed handlers and replaces `print()` / `traceback.print_exc()` with logger calls, while keeping the tuple shape `get_fail_list()` returns so the existing GUI error dialog is untouched. To try it: install a deliberately broken third-party add-on and watch the failure surface through the logger instead of stdout, with `KeyboardInterrupt` / `SystemExit` still propagating.

## Root cause
Bare `except:` clauses in `gramps/gen/plug/_manager.py` and `gramps/gen/plug/_pluginreg.py` catch everything — including `KeyboardInterrupt` / `SystemExit` — and plugin scan/load diagnostics are emitted to `stdout` via `print()` / `traceback.print_exc()`, so a single broken add-on can silently swallow diagnostics or risk propagating an unrelated exception through the scanner.

## Fix
- Replace the bare `except:` clauses in `gramps/gen/plug/_manager.py` and `gramps/gen/plug/_pluginreg.py` with typed exception handlers (`Exception`, `TypeError`, `(OSError, UnicodeDecodeError)`), so genuine programming errors and `KeyboardInterrupt` propagate naturally.
- Route all plugin scan and load diagnostics through module loggers (`LOG.error`, `LOG.exception`, `LOG.warning`) instead of `print()` / `traceback.print_exc()`.
- Preserve the `(filename, sys.exc_info(), pdata)` shape returned by `BasePluginManager.get_fail_list()` so the existing GUI error dialog (`gramps/gui/viewmanager.py`, `gramps/gui/plug/_windows.py`) keeps working unchanged.
- Translatable strings: several scan diagnostics in `_pluginreg.py` / `_manager.py` that were previously wrapped in `_()` and `print()`ed are now unwrapped `LOG.*` messages (diagnostic log output is not localized). `_pluginreg.py` keeps other translatable strings, so no `po/POTFILES.in` change is required for it; only the new test module is added to `po/POTFILES.skip`. The touched function bodies (`reg_plugins`, `load_plugin`, `import_plugin`, `reload`, `reload_plugins`, `get_plugin_data`, `process_plugin_data`, `scan_dir`) already carry no type hints; retrofitting hints is intentionally left out to keep this to one concern.

## Verification
- **Claim:** plugin scan/load failures are reported through module loggers and no longer suppress `KeyboardInterrupt` / `SystemExit`, while the `get_fail_list()` tuple contract is unchanged.
- **Checked:** `gramps/gen/plug/_manager.py`, `gramps/gen/plug/_pluginreg.py` on maintenance/gramps60 (the branch this PR targets) — typed excepts plus `LOG.*` calls; `(filename, sys.exc_info(), pdata)` preserved for the GUI dialog.
- **Test:** `gramps/gen/plug/test/_pluginreg_errors_test.py` (new) exercises the error paths in `PluginRegister.scan_dir` for `RuntimeError`, `ValueError`, and non-UTF-8 bytes in a `gpr.py`, asserting `LOG.exception` fires via `assertLogs("._manager", level="ERROR")` for each arm. Full suite: 32,502 tests, only the two pre-existing unrelated `test_WebCal` / `test_navwebpage` CSS failures; `mypy` and `black --check` clean on the touched files.

No Mantis tracker ticket — this is an internal robustness / logging change.

### DIFF
--- a/gramps/gen/plug/_manager.py
+++ b/gramps/gen/plug/_manager.py
@@ -187,9 +187,13 @@ class BasePluginManager:
                             plugins_to_load.remove(plugin)
                 count += 1
                 if count > max_count:
-                    print("Cannot resolve the following plugin dependencies:")
+                    LOG.error("Cannot resolve the following plugin dependencies:")
                     for plugin in plugins_to_load:
-                        print(f"   Plugin '{plugin.id}' requires: {plugin.depends_on}")
+                        LOG.error(
+                            "   Plugin '%s' requires: %s",
+                            plugin.id,
+                            plugin.depends_on,
+                        )
                     break
             # now load them:
             for plugin in plugins_sorted:
@@ -201,16 +205,15 @@ class BasePluginManager:
                     # LOG.warning("\nRun %s at registration", plugin.id)
                     try:
                         results = mod.load_on_reg(dbstate, uistate, plugin)
-                    except:
-                        import traceback
-
-                        traceback.print_exc()
-                        print(f"Plugin '{plugin.name}' did not run; continuing...")
+                    except Exception:
+                        LOG.exception(
+                            "Plugin '%s' did not run; continuing...", plugin.name
+                        )
                         continue
                     try:
                         iter(results)
                         plugin.data += results
-                    except:
+                    except TypeError:
                         plugin.data = results
         # Get the addon rules and import them and make them findable
         for plugin in self.__pgr.rule_plugins():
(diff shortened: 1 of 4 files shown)


eduralph:
Retargeted the enhancement to gramps60 as per rule

PR #2244, opened 2026-04-18
Make plugin scan and import transactional against failures

eduralph:
> **Note:** This PR is stacked on top of #2243. Merging #2243 first and then rebasing this PR onto `master` is recommended. Only the commit titled *"Make plugin scan and import transactional against failures"* is new here.

## Summary
**User impact:** A single broken add-on can leave Gramps' internal list of installed add-ons half-updated, or permanently pollute the set of places Gramps looks for code. Both cause confusing, hard-to-diagnose problems in setups with third-party add-ons — sometimes long after the broken add-on itself is gone.

This makes scanning and importing each add-on all-or-nothing, so one broken add-on can no longer leave partial state or a lingering side effect behind.

Origin: robustness / code-quality change to the plugin loader — no Mantis tracker ticket (stacked on #2243).

## What to look at
`PluginRegister.scan_dir` now snapshots the registry per `.gpr.py` and rolls it back on any failure; `import_plugin` routes its search-path insertion through a context manager so the entry is always removed. To try it: point Gramps at a `.gpr.py` that raises mid-registration and confirm the registry and the module search path are exactly as they were before the attempt.

## Root cause
A failing `.gpr.py` today can leave the registry inconsistent: `__plugindata` is trimmed on exception but `__id_to_pdata` is not, and the validation loop is not wrapped in a try/except, so a raise there leaves partial data in both. Separately, an uncaught exception from an imported plugin module (e.g. a `RuntimeError` at import time) skips the manual `sys.path.pop(0)` at the end of `import_plugin`, permanently leaking `pdata.fpath` into `sys.path` — a source of hard-to-diagnose side effects in third-party plugin setups.

## Fix
- `PluginRegister.scan_dir` is now transactional per `.gpr.py`: before processing each file it snapshots `__plugindata` and `__id_to_pdata`; any failure in read / exec / id-mapping / validation restores both snapshots. The per-file body is extracted into a private `__scan_file` helper.
- `BasePluginManager.import_plugin` routes `sys.path` manipulation through a new `_prepended_sys_path` context manager so the inserted entry is removed on exit regardless of how the `with` block terminates. The Windows retry branch is hardened with the same pattern, removing a nested double-pop risk.
- The external contract of `get_fail_list()` and `scan_dir()` return values is preserved; no user-visible behaviour changes for correctly registered plugins.

## Verification
- **Claim:** a failing `.gpr.py` or plugin import leaves the registry (`__plugindata` and `__id_to_pdata`) and `sys.path` exactly as they were before the attempt.
- **Checked:** `gramps/gen/plug/_pluginreg.py`, `gramps/gen/plug/_manager.py` on maintenance/gramps60 (the branch this PR targets).
- **Test:** `gramps/gen/plug/test/_pluginreg_transaction_test.py` (rollback on mid-file `RuntimeError` / `ValueError`, plus a successful-scan sanity check) and `gramps/gen/plug/test/_manager_import_test.py` (`sys.path` invariants after successful / `ImportError` / uncaught-exception imports, plus direct unit tests for the `_prepended_sys_path` context manager). Full suite: 32,512 tests, only the two pre-existing unrelated `test_WebCal` / `test_navwebpage` CSS failures; `mypy` and `black --check` clean on the touched files.

No Mantis tracker ticket — internal robustness change stacked on #2243.

### DIFF
--- a/gramps/gen/plug/_manager.py
+++ b/gramps/gen/plug/_manager.py
@@ -187,9 +187,13 @@ class BasePluginManager:
                             plugins_to_load.remove(plugin)
                 count += 1
                 if count > max_count:
-                    print("Cannot resolve the following plugin dependencies:")
+                    LOG.error("Cannot resolve the following plugin dependencies:")
                     for plugin in plugins_to_load:
-                        print(f"   Plugin '{plugin.id}' requires: {plugin.depends_on}")
+                        LOG.error(
+                            "   Plugin '%s' requires: %s",
+                            plugin.id,
+                            plugin.depends_on,
+                        )
                     break
             # now load them:
             for plugin in plugins_sorted:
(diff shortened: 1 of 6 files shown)


eduralph:
Retargeted to gramps60 branch

PR #2245, opened 2026-04-18
Validate addon listing JSON gracefully in get_addons (Bug #12974)

eduralph:
## Summary
**User impact:** If the add-on server sends back a broken or incomplete list of available add-ons, the Add-on/Plugin Manager window crashes instead of simply skipping the bad entries and carrying on.

This makes the add-on list load defensively — skipping and logging malformed entries — so the manager never crashes on a bad server response.

Reported in Mantis [#12974](https://gramps-project.org/bugs/view.php?id=12974).

## What to look at
`get_addons()` now validates the downloaded listing before trusting it, and the bad-listing paths are logged and skipped rather than raised. To try it: point the manager at a `file://` listing that is malformed (bad JSON, not an array, or entries missing fields) and confirm the window stays up. Seven new tests drive exactly these cases through a `file://` fixture, so no network is involved.

## Root cause
`gramps/gen/plug/utils.py::get_addons` parses the downloaded addon listing without validation, so a malformed or partial listing on the remote side raises and crashes the plugin-manager window. A bare `except:` on the per-language fetch also swallowed `KeyboardInterrupt` / `SystemExit`.

## Fix
- `get_addons` now parses defensively: `json.load` is wrapped with explicit handling for `JSONDecodeError`, `OSError`, and `UnicodeDecodeError`; the top level must be a JSON array; individual entries must be objects containing the required keys `i`, `n`, `v`, `d`, `t`, `z`. Anything else is logged at an appropriate level and skipped.
- Narrowed the bare `except:` used when fetching per-language listings to `except Exception:`, keeping `KeyboardInterrupt` / `SystemExit` propagating.
- Strictly additive: well-formed listings behave exactly as before, and the returned dicts still carry the same keys the plugin-manager GUI consumes (including the defaulted `a` / `s` / `h` and the injected `_p` / `_u` / `_v`).

## Verification
- **Claim:** a malformed or partial addon listing is logged and skipped and never crashes the plugin manager, while well-formed listings behave unchanged.
- **Checked:** `gramps/gen/plug/utils.py` on maintenance/gramps60 (the branch this PR targets) — validation guard on the top-level array and per-entry required keys.
- **Test:** `gramps/gen/plug/test/_utils_addons_test.py` (new) — seven tests covering the happy path, malformed JSON body, non-array top level, non-dict entries, entries missing required fields, unsupported URL schemes, and a missing listing file, all via a `file://` fixture. Full suite: 32,506 tests, only the two pre-existing unrelated `test_WebCal` / `test_navwebpage` CSS failures; `mypy` and `black --check` clean on the touched files.

Fixes [#12974](https://gramps-project.org/bugs/view.php?id=12974)

### DIFF
--- a/gramps/gen/plug/test/_utils_addons_test.py
+++ b/gramps/gen/plug/test/_utils_addons_test.py
@@ -1,0 +1,147 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Eduard Ralph
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""
+Tests that ``gramps.gen.plug.utils.get_addons`` validates addon listing
+JSON and fails gracefully rather than crashing the plugin manager UI.
+
+See Gramps bugtracker issue #12974.
+"""
+
+# ------------------------
+# Python modules
+# ------------------------
+import json
+import os
+import pathlib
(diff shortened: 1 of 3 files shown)


eduralph:
Force-pushed (9828b150c6 → 205dbd47be) — commit-message-only change, no code touched:

Moved the bug reference to the commit body's last line as "Bug #12974." per AGENTS.md §Commit Messages ("The last line must link to or resolve a bug report using the Mantis BT integration keywords"). The reference was previously only in the title.
Dropped the redundant "(Bug #12974)" from the title to keep the summary focused.

eduralph:
retargeted to gramps60 branch

PR #2246, opened 2026-04-18
Warn on plugins that target the previous minor Gramps version

eduralph:
## Summary
**User impact:** When an add-on is written for the previous Gramps version, Gramps silently refuses to load it and buries the reason among the other start-up messages. Add-on authors usually only find out from user bug reports — with no early warning before their add-on stops working.

This loads such one-version-behind add-ons for a one-release grace period while emitting a clear deprecation warning, giving authors a window to update the target before the add-on is dropped entirely.

Origin: feature enhancement to the plugin loader — no Mantis tracker ticket (opened as a draft because it is a feature, not a bug fix).

## What to look at
Plugins whose `gramps_target_version` is exactly one minor behind the running Gramps are now loaded, but with a `LOG.warning` telling the author to bump the target; two-minors-behind, wrong-major, or future versions are still rejected and skipped exactly as before. To try it: drop a `.gpr.py` targeting the previous minor version into a plugin directory and confirm it registers *and* emits a deprecation warning naming that version.

## Root cause
In `gramps/gen/plug/_pluginreg.py`, a `gpr.py` that still targets the previous minor version is silently rejected with an error message mixed in with the rest of the plugin-registration output. There is no deprecation window and no structured signal, so authors typically learn about it only from user bug reports.

## Fix
- Introduce a three-way `PluginVersionStatus` (`CURRENT` / `DEPRECATED` / `INVALID`) and a `plugin_version_status()` classifier in `gramps/gen/plug/_pluginreg.py`.
- `PluginRegister.scan_dir` now loads plugins one minor behind the running Gramps but emits a `LOG.warning` through the `._manager` logger; `INVALID` (two minors behind, wrong major, or future) still routes through the existing print-and-skip path — no behaviour change for those.
- `valid_plugin_version()` is kept as a thin bool compat wrapper that delegates to `plugin_version_status()` and treats `DEPRECATED` as valid, so external callers do not regress.
- Adds `gramps/gen/plug/test/__init__.py` (new, empty) test-package marker.

## Verification
- **Claim:** a plugin one minor behind the running Gramps loads and emits a deprecation warning; two-minors-behind / wrong-major / future remain rejected; the `valid_plugin_version()` bool contract is preserved.
- **Checked:** `gramps/gen/plug/_pluginreg.py` on master (the branch this PR targets) — `scan_dir` routes `DEPRECATED` through `LOG.warning` while still registering the plugin.
- **Test:** `gramps/gen/plug/test/_pluginreg_version_status_test.py` (new) — 17 tests: 13 classifier units (CURRENT / DEPRECATED / INVALID including patch handling, zero-minor edge case, non-string inputs, malformed strings), 3 compat-wrapper tests, and an integration test that writes a real `.gpr.py` targeting the previous minor, invokes `scan_dir`, and asserts the plugin is registered *and* a warning containing "deprecated" and the deprecated version number is emitted. `mypy` and `black --check` clean; full suite 32,516 tests, only the two pre-existing unrelated `test_WebCal` / `test_navwebpage` failures. As a new feature there is no pre-existing bug to reproduce; the integration test guards the new deprecation-window behaviour.

No Mantis tracker ticket — feature enhancement, opened as a draft.

### DIFF
--- a/gramps/gen/plug/_pluginreg.py
+++ b/gramps/gen/plug/_pluginreg.py
@@ -212,23 +212,84 @@ def version(string_version):
     return tuple(myint(x or "0") for x in (f"{string_version}..").split("."))
 
 
-def valid_plugin_version(plugin_version_string):
+# ------------------------------------------------------------
+#
+# PluginVersionStatus
+#
+# ------------------------------------------------------------
+class PluginVersionStatus:
     """
-    Checks to see if string is a valid version string for this version
-    of Gramps.
+    Classification of a plugin's ``gramps_target_version`` relative to the
+    currently running Gramps version.
     """
+
+    #: Plugin targets the current major.minor (and, if three-part, a patch
+    #: level no later than the running Gramps).
+    CURRENT = "current"
+    #: Plugin targets the immediately previous minor within the same major.
+    #: It is still loaded but its author should refresh the target version.
+    DEPRECATED = "deprecated"
+    #: Anything else: malformed, cross-major, future version, or a minor
+    #: more than one release behind.
+    INVALID = "invalid"
+
+
+def plugin_version_status(
+    plugin_version_string: object,
+    current_version: tuple[int, int, int] | None = None,
+) -> str:
+    """
+    Classify *plugin_version_string* against *current_version*.
+
+    :param plugin_version_string: The value a plugin sets for
+                                  ``gramps_target_version``.
+    :type plugin_version_string: object
+    :param current_version: A ``(major, minor, patch)`` tuple describing the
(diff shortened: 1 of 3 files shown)


eduralph:
Force-pushed (cfca9eba07 → 43c0232cec) to address AGENTS.md requirements before lifting draft:

Added Python 3.10+ type hints to plugin_version_status and valid_plugin_version
Added the class navigation header above PluginVersionStatus
Expanded both docstrings to include :type: and :rtype: entries per the Sphinx example in AGENTS.md
No behavior change.

PR #2247, opened 2026-04-18
Classify Addon Manager project-path fetches for a validity indicator

eduralph:
## Summary
**User impact:** In the Add-on Manager, when a configured project's address cannot be reached — a wrong or missing local path, no internet, a server error, or a list left over from an older Gramps — it looks exactly the same as "no new add-ons available." Users get no hint that anything is actually wrong.

This adds the backend that classifies each project fetch into a specific outcome, so a follow-up can attach a per-project validity indicator (icon + tooltip) to each row of the Projects tab.

Reported in Mantis [#13069](https://gramps-project.org/bugs/view.php?id=13069).

## What to look at
A new `classify_project_fetch()` performs the same language-chain lookup as `get_addons()` and maps the result to one distinct status (OK / INVALID_URL / NETWORK_ERROR / HTTP_ERROR / JSON_ERROR / EMPTY_LISTING / NO_COMPATIBLE_ADDONS / LANG_FALLBACK_EN), alongside a `normalize_project_url()` helper. The change is backend-only — the existing fetch functions are untouched and the GTK wiring is a separate follow-up. To try it: call `classify_project_fetch()` against a bad URL and inspect the returned status (and `http_code` on HTTP errors).

## Root cause
Bug #13069 (*[6.1 Addon Manager] Add 'valid' indicator for 'project' paths*) lists several failure modes — invalid local paths, missing JSON for the current language, lost internet, HTTP 4xx/5xx, and add-on lists incompatible with the running version (the 2025-04-06 comment about 5.2 projects left over after upgrade to 6.0) — that all appear identical to the user today, because `gramps/gen/plug/utils.py` exposes no way to distinguish them.

## Fix
- Add to `gramps/gen/plug/utils.py`: `normalize_project_url()` (strips surrounding whitespace and trailing slashes, rejects non-`http(s)://` / `file://` values), `ProjectFetchStatus` constants, a `ProjectFetchResult` dataclass, `classify_project_fetch(url, current_version=None, languages=None)`, and two private helpers (`_candidate_languages`, `_try_fetch_listing`).
- `classify_project_fetch` runs the same language-chain lookup as `get_addons()`, then classifies the outcome as OK, INVALID_URL, NETWORK_ERROR (`urlopen` raised `URLError`), HTTP_ERROR (`HTTPError`, with `http_code` populated), JSON_ERROR (body not valid JSON / not an array), EMPTY_LISTING, NO_COMPATIBLE_ADDONS (no entry targets the running major.minor), or LANG_FALLBACK_EN (requested language unavailable, English fallback used).
- `get_addons()`, `get_all_addons()`, `available_updates()`, `load_addon_file()` are unchanged — the classifier is purely additive. Adds `gramps/gen/plug/test/__init__.py` (new, empty) marker. The GTK-side wiring (status icon / tooltip per Projects-tab row in `gramps/gui/plug/_windows.py`) is deliberately deferred to a follow-up PR.

## Verification
- **Claim:** each project-fetch outcome is classified into a distinct `ProjectFetchStatus` (populating `http_code` on HTTP errors), with URL normalization (whitespace / trailing slash / scheme) applied first.
- **Checked:** `gramps/gen/plug/utils.py` on master (the branch this PR targets) — additive classifier; existing fetch functions unchanged.
- **Test:** `gramps/gen/plug/test/_utils_project_status_test.py` (new) — 20 tests: 8 `NormalizeProjectUrlTest` (whitespace / trailing-slash stripping, scheme validation, non-string / empty rejection) and 12 `ClassifyProjectFetchTest` covering every `ProjectFetchStatus` value plus the "trailing slash is normalized before the listing URL is built" invariant, with all network access mocked via `urlopen_maybe_no_check_cert`. `mypy` and `black --check` clean; full suite 32,519 tests, only the two pre-existing unrelated `test_WebCal` / `test_navwebpage` failures. This PR is backend-only groundwork for the indicator; the user-facing UI is a follow-up, so it does not close the bug.

Bug [#13069](https://gramps-project.org/bugs/view.php?id=13069)

### DIFF
--- a/gramps/gen/plug/test/_utils_project_status_test.py
+++ b/gramps/gen/plug/test/_utils_project_status_test.py
@@ -1,0 +1,283 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Eduard Ralph
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""
+Tests for ``normalize_project_url`` and ``classify_project_fetch``
+(bug #13069 — Addon Manager "project" path validity indicator).
+"""
+
+# ------------------------
+# Python modules
+# ------------------------
+import io
+import json
+import unittest
+from email.message import Message
+from unittest import mock
+from urllib.error import HTTPError, URLError
+
+# ------------------------
+# Gramps modules
+# ------------------------
+from .. import utils as plug_utils
+from ..utils import (
+    ProjectFetchResult,
+    ProjectFetchStatus,
+    classify_project_fetch,
+    normalize_project_url,
+)
+
+
+# ------------------------------------------------------------
+#
+# _FakeResponse
+#
+# ------------------------------------------------------------
+class _FakeResponse:
+    """Minimal stand-in for the object returned by :func:`urlopen`."""
+
+    def __init__(self, payload: bytes, code: int = 200, is_file: bool = False) -> None:
+        self._buffer = io.BytesIO(payload)
+        self._code = code
+        self.file = self if is_file else None
+
+    def read(self, size: int = -1) -> bytes:
+        return self._buffer.read(size)
+
+    def getcode(self) -> int:
+        return self._code
+
+    def close(self) -> None:
+        self._buffer.close()
+
+
+def _json_response(records: object, code: int = 200) -> _FakeResponse:
+    return _FakeResponse(json.dumps(records).encode("utf-8"), code=code)
+
+
+# ------------------------------------------------------------
+#
+# NormalizeProjectUrlTest
+#
+# ------------------------------------------------------------
+class NormalizeProjectUrlTest(unittest.TestCase):
+    """Unit tests for :func:`normalize_project_url`."""
+
+    def test_strips_surrounding_whitespace(self) -> None:
+        self.assertEqual(
+            normalize_project_url("  https://example.com/path  "),
+            "https://example.com/path",
+        )
+
+    def test_strips_trailing_slashes(self) -> None:
+        self.assertEqual(
+            normalize_project_url("https://example.com/path///"),
+            "https://example.com/path",
+        )
+
+    def test_keeps_scheme_double_slash(self) -> None:
+        self.assertEqual(normalize_project_url("file://"), "file://")
+
+    def test_http_scheme_accepted(self) -> None:
+        self.assertEqual(
+            normalize_project_url("http://example.com"), "http://example.com"
+        )
+
+    def test_file_scheme_accepted(self) -> None:
+        self.assertEqual(
+            normalize_project_url("file:///tmp/addons/"), "file:///tmp/addons"
+        )
+
+    def test_empty_string_rejected(self) -> None:
+        self.assertIsNone(normalize_project_url(""))
+        self.assertIsNone(normalize_project_url("   "))
+
+    def test_unsupported_scheme_rejected(self) -> None:
+        self.assertIsNone(normalize_project_url("ftp://example.com"))
+        self.assertIsNone(normalize_project_url("gopher://example.com"))
+        self.assertIsNone(normalize_project_url("example.com"))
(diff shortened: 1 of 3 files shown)


Nick-Hall:
See also the related PR #2211.

emyoulation:
Thanks for looking at the Addon Manager's project-paths. 

Please **_do_** take a moment to review PR #2211 (currently in draft).

The feature is functionally complete and _works_ with 5.2 or 6.0, but the UI/UX aspects will likely need refinement before it could be considered for inclusion in Gramps. It would benefit from a contributor willing to help sponsor the design direction and act as a change agent.

The implementation was initially generated with Claude AI, based on a prompt designed to minimize unsolicited network activity. In particular, validation is handled within the Edit Project dialog, so there is no external traffic unless explicitly triggered by the user. 

image

There are also broader reuse possibilities here: several modules could benefit from shared URI validation (local, LAN, or internet). It may be worth considering whether parts of this should be promoted into the Gramps Engine, with separate GTK handling for interface and feedback.

This design choice was intentional. For comparison, the current Addon Manager fetches JSON data during Gramps startup, which may give some users the impression of background network activity occurring without clear user initiation. The draft PR aims to make such interactions more transparent and explicitly user-driven.

eduralph:
Force-pushed (269bc4b718 → 4af01f0052) for two AGENTS.md conformance fixes, no behavior change:

Added the class navigation header blocks above ProjectFetchStatus and ProjectFetchResult in gramps/gen/plug/utils.py
Amended the commit body: replaced "Part of Feature #13069." with "Issue #13069." so the Mantis BT integration picks up the reference — "Feature" isn't one of the parsed keywords, but "Issue" is

eduralph:
@emyoulation , I took a look and you are right, there is some overlap. I would suggest to land #2247 first, then rebase #2211 to call classify_project_fetch inside _test_url — removes ~200 lines of duplication, keeps #2211's (ie your) GUI work (dialog, test button, lozenge toggles).

Alternatively we can merge the two PRs into one stack.

What do you think?

emyoulation:
I am good with whichever way you choose to proceed.  

Your PR is likely to be approved, mine was just shared for reference or inspiration.  I needed the features to ease my testing of 3rd party curated Addon collections. And it knocked down the barrier-to-entry in using Addon Manager Projects enormously. So much so that I thought others would want it.

But my coding skills are very outdated: Pascal, Fortran, Basic, script kiddie. Rusty enough that I do not proofread Claude code well enough to comment with confidence. So being the lead submitting my 'wish-coding' for a core code change is not happening.  

<hr />

In another vein, I have already noticed some shortfalls in my URL/URI browsing validation. 

Users can understand how to Browse and select the desired JSON file.  But they do not understand how to covert the fully qualified JSON filepath (or the `listings` folderpath) to a valid `file:///` or `https://raw.githubcontent.com` interpretation. (Which includes the need to trim back to the `listings` folder's parent folder. And whether that means including the trailing slash or not.)  The idea that the path is to a folder _containing_ the various localized JSONs rather than the JSON _itself_ is too abstract for most users. 

Then there is a possibility that some users will host their curated collections outside of the GitHub file management server. So, the `raw` idiosyncrasy would go away.

eduralph:
Sounds good. I'm happy to do the integration work on your PR, assuming my PR does get pushed.

emyoulation:
The path to integration is unclear. How can we get this back in motion?

Claude and I have continued to work on the replacement `_windows.py` file for experimental use. The original base was the v5.2 but this latest version has all the master branch commits. (It is growing bloated.)

Furthermore, this version has:
* new _default_ filter level for the **Addons** tab to find **New or Updated** addons. (Needs a lozenge for this filter level. Also needs a special case "No addons found" row with lozenges for accelerated clearing of filters) 
* a status bar readout that lists the number of matching addons and the total number of addons in the active Projects. (e.g., "4 of 156 plugins")
* a manual override to the Gramps locale for each Edit Projects (not persistent between sessions, cleared on next Gramps restart)
* eliminate the suppression of installing due to Gramps version mismatch
* supports both the "PyPl wheel installer" (preferred) with fallback pip3 to allow use with Gramps 5.2+ <br />(This needs a diagnostic logging for when Fallback is used. This would allow field validation of [v6.1 PR#2308 PyPI wheel installer](https://github.com/gramps-project/gramps/pull/2308). But should be eliminated for 6.2 to reduce bloat.)
* validation diagnostics for macOS SSL problem
* reduced load when Servers or net connection unavailable. 
* color coding of Version lozenge if "Update" would be to a lesser version tuple. With button relabeled "Downgrade" instead of Update
* hotlinking in <kbd>Test Location</kbd> of the **Edit Projects** for quick manual inspection of the curated "Project" repository, `listings` subfolder, and the selected `addons-xx.json`  
* now will parse a full path to a fully qualified GitHub.com `.json` filepath to a standardized "Project" URL. (e.g., `https://github.com/Taapeli/isotammi-addons/blob/master/addons/gramps61/listings/addons-fi.json` becomes `https://raw.githubusercontent.com/Taapeli/isotammi-addons/master/addons/gramps61`) 

[_windows.py](https://github.com/user-attachments/files/30281560/_windows.py)

[gramps.css](https://github.com/user-attachments/files/30281739/gramps.css)

image

image

PR #2248, opened 2026-04-18
Follow symlinks when scanning plugin directories (#10436)

eduralph:
## Summary

Fixes Mantis [bug #10436](https://gramps-project.org/bugs/view.php?id=10436) (open since 2018-02-16, tagged PATCH).

`BasePluginManager.reg_plugins` walked plugin directories with `os.walk(..., topdown=True)`, leaving `followlinks` at the default of `False`. Any plugin directory reached via a symlink was silently skipped. Users who wanted to organize their plugin library by symlinking individual plugins into the Gramps plugin path had to work around this with a separate loader addon (the `pluginloader.py` community addon by kku does exactly this).

This PR flips `followlinks=True` on the walk and adds a small guard to prevent infinite recursion if a symlink points back at an ancestor.

## Change

`gramps/gen/plug/_manager.py`:

- Pass `followlinks=True` to `os.walk()`.
- Track the resolved path (`os.path.realpath`) of each visited directory within the walk and prune recursion if we re-enter a directory we've already seen. This terminates on pathological symlink loops without relying on the process being killed.

## Tests

New test file `gramps/gen/plug/test/_manager_symlinks_test.py` with two cases:

1. `test_symlinked_subdir_is_scanned` — creates a plugin directory that is only reachable via a symlink and asserts `scan_dir` is invoked against it.
2. `test_self_referential_symlink_loop_terminates` — creates a self-referential symlink loop and asserts `reg_plugins` terminates with a bounded number of `scan_dir` calls.

Full test suite: 32499 baseline + 2 new = 32501 tests, pass (the two pre-existing `test_WebCal` / `test_navwebpage` failures on `master` are unrelated to this change).

## Policy note — unconditional vs. dev-mode gated

In the 2018 discussion on #10436, admin Sam888 suggested that symlink-following might only be enabled in developer mode. The reporter (Pander) pushed back, observing that Gramps runs as the user and a symlink inside a plugin directory cannot expose anything the user does not already have access to — there is no security delta.

I agree with the reporter and have implemented this unconditionally:

- **No security delta.** Symlinks honor the user's own filesystem permissions.
- **Dev-mode gating would defeat the request.** The bug was filed by an end user who wants to organize their own plugins by symlinking. Hiding the feature behind developer mode leaves that exact user unable to use it.
- **The only real hazard — symlink loops — is handled in code** via the `realpath` dedup, not by a flag.

Happy to move this behind a config option if reviewers prefer; that would be a small conditional around the walk call.

### DIFF
--- a/gramps/gen/plug/test/_manager_symlinks_test.py
+++ b/gramps/gen/plug/test/_manager_symlinks_test.py
@@ -1,0 +1,95 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Eduard Ralph
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""Tests that plugin scanning follows symlinks and survives symlink loops."""
+
+# ------------------------
+# Python modules
+# ------------------------
+import os
+import tempfile
+import unittest
+from unittest.mock import patch
+
+# ------------------------
+# Gramps modules
+# ------------------------
+from .._manager import BasePluginManager
+from gramps.gen.constfunc import win
+
+
+# ------------------------------------------------------------
+#
+# RegPluginsSymlinkTest
+#
+# ------------------------------------------------------------
+@unittest.skipIf(win(), "Symlinks not supported or tested by default on Windows")
+class RegPluginsSymlinkTest(unittest.TestCase):
+    """Exercise the os.walk(followlinks=True) behavior in reg_plugins."""
+
+    def setUp(self) -> None:
+        self.mgr = BasePluginManager.get_instance()
+        self.pgr = self.mgr._BasePluginManager__pgr  # type: ignore[attr-defined]
+
+    def test_symlinked_subdir_is_scanned(self) -> None:
+        """A directory reached only via a symlink must be scanned."""
+        with tempfile.TemporaryDirectory() as tmp:
+            target = os.path.join(tmp, "target_plugin")
+            os.makedirs(target)
+            # A marker filename so we can confirm scan_dir saw target's contents.
+            open(os.path.join(target, "marker.gpr.py"), "w").close()
+
+            plugin_root = os.path.join(tmp, "plugins")
+            os.makedirs(plugin_root)
+            link = os.path.join(plugin_root, "linked_plugin")
+            os.symlink(target, link, target_is_directory=True)
+
+            with patch.object(self.pgr, "scan_dir") as mock_scan:
+                self.mgr.reg_plugins(plugin_root)
+
+            visited = [call.args[0] for call in mock_scan.call_args_list]
+            self.assertIn(link, visited)
+            # scan_dir must have received the marker filename for the link.
+            link_call = next(
+                call for call in mock_scan.call_args_list if call.args[0] == link
+            )
+            self.assertIn("marker.gpr.py", link_call.args[1])
+
+    def test_self_referential_symlink_loop_terminates(self) -> None:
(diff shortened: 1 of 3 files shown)


kulath:
Good. Another vote to keep it unconditional (not hide it in dev-mode)

Nick-Hall:
I remember the previous discussions and agree that symlinks should be followed.

eduralph:
Force-pushed to bring the new test file into line with AGENTS.md and the style used on the other plugin-test additions (7cd9f5d145 → 6e05b08f13):

Swapped the short-form GPL header for the full FSF-address form in AGENTS.md, with author attribution
Added the three-section import grouping comment headers (Python modules / Gramps modules)
Added the class navigation header block above RegPluginsSymlinkTest
No test logic changed.

stevenyoungs:
Was this tested on Windows as well? If not, I can do so.
symlinks behave differently on Windows, so I'm interested to see what the test does.
The Windows equivalent of a symlink is a junction. Gramps already correctly followed junctions.

stevenyoungs:
~~Is there any value in testing the `followlinks=False` behaviour as well?~~
Ignore that, now I've read the actual code change!

stevenyoungs:
@eduralph unfortunately the test fails on Windows
```bash
$ LANGUAGE= LANG=en_US.utf-8 GRAMPS_RESOURCES=. python -m unittest gramps.gen.plug.test._manager_symlinks_test.py
E
ERROR: test_self_referential_symlink_loop_terminates (gramps.gen.plug.test._manager_symlinks_test.RegPluginsSymlinkTest.test_self_referential_symlink_loop_terminates)
A symlink pointing at an ancestor must not cause infinite recursion.
  File "C:/msys64/home/steve/gramps2/gramps/gen/plug/test/_manager_symlinks_test.py", line 60, in test_symlinked_subdir_is_scanned
    os.symlink(target, link, target_is_directory=True)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [WinError 1314] A required privilege is not held by the client: 'C:/msys64/tmp/tmp18vxnfrh/target_plugin' -> 'C:/msys64/tmp/tmp18vxnfrh/plugins/linked_plugin'

Ran 2 tests in 0.010s

FAILED (errors=2)
```

symlink creation on Windows require additional privileges which are not granted by default to a normal user. See [Create symbolic links](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/security-policy-settings/create-symbolic-links). I'll propose skipping this test on Windows as I don't think it offers any significant value.

Tested on `maintenance/gramps61` with this branch merged in.

gramps/gen/plug/test/_manager_symlinks_test.py
```diff
@@ -0,0 +1,93 @@
...
+# ------------------------
+import os
+import tempfile
+import unittest
+from unittest.mock import patch
+
+# ------------------------
+# Gramps modules
+# ------------------------
+from .._manager import BasePluginManager
```
stevenyoungs, line 34:
```suggestion
from .._manager import BasePluginManager
from gramps.gen.constfunc import win
```

gramps/gen/plug/test/_manager_symlinks_test.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+# ------------------------------------------------------------
+#
+# RegPluginsSymlinkTest
+#
+# ------------------------------------------------------------
+class RegPluginsSymlinkTest(unittest.TestCase):
```
stevenyoungs, line 44:
```suggestion
@unittest.skipIf(win(), "Symlinks not supported or tested by default on Windows")
class RegPluginsSymlinkTest(unittest.TestCase):
```

Whilst it would be possible to improve this and only skip if
1. the underlying filesystem does not support symbolic links
or
2. the current process token does not have SE_CREATE_SYMBOLIC_LINK_NAME privilege

python code for number 2 [here](https://stackoverflow.com/a/2129589) [which I've not tested]

It's not, quite honestly, worth it, in my view.

stevenyoungs (changes requested):

gramps/gen/plug/test/_manager_symlinks_test.py
eduralph: Fortgot to report back: implemented in PR head 0a792062d3

gramps/gen/plug/test/_manager_symlinks_test.py
eduralph: Same here, forgot to report back - implemented

dsblank:
Paraphrasing @Nick-Hall:

"If a developer targets the current maintenance branch with a bugfix then it is considered for fairly quick release"

eduralph:
@Nick-Hall - rebased it to gramps61 as asked in #2275 
@stevenyoungs  - do you want to review the changes I made in d9aedd6?

PR #2250, opened 2026-04-18
Allow requires_mod entries to specify pip package name and version

dsblank:
## Summary

- `requires_mod` entries (and addon `rm` list entries) can now be either a plain string (existing behaviour, fully backwards compatible) or a `(import_name, pip_spec)` tuple, e.g. `("gramps_gedcom7", "gramps-gedcom7>=1.0")`.
- `check_mod()` uses the import name for `find_spec()`.
- `install()` returns the pip spec so the correct package name and optional version constraint are passed to pip.
- `info()` displays `import_name (pip_spec)` when the two differ.
- 17 unit tests added in `gramps/gen/utils/test/requirements_test.py`.

## Test plan

- [ ] Run `GRAMPS_RESOURCES=. python3 -m unittest gramps.gen.utils.test.requirements_test` — all 17 tests pass.
- [ ] Run `mypy gramps/gen/utils/requirements.py` — no errors.
- [ ] Verify existing plugins that use plain-string `requires_mod` entries are unaffected.
- [ ] Verify a plugin using the new tuple form (e.g. `requires_mod = [("gramps_gedcom7", "gramps-gedcom7>=1.0")]`) is correctly checked and installed.

### DIFF
--- a/gramps/gen/plug/_pluginreg.py
+++ b/gramps/gen/plug/_pluginreg.py
@@ -299,7 +299,10 @@ class PluginData:
        The URL where documentation for the plugin can be found
     .. attribute:: requires_mod
        A list of required modules that should be importable using the python
-       `import` statement.
+       `import` statement.  Each entry may be either a plain string (the
+       import name) or a ``(import_name, pip_spec)`` tuple where *pip_spec*
+       is the pip package name with an optional version constraint, e.g.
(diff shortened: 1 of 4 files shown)


Nick-Hall:
@dsblank Please don't create branches in _origin_.  Use you own fork when creating pull requests.

PR #744, opened 2018-12-20
Update debian files after 5.0.1 release

RossGammon:
Debian copyright file updated & Debian changelog entry for the 5.0.1 release.

PR #2252, opened 2026-04-18
Add sniff_function support to ImportPlugin for version-aware dispatch

dsblank:
## Summary

- Adds an optional `sniff_function` attribute to `PluginData` and `ImportPlugin` so multiple importers can share a file extension and be selected by file content.
- Updates both GUI (`dbloader.py`) and CLI (`arghandler.py`) dispatch to prefer a sniff-matched plugin over the first-match fallback.
- Adds 10 unit tests covering all dispatch paths.

## Motivation

The `.ged` extension is currently mapped to a single GEDCOM importer.
The [GEDCOM 7 addon (PR #744)](https://github.com/gramps-project/addons-source/pull/744) needs to register a second importer for `.ged` files and have it chosen automatically when the file header identifies the file as GEDCOM 7.  GEDCOM 7 requires a version string in the header, so detection is trivial — the only missing piece was a hook in core.

## How it works

A plugin declares a `sniff_function` by name in its `.gpr.py`:

```python
plg.sniff_function = "sniffData"
```

The function accepts a filename and returns `True` when the plugin should handle it:

```python
def sniffData(filename):
    with open(filename, "rb") as f:
        for line in itertools.islice(f, 20):
            if b"VERS 7." in line:
                return True
    return False
```

Plugins without a `sniff_function` continue to work exactly as before.

## Test plan

- [x] `GRAMPS_RESOURCES=. python3 -m unittest discover -p "*_test.py" gramps/gen/plug/test` — all 10 tests pass
- [x] `black` run on all changed files — no reformatting needed
- [ ] Manual smoke test: import a GEDCOM 5.5 file and verify the built-in importer is still selected
- [ ] Manual smoke test: install GEDCOM 7 addon and verify a GEDCOM 7 file routes to it

### DIFF
--- a/gramps/cli/arghandler.py
+++ b/gramps/cli/arghandler.py
@@ -147,6 +147,43 @@ def _split_options(options_str):
     return options_str_dict
 
 
+def _choose_import_plugin(plugins, extension, filename):
+    """
+    Return the best import plugin for *filename* among *plugins*.
+
+    Plugins that declare a ``sniff_function`` are tested first; the first one
+    whose sniff function returns ``True`` for *filename* is returned.  If none
+    matches, the first plugin whose extension equals *extension* and that has
+    no sniff function is returned as a fallback.  Returns ``None`` when no
+    suitable plugin is found.
+
+    This allows multiple importers to share the same file extension while each
+    handling a distinct file version (e.g. GEDCOM 5.5 vs GEDCOM 7.0).
+
+    :param plugins: list of :class:`~gramps.gen.plug.ImportPlugin` instances
+    :type plugins: list
+    :param extension: lower-case file extension without leading dot
+    :type extension: str
+    :param filename: full path of the file to import
+    :type filename: str
+    :returns: the chosen plugin, or ``None``
+    :rtype: :class:`~gramps.gen.plug.ImportPlugin` | None
+    """
+    candidates = [p for p in plugins if extension == p.get_extension()]
+    for plugin in candidates:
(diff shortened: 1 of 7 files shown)


kulath:
It would be really nice if the Import code were converted to support GEDCOM extension addons as [#0010550: Refactor Gramps GEDCOM import so as to support GEDCOM extension addons](https://gramps-project.org/bugs/view.php?id=10550) at the same time as this change, so that it is simpler to write alternative importers.

emyoulation:
I've had updated vCard importer and export plugins wish-coded that add support beyond the original v3.0 format for 5.2 and 6.0 Gramps. (It seems most smart phones still use version 2.1 .vcf files.)  Do you want to use them to exercise the sniffer? 

I'd like to roll them into 6.1 but testing as addons might tell us something about integrating. The AI tried building automatic hiding of the superseded into the .gpr.py file.  But that did not work.

https://github.com/emyoulation/CuratedGrampsPlugins/tree/main/gramps60/download

* `.vcf` format: Enhanced version Supports vCard versions 1, 2.1, 3.0, and 4.0. Unhandled properties are preserved as Person attributes. ( .vcf Variant Call Format (DNA) files are detected and rejected.) 
* new `.webp` thumbnailer 
* future `.avif` thumbnailer
* future `.vcf` that supports more fields. (Such as embedded avatar bitmaps.)  
* future `.ged` 5.5.1 that does not mangle unhandled GEDCOM tags as formatted Notes. Preserve them as raw Attributes for easier post-processing and dialect export.

PR #2259, opened 2026-04-20
Fix mypy errors on windows

stevenyoungs:
Some python module attributes are not available on all platforms. 
If an attribute is not available, mypy reports an error. For example, On Windows
```bash
gramps/gen/datehandler/_grampslocale.py:91: error: Module has no attribute "nl_langinfo"  [attr-defined]
```

This PR fixes mypy's `attr-defined` error for
- [`locale.nl_langinfo`](https://docs.python.org/3/library/locale.html#locale.nl_langinfo)
- [`signal.SIGCHLD`](https://docs.python.org/3/library/signal.html#signal.SIGCHLD)

neither of which is available on Windows

Two approaches were considered and I settled on the second
1. Add `# type: ignore[attr-defined]` to relevant lines
    there are many lines to change and it clutters the code
    it suppresses the error for all platforms
2. protect relevant code with `sys.platform` checks
    fewer lines to change
    recognised by mypy - but not if Gramps `win()` function is used

### DIFF
--- a/gramps/gen/datehandler/_grampslocale.py
+++ b/gramps/gen/datehandler/_grampslocale.py
@@ -48,7 +48,7 @@ codeset = glocale.encoding  # TODO I don't think "codeset" is used anymore
 
 try:
     if sys.platform == "win32":
-        raise locale.Error # local.nl_langinfo is not supported on Windows
+        raise locale.Error  # local.nl_langinfo is not supported on Windows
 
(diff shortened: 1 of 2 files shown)


stevenyoungs:
Variants such as
`if win():`
or
`if sys.platform in LINUX`
were not detected by mypy and the `attr-defined` error was still emitted

PR #2262, opened 2026-04-24
Fix AIO workflow for Gramps 6.1

stevenyoungs:
The gramps 6.1 workflow is failing to build, with errors compiling orjson 3.11.8. See below for selected log messages.
Fix this by pinning orjson to version 3.11.7 (for Windows AIO builds only)

Update the AIO unit test workflows so that the unit test suite
- is only run if the python venv is available
- activates the venv before running the tests

```
Building wheels for collected packages: orjson, pydotplus, pygraphviz
  Building wheel for orjson (pyproject.toml): started
  Building wheel for orjson (pyproject.toml): finished with status 'error'
  error: subprocess-exited-with-error
  
  Building wheel for orjson (pyproject.toml) did not run successfully.
  exit code: 1
```

```
For more information about this error, try `rustc --explain E0554`.
  error: could not compile `orjson` (lib) due to 1 previous error
  💥 maturin failed
    Caused by: Failed to build a native library through cargo
    Caused by: Cargo build finished with "exit code: 101": `"cargo" "rustc" "--profile" "release" "--message-format" "json-render-diagnostics" "--manifest-path" "D:\\a\\_temp\\msys64\\tmp\\pip-install-q7sxd6tt\\orjson_7d41824dfa354d46a6bb6b4952d6757b\\Cargo.toml" "--lib"`
  Error: command ['maturin', 'pep517', 'build-wheel', '-i', 'D:/a/_temp/msys64/tmp/grampspythonenv/bin/python.exe', '--compatibility', 'off'] returned non-zero exit status 1
  [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for orjson
```

### DIFF
--- a/.github/workflows/windows-aio.yml
+++ b/.github/workflows/windows-aio.yml
@@ -50,8 +50,10 @@ jobs:
         cd aio
         ./build.sh ${{ inputs.cleanup }} ${{ inputs.build-number }}
     - name: Run unit test suite
+      if: ${{ ! inputs.cleanup }} # can only run unit tests if python venv was preserved
       run: |
         export GRAMPS_RESOURCES=build/share
+        source /tmp/grampspythonenv/bin/activate
         python3 -m unittest discover -p "*_test.py"
     - uses: actions/upload-artifact@v6
       with:
--- a/aio/build.sh
+++ b/aio/build.sh
@@ -80,7 +80,7 @@ source $pythonvenv/bin/activate
 
 ## prerequisites in pip packages
 python -m pip install --upgrade pip
-pip install --upgrade orjson pydot pydotplus pygraphviz requests selenium
+pip install --upgrade orjson==3.11.7 pydot pydotplus pygraphviz requests selenium
 
 ## download dictionaries
 mkdir -p /ucrt64/share/enchant/hunspell


stevenyoungs:
I cannot explain why orjson 3.11.8 fails in CI when it works correctly on my local PC.
I can only assume that the difference in Windows platform (2025 server vs 11 Pro) or MSYS configuration causes an unknown difference. Ostensibly, both systems have a fully updated MSYS environment.

.github/workflows/windows-aio.yml
```diff
@@ -50,8 +50,10 @@ jobs:
         cd aio
         ./build.sh ${{ inputs.cleanup }} ${{ inputs.build-number }}
     - name: Run unit test suite
+      if: ${{ ! inputs.cleanup }} # can only run unit tests if python venv was preserved
```
hgohel, line 53: In a recent build UTs were being run despite the build step having failed. Now that the UT step occurs first, test for success of the build step (EXE creation) should be done earlier.

hgohel:
> I cannot explain why orjson 3.11.8 fails in CI when it works correctly on my local PC.

With python packages involved, a debugging option would be to ignore cached packages on your system. That may make the two builds more similar and expose the problem locally.

If it were an MSYS package issue, I would have suggested doing a diff on the package versions local v/s CI.

Nick-Hall:
Let me know when this is ready to merge.

hgohel:
Don't let my comments hold up this PR if it fixes the build; they can be addressed separately.

jralls:
> I cannot explain why orjson 3.11.8 fails in CI when it works correctly on my local PC.

I encountered the orjson build failure yesterday due to a Rust error, `feature` (used [here](https://github.com/ijl/orjson/blob/5cbb3d0398a2f42de51210270286fecd798c5d78/src/lib.rs#L4) and a few places after) is allowed only on Rust's "nightly build" channel and not on production channels.

.github/workflows/windows-aio.yml
stevenyoungs: Ah, I see what you mean. `build.sh` returns 0 (unless the new smoke test fails). That's not nice.
Then yes, I agree we should elevate the `if-no-files-found` check into the `Run unit test suite` step. Really we want to fix `build.sh` to correctly return a non-zero exit code on error. Then we don't need the check in the workflow at all.

stevenyoungs: Let's defer this for a later PR. I think it is more valuable to get an AIO build out and hopefully get some feedback on the UCRT64 change.

stevenyoungs:
> Let me know when this is ready to merge.

@Nick-Hall please merge when you can and run an AIO build. I get a good build on my own github org so 🤞 
We need to remember to unpin orjson at some point.

.github/workflows/windows-aio.yml
stevenyoungs: See #2263

Nick-Hall:
The build ran successfully in the _aio-6.1.0-beta1_ branch, and I have uploaded the AIO.

hgohel (approved):
Good to merge.

Nick-Hall:
OK. I'll review and merge tomorrow.

PR #2263, opened 2026-04-24
Trap AIO Build Errors

stevenyoungs:
Use a trap to catch all non-zero exit codes during execution of `aio\build.sh`. 

If an error is trapped:
- log information to help diagnose the problem
- exit the script with the non-zero exit code

The non-zero exit code will cause subsequent workflow steps to not run

Example workflow output using this PR and orjson 3.11.8, which at the time of writing, fails to compile on Window MSYS UCRT64
image

### DIFF
--- a/aio/build.sh
+++ b/aio/build.sh
@@ -23,6 +23,24 @@
 #   clean-up     : [true|false]. clean the python venv on completion
 #   build-number : the build number to use when DEV_VERSION is not True
 #
+handle_error() {
+    # Get information about the error
+    local error_code=$?
+    local error_line=$BASH_LINENO
+    local error_command=$BASH_COMMAND
+
+    # Log the error details
+    echo "$BASH_ARGV0: Error on line $error_line"
+    echo "    command: $error_command"
+    echo "    exit code: $error_code"
+
+    # exit with the same error code
+    exit $error_code
+}
+
+# trap any non-zero exit code from commands and handle
+trap 'handle_error' ERR
+
 # install prerequisites
 ## prerequisites in msys packages
 pacman -S --needed --noconfirm \


stevenyoungs:
I'm slightly concerned that this will partially nullify the recently introduced "pip smoke test" as pip will now exit with a non-zero exit code.
https://github.com/gramps-project/gramps/blob/3fa5aead234e2f74ec9b17d02cfa9d857477c3d2/aio/build.sh#L195-L202

hgohel:
After reflecting on the recent addition of unit testing during the installer build step, I still feel like it's not the right time or place to run them for two reasons:
1. The tests have already run when the code was committed, so any failures were already caught
2. This step is building the installer so smoke tests are more appropriate. The unit tests simply add to run time.
The reason I'm bringing this up again here is that this pr was necessitated by ut being run when the build failed. While trapping errors in the script is a good idea, we ought to remove the extra UT run.

stevenyoungs:
> 1. The tests have already run when the code was committed, so any failures were already caught

We know the tests ran on Linux. This ensures they also run on Windows. As we saw during Gramps 6.0 with bsddb support, different platforms may behave differently.
Ideally we'd run the unit tests on each supported platform for each workflow run.

hgohel:
Thanks for reminding me that that the PR-triggered UTs run only on Linux, not Windows (MSYS2 environment) or MacOS (as far as I can see).

> Ideally we'd run the unit tests on each supported platform for each workflow run.

Agree with you, and when that's implemented it can be removed from the installer generation step.

stevenyoungs:
> Thanks for reminding me that that the PR-triggered UTs run only on Linux, not Windows (MSYS2 environment) or MacOS (as far as I can see).
> 
> > Ideally we'd run the unit tests on each supported platform for each workflow run.
> 
> Agree with you, and when that's implemented it can be removed from the installer generation step.

Great. We are both aligned.

PR #2264, opened 2026-04-25
Ensure date columns are sorted consistently

iand:
The original report noted that date columns sort without grouping dates with qualifiers like about, after, before so they appear in an arbitrary order.

This happens because get_sort_value() returns only the Julian Day Number with no modifier or quality information.

This change adds Date.get_sort_key() which creates a sort key based on the JDN adjusted with scores for modifiers and quality

It also updates date sort columns in the list view treemodels (peoplemodel, familymodel, eventmodel, mediamodel, citationbasemodel) to call get_sort_key() instead of get_sort_value().

The key is 'JDN * 12 + modifier_score * 3 + quality_score'

Modifier scores:

  0 - MOD_BEFORE
  1 - MOD_NONE, MOD_FROM, MOD_TO, MOD_SPAN, MOD_RANGE
  2 - MOD_ABOUT
  3 - MOD_AFTER

Quality scores:

  0 - QUAL_NONE
  1 - QUAL_ESTIMATED
  2 - QUAL_CALCULATED

For year-only after-dates the position advances to the first JDN of the following year, which is correct for all supported calendars. This ensures 'after 1823' sorts after '1823-12-31'.

The max sort key for 31 Dec 9999 is 64481819 (JDN 5373484 * 12) which fits in the format string "%09d" used by treemodels.

get_sort_value() is unchanged and continues to be used for date arithmetic, calendar conversion and span calculations throughout the rest of the codebase.

Fixes [#7761](https://gramps-project.org/bugs/view.php?id=7761).

### DIFF
--- a/gramps/gen/lib/date.py
+++ b/gramps/gen/lib/date.py
@@ -1211,6 +1211,69 @@ class Date(BaseObject):
         """
         return self.sortval
 
+    def get_sort_key(self):
+        """
+        Return an integer sort key that incorporates modifier and quality.
+
+        Unlike :meth:`get_sort_value`, which returns the raw Julian Day Number
+        used for date arithmetic, this value is adjusted so that dates with
+        different modifiers and qualities group correctly in list view columns.
+
+        The return value is ``pos * 12 + modifier_score * 3 + quality_score``.
+
+        Modifier scores:
+
+          0 - MOD_BEFORE
+          1 - MOD_NONE, MOD_FROM, MOD_TO, MOD_SPAN, MOD_RANGE
+          2 - MOD_ABOUT
+          3 - MOD_AFTER
+
+        Quality scores:
+
+          0 - QUAL_NONE
+          1 - QUAL_ESTIMATED
+          2 - QUAL_CALCULATED
+
+        Modifier and quality scores combine independently, so "estimated about
+        1900" (modifier score 2, quality score 1) sorts between plain "about 1900"
+        (2, 0) and "calculated about 1900" (2, 2).
+
+        For year-only MOD_AFTER dates the position advances to the first JDN of
+        the following year so that "after 1900" sorts after every exact date
+        within 1900.  All other modifiers use sortval as the position.
+
+        Use this method when building sort strings for list views.
+        """
+        if self.sortval == 0:
+            return 0
+
+        if self.modifier == Date.MOD_BEFORE:
+            mod_score = 0
+        elif self.modifier == Date.MOD_ABOUT:
+            mod_score = 2
+        elif self.modifier == Date.MOD_AFTER:
+            mod_score = 3
+        else:
+            mod_score = 1
+
+        if self.quality == Date.QUAL_ESTIMATED:
+            qual_score = 1
+        elif self.quality == Date.QUAL_CALCULATED:
+            qual_score = 2
+        else:
+            qual_score = 0
+
+        pos = self.sortval
+        if self.modifier == Date.MOD_AFTER:
+            yr = self.dateval[Date._POS_YR]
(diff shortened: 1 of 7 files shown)


iand:
This is my first contribution. I picked a problem that's been bugging me for a while and was marked confirmed and major with high priority.

dsblank (approved):
*Review by Claude (claude-sonnet-4-6), posted on behalf of @dsblank.*

## PR #2264 — "Ensure date columns are sorted consistently"

First PR — and a very strong one. The code is clean, the approach is sound, and it includes more test coverage than many experienced contributors write.

---

### What It Does

Introduces `Date.get_sort_key()` — a new method that produces a composite integer sort key (`pos * 12 + modifier_score * 3 + quality_score`) so that dates with qualifiers like *before*, *about*, and *after* sort in a semantically meaningful order in all list-view columns. The five treemodels (`peoplemodel`, `familymodel`, `eventmodel`, `mediamodel`, `citationbasemodel`) are updated to call `get_sort_key()` instead of `get_sort_value()`. The raw `get_sort_value()` is untouched, preserving all existing date arithmetic.

---

### Strengths

- **Clean separation of concerns.** Adding a new method rather than changing `get_sort_value()` is exactly the right call — arithmetic users are unaffected.
- **Well-documented formula.** The PR description and the docstring both spell out the scoring table clearly.
- **Good test coverage.** 9 tests covering ordering, all 12 modifier×quality combinations, span/range equality, the year-only `MOD_AFTER` edge case, the empty-date case, and a regression guard confirming `get_sort_value()` is not affected. All pass.
- **Max value verified.** The max sort key (64,481,819 for Dec 31 9999 + max scores) fits within the existing `"%09d"` format string used by treemodels.

---

### Issues & Suggestions

**1. `MOD_TEXTONLY` is silently correct but untested.**

Text-only dates have `sortval == 0`, so the early-return `if self.sortval == 0: return 0` handles them correctly. But this is implicit — a future reader won't know why. Consider either a comment or a test:

```python
def test_text_only_date_returns_zero(self):
    d = Date()
    d.set_as_text("circa 1900")
    self.assertEqual(d.get_sort_key(), 0)
```

**2. `MOD_FROM` and `MOD_TO` are not tested.**

These two modifiers fall into the `else` branch (score 1, same as `MOD_NONE`). The test `test_span_range_same_modifier_score_as_none` covers `MOD_SPAN` and `MOD_RANGE` but not `MOD_FROM`/`MOD_TO`. Worth adding for completeness.

**3. Month-precision `MOD_AFTER` dates are not advanced.**

The year-only advancement (`day == 0 and mon == 0`) correctly handles "after 1902". But for a month-precision date like "after Jun 1902" (`day == 0, mon == 6`), `pos` stays at the JDN for June 1 — meaning it sorts at the *start* of June, not after the *end*. The docstring is honest that only year-only dates are advanced, but it might be worth a note on the month case or a follow-up issue, since the semantic imprecision could surprise users.

**4. Minor: `yr + 1` when year is at calendar maximum.**

`Date._calendar_convert[self.calendar](yr + 1, 1, 1)` is called for year-only `MOD_AFTER` dates. If `yr` is the maximum supported year for a given calendar, `yr + 1` could produce an unexpected JDN. This is an extremely unlikely edge case in practice, but a boundary check or a comment would be defensive.

---

### Summary

This is a well-executed fix for a real UX problem. The algorithm is correct, the implementation is minimal and focused, and the tests are solid. The issues above are minor — items 1 and 2 are easy wins for test coverage before merge, item 3 is documented already and reasonable to defer, and item 4 is very low risk.

Excellent first contribution — the code is better than many PRs from regular contributors.

Nick-Hall:
Rebased.

PR #2266, opened 2026-04-26
Ensure family order is unaffected by family edits

iand:
The original report stated that making edits to a family such as adding a child would abitrarily re-order family order in the relationship view. This is problematic because family order is significant in many reports.

In the transaction to save the family edits the comparison between the original and new family lists was using identity comparison which resulted in all the families being removed and then re-added, losing any ordering.

This change fixes the problem by using handles to compare the two lists. In the case of the reported issue this results in no remove/add cycle being made since the family handles are unchanged.

Fixes [#13819](https://gramps-project.org/bugs/view.php?id=13819).

### DIFF
--- a/gramps/gui/editors/editfamily.py
+++ b/gramps/gui/editors/editfamily.py
@@ -1315,18 +1315,18 @@ class EditFamily(EditPrimary):
                     original.get_mother_handle(), self.obj.get_mother_handle(), trans
                 )
 
-                orig_set = set(original.get_child_ref_list())
-                new_set = set(self.obj.get_child_ref_list())
+                orig_handles = set(ref.ref for ref in original.get_child_ref_list())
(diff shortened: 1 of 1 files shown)


dsblank (approved):
Looks good to me! Thanks, and welcome to Gramps dev!

stevenyoungs (approved):
Looks good. A nice change that will prevent accidental and subtle changes that users may not immediately spot.

Nit-picking, there's a typo in the commit message
abitrarily -> arbitrarily

stevenyoungs:
FWIW claude proposed the following unittest
`gramps/gui/editors/test/editfamily_test.py`
[editfamily_test.py](https://github.com/user-attachments/files/27106295/editfamily_test.py)

I've not tested it.

iand:
Updated the typo in the commit message :)

Nick-Hall:
Rebased.

PR #2267, opened 2026-04-26
Add 'New' button to Place, Source, and Citation selector dialogs

dsblank:
## Motivation

The "Select Place", "Select Source", and "Select Source or Citation" dialogs are
unfriendly when the object a user needs does not yet exist.  The current
workflow forces them to:

1. Open the selector and search — find nothing.
2. Cancel the selector.
3. Navigate elsewhere to create the object.
4. Return and re-open the selector.

This is particularly disruptive because users often do not know in advance
whether an object exists; they open the selector to find out.  The fix is
to let users create the missing object without ever leaving the selector.

## Design (Phase 1 of planned improvement)

`BaseSelector` gains an optional **New** button (hidden by default, shown
only by subclasses that opt in).  Clicking it opens the appropriate editor
as a child window.  When the editor is saved, a callback rebuilds the
selector tree and pre-selects the newly created object, so the user can
confirm with **OK** in one uninterrupted flow.  If the editor is cancelled,
the selector simply stays open.

Here's where the New button will appear:                                                                                              
   
**Select Place dialog opens from:**                                                                                                       
  1. Event editor — the place field (via PlaceEntry in objectentries.py). When you click the select-place button next to the place field on an event.                                                                                                                         
  2. Place editor → Enclosed By tab — selecting a parent place in the place hierarchy (placerefembedlist.py).
  3. PlaceEntry widget (placeentry.py) — any inline place entry field across the app (address tabs, etc.).                              
  4. Report/filter options (_guioptions.py) — place picker in report dialogs.                                                           
                                                                                                                                        
**Select Source or Citation dialog opens from:**                                                                                          
  1. Any editor with a Citations tab — the "Share" (link) button in the citations embed list (citationembedlist.py). This appears on Person, Family, Event, Place, Media, and Repository editors.                                                                          
                                                            
**Select Source dialog opens from:**                                                                                                      
  1. Source entry fields (objectentries.py) — less common, used in specific contexts where only a source (not a citation) is being linked.                                                                                                                               
         
In all cases the New button appears at the bottom-left of the dialog alongside the existing Help button, leaving Cancel and OK on the right unchanged.           

Screenshot from 2026-04-26 07-23-28

image

For the citation selector, a small prompt first asks whether the user wants
a new **Source** or a new **Citation** (since citations are always children
of a source).

A later phase will add comma-separated place string parsing (e.g.
"Paris, France") directly in the Place selector and in inline place
entry fields.

## Summary of changes

- Adds a hidden `newbutton` to `baseselector.glade` and wires it up in `BaseSelector` via `enable_new_button()` and `cb_new_button_clicked()`.
- `SelectPlace` and `SelectSource` enable the button and open `EditPlace` / `EditSource` on click; on save the tree is rebuilt and the new object is pre-selected.
- `SelectCitation` prompts the user to choose between a new Source or new Citation before opening the appropriate editor.
- All new methods follow AGENTS.md conventions: `cb_` callback prefix, type hints, Sphinx docstrings, import section headers.
- Tests added in `gramps/gui/selectors/test/baseselector_test.py`; they skip gracefully when the GTK3 widget stack is unavailable (GTK4 environment).

## Test plan

- [ ] Open an event editor and click the place selector — confirm a **New** button appears at the bottom left.
- [ ] Click **New**, fill in a place name, save — confirm the selector tree refreshes and the new place is highlighted.
- [ ] Click **OK** — confirm the new place is applied to the event.
- [ ] Repeat for the Source and Citation selectors.
- [ ] In the Citation selector, click **New** and choose **Cancel** in the prompt — confirm the selector stays open unchanged.
- [ ] Open a second selector while an editor is already open — confirm `WindowActiveError` is swallowed silently.
- [ ] Run `GRAMPS_RESOURCES=. python3 -m unittest gramps.gui.selectors.test.baseselector_test -v` — all 19 tests should show as skipped (not failed) on GTK4.

### DIFF
--- a/gramps/gui/glade/baseselector.glade
+++ b/gramps/gui/glade/baseselector.glade
@@ -66,6 +66,23 @@
                 <property name="secondary">True</property>
               </packing>
             </child>
+            <child>
+              <object class="GtkButton" id="newbutton">
+                <property name="label" translatable="yes">_New</property>
+                <property name="visible">False</property>
+                <property name="no-show-all">True</property>
+                <property name="can-focus">True</property>
+                <property name="can-default">True</property>
+                <property name="receives-default">True</property>
+                <property name="use-underline">True</property>
+              </object>
+              <packing>
+                <property name="expand">False</property>
+                <property name="fill">False</property>
+                <property name="position">3</property>
+                <property name="secondary">True</property>
+              </packing>
+            </child>
           </object>
           <packing>
             <property name="expand">False</property>
--- a/gramps/gui/selectors/baseselector.py
+++ b/gramps/gui/selectors/baseselector.py
@@ -125,6 +125,8 @@ class BaseSelector(ManagedWindow):
         self.define_help_button(
             self.glade.get_object("help"), self.WIKI_HELP_PAGE, self.WIKI_HELP_SEC
         )
+        self.new_button = self.glade.get_object("newbutton")
+        self.new_button.connect("clicked", self.cb_new_button_clicked)
 
         # connect to signal for custom interactive-search
         self.searchbox = InteractiveSearchBox(self.tree)
@@ -261,6 +263,24 @@ class BaseSelector(ManagedWindow):
 
     def _on_row_activated(self, treeview, path, view_col):
         self.window.response(Gtk.ResponseType.OK)
+
+    def enable_new_button(self) -> None:
+        """
+        Show the 'New' button. Call from ``_local_init()`` in subclasses
+        that want to offer in-place object creation.
+        """
+        self.new_button.show()
+
+    def cb_new_button_clicked(self, obj: Gtk.Button) -> None:
+        """
+        Handle a click on the 'New' button.
+
+        The default implementation is a no-op.  Subclasses should override
+        this to open the appropriate editor for creating a new object.
+
+        :param obj: The 'New' button that was clicked.
+        :type obj: Gtk.Button
+        """
 
     def _local_init(self):
         # define selector-specific init routine
--- a/gramps/gui/selectors/baseselector.py
+++ b/gramps/gui/selectors/baseselector.py
@@ -45,6 +45,7 @@ from ..glade import Glade
 from ..widgets.interactivesearchbox import InteractiveSearchBox
 from ..display import display_help
 from gramps.gen.const import URL_MANUAL_PAGE
+from gramps.gen.dbstate import DbState
 from gramps.gen.filters import GenericFilter
 from gramps.gui.widgets.persistenttreeview import PersistentTreeView
 from gramps.gui.widgets.multitreeview import MultiTreeView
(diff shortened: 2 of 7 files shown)


stevenyoungs:
@dsblank for symmetry, will you extend this concept to SelectPerson e.g. when adding a child to a Family it would allow the new child Person to be created.

stevenyoungs:
If `allow_multiple_selection` is enabled, I _think_ the better user experience is to _add_ the newly created object to the existing selection (but still scroll so that the newly created object is visible) rather than replace the existing selection with the new object.
#2028 introduces the concept of multiple selection for additional object types.

emyoulation:
Wonderful 'Add' ... in multiple ways! Thank you

Inserting an "Add" button to the Object Selector to eliminate retracing the Share drill-down has been a long-standing request. It may resolve some issues MantisBT. (There may be duplicate requests. But only found 1 on a quick search using a smart phone. Will search again when have a desktop available.)

Item 1.2 of :
https://gramps-project.org/bugs/view.php?id=8698

(This particular decade-old MantisBT issue 8698 has multiple UX requests. That makes it difficult to resolve.)

A question... hierarchical Object Selectors can make for ambiguous indicating of an Insert point for a new object. When you click an Add button, does it add at the top of Object hierarchy? 

When there's a selected object in the dialog, there's a focus. But would the insertion be as a child of the selected object? Or as a sibling? What if nothing is selected?

e.g., when trying to 'share' an existing "county" place, the user drills-down through Continent, Reion, Nation, State and discovers the county does not exist. Do they select the parent State to insert the new County under the State? Or select one of the Sister counties to add at their level? (And sister counties might not exist yet?) And if the user used the search instead of a browse, a focus in the place hierarchy might not be selected yet.

I look forward to exploring it!

kulath:
Just a minor point on such a nice change.

It would be nice to move the glocale import to be last, because then the global variable 
`_ = glocale.translation.sgettext`
 
could be placed after the imports and still keep the global ind import together.


I appreciate that this was how it was originally, but the imports made me notice this point, and it is in PEP8, which seems to be referenced from our guidelines.

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

dsblank:
Great question — here is how it works (by claude):

**No hierarchy is implied by the current selection.** Clicking *New* opens a blank `EditPlace` editor. The new place has no enclosing place pre-set, so the user fills in all the details — including the *Enclosed by* field — inside the editor itself. That is the canonical place for building the place hierarchy, so nothing ambiguous happens at the selector level.

Once the editor is saved, the selector tree is rebuilt and the view scrolls to the newly created place, wherever it ended up in the hierarchy.

**Practical answer to your county example:** the user would click *New*, type the county name, then set *Enclosed by → State* inside the editor, and save. They are dropped straight back into the selector with the new county selected, ready to confirm with OK.

There is no "insert as child of selected" or "insert as sibling" concept — the editor always gives the user full control over placement, which avoids the ambiguity you described.

dsblank:
@emyoulation see answers above.

PR #2269, opened 2026-04-26
Add Quick Add bar to Select Place dialog for comma-separated place hierarchy entry

dsblank:
## Summary

Adds a **Quick Add** bar to the Select Place selector dialog, allowing
users to type a comma-separated place hierarchy (e.g. ``Springfield,
Illinois, USA``) and create the missing places without leaving the
selector.

Screenshot from 2026-04-26 08-11-42

Screenshot from 2026-04-26 08-14-43

### New files

- ``gramps/gui/selectors/quickplaceparser.py`` — pure-logic module
  (no GTK) with three public functions:
  - ``parse_place_hierarchy(db, text)`` resolves existing places and
    marks new ones, honouring parent-child relationships.
  - ``commit_place_hierarchy(db, parsed, transaction)`` creates new
    Place objects, wires ``PlaceRef`` links, and returns the leaf handle.
  - ``_find_place_child(db, name, parent_handle)`` case-insensitive
    lookup with optional parent constraint.

- ``gramps/gui/selectors/quickplacedialog.py`` — ``QuickPlaceDialog``
  (``Gtk.Dialog`` subclass) that shows the parsed rows in a table:
  existing rows are read-only, new rows have a PlaceType combo.
  On OK the chosen ``PlaceType`` is written back into each entry dict.

- ``gramps/gui/selectors/test/quickplaceparser_test.py`` — 20 unit
  tests covering happy path, edge cases, and error cases for all three
  parser functions.

### Modified files

- ``gramps/gui/selectors/selectplace.py`` — adds ``_build_quick_add_bar()``
  and ``cb_quick_add()``; the bar is packed below the place tree.
- ``po/POTFILES.in`` — adds ``quickplacedialog.py``.
- ``po/POTFILES.skip`` — adds ``quickplaceparser.py`` and the test file.

## User-visible change

The Select Place dialog gains a "Quick Add:" row at the bottom with a
text entry (placeholder *City, Region, Country*) and a **Quick Add**
button.  Pressing Enter or clicking the button:

1. Resolves any tokens that already exist in the database.
2. Shows a confirmation table where the user assigns a PlaceType to
   each new place.
3. Commits all new places in a single undo-able transaction.
4. Rebuilds the tree and scrolls to the newly created leaf place,
   ready to select with OK.

## Test plan

- [ ] Run unit tests: ``GRAMPS_RESOURCES=. python3 -m unittest discover -p "*_test.py"``
- [ ] Open the Select Place dialog; verify the Quick Add bar appears at the bottom.
- [ ] Type ``Springfield, Illinois, USA`` where none exist; confirm all three are created with correct parent links.
- [ ] Repeat with a hierarchy where some levels already exist; confirm existing rows show as "Existing" and no duplicate is created.
- [ ] Cancel the confirmation dialog; confirm no places are added.
- [ ] Press Enter in the entry field; confirm it triggers the same flow as the button.
- [ ] After a successful quick add, confirm the tree scrolls to the new leaf and it is pre-selected.


gramps/gui/selectors/quickplacedialog.py
```diff
@@ -0,0 +1,239 @@
...
+    :param parsed: List of entry dicts from
+        :func:`~gramps.gui.selectors.quickplaceparser.parse_place_hierarchy`.
+    :type parsed: list[dict]
+    :param parent: Optional transient parent window.
+    :type parent: Gtk.Window | None
+    """
+
+    def __init__(
+        self,
+        db: "DbGeneric",
```
stevenyoungs, line 103:
```suggestion
        db: DbGeneric,
```
My understanding is that the general advice is to not quote type names by default (you'll need the relevant import though)

> Whether quotes are needed depends entirely on Python version and context:
> No quotes needed (the normal case) in Python 3.10+ with `from __future__ import annotations` at the top of the file, or in Python 3.12+ unconditionally. All annotations are treated as strings lazily, so you write them plainly
> 
> Quotes are required only for forward references — when a type is referenced before it's defined in the same file, and you're not using from `__future__ import annotations`
> 
> The short rule: if you find yourself adding quotes around a type name, ask whether `from __future__ import annotations` or a Python version upgrade would remove the need. Nine times out of ten, it would, and that's the cleaner solution.

source: Claude

Nick-Hall: For v6.2, I think that we can require python 3.11+, but not 3.12+.

I agree that the `__future__` import is neater than quoting.

kulath:
As well as allowing the specification of a type, would it be possible to say that a particular level should be omitted?

PR #2270, opened 2026-04-26
Fix calls to get_relationship_calculator

stevenyoungs:
`get_relationship_calculator` is defined as
`def get_relationship_calculator(reinit=False, clocale=glocale):`

Adjust the code so that `glocale` is not passed as the first argument, avoiding unnecessary reinitialisation.
As the default value for `clocale` is `glocale`, simply remove the argument


dsblank (approved):
LGTM!

Nick-Hall:
Rebased.

PR #2271, opened 2026-04-26
Allow current release of GExiv2

jralls:
GExiv2's import is currently locked at version 0.10 from 2019. While I can't say that it's completely unavailable, it certainly isn't widely available anymore: The current release is 0.16.0 and most distributions (including UCRT64) [provide either that or 0.14.x.](https://repology.org/project/gexiv2/versions).

### DIFF
--- a/gramps/grampsapp.py
+++ b/gramps/grampsapp.py
@@ -356,8 +356,9 @@ def show_settings():
         import gi
 
         repository = gi.Repository.get_default()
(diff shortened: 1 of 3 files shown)


jralls:
[Companion addons-source PR](https://github.com/gramps-project/addons-source/pull/829) for EditExifMetadata and Prerequisites Checker.

PR #2272, opened 2026-04-27
Re-enable personal map tile provider in geography view

iand:
Restores support for a user-supplied tile URL in the embedded geography view. The custom URL is entered in Geography preferences using #Z/#X/#Y as coordinate placeholders. Example providers include MapTiler NLS historical Ordnance Survey maps (uk-osgb1888, uk-osgb63k1885, etc.).

OsmGpsMap requires a non-null map-source enum value to honour a custom repo_uri; using map_source=0 (NULL) causes it to render gray tiles and ignore the URL entirely. Fix this by passing OPENSTREETMAP_RENDERER, whose built-in URI is NULL (service defunct), so OsmGpsMap keeps the supplied repo_uri and fetches tiles from it.

Also fixes the image-format lookup for custom URLs, debounces the url entry field so it only fires on Enter or focus-out rather than on every keystroke, and widens the URL input field to span the full grid width.

### DIFF
--- a/gramps/plugins/lib/maps/constants.py
+++ b/gramps/plugins/lib/maps/constants.py
@@ -62,7 +62,7 @@ VIRTUAL_EARTH_HYBRID = 13
 YAHOO_STREET = 14
 YAHOO_SATELLITE = 15
 YAHOO_HYBRID = 16
-# PERSONAL = 30
+PERSONAL = 30
 
 TILES_PATH = {
     OPENSTREETMAP: "openstreetmap",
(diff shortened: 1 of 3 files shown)


iand:
The National Library of Scotland has quite a few maps of the UK and Ireland available via maptiler that work well with this change. These maps require an API key from map tiler, so the user would have to obtain that for themselves.

If this gets accepted then I can look at the wiki documentation for the re-enabled feature before next release.

Nick-Hall:
@SNoiraud What is your opinion on this PR?

SNoiraud:
I removed this feature when API keys were required on some sites.
I don’t like it. But people are happy to share all their data with Google or others.
So no problem for me.

Nick-Hall:
Rebased.

PR #2273, opened 2026-04-27
Persist clipboard across sessions; filter view by active database

dsblank:
## Summary

- Clipboard items are saved to `~/.local/share/gramps/gramps60/clipboard.json` on every change and reloaded on the next startup, so the clipboard survives application restart.
- Items from other databases are preserved in the backing store. When a different database is opened the view filters to show only its own items; switching back reveals the originals.
- The clipboard is never cleared automatically when switching databases.

## Details

- **`_build_drag_type_map()`** — maps drag-type strings to wrapper classes for serialisation.
- **`ClipboardWindow.save_clipboard()`** — serialises each row using the original raw bytes (`_pickle` / `_obj`), bypassing `pack()`/`is_valid()`, so cross-database items are stored faithfully. Triggered by `row-inserted` / `row-deleted` signals.
- **`ClipboardWindow.load_clipboard()`** — restores items on startup. `refresh()` is suppressed during wrapper construction to avoid `HandleError` for handles that belong to another database; `_title`, `_value`, `_dbid`, and `_dbname` are restored from JSON afterwards.
- **`Gtk.TreeModelFilter`** wraps the backing `ClipboardListModel` so only current-database rows are displayed. `refilter()` is called in `database_changed()` for an instant icon update when switching databases.
- **`refresh_objects()`** no longer removes rows; it only refreshes the display of items that are valid in the current database.
- Tests in `gramps/gui/test/clipboard_test.py` cover save, load, round-trip, error cases, and the drag-type map (17 tests).

## Test plan

- [ ] Add items from two different databases to the clipboard.
- [ ] Close and reopen Gramps; confirm both sets of items are present.
- [ ] Switch databases; confirm only the active database's items are shown, and the others reappear when switching back.
- [ ] Delete an object that has a clipboard entry; confirm the entry is removed.
- [ ] Run `python -m pytest gramps/gui/test/clipboard_test.py` — all 17 tests pass.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/gramps/gui/clipboard.py
+++ b/gramps/gui/clipboard.py
@@ -84,7 +84,7 @@ CLIPBOARD_FILE = os.path.join(USER_DATA_VERSION, "clipboard.json")
 # -------------------------------------------------------------------------
 
 theme = Gtk.IconTheme.get_default()
-LINK_PIC = theme.load_icon("stock_link", 16, 0)
+LINK_PIC = theme.load_icon("stock_link", 16, 0) if theme else None
 OBJ2ICON = {
     "media": "gramps-media",
     "note": "gramps-notes",
@@ -110,8 +110,9 @@ def obj2icon(target):
 
 
 ICONS = {}
-for name, icon in OBJ2ICON.items():
-    ICONS[name] = theme.load_icon(icon, 16, 0)
+if theme:
+    for name, icon in OBJ2ICON.items():
+        ICONS[name] = theme.load_icon(icon, 16, 0)
 
 
 # -------------------------------------------------------------------------
--- a/gramps/gui/clipboard.py
+++ b/gramps/gui/clipboard.py
@@ -323,7 +323,7 @@ class ClipObjWrapper(ClipWrapper):
 class ClipAddress(ClipObjWrapper):
     DROP_TARGETS = [DdTargets.ADDRESS]
     DRAG_TARGET = DdTargets.ADDRESS
-    ICON = ICONS["address"]
+    ICON = ICONS.get("address")
 
     def __init__(self, obj):
         super(ClipAddress, self).__init__(obj)
@@ -344,7 +344,7 @@ class ClipAddress(ClipObjWrapper):
 class ClipLocation(ClipObjWrapper):
     DROP_TARGETS = [DdTargets.LOCATION]
     DRAG_TARGET = DdTargets.LOCATION
-    ICON = ICONS["location"]
+    ICON = ICONS.get("location")
 
     def __init__(self, obj):
         super(ClipLocation, self).__init__(obj)
@@ -362,7 +362,7 @@ class ClipLocation(ClipObjWrapper):
 class ClipEvent(ClipHandleWrapper):
     DROP_TARGETS = [DdTargets.EVENT]
     DRAG_TARGET = DdTargets.EVENT
-    ICON = ICONS["event"]
+    ICON = ICONS.get("event")
 
     def __init__(self, obj):
         super(ClipEvent, self).__init__(obj)
@@ -381,7 +381,7 @@ class ClipEvent(ClipHandleWrapper):
 class ClipPlace(ClipHandleWrapper):
     DROP_TARGETS = [DdTargets.PLACE_LINK]
     DRAG_TARGET = DdTargets.PLACE_LINK
-    ICON = ICONS["place"]
+    ICON = ICONS.get("place")
 
     def __init__(self, obj):
         super(ClipPlace, self).__init__(obj)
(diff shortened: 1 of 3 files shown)


emyoulation:
Is this a necessary change? Or does it layer in uncessary complexity?

There is already an addon (Collections Clipboard Gramplet) that provides persistent clipboards for each category.

And the recovery is simple ... if something weird clutters the clipboard caching.

dsblank:
> Is this a necessary change? Or does it layer in uncessary complexity?

Always a good question to ask. But if you are following the thread:

https://gramps.discourse.group/t/what-do-did-you-find-to-be-the-most-complicated-parts-of-gramps/9392

you see that it was requested. This PR:

1. Persists the clipboard between uses. Saves a lot of clicks!
2. Doesn't show items from other databases (that would clutter up the view quickly)

So, for all intents and purposes, there is no difference in use, except for the requested feature.

> There is already an addon (Collections Clipboard Gramplet) that provides persistent clipboards for each category.

Yes, I wrote that 16 years ago. It serves a different purpose. But the persistence was shown to be useful. The Clipboard and ClipboardGramplet continue to function independently.

> And the recovery is simple ... if something weird clutters the clipboard caching.

I don't understand your point.

Anyway, my goal is to remove all of the annoyances one at a time.

emyoulation:
I love both the Clipboard and Collections Clipboard... and rely on them heavily. (For Drag'n'drop, setting active objects that are other than people, creating filters from selected items.)  

But they are _always_ getting in the way and have to be constantly shuffled. When I'm using them for Gramps, I often need the clipboards to be on the second screen to keep from obscuring stuff in the main view. But when using a digital Reference document, I need the second screen for that. Or when comparing the drill-down data of objects (people or families typically), the 2nd screen space is needed for the object editor being referenced. So then the clipboards need to be minimized or dragged downwards to show only a few lines.

So maybe we can explore outside of the box. Gramps has hideable sidebars and bottombars for the main category views. Maybe this interface feature could be extended to _**all**_ the Object Editor dialogs? Add a standard slideout clipboard sidebar with multiple tabs? So the clipboard is there (and near) when and where you need it. And not cluttering things up otherwise.

(My android smartphone has slideouts on its small screen. There is a vertical semitransparent gadget that can be clicked or dragged to show occasionally used pinned apps. Maybe we could use that GUI indicator to make them more intuitive and conspicuous than the splitbar options in the View menus checkboxes? If there is room to the right virtual screen edge, it could slide `outward`. If there it is blocked to the right, the slider goes 'inward' like the view Sidebar does now.) 

The initial version could have 2 tabs on the clipboard slideout: the current clipboard and a `Recent` objects tab. (A future version might filter in a  context sensitive manner to the input focus : filtering the Clipboard list to the object types appropriate to the focus. Which would open the interface to tabbing through the Object Editor fields and double-clicking to apply the clipboarded object -- instead of a fussy drag'n'drop. That would be more touch interface friendly too.)

Another future context-sensitive enhancement could be a Proximity tab.  So it shows all the secondary objects of a type that are Referenced by the active person (or the family members of families where the Active Person is a spouse or child).  Sorting of hierarchical listed objects would be a hassle. I often do not recall which township has a cemetery or which first/call/nick name goes with a particular maiden name.

emyoulation:
While the Clipboard is being reviewed, could you add a Context menu for Ref objects? 

One of the reasons people may not understand about Ref items with the Link icon is that the simple lack of a Context menu deprives the user of hints. They see no action list about what can be done with that type of object. It makes "Ref" clipboard rows look like glitches rather than the added way to copy the "Reference information" in addition to the "Shared information."

The menu items themselves give hint about the functionality of that Type of Clipboard object. If the WikiContributors could add an anchor for the Ref clipboard objects, maybe the context menu could have a "Ref object help" linked to the wiki?

The main wiki anchor for explaining the Clipboard is not the right location. It is in the Navigation chapter... strange in itself but particularly problematic for Ref object because there is NO Navigation functionality.
https://gramps-project.org/wiki/index.php/Gramps_6.0_Wiki_Manual_-_Navigation#Using_the_Clipboard

The Glossary entry for "clipboard"  guidelines. Instead of a brief definition, it tries (and falls short) to be comprehensive. It strays into usage/introduction/tutorial information. It is SUPPOSED to be a gross summary and link to such material in the wiki chapters. (I will shuffle that content.) 
https://gramps-project.org/wiki/index.php/Gramps_Glossary#clipboard

There are "reference" and "references tab Glossary entries too. But no good anchor for this purpose.
https://gramps-project.org/wiki/index.php/Gramps_Glossary#r

But a really useful Context menu item that would be one that clipboards the "reference information" object. That would give unambiguous confirmation that the Ref (Shared information) is distinct from the "Reference information". And that a separate clipboard type was intentional.

dsblank:
@emyoulation you have a few ideas mentioned above that are outside the scope of this PR. Can you create a Github discussion for any items that require decisions to be made, or file a feature request otherwise.

IstvanZachar:
Thanks a lot Douglas!

PR #2281, opened 2026-05-01
Change error message in post-crash Error on restart Dialog

emyoulation:
The dialog title "Error Processing Arguments" (CLI-oriented dialog) when restarting Gramps after a crash is related to a locked Tree.
image

Suggest patch "Cannot process as requested" until the branching can changed to where the user can immediately address the 'lock' problem. 

See:
Bug 0012640: Add better message when “Error message on Locked Database is not actionable”

Ideal fix would be to show the GUI dialog instead of the CLI dialog. Which selects the locked file row in the Manage Family Trees dialog with the dialog that actually allows the problem to be resolved:
https://gramps.discourse.group/t/error-parsing-arguments-dialog/9378
image

### DIFF
--- a/gramps/cli/arghandler.py
+++ b/gramps/cli/arghandler.py
@@ -177,6 +177,7 @@ class ArgHandler:
         self.open_gui = parser.open_gui
         self.imp_db_path = None
         self.dbman = CLIDbManager(self.dbstate)
+        self.locked = False
         self.force_unlock = parser.force_unlock
         self.cl_bool = False
         self.imports = []
@@ -225,7 +226,8 @@ class ArgHandler:
             # We have a potential database path.
             # Check if it is good.
             if not self.check_db(db_path, self.force_unlock):
-                sys.exit(1)
+                if not (self.gui and self.locked):
+                    sys.exit(1)
             if create:
                 self.__error(
                     _(
@@ -571,6 +573,7 @@ class ArgHandler:
         if force_unlock:
             self.dbman.break_lock(dbpath)
         if self.dbman.is_locked(dbpath):
+            self.locked = True
             self.__error(
                 (_("Database is locked, cannot open it!") + "\n" + _("  Info: %s"))
                 % find_locker_name(dbpath)
--- a/gramps/gui/grampsgui.py
+++ b/gramps/gui/grampsgui.py
@@ -632,7 +632,7 @@ class Gramps:
         if hasattr(self, "_vm"):
             if hasattr(self._vm, "window"):
                 parent = self._vm.window
-        ErrorDialog(_("Error parsing arguments"), string, parent=parent)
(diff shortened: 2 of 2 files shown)


emyoulation:
This hack just alleviates some user complaints about a confusing dialog message. (I can hack a string, but changing which GUI dialogs appears on an error is beyond my skillset.)

Nick-Hall:
> Ideal fix would be to show the GUI dialog instead of the CLI dialog.

This is a GUI dialog used to display errors from the argument handler.  If running from the command line the errors will be printed to _stderr_.

emyoulation:
Ok.  Then I am baffled about how this would be corrected the right way.  If the messaging presented for 'Locked file on restart' could be rerouted, then this dialog's text need not be changed.  Perhaps the method that "Remember last Family Tree" uses is built on a batch file?

A locked file on reload is something that MUST be resolved.  And the only way to resolve it is with the Manage Family Tree dialog. (By breaking the lock and loading the tree, choosing to load another tree, or creating a new tree to load.) 

Opening to a dialog basically says 'Nope. Not gonna do it.'  (with only a 'close' option) just frustrates the user. It not only adds several extraneous clicks between them and resolving the problem, it also expect the user (who just suffered a crash) to be experienced enough to know that the next step is to open the Manage Family Tree, browse to find the locked file, and double-click that row to show the option to break the lock. And either way                                                expects them to realize when breaking the lock is OK. That's a lot to expect from 'Aunt Martha'.

Nick-Hall:
I've added a commit which opens the Family Tree Manager after the error messages have been displayed.

emyoulation:
Thank you! I was too uncertain to try anything like that myself. It looked like that dialog was also used for other failures than Locked Trees. So I was too likely to break something vital elsewhere.

Hopefully your change will help the (significant) percentage of users who experience an abnormal exit.

Thanks again, sincerely.

samusz:
I would add to this issue that it would be nice if the error message stated the (full) path of the database and lock file, if possible. 
Best

kulath:
> I've added a commit which opens the Family Tree Manager after the error messages have been displayed.

Surely switching from CLI to GUI is not a good idea. There must be many scenarios where that is undesirable for the user. I use the CLI in a script that opens a tree and then in more of the script does more things. Suddenly opening the GUI would be very confusing. 

If the user is knowledgeable enough to use the CLI he is likely to be able to figure out an error message on the console or wherever. 

Isn’t there a force option on the CLI that will open the database anyway? The error message should be enhanced to mention that. (Then I could add an optional force parameter to my script to resolve the issue if I am sure I haven’t got the tree open elsewhere - it is actually quite likely I haven’t got the the tree open already and am trying to do my script processing after having just done an edit to the tree, and have accidentally left it open, i.e. two copies of Gramps open at the same time) in which case it is really confusing, does the error pop up relate to my GUI or to the CLI script.

PR #2282, opened 2026-05-02
Add tests for referential integrity of deletes and fix 3 bugs

iand:
This is based on issue [#9895](https://gramps-project.org/bugs/view.php?id=9895) which reported that not all references are removed when some objects are deleted. This issue is nine years old so probably some of these situations have been fixed in the meantime.

dsblank suggested writing some tests to uncover any remaining bugs and that's what this change does. It adds a range of tests for get_referenced_handles_recursively() and remove_handle_references() and in the process found three bugs which I have tried to fix:

Bug 1: Family.get_referenced_handles_recursively() double-counts child person handles.

Family.get_referenced_handles() explicitly included ("Person", ref.ref) for each entry in child_ref_list. But child_ref_list is also returned by Family.get_handle_referents(), and each
ChildRef.get_referenced_handles() independently returns ("Person", self.ref). The recursive traversal therefore sees each child person handle twice. Fixed by removing [ref.ref for ref in self.child_ref_list] from Family.get_referenced_handles().

Bug 2: Place.remove_handle_references("Place", ...) silently did nothing.

_remove_handle_references was never implemented on Place so calles fall through to the base class no-op, leaving stale PlaceRef entries behind when a place is deleted. Fixed by implementing remove_handle_references

Bug 3: get_referenced_handles_recursively() returns duplicate handles when the same handle is referenced at multiple levels

The accumulation in BaseObject.get_referenced_handles_recursively() is a plain list append with no deduplication. If the same handle appeared at both the top-level object (e.g. person.note_list) and in a child secondary object (e.g. EventRef.note_list), or in two sibling secondary objects, it was emitted once per occurrence. The database backend wraps the result in set() so it was unaffected, but the function's contract implies uniqueness.

Fixed by replacing the list accumulation in
BaseObject.get_referenced_handles_recursively() with a seen set that filters duplicates as they are collected, preserving insertion order.


iand:
The dedup fix in `baseobj.py` makes `get_referenced_handles_recursively()` return each (classname, handle) tuple at most once per object traversal but the `test_tcg_and_check_and_repair test` had baked in numbers including duplicates.

Nick-Hall:
@iand Are you _aviansid_ on the bug tracker?

iand:
@Nick-Hall yes, that's me

kulath:
Does the Check and Repair tool detect and fix these incorrect references? If not a check should probably be added.

iand:
I'm not very familiar with the Check and Repair tool's operation so I asked Claude to help

> There are two separate things to address:
> 
> Stale PlaceRef entries (Place._remove_handle_references): check_place_references() (line 1553 of check.py) does scan for exactly this - PlaceRef entries whose target handle doesn't exist in the database. But the remediation is different from removal: it calls make_unknown() to create a stub "unknown" place and keeps the dangling reference pointing at it. So Check and Repair detects the symptom but can't reverse it to the intended state (no reference). Your fix prevents the stale reference being created in the first place when a place is deleted through normal UI operations.
> 
> Duplicate handles in get_referenced_handles_recursively(): Check and Repair calls this method on every primary object throughout its checks (lines 1381, 1698, 1712, 1726, ...). Before the fix it was receiving duplicate handles, which inflated its "N citations/sources were referenced but not found" counts.
> 
> So: for the PlaceRef bug, Check and Repair detects the damage but heals it differently (stub object vs. clean removal). For the duplicate handle bug, Check and Repair was a victim of the bug, not a detector of it.

kulath:
I'm not sure I understand Claude's comments or if I am just confused. I am concerned whether C&D detects and fixes the problems caused by the bugs you have identified BEFORE THEY WERE FIXED.

Claude seems to be saying that C&R fixes problems in Places (albeit by creating new stub objects - which is safer because C&R doesn't know whether the problem was caused by a delete that didn't delete everything, or by some other bug where something that was supposed to be there has got lost).

However, Claude doesn't seem to be saying that problems in other objects where there is a dangling reference get fixed, Claude is just talking about bugs in that case - not actually saying that the dangling reference is actually detected and fixed.

iand:
My understanding is that, as you point out,  C&D works around the problems in places by creating stubs. My fix prevents this from being necessary.

The problem with duplicates in get_referenced_handles_recursively doesn't really affect C&D's operation. My fix is just tightening up the correctness of the function.

Nick-Hall:
The new unit test file needs to be listed in `po/POTFIELS.skip`.

dsblank:
> The new unit test file needs to be listed in po/POTFIELS.skip.

If you are using an AI agent, then this is mentioned in AGENTS.md.

Nick-Hall:
Rebased.

PR #2283, opened 2026-05-03
Web Report - improvements to dates in family tables on the individual page

tepreece:
This PR makes two small adjustments to the way dates are displayed in the family tables on the individual page of the Narrated Web Site report.

1. In the spouse and children table, fallback birth dates are now included for children that don't have a Birth event.
2. In all family tables (parents and siblings / spouse and children), fallback birth and death dates are rendered in italics.


Nick-Hall:
@SNoiraud Did you notice this?

PR #2289, opened 2026-05-05
Specify the required Gtk version

stevenyoungs:
This prevents a runtime warning
`/gramps/gui/plug/_guioptions.py:44: PyGIWarning: Gtk was imported without specifying a version first. Use gi.require_version('Gtk', '3.0') before import to ensure that the right version gets loaded.`

when running the unittest
`python -m unittest gramps.plugins.export.test.exportvcard_test.VCardCheck`

### DIFF
--- a/gramps/gui/__init__.py
+++ b/gramps/gui/__init__.py
@@ -21,6 +21,10 @@
 Package init for the gui package.
 """
 
+import gi
+
+gi.require_version("Gtk", "3.0")
+
 # DO NOT IMPORT METHODS/CLASSES FROM src/gui HERE ! Only __all__
 
 __all__ = ["editors", "filters", "logger", "merge", "selectors", "views", "widgets"]


hgohel (commented):
OK for the change. I wonder if this could be done at a high level/early enough in the UT execution so that no matter which set of UTs are run, the require statement would be executed early on?

stevenyoungs:
I assumed that this was core Gramps that was being changed. The VCardTest unit test invokes Gramps via the command line

hgohel:
If you feel like experimenting: Google's Gemma 4 suggests making this change in `gramps\gui\__init__.py` to lock the version the first time any gui component is touched.

stevenyoungs:
> If you feel like experimenting: Google's Gemma 4 suggests making this change in `gramps\gui\__init__.py` to lock the version the first time any gui component is touched.

That works and CoPilot also agrees that it is a better approach. Code amended. Thanks.

hgohel:
@stevenyoungs I see a force push but code change is still in `gramps/gui/plug/__init__.py` rather than `gramps/gui/__init__.py`

stevenyoungs:
> @stevenyoungs I see a force push but code change is still in `gramps/gui/plug/__init__.py` rather than `gramps/gui/__init__.py`

I've put the change in the correct place this time 🤦 
There are several other places within `gui` where `gi.require_version("Gtk", "3.0")` is called. I've not investigated if these could now be removed
image

hgohel (approved):
Looks good to me.

PR #2294, opened 2026-05-08
Add test to validate example .gramps files against the RelaxNG schema

iand:
Adds a test that scans the examples directory for .gramps files and validates them against the current RelaxNG schema so we don't ship out of date or invalid examples.

Also updates data.gramps to the latest schema version, which was missed on last schema update.


Nick-Hall:
The new unit test file should be added to `po/POTFILES.skip`.

dsblank (approved):
Looks good; thank you!

PR #2298, opened 2026-05-09
Remove checks for PyICU identifier, replaced and deprecated in 2015

jralls:
and removed in 2022.

### DIFF
--- a/gramps/gen/utils/grampslocale.py
+++ b/gramps/gen/utils/grampslocale.py
@@ -63,18 +63,13 @@ try:
     from icu import Locale, Collator, ICUError
 
     HAVE_ICU = True
-except ImportError:
-    try:
-        from PyICU import Locale, Collator, ICUError
-
-        HAVE_ICU = True
-    except ImportError as err:
-        ERRSTR = str(err)
-        # No logger, save the warning message for later.
-        _ICU_ERR = (
-            f"ICU not loaded because {ERRSTR}. Localization will be "
-            "impaired. Use your package manager to install PyICU"
-        )
+except ImportError as err:
+    ERRSTR = str(err)
+    # No logger, save the warning message for later.
+    _ICU_ERR = (
+        f"ICU not loaded because {ERRSTR}. Localization will be "
+        "impaired. Use your package manager to install PyICU"
+    )
 
 ICU_LOCALES = None
 if HAVE_ICU:
--- a/gramps/gen/utils/test/grampslocale_test.py
+++ b/gramps/gen/utils/test/grampslocale_test.py
@@ -26,12 +26,7 @@ try:
 
     HAVE_ICU = True
 except ImportError:
-    try:
-        from PyICU import Collator
-
-        HAVE_ICU = True
-    except ImportError:
-        HAVE_ICU = False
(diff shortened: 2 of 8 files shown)


dsblank (approved):
Looks good!

stevenyoungs:
Do either of these files need updating as well?

https://github.com/gramps-project/gramps/blob/ecbab0514135a974c96b3a259a4b822aa6f536b0/mypy.ini#L44-L45

https://github.com/gramps-project/gramps/blob/ecbab0514135a974c96b3a259a4b822aa6f536b0/pyproject.toml#L110

jralls:
> Do either of these files need updating as well?

I think mypy.ini does, pyproject.toml not: The [package name](https://pypi.org/project/pyicu/) is 'PyICU' but the import name is 'icu'. IIUC mypy.ini is telling mypy not to worry if `import PyICU` fails (those are the imports we're getting rid of in this PR), while pyproject.toml's [`optional-dependencies`](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#dependency-specifiers) block is about telling `pip` what to install. Is that right?

eduralph:
@jralls 

Heads-up on the webreport/common.py change: collapsing the outer except ImportError: to pass reintroduces a latent NameError.

After this PR, on a system with no icu:

```python
try:
    from icu import Locale
    ...
except ImportError:
    pass
```

localAlphabeticIndex is never bound, but further down the module still does AlphabeticIndex = localAlphabeticIndex when HAVE_ALPHABETICINDEX is false — so importing gramps.plugins.webreport.common raises NameError: name 'localAlphabeticIndex' is not defined at module load, and every plugin importing from webreport.common fails to register.

This is the same bug #2302 fixes on master (it surfaced via the TimePedigreeHTML addon in addons-source plugin-registration smoke tests). #2302 changes the innermost except ImportError: pass to import the in-tree fallback; this PR removes that branch entirely, but the pass carries the bug along.

Suggest the except ImportError: here keep the fallback import rather than pass:

```python
except ImportError:
    from gramps.plugins.webreport.alphabeticindex import (
        AlphabeticIndex as localAlphabeticIndex,
    )
```

Since this PR targets maintenance/gramps61 and #2302 targets master, the bug otherwise gets fixed on master but re-added on gramps61 — folding the one-liner in here keeps both branches consistent.

jralls:
@eduralph Thanks for the heads-up about the `localAlphabeticIndex` error. I've added a better fix for it so you might as well close #2302. For future reference you should base bug-fix and code-cleanup PRs on the current production branch. The master branch is for new features and doesn't get into users' hands until the next major release. We obviously want bug fixes published as soon as possible.

@Nick-Hall Are we going to have a 6.0.9? If so I'll rebase this to gramps60 since it's now a bug fix as well as a cleanup.

eduralph:
@jralls  - closed #2302 as suggested. Thanks for the pointer, I'll make sure to do it on the right branch in future

Nick-Hall:
> Are we going to have a 6.0.9?

I doubt it.  You can leave this fix in the _gramps61_ branch and I'll merge it.

jralls:
@Nick-Hall OK, thanks.

PR #2299, opened 2026-05-10
Allow gui.clipboard to import without a default screen

eduralph:
## Summary
**User impact:** On a computer with no graphical display available — for example an automated test machine — Gramps' clipboard code fails to load. Any add-on that depends on it then fails to register. In practice this breaks the Collections Clipboard add-on in headless plugin-registration tests.

This lets that clipboard code load without a display by falling back to empty icons, so the add-ons that depend on it register cleanly in headless runs.

Origin: fix driven by addons-source CI work — no Mantis tracker ticket (see gramps-project/addons-source#820).

## What to look at
Three module-level `theme.load_icon(...)` calls in `gramps/gui/clipboard.py` run at import time and crash when there is no default icon theme. The fix guards them so the module imports with `None` icon placeholders instead. To try it: import `gramps.gui.clipboard` with no `DISPLAY` (or with `Gtk.IconTheme.get_default` patched to return `None`) and confirm it loads.

## Root cause
Three module-level statements in `gramps/gui/clipboard.py` (lines 81-109) call `theme.load_icon(...)` where `theme = Gtk.IconTheme.get_default()`. In headless contexts (no `DISPLAY`, plugin-registration smoke tests run outside an X server, etc.) `Gtk.IconTheme.get_default()` returns `None`, so the calls raise `AttributeError: 'NoneType' object has no attribute 'load_icon'` at `clipboard.py:82` and the whole module fails to load. Every addon that does `from gramps.gui.clipboard import ...` then fails at registration. The most user-visible consequence is the `Collections Clipboard Gramplet` addon failing in plugin-registration smoke tests run by addons-source CI (gramps-project/addons-source#820).

## Fix
Guard the icon-loading paths with `if theme is not None`. When the default theme is unavailable, `LINK_PIC = None` and `ICONS` is populated with `None` values for every key in `OBJ2ICON`. The class bodies later in the file look up `ICONS[name]` at module load to set `ICON` class attributes (`ClipAddress.ICON = ICONS["address"]` etc.); the `None` placeholders keep those lookups from raising `KeyError`. Real icon rendering is gated by GTK at the widget level, which itself requires a display, so the fallback values are never used to draw anything. Behaviour with a default screen present is unchanged. `gramps/gui/test/clipboard_test.py` is added to `po/POTFILES.skip` alongside the sibling `gui/test` files.

## Verification
- **Claim:** `gramps.gui.clipboard` imports without a default screen (falling back to `None` icons); with a screen present behaviour is unchanged.
- **Checked:** `gramps/gui/clipboard.py:81-109` on master (the branch this PR targets) — icon loads guarded on `theme is not None`, with `LINK_PIC = None` and all-`None` `ICONS`.
- **Test:** `gramps/gui/test/clipboard_test.py` (new) patches `Gtk.IconTheme.get_default` to return `None`, re-imports the module, and asserts `LINK_PIC` is `None` and `ICONS` is keyed identically to `OBJ2ICON` with all-`None` values. Fails on the unfixed tree with the exact `AttributeError: 'NoneType' object has no attribute 'load_icon'` at `clipboard.py:82`, and passes after the patch. The test pins `gi` to GTK 3 / Gdk 3 (`clipboard.py` uses `Gdk.atom_intern`, removed in GTK 4) and skips cleanly on hosts without GTK 3 / PyGObject. `GRAMPS_RESOURCES=. python3 -m unittest gramps.gui.test.clipboard_test -v` → 1 test, OK; `black` and `mypy` clean on both files.

No Mantis tracker ticket — fix originates from addons-source CI work (gramps-project/addons-source#820).

### DIFF
--- a/gramps/gui/clipboard.py
+++ b/gramps/gui/clipboard.py
@@ -79,7 +79,13 @@ clipdb = None  # current db to avoid different transient dbs during db change
 # -------------------------------------------------------------------------
 
 theme = Gtk.IconTheme.get_default()
-LINK_PIC = theme.load_icon("stock_link", 16, 0)
+# Gtk.IconTheme.get_default() returns None when no default screen is
+# available (headless containers, plugin-registration smoke tests run
+# without a display, etc.). Fall back to a None LINK_PIC and an empty
+# ICONS dict in that case so this module imports cleanly. Any actual
+# UI rendering still requires a display, but that is gated by GTK
+# elsewhere — module import on its own should not raise.
+LINK_PIC = theme.load_icon("stock_link", 16, 0) if theme is not None else None
 OBJ2ICON = {
     "media": "gramps-media",
     "note": "gramps-notes",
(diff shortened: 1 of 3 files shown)


Nick-Hall:
Please rebase this on the _master_ branch.

eduralph:
> Please rebase this on the _master_ branch.

Done

PR #2302, opened 2026-05-11
Define localAlphabeticIndex in the no-ICU fallback branch

eduralph:
## Summary

[`gramps/plugins/webreport/common.py:46-75`](https://github.com/gramps-project/gramps/blob/maintenance/gramps60/gramps/plugins/webreport/common.py#L46-L75) has a three-level try/except chain trying to import an ICU `AlphabeticIndex` (from `icu`, then `PyICU`, falling back to an in-tree pure-Python implementation). The bottom-most branch — when **neither** `icu` nor `PyICU` is installed — was an empty `pass`, so `localAlphabeticIndex` was never bound:

```python
HAVE_ICU = False
HAVE_ALPHABETICINDEX = False
try:
    from icu import Locale
    HAVE_ICU = True
    try:
        from icu import AlphabeticIndex as icuAlphabeticIndex
        HAVE_ALPHABETICINDEX = True
    except ImportError:
        from gramps.plugins.webreport.alphabeticindex import (
            AlphabeticIndex as localAlphabeticIndex,
        )                                                      # ┐ sets localAlphabeticIndex
except ImportError:
    try:
        from PyICU import Locale
        HAVE_ICU = True
        try:
            from PyICU import AlphabeticIndex as icuAlphabeticIndex
            HAVE_ALPHABETICINDEX = True
        except ImportError:
            from gramps.plugins.webreport.alphabeticindex import (
                AlphabeticIndex as localAlphabeticIndex,
            )                                                  # ┘ sets localAlphabeticIndex
    except ImportError:
        pass                                                   # ← never sets localAlphabeticIndex
```

Later in the same module (around line 681):

```python
if HAVE_ALPHABETICINDEX:
    class AlphabeticIndex(icuAlphabeticIndex): ...
else:
    AlphabeticIndex = localAlphabeticIndex   # ← NameError if neither icu/PyICU was available
```

On a host lacking both icu bindings, `HAVE_ALPHABETICINDEX` is False, the `else` branch runs, and importing the module raises:

```
NameError: name 'localAlphabeticIndex' is not defined
```

Every plugin that does `from gramps.plugins.webreport.common import ...` then fails to load. The most visible victim so far is the `TimePedigreeHTML` addon (flagged in the [addons-source plugin-registration smoke tests](https://github.com/gramps-project/addons-source/pull/820)), which imports `get_gendex_data` from this module.

## Fix

Mirror the two middle fallback branches in the bottom one — import the in-tree pure-Python `AlphabeticIndex` when both icu modules are absent:

```diff
     except ImportError:
-        pass
+        # Neither icu nor PyICU is available. Fall back to the
+        # in-tree pure-Python AlphabeticIndex so the ``else`` branch
+        # later in this module (``AlphabeticIndex = localAlphabeticIndex``)
+        # finds a defined name. Without this, importing
+        # ``gramps.plugins.webreport.common`` on a system lacking
+        # both icu bindings raises ``NameError: name
+        # 'localAlphabeticIndex' is not defined`` at module load.
+        from gramps.plugins.webreport.alphabeticindex import (
+            AlphabeticIndex as localAlphabeticIndex,
+        )
```

Behaviour unchanged on hosts where either `icu` or `PyICU` is available; on hosts where neither is, the module now imports cleanly and `AlphabeticIndex` falls back to the in-tree implementation.

## Regression test

Added `gramps/plugins/webreport/test/common_test.py` that subprocesses an import with both `icu` and `PyICU` blocked via `sys.modules` tampering, then verifies the module loads with `AlphabeticIndex` defined and `HAVE_ICU`/`HAVE_ALPHABETICINDEX` set to `False`.

Verified locally:

- **Before fix**: `NameError: name 'localAlphabeticIndex' is not defined` (FAIL)
- **After fix**: 1 test, OK

## Verification

- `GRAMPS_RESOURCES=. python3 -m unittest gramps.plugins.webreport.test.common_test -v` → 1 test, OK.
- `black gramps/plugins/webreport/common.py gramps/plugins/webreport/test/common_test.py` → no formatting changes needed.
- `mypy gramps/plugins/webreport/common.py gramps/plugins/webreport/test/common_test.py` → no issues found.

## Translation files

`gramps/plugins/webreport/test/common_test.py` added to `po/POTFILES.skip`.

### DIFF
--- a/gramps/plugins/webreport/common.py
+++ b/gramps/plugins/webreport/common.py
@@ -72,7 +72,16 @@ except ImportError:
                 AlphabeticIndex as localAlphabeticIndex,
             )
     except ImportError:
-        pass
+        # Neither icu nor PyICU is available. Fall back to the
+        # in-tree pure-Python AlphabeticIndex so the ``else`` branch
+        # later in this module (``AlphabeticIndex = localAlphabeticIndex``)
+        # finds a defined name. Without this, importing
+        # ``gramps.plugins.webreport.common`` on a system lacking
+        # both icu bindings raises ``NameError: name
+        # 'localAlphabeticIndex' is not defined`` at module load.
+        from gramps.plugins.webreport.alphabeticindex import (
+            AlphabeticIndex as localAlphabeticIndex,
+        )
 
 LOG = logging.getLogger(".NarrativeWeb")
 
--- a/gramps/plugins/webreport/test/common_test.py
+++ b/gramps/plugins/webreport/test/common_test.py
@@ -1,0 +1,97 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Gramps Development Team
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""Unittest for webreport/common.py module-load resilience."""
+
+import subprocess
+import sys
+import textwrap
+import unittest
+
+
+# ------------------------------------------------------------
+#
+# TestCommonImportsWithoutICU
+#
+# ------------------------------------------------------------
+class TestCommonImportsWithoutICU(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


eduralph:
@Nick-Hall Cross-reference: #2298 (PyICU removal, targeting maintenance/gramps61) edits the same except ImportError: block in webreport/common.py. As currently written it collapses that block to pass, which reintroduces the localAlphabeticIndex NameError this PR fixes — I've left a note on #2298 suggesting it keep the in-tree fallback import.

Worth coordinating: this PR fixes the bug on master (v6.2); #2298 is gramps61 (v6.1). The bug exists on both branches, so the fix needs to reach both — either #2298 incorporates it, or this fix is also backported to gramps61.

jralls:
@Nick-Hall This is a bug fix not a feature. It should be on maintenance/gramps60 if we're going to do a 6.0.9 or gramps61 otherwise.

More generally, ICU 4.6 was released 16 years ago. Nobody is using anything older anymore so the separate check is obsolete. I've pushed a better fix to #2298.

eduralph:
As discussed with @jralls closing as he fixed it in #2298

PR #2303, opened 2026-05-11
Fix requires_mod addon module installation on Windows and Mac

dsblank:
## Problem

Addons can declare Python module dependencies via \`requires_mod\` in their plugin registration. When a user installs such an addon, Gramps is supposed to \`pip install\` the missing modules into \`LIB_PATH\` (\`~/.gramps/gramps6x/plugins/lib/\`), which is already on \`sys.path\`.

This did not work on Windows or Mac bundles.

### Root cause — invocation

\`gramps/gui/plug/_windows.py\` called pip as a bare shell command:

\`\`\`python
"pip.exe" if win() else "pip"
\`\`\`

- **Windows AIO**: \`pip.exe\` is a cx_Freeze-frozen executable that runs in its own Python context, separate from the running Gramps process. Installing to \`LIB_PATH\` via that executable does not reliably map back to the Gramps interpreter's import search path.
- **Mac bundle**: \`sys.executable\` is the Gramps C launcher binary, not a Python interpreter, so spawning a subprocess with \`sys.executable -m pip\` raises a \`PermissionError\`. The \`pip\` shell command is also not on the bundle's \`PATH\`.
- **Linux standard install**: works, since \`pip\` is usually on \`PATH\` and points to the right Python.

On both Mac and Windows, pip is actually importable — Mac bundles include it via the \`*.py\` glob that covers all of \`site-packages/\`, and Windows AIO includes it in \`lib/pip/\` which is on \`sys.path\`. The problem is solely the invocation mechanism.

### Root cause — check_mod not verifying importability

\`Requirements.check_mod()\` used \`find_spec()\` to test whether a module was available. \`find_spec()\` only checks whether a module file exists on \`sys.path\` — it does not attempt to load it. A compiled extension (\`.so\` / \`.pyd\`) whose native library dependencies are missing (common in bundled apps) passes \`find_spec()\` but raises \`ImportError\` when actually imported. This caused the addon manager to report a broken module as satisfied and then fail silently when the addon tried to use it.

### Invocation — \`gramps/gui/plug/_windows.py\`

Replace the subprocess-based pip call with \`_install_module()\`, which runs pip via \`runpy.run_module("pip", run_name="__main__")\`. This executes pip's \`__main__\` in the current interpreter with no external binary required — equivalent to \`python -m pip\` but entirely in-process. Output is captured via \`contextlib.redirect_stdout/stderr\` for display in error dialogs.

### check_mod — \`gramps/gen/utils/requirements.py\`

After \`find_spec()\` confirms a module file exists, attempt \`importlib.import_module()\`. If that raises \`ImportError\` (e.g. a compiled extension with a missing native library), return \`False\` so the addon manager correctly treats the module as unavailable.

## Test plan

- [ ] Install an addon whose \`requires_mod\` lists a pure-Python PyPI package not already installed
- [ ] Verify the install succeeds on Linux (regression check)
- [ ] Verify the install succeeds on Windows AIO
- [ ] Verify the install succeeds on a Mac \`.app\` bundle

### DIFF
--- a/gramps/gen/utils/requirements.py
+++ b/gramps/gen/utils/requirements.py
@@ -22,6 +22,7 @@
 # Python modules
 #
 # -------------------------------------------------------------------------
+from importlib import import_module
 from importlib.util import find_spec
 
 # -------------------------------------------------------------------------
@@ -47,17 +48,25 @@ class Requirements:
         self.gi_list = []
         self.exe_list = []
 
-    def check_mod(self, module):
+    def check_mod(self, module: str) -> bool:
         """
-        Check to see if a given module is available.
+        Check to see if a given module is available and importable.
+
+        Uses find_spec first (fast path), then actually imports the module to
+        catch compiled extensions whose native libraries are missing — a common
+        failure mode in bundled apps where pip can install the .so/.pyd file
+        but the dynamic linker cannot resolve its dependencies at load time.
         """
         if module in self.mod_list:
             return True
-        if find_spec(module) is not None:
-            self.mod_list.append(module)
-            return True
-        else:
+        if find_spec(module) is None:
             return False
+        try:
+            import_module(module)
+        except ImportError:
+            return False
+        self.mod_list.append(module)
+        return True
 
     def check_gi(self, module_spec):
         """
--- a/gramps/gui/plug/_windows.py
+++ b/gramps/gui/plug/_windows.py
@@ -293,7 +293,9 @@ class AddonRow(Gtk.ListBoxRow):
             try:
                 subprocess.check_output(
                     [
-                        "pip.exe" if win() else "pip",
+                        sys.executable,
+                        "-m",
+                        "pip",
                         "install",
                         "--target",
                         LIB_PATH,
--- a/gramps/gui/plug/_windows.py
+++ b/gramps/gui/plug/_windows.py
@@ -25,13 +25,15 @@
 # Python modules
 #
 # -------------------------------------------------------------------------
-import traceback
-import os
-from html import escape
-import threading
-import sys
-import subprocess
+import contextlib
 import importlib
+import io
+import os
+import runpy
+import sys
+import threading
+import traceback
+from html import escape
 
 # -------------------------------------------------------------------------
 #
@@ -192,6 +194,38 @@ class ProjectRow(Gtk.ListBoxRow):
 
 # -------------------------------------------------------------------------
 #
+# _install_module
+#
+# -------------------------------------------------------------------------
+def _install_module(package: str, target: str) -> tuple[int, str]:
+    """
+    Install a Python package into target using pip's __main__ in-process.
+
+    Runs pip via runpy so no external Python binary is required — essential
+    for Mac .app bundles and Windows AIO where sys.executable is the Gramps
+    launcher, not a Python interpreter.  Output (stdout + stderr) is captured
+    and returned for display in error dialogs.
+    """
+    buf = io.StringIO()
+    old_argv = sys.argv[:]
+    sys.argv = ["pip", "install", "--target", target, package]
+    try:
+        with contextlib.redirect_stdout(buf), contextlib.redirect_stderr(buf):
+            runpy.run_module("pip", run_name="__main__")
+        return 0, buf.getvalue()
+    except SystemExit as exc:
+        code = exc.code if isinstance(exc.code, int) else 1
+        return code, buf.getvalue()
+    except ModuleNotFoundError:
+        return 1, _("pip is not available in this Python installation.")
+    except Exception as exc:
+        return 1, str(exc)
+    finally:
+        sys.argv = old_argv
+
+
+# -------------------------------------------------------------------------
+#
 # AddonManager
 #
 # -------------------------------------------------------------------------
@@ -288,26 +322,13 @@ class AddonRow(Gtk.ListBoxRow):
         """
         Install the addon and possibly some required python modules.
         """
-        # Install required modules
         for package in self.req.install(addon):
-            try:
-                subprocess.check_output(
-                    [
-                        sys.executable,
-                        "-m",
-                        "pip",
-                        "install",
-                        "--target",
-                        LIB_PATH,
-                        package,
-                    ],
-                    stderr=subprocess.STDOUT,
-                )
-            except subprocess.CalledProcessError as err:
+            returncode, output = _install_module(package, LIB_PATH)
+            if returncode != 0:
                 button.set_sensitive(False)
                 InfoDialog(
                     _("Module installation failed"),
-                    err.output.decode("utf-8"),
+                    output or _("pip exited with code %d") % returncode,
                     parent=self.window,
                 )
                 return
--- a/mac/gramps.bundle
+++ b/mac/gramps.bundle
@@ -164,13 +164,6 @@
     ${prefix}/lib/python3.13/*.py
   </data>
 
-  <!-- pip is needed so that addons can install required Python modules
-       at runtime via "requires_mod". The *.py glob above does not cover
-       site-packages subdirectories, so pip must be listed explicitly. -->
-  <data>
-    ${prefix}/lib/python3.13/site-packages/pip/
-  </data>
-
   <data>
     ${prefix}/lib/python3.13/config-3.13-darwin/
   </data>
--- a/mac/gramps.bundle
+++ b/mac/gramps.bundle
@@ -164,6 +164,13 @@
     ${prefix}/lib/python3.13/*.py
   </data>
 
+  <!-- pip is needed so that addons can install required Python modules
+       at runtime via "requires_mod". The *.py glob above does not cover
+       site-packages subdirectories, so pip must be listed explicitly. -->
+  <data>
+    ${prefix}/lib/python3.13/site-packages/pip/
+  </data>
+
   <data>
     ${prefix}/lib/python3.13/config-3.13-darwin/
   </data>


dsblank:
CC: @jralls @Nick-Hall @stevenyoungs: would appreciate any feedback on this. Having pip-installable packages available for addons is a game changer.

dsblank:
See also: https://github.com/gramps-project/gramps/pull/2250

Nick-Hall:
@dsblank Yes.  This was the original intention of the `requires_mod` field.  It was disappointing when it didn't work.

You should test with both pure python and non-pure python modules.  We could only get pure python modules to work in the past on Windows.

The requirements code should also handle the case when a required module is removed.  The addon should just fail to register when its requirements are no longer met.

GaryGriffin:
How can we also allow Addon unittests to use the additional modules? For instance, the TMGimporter addon requires `dbf` (which requires `aenum`). The unittests for this addon are:

```
python3 -m unittest TMGimporter.tests.test_integration -v
python3 -m unittest TMGimporter.tests.test_libtmg -v
```

dsblank:
> You should test with both pure python and non-pure python modules. We could only get pure python modules to work in the past on Windows.

I'm at a bit of blocker here because I have neither a Mac nor Windows computer. I'll need some help from the team. If you give me a an error message I can ask Claude.

> The requirements code should also handle the case when a required module is removed. The addon should just fail to register when its requirements are no longer met.

That sounds like it is outside the scope of this PR.

Nick-Hall:
I don't have Windows or Mac either, so I've also had to rely on others for testing.

On Linux, it is robust and you can actually do things like remove a required dependency of an installed plugin without getting errors.

jralls:
> Root cause — Mac bundle missing pip

> The Mac .app bundle copies Python files via:
```
<data recurse="True">
  ${prefix}/lib/python3.13/*.py
</data>
```
> This glob covers files directly under lib/python3.13/ (the standard library) but does not descend into site-packages/ subdirectories. As a result, site-packages/pip/ is never copied into the bundle, so even a corrected invocation would fail.

It does recurse into the `site-packages` directory. If it didn't it wouldn't even install gramps, never mind all of the other dependencies. The `pip` directory *is* in the macOS bundle.

What's missing are: The `pip` executable, because `ensurepip` is never run to create it, and a shell from which to run it. There's also no python cli to use `python -mpip`, which always works even when `ensurepip` hasn't run. `Gramps.app/Contents/MacOSX/Gramps` is a small C program that's linked to libpython and runs `Gramps.app/Contents/Resources/gramps_launcher.py` that sets up the environment and calls gramps's `main()`.

But `pip` is a module, so we could just `import pip` and run `pip.main(["install", "--user", "pypdf"]), right?
I tried that in the Python Console gramplet:
```
> import pip
> pip.main(["install", "--user", "pypdf"])
2
> 
```
Not exactly the result I hoped for. Gemini tells me that an rv of 2 means a command-line syntax error. I tried again without the `--user` arg and got the same result so I need to dig around in the pip code some more. This is definitely not an intended use case.

jralls:
I pretty quickly found [this document](https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program) with a recommended way to run pip from inside a python program (after a couple of paragraphs explaining why you probably shouldn't):
```
subprocess.check_call([sys.executable, '-m', 'pip', 'install', 'my_package'])
```
I tried that in the Python Shell gramplet and got what I guess is a predictable exception:
```
File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/subprocess.py", line 1990, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
PermissionError: [Errno 13] Permission denied: '/Users/john/Desktop/Gramps.app'
```
Interesting that `sys.executable` is the outer bundle name instead of the actual executable. Anyway, this looks like a workable direction. We need the python executable in the bundle and the path available to the addons dialog. Since I can only find `pip` used in gramps/gui/plug/_windows.py I guess that Unix is relying on something in importlib and that we'll need a similar custom downloader for macOS.

dsblank:
Thank you @jralls for the detailed investigation and correction — I apologize for the incorrect analysis in the original description. I claimed that the `*.py` glob didn't recurse into `site-packages/`, but as you rightly pointed out, it does (pip is already in the bundle). The real problem is that `sys.executable` is the Gramps C launcher rather than a Python interpreter, so neither a subprocess call nor `sys.executable -m pip` can work.

The updated fix uses `runpy.run_module("pip", run_name="__main__")` which runs pip's `__main__` in-process — no external binary needed. I've also updated the PR description to reflect the correct understanding.

Thanks also to @Nick-Hall and @stevenyoungs for the helpful context on what has and hasn't worked in the past, and for the pointer to the `check_mod` importability issue.

dsblank:
## Should we build our own installer?

It's a spectrum of complexity depending on how much of pip's behaviour you need to replicate.

### The easy part — pure-Python packages

A wheel is just a zip file. PyPI has a JSON API. Everything needed is in the stdlib:

```python
import hashlib, json, io, urllib.request, zipfile

def install(package, target):
    # 1. Fetch metadata from PyPI
    with urllib.request.urlopen(f"https://pypi.org/pypi/{package}/json") as r:
        meta = json.loads(r.read())

    # 2. Find a py3-none-any wheel (pure Python)
    wheel = next(
        f for f in meta["urls"]
        if f["packagetype"] == "bdist_wheel"
        and f["filename"].endswith("-py3-none-any.whl")
    )

    # 3. Download and verify
    data = urllib.request.urlopen(wheel["url"]).read()
    assert hashlib.sha256(data).hexdigest() == wheel["digests"]["sha256"]

    # 4. Extract into LIB_PATH
    with zipfile.ZipFile(io.BytesIO(data)) as whl:
        whl.extractall(target)
```

That's the whole happy path for something like `pypdf` or `markdown`. About 15 lines, zero external dependencies, works in any bundle.

### Where it gets hard

**Dependency resolution** is the main difficulty. Every `Requires-Dist` line in the wheel's `METADATA` is itself a package with version constraints, extras, and environment markers. Resolving a full dependency graph without conflicts is essentially a constraint satisfaction problem — it's why pip's resolver is thousands of lines.

For Gramps's use case you could take a shortcut: install direct dependencies without conflict checking, and let `check_mod` catch failures. That works fine for simple trees.

**Platform tag matching** for compiled wheels requires building the right tag string from `sys.version_info`, `sysconfig.get_platform()`, and the ABI flags. Python's own `packaging` library (which is part of pip) does this, but reimplementing it for compiled packages is fiddly.

**SSL certificates** — `urllib` needs the system CA bundle. Bundled apps handle this differently (the AIO setup includes its own cert file via `SSL_CERT_FILE`).

### Realistic scope

| Goal | Lines of code | Handles |
|------|--------------|---------|
| Pure-Python, latest version, no deps | ~30 | `pypdf`, `markdown`, `svgwrite` |
| Pure-Python, with direct deps | ~150 | Most simple addons |
| Compiled + pure, with deps | ~500+ | Still no conflict resolution |
| Full pip replacement | ~50,000 | Everything |

### Is it worth it?

Given that `runpy.run_module("pip")` works for the common cases (pip is present in both the Mac bundle and Windows AIO), the main reason to build a mini-installer would be robustness — not depending on pip's internal structure — and being explicit about what "pure Python only" means.

A 150-line installer that handles pure-Python packages with their direct dependencies would cover the vast majority of real addon requirements (`pypdf`, `litellm`, `markdown`, `dbf`, `svgwrite`). It would be self-contained, auditable, and immune to pip API changes.

The question is whether the edge cases it can't handle (complex version conflicts, compiled deps, non-PyPI sources) are worth the ongoing maintenance cost compared to just wrapping pip.

jralls:
@dsblank The [pip docs](https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program) warn against running pip inside a running python process, saying that it relies on python exiting to do a thorough cleanup. They're especially concerned about running pip twice in the same process. I can't speak for the Windows AIO, but fit's not a big deal to add the python executable to the macOS bundle so that we can do `subprocess.check_call` to fork & exec a new python instance in which to run pip.

dsblank:
> but fit's not a big deal to add the python executable to the macOS bundle

Ok, let's do that!

dsblank:
## Is running pip inside a running Python process problematic?

The concern is legitimate in general, but the severity for Gramps's specific use case is low.

### What the pip docs are actually worried about

1. **`sys.modules` pollution** — pip imports its own dependencies (requests, packaging, urllib3, etc.) during a run and caches them. A second pip invocation reuses those cached versions. If pip upgraded one of its own deps during the first run, the second run gets a stale cached module.

2. **`sys.path` mutation** — pip may temporarily add paths during operation; these persist in the running process.

3. **Temp file cleanup** — pip creates temp dirs for downloads/extraction. Normally cleaned up on process exit, but `runpy` keeps the process alive.

### Why the risk is low for Gramps

- Gramps uses `--target LIB_PATH`, not system `site-packages`. This isolates the installed files away from pip's own dependencies.
- Addon installation is interactive, infrequent, and one-at-a-time. The "run pip twice" scenario does exist (a user installing two addons back-to-back) but the shared state from the first run is unlikely to corrupt the second.
- The newly installed modules aren't imported immediately — they're picked up when the addon loads later, so `sys.modules` caching isn't a concern for the installed package itself.

### The real trade-off

The `subprocess` alternative (recommended by pip docs) is the correct long-term answer, but it requires a real Python binary path — which is unavailable on Mac (`.app` launcher) and Windows AIO (cx_Freeze). The `runpy` approach is the best available option under those platform constraints.

### If the concern grows

The mitigation would be running pip in a `subprocess` using the *embedded* Python binary path (not `sys.executable`), found via `sysconfig.get_config_var("BINDIR")` or a platform-specific search — but that's significantly more complex and platform-specific. For now, `runpy` is a reasonable pragmatic compromise.

jralls:
@dsblank Built and bundled this. It doesn't appear to have made any difference:

Screenshot 2026-05-12 at 2 01 38 PM

Nick-Hall:
@dsblank I can build a Windows AIO for you, but cannot test it myself.

dsblank:
> Built and bundled this. It doesn't appear to have made any difference:

@jralls can you try a pure-python library, like [pypdf](https://pypi.org/project/pypdf/) ?

jralls:
@dsblank It seems that the setuptools install ignored my feature branch with this PR in it and installed maintenance/gramps61 and that's what got bundled. So I copied the two files into the installation directory and bundled again. When trying to install TMG Project Backup (whose dependency is [dbf](https://pypi.org/project/dbf/), which claims to be pure python) I get the below dialog instead of the previous error dialog with a stack trace:
Screenshot 2026-05-13 at 9 32 18 PM

Edit: And it's right. pip isn't in the bundle. I'll investigate that tomorrow.

jralls:
I'm still not sure what happened with pip. It only installed the `__pycache__`s in my git build directory so `*.py` in the bundler didn't match anything. The tarball build (git and tarball are just about where gramps code is obtained, everything else should be the same) has the python files and so the bundle had them too. I copied the tarball's installation to the git build and bundled again. This time I got the "Module installation failed" dialog box containing
```
WARNING: There was an error checking the latest version of pip.
ERROR: Exception:
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/site-packages/pip/_internal/cli/base_command.py", line 107, in _run_wrapper
    status = _inner_run()
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/site-packages/pip/_internal/cli/base_command.py", line 98, in _inner_run
    return self.run(options, args)
           ~~~~~~~~^^^^^^^^^^^^^^^
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/site-packages/pip/_internal/cli/req_command.py", line 96, in wrapper
    return func(self, options, args)
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/site-packages/pip/_internal/commands/install.py", line 340, in run
    session = self.get_default_session(options)
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/site-packages/pip/_internal/cli/index_command.py", line 82, in get_default_session
    self._session = self.enter_context(self._build_session(options))
                                       ~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/site-packages/pip/_internal/cli/index_command.py", line 101, in _build_session
    ssl_context = _create_truststore_ssl_context()
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/site-packages/pip/_internal/cli/index_command.py", line 52, in _create_truststore_ssl_context
    ctx.load_verify_locations(certifi.where())
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/site-packages/pip/_vendor/truststore/_api.py", line 160, in load_verify_locations
    return self._ctx.load_verify_locations(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        cafile=cafile, capath=capath, cadata=cadata
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
FileNotFoundError: [Errno 2] No such file or directory
```
Looks like I also need to install $PREFIX/etc/ssl/cert.pm and call [`set_default_verify_paths](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.set_default_verify_paths) in gramps_launcher.py.

dsblank:
## Root cause of the Mac SSL failure

Thanks @jralls for testing and diagnosing. The new traceback pinpoints the actual problem:

```
FileNotFoundError: [Errno 2] No such file or directory
  File "pip/_internal/cli/index_command.py", line 52, in _create_truststore_ssl_context
      ctx.load_verify_locations(certifi.where())
```

pip's `_create_truststore_ssl_context()` calls `certifi.where()`, which returns a path to `cacert.pem` inside the certifi package directory. The Mac bundle includes certifi's `.py` files but **not its `cacert.pem` data file**, so the path returned by `certifi.where()` doesn't exist.

### Why the usual workarounds don't help

`--cert /path/to/cert` and the `SSL_CERT_FILE` environment variable are processed later in pip's startup — after `_build_session()` is called, which is where `_create_truststore_ssl_context()` runs. The `FileNotFoundError` fires before pip reads those options.

### Options

1. **Add `cacert.pem` to the Mac bundle** — Add a `<data>` entry in `gramps.bundle` for `site-packages/certifi/cacert.pem`. This is a one-line fix in the bundler config that gives pip a valid cert file.

2. **Mini-installer** — Bypass pip entirely. Python's built-in `urllib` used with `ssl.create_default_context()` uses the macOS Security framework (system keychain) instead of certifi, so there's no `cacert.pem` dependency at all. As discussed in the earlier comment, this is ~150 lines for pure-Python packages with direct deps. Eliminates the certifi problem, the in-process pip concerns, and the Windows/Mac `sys.executable` issues in one shot.

3. **Pre-flight patch** — Before calling pip, check whether `certifi.where()` exists; if not, symlink or copy `/etc/ssl/cert.pem` (macOS system cert bundle) into the expected location. Hacky and fragile.

Option 1 is the smallest code change. Option 2 is the most principled and would permanently close off this category of platform-specific pip failure, at the cost of building and maintaining a small wheel downloader.

jralls:
@dsblank No, the bundle doesn't include certifi at all. That's usually needed only on very old systems where the OS certificates are expired. The build includes OpenSSL, which provides a `cert.pem`. The compiled-in path isn't in the bundle, so python needs to be told to where to look.

OTOH if you think that it's cleaner to have our own download/install function instead of pip--which indeed has a lot of features that we don't need--I'm OK with that. I'm going to try the `cert.pem` route anyway because it won't take much time.

dsblank:
Thanks @jralls! I'll take a look at the devil in the details of building our own downloader/installer.

dsblank:
I've opened a related PR #2308 that takes the mini-installer approach we discussed: a stdlib-only wheel downloader that replaces the pip invocation entirely. It uses `ssl.create_default_context()` (system trust store, no certifi) and runs in-process with no subprocess or external binary, which should resolve both the Mac `cacert.pem` issue and the Windows AIO context problem. Includes 33 unit tests (no network) and 13 end-to-end tests against real PyPI.

dsblank:
We've also pulled the `check_mod()` importability fix into PR #2308, so that the complete solution lands together. Thanks for the careful analysis of that failure mode.

dsblank:
Closing in favor of #2308

PR #2305, opened 2026-05-12
Add order change UI for rules in filter editor

dsblank:
This PR adds the ability to change the order of rules in the filter editor by dragging/dropping, up/down arrow buttons, and keyboard controls.

Visual info is given by a textual hint below the rule list (requires translation), and via changing the mouse-over cursor.

image
 
Thanks to @emyoulation for bringing this issue up.

Note: This PR was written with help from Cursor (various models) and Copilot.

---
*Originally opened on 2025-12-31 as PR #2161.*

### DIFF
--- a/gramps/gui/editors/filtereditor.py
+++ b/gramps/gui/editors/filtereditor.py
@@ -82,6 +82,7 @@ from ..widgets import DateEntry
 from gramps.gen.datehandler import displayer
 from gramps.gen.config import config
 from gramps.gen.constfunc import mac
+from ..utils import match_primary_mask
 from gramps.gui.widgets.persistenttreeview import PersistentTreeView
 
 # -------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


hgohel (commented):
I reviewed the commit to use Cmd instead of Ctrl and it looks like it should work and feel native on a Mac. Since I don't work on a Mac, I'll ping @jralls for the final word, and a note that it looks good to me.

jralls:
@hgohel, @dsblank 
That will work, but it's nicer to use [`match_primary_mask`](https://github.com/gramps-project/gramps/blob/7dc332d304a879bb1c93cafcea86805cb5e85459/gramps/gui/utils.py#L925), at least until we migrate to Gtk4 where `GdkModifierIntent` has been removed.

dsblank:
Thanks @jralls for the hint! We've updated the code to use `match_primary_mask()` instead of the manual `META_MASK`/`CONTROL_MASK` detection.

PR #2307, opened 2026-05-14
Gramps is using GTK3 not GTK4

hgohel:
AGENTS.md says Gramps is a GKT4 app, but we're still on GTK3 so correct that.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -43,7 +43,7 @@ black --check .
 
 ## Architecture
 
-Gramps is a GTK4 genealogy application. The codebase is split into four main packages:
+Gramps is a GTK3 genealogy application. The codebase is split into four main packages:
 
 ### `gramps/gen/` — Core (no GUI dependency)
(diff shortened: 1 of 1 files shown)


dsblank (approved):
Good catch!

stevenyoungs:
I saw this the other day and did not submit a PR as the problem was not present in `maintenance\gramps61`. I wrongly assumed it was just waiting for Nick to merge. I was wrong. `master` is ahead of `maintenance\gramps61` for this file 🤦

hgohel:
Maybe AI already migrated Gramps to GTK4 and I'm not aware!? 😉

dsblank:
AI is now using GTK62, but there are no more humans to worry about genealogy :(

dsblank:
@hgohel it may be that the version in maintenance/gramps61 hasn't been copied over to master yet.

Nick-Hall:
I'll do another branch merge tomorrow.

hgohel:
Hmm. [AGENTS.md in gramps61](https://github.com/gramps-project/gramps/blob/maintenance/gramps61/AGENTS.md) doesn't mention GTK at all. I'll wait for the branch merge.

PR #2308, opened 2026-05-15
Add Python PyPI wheel installer for addon module dependencies

dsblank:
## Problem

Addon plugins can declare Python module dependencies via `requires_mod`. When a user installs such an addon, Gramps is supposed to install the missing modules into `LIB_PATH` (`~/.gramps/gramps6x/plugins/lib/`). This did not work on Windows AIO or Mac `.app` bundles:

- **Windows AIO**: `pip.exe` is a cx_Freeze binary running in its own Python context, separate from Gramps's interpreter.
- **Mac bundle**: `sys.executable` is the Gramps C launcher, not a Python interpreter, so subprocess calls raise `PermissionError`. pip is present in the bundle but its SSL layer calls `certifi.where()`, which returns a path to `cacert.pem` — a data file that was never bundled — causing `FileNotFoundError` before any `--cert` flag is read.

A second independent bug: `Requirements.check_mod()` used `find_spec()` to test whether a module was available. `find_spec()` only checks whether a module file exists on `sys.path` — it does not attempt to load it. A compiled extension (`.so` / `.pyd`) whose native library dependencies are missing passes `find_spec()` but raises `ImportError` when actually imported. This caused the addon manager to report a broken module as satisfied and then fail silently.

Two additional Addon Manager bugs are also fixed in this PR:

- **Shared-tgz install conflict**: Multiple addons sharing the same `.addon.tgz` (e.g. SQLite Import + Export, PDFForms) each appear as separate installable entries. Installing the second entry re-extracted the tgz into an already-existing directory, causing an `OSError` on Mac. The return value of `load_addon_file` was never checked so the failure was silent.

- **Sibling plugins not marked installed**: After installing one entry from a shared tgz, the other entries from the same tgz stayed showing an "Install" button (only the clicked row was refreshed). Clicking Install again triggered the conflict above.

See also: #2303

### Installer — `gramps/gen/utils/pypi.py` (new) and `gramps/gui/plug/_windows.py`

The install path is now split based on whether Gramps is running inside a frozen bundle (`sys.frozen`):

**Frozen bundles (Windows AIO, Mac .app):** A new stdlib-only module `gramps/gen/utils/pypi.py` downloads and installs wheels from PyPI directly, with no subprocess or external binary required:

- **`install_package(package, target)`** — public API, drop-in replacement for the old pip call
- Fetches wheel metadata from the PyPI JSON API
- Verifies SHA-256 of the downloaded wheel before extracting
- Extracts using `zipfile` (wheels are zips)
- Recursively installs direct dependencies (`Requires-Dist` entries), and checks version constraints — if a dep is already importable but its installed version does not satisfy the declared spec (e.g. `>=2.0`) it is reinstalled
- **Network requests time out after 30 s** — a slow or unreachable PyPI will raise an error rather than hanging the UI indefinitely
- **SSL uses `ssl.create_default_context()`** — system trust store on all platforms, no certifi or `cacert.pem` required
- **Supports both pure-Python and compiled wheels.** Pure wheels (`py3-none-any`) are preferred; if none is available, the installer selects a compiled wheel matching the current Python version and platform (Windows: `win_amd64` / `win_arm64` / `win32`; macOS: `macosx_X_Y_{arm64,x86_64,universal2}` spanning all versions back to 10.4).

**Source / snap installs with pip:** `_pip_available()` probes whether `sys.executable -m pip --version` succeeds (result cached). When pip is available, it is used via `sys.executable -m pip install --target`; pip handles all wheel formats natively.

**Flatpak / pip-less environments:** Flathub and other sandboxed runtimes strip `pip` from the Python installation. `_pip_available()` returns `False` and the stdlib installer is used instead — exactly as for frozen bundles.

The stdlib installer now supports compiled Linux wheels as well: `_compatible_tags()` detects glibc (via `ctypes.gnu_get_libc_version()`, falling back to `platform.libc_ver()`) and musl (by probing `/lib/ld-musl-{arch}.so.1`) and generates the full manylinux / musllinux tag set — PEP 600 perennial tags plus the three legacy aliases (manylinux1, manylinux2010, manylinux2014) where the architecture qualifies. The bare `linux_{arch}` tag is still included as a fallback.

**Platform-conditional `Requires-Dist` entries evaluated, not skipped:** Previously every dependency line containing a `;` marker was unconditionally dropped. Now a full PEP 508 marker evaluator is used: `sys_platform`, `python_version`, `platform_system`, `implementation_name`, `extra`, and all other standard marker variables are supported; comparisons of version variables (`python_version`, `python_full_version`, `implementation_version`) use tuple ordering rather than lexicographic string ordering; `and`, `or`, `not`, parentheses, and the `in` / `not in` operators are all handled. Markers that fail to parse are treated as `True` (dep included) to avoid silently dropping dependencies. Examples: `pywin32; sys_platform == "win32"` is installed on Windows and skipped on Linux; `importlib-metadata>=3.6; python_version < "3.10"` is skipped on Python 3.10+.

**Extras on transitive deps propagated:** A dependency declared as e.g. `requests[security]>=2.0` previously had its extras silently dropped, and the version spec was parsed incorrectly as `"[security]>=2.0"`. Now `_direct_dependencies()` uses a compiled PEP 508 regex to correctly separate name, bracket-extras, version spec, and marker. Dep-extras are forwarded to recursive `_install_one()` calls so that `extra == "foo"` markers within the transitive dependency are evaluated correctly. `install_package()` also accepts a top-level `extras` parameter and parses extras embedded in the package name (e.g. `install_package("requests[security]", target)` activates the `security` extra automatically).

**Version-conflict resolution (tier 1):** `install_package` now performs a three-pass install instead of a single DFS:

1. **Gather** (`_gather_requirements`): traverses the full dependency graph using `meta["info"]["requires_dist"]` from the PyPI JSON API — no wheels are downloaded in this pass. Collects all version constraints from every path through the graph into a `dict[str, list[str]]`. Already-importable packages are skipped (they will also be skipped in the install pass).

2. **Resolve** (`_find_satisfying_version`): for each package that has at least one version constraint, scans `meta["releases"]` newest-first and returns the newest **stable** release satisfying the combined constraint string. Pre-releases (`aN`, `bN`, `rcN`, `.devN`) and fully-yanked releases (all files carry `"yanked": true`) are skipped. Raises `PyPIInstallError` immediately when no version satisfies the intersection (e.g. one dep requires `>=2.0` while another requires `<1.5`).

3. **Install**: unchanged DFS pass, but `_pick_wheel` now accepts an optional `version=` parameter so the install pass uses the exact version chosen in the resolve pass rather than always picking the latest.

`_pypi_metadata` is now decorated with `@functools.cache` so that the gather and install passes share the same PyPI response without a second network round-trip.

Known limitation: constraint gathering reads the *latest* release's dep list; if an older version is resolved, its actual transitive dependencies may differ slightly (addressed in a future tier).

**Wheel quality filters:** `_pick_wheel` now skips files that should not be installed on the running interpreter:

- **Yanked files** (`"yanked": true` in the PyPI API response) are excluded from all search paths — latest release, older releases, and pinned-version installs. A release is used only if at least one of its files is not yanked.
- **`Requires-Python` constraint** (`requires_python` field on each file entry) is validated against the running interpreter's `major.minor` version before a wheel is selected. A wheel declaring `Requires-Python: >=3.12` will not be chosen on Python 3.9.

Both filters apply in `_pick_wheel` regardless of whether the call comes from the unversioned (latest) path or the pinned-version path.

**Import name → PyPI name mapping:** Addon `requires_mod` entries use Python import names (e.g. `"PIL"`, `"yaml"`, `"bs4"`) which often differ from PyPI distribution names (`"Pillow"`, `"PyYAML"`, `"beautifulsoup4"`). `resolve_pypi_name()` translates known mismatches via a built-in `_IMPORT_TO_PYPI` table before any install attempt. When a mapped name is encountered a `LOG.warning` is emitted so addon authors know to update `requires_mod` to use the correct PyPI name directly.

### check_mod — `gramps/gen/utils/requirements.py`

After `find_spec()` confirms a module file exists, attempt `import_module()`. If that raises `ImportError` (e.g. a compiled extension with a missing native library), return `False` so the addon manager correctly treats the module as unavailable. (Ported from #2303.)

### Addon Manager install fixes — `gramps/gui/plug/_windows.py`

- **Check `load_addon_file` return value**: if extraction fails (e.g. OSError on Mac when re-extracting into an existing directory), show an error dialog and abort instead of proceeding into a broken state.

- **Call `manager.refresh()` after install** instead of only rebuilding the clicked row. This marks all sibling plugins sharing the same tgz as installed and prevents the spurious second-install attempt.

- **Auto-install `depends_on` addons**: before installing an addon, check its `"do"` catalog field (populated from `depends_on` in the gpr.py by the build system) and automatically install any declared dependencies that are not yet registered.  Python module requirements (`requires_mod`) declared by those dependency addons are now installed as well, via the new shared `_install_python_deps()` helper.

## Bundle detection (`_is_frozen`)

`_is_frozen()` determines which install path to take. Most bundlers set `sys.frozen`, but some do not — notably older Briefcase versions on macOS and Nuitka/PyOxidizer on Windows. The function now uses three checks:

| Bundler | Platform | `sys.frozen` set? | Detected by |
|---|---|---|---|
| cx_Freeze | Win/Mac/Linux | ✓ `True` | `sys.frozen` check |
| PyInstaller | Win/Mac/Linux | ✓ `True` | `sys.frozen` check |
| py2app | macOS | ✓ `'macosx_app'` | `sys.frozen` check (truthy) |
| py2exe | Windows | ✓ `True`/`'dll'` | `sys.frozen` check |
| Briefcase (new) | Win/Mac | ✓ `True` | `sys.frozen` check |
| Briefcase (old) / unknown | macOS | ✗ | `.app` path check |
| Nuitka | Windows | ✗ | `basename != python*.exe` check |
| PyOxidizer | Windows | ✗ | `basename != python*.exe` check |

The `.app` path check catches bundles where `sys.executable` ends with `.app` or contains `.app/` (e.g. a user running from `/Users/john/Desktop/Gramps.app`). The Windows basename check catches bundles where `sys.executable` is a named app binary (`Gramps.exe`) rather than a Python interpreter (`python.exe`, `python3.12.exe`, etc.).

## Tests

**Unit tests** (`gramps/gen/utils/test/pypi_test.py`) — 200 tests, no network required, all logic mocked:
- `_is_pure_wheel`: 7 cases covering both pure tags, platform wheels, abi3, tarballs, malformed filenames
- `_compatible_tags`: 9 existing cases (Windows AMD64/ARM64, macOS arm64, macOS x86_64, cache behaviour, always-present tags) + 8 new Linux cases (manylinux tag generation, musllinux tag generation, bare linux fallback, aarch64/ppc64le arch coverage)
- `_glibc_version`: ctypes path, `platform.libc_ver()` fallback, returns None on non-glibc, caching
- `_musl_version`: linker found + version parsed, linker absent, subprocess error, ARM arch normalisation
- `_manylinux_platform_tags`: PEP 600 tag range, all three legacy aliases, arch restrictions, glibc != 2 guard
- `_is_compatible_wheel`: 8 cases — matching compiled wheel, platform mismatch, Python version mismatch, abi3 wheel, dot-joined multi-tag platform field, tarball, too-few parts, empty compat sets
- `_is_prerelease`: 6 cases — alpha (`1.0a1`), beta (`1.0b1`), release candidate (`2.0rc1`), dev release (`1.0.dev0`), stable release (`1.0.0`), post-release (`1.0.post1`) correctly classified as stable
- `_is_yanked`: 4 cases — boolean `True`, truthy reason string, boolean `False`, absent key
- `_requires_python_ok`: 6 cases — no constraint (always True), satisfied `>=` constraint, unsatisfied `>=` constraint, satisfied range, unsatisfied upper bound, malformed spec treated as satisfied
- `_verify_sha256`: correct hash, mismatch, empty bytes
- `_wheel_metadata`: name/version parsing, multi-value `Requires-Dist`, missing METADATA
- `_tokenize_marker`: 7 cases — two-char operators, `not in` merging, single-quoted strings, whitespace
- `_eval_marker`: 17 cases — `sys_platform` equality, `python_version` numeric ordering (catches `"3.10" > "3.9"` lexicographic trap), `extra ==` with/without matching extras, `and`/`or`/`not`, parentheses, parse-error safety
- `_direct_dependencies`: returns `(name, spec, dep_extras)` triples — bare names, preserved version constraints, platform markers evaluated for current/non-current platform, extras markers with/without requested extras, dep-extras parsed (single and multiple), compound constraints
- `_gather_requirements`: 8 cases — versioned dep added to specs, unversioned dep not added, multiple constraints accumulated across packages, circular dep guard, platform-filtered dep not gathered, PyPIInstallError handled gracefully, null/missing `requires_dist`
- `_find_satisfying_version`: 10 cases — newest for empty spec, newest satisfying `>=`, range `>=X,<Y`, no version satisfies, PyPI unreachable, empty releases, pre-release skipped in favour of older stable release, fully-yanked release skipped, partially-yanked release used (non-yanked file present), pre-release selected when it is the only version satisfying the constraint
- `_pick_wheel`: selects pure from `urls`, compiled wheel fallback, falls back to `releases`, raises when nothing found; pinned-version variants: selects from release files, raises when pinned release has no compatible wheel; yanked file skipped, `Requires-Python`-incompatible wheel skipped
- `_satisfies_spec`: all seven PEP 440 operators, wildcard `==X.Y.*` / `!=X.Y.*`, compatible-release `~=` (two-part and three-part), compound specs
- `_version_satisfies`: satisfied, not satisfied, hyphenated-name fallback, missing dist-info returns True
- `resolve_pypi_name`: all mapped names, LOG.warning emitted for mapped names, no warning for pass-through names
- `install_package`: full happy path, skips importable modules, SHA-256 mismatch, no compatible wheel, network error, direct dep install, circular dep guard, platform-conditional dep skipped, extra dep installed/skipped based on requested extras, extras parsed from package name, version conflict resolved to correct release, unsatisfiable conflict raises PyPIInstallError

**End-to-end tests** (`gramps/gen/utils/test/pypi_e2e_test.py`) — 12 tests, hit real PyPI, auto-skipped when offline:
- SSL connectivity and JSON parsing without certifi
- Wheel selection on real `iniconfig` and `numpy` metadata
- Full download, SHA-256 verification, and extraction of a real wheel
- Package importable after adding extract target to `sys.path`
- `install_package()` skips already-importable packages without touching the filesystem

## Test plan

- [ ] Run unit tests: `python3 -m unittest gramps.gen.utils.test.pypi_test -v`
- [ ] Run e2e tests: `python3 -m unittest gramps.gen.utils.test.pypi_e2e_test -v`
- [ ] Install an addon with a pure-Python `requires_mod` dependency from source (pip path)
- [ ] Install an addon whose dependency already has an older version installed — verify the newer version satisfying the spec is installed
- [ ] Install an addon with conflicting transitive version constraints — verify the resolver picks a version satisfying all constraints, or shows a clear error if none exists
- [ ] Install an addon with a pure-Python `requires_mod` dependency on Flatpak / pip-less Linux
- [ ] Install an addon with a pure-Python `requires_mod` dependency on Mac `.app` bundle
- [ ] Install an addon with a pure-Python `requires_mod` dependency on Windows AIO
- [ ] Install an addon with a compiled `requires_mod` dependency on Mac `.app` bundle
- [ ] Install an addon with a compiled `requires_mod` dependency on Windows AIO
- [ ] Install an addon whose `requires_mod` entry uses a Python import name (e.g. `PIL`) — verify it installs correctly and a warning is logged
- [ ] Install an addon that has `requires_mod` with a platform-conditional dep (e.g. `pywin32; sys_platform == "win32"`) — verify it is installed on Windows but skipped on Linux/macOS
- [ ] Install an addon with `depends_on` that pulls in a package with extras (e.g. `requests[security]`) — verify the security extras' deps are also installed
- [ ] Install one addon from a shared-tgz pair — verify both show as installed after refresh
- [ ] Install an addon with `depends_on` — verify the dependency is auto-installed first

### DIFF
--- a/.github/workflows/gramps-ci.yml
+++ b/.github/workflows/gramps-ci.yml
@@ -85,3 +85,20 @@ jobs:
           echo "ERROR - A new file is not listed in either POTFILES.in or POTFILES.skip";
           exit 1;
         fi
+
+  pypi-installer:
+    name: PyPI installer tests (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, windows-latest, macos-latest]
+    steps:
+    - uses: actions/checkout@v6
+    - uses: actions/setup-python@v5
+      with:
+        python-version: "3.12"
+    - name: Create minimal resource directory
+      run: python -c "import os, shutil; os.makedirs('build/share/gramps', exist_ok=True); shutil.copy('data/authors.xml', 'build/share/gramps/authors.xml')"
+    - name: Run PyPI installer tests
+      run: python -m unittest gramps.gen.utils.test.pypi_test -v
--- a/gramps/gen/utils/pypi.py
+++ b/gramps/gen/utils/pypi.py
@@ -1,0 +1,1312 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026       Doug Blank
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+
+"""
+Stdlib-only wheel installer for PyPI packages.
+
+Downloads and installs wheels from PyPI into a target directory using only the
+Python standard library.  No pip, no certifi, no external dependencies.  SSL
+uses the system trust store via ssl.create_default_context().
+
+This installer is used in two distinct situations:
+
+* **Frozen bundles** (cx_Freeze Windows AIO, macOS .app): pip is not available
+  as a subprocess.  Both pure-Python and compiled wheels are supported; the
+  compiled-wheel path selects the best wheel for the current Python version and
+  platform (Windows ``win_amd64``/``win_arm64``/``win32``; macOS
+  ``macosx_X_Y_arm64|x86_64|universal2``).
+
+* **Source / snap installs with pip**: pip is available and is used instead
+  (see ``gramps/gui/plug/_windows.py``).  ``install_package()`` is not called
+  in this code path.
+
+* **Flatpak / pip-less environments**: Flathub and other sandboxed runtimes
+  strip ``pip`` from the Python installation.  ``_pip_available()`` detects
+  this and routes to ``install_package()`` exactly as for frozen bundles.
+
+Limitations:
+  - No support for VCS URLs or local paths.
+  - Constraint gathering uses the *latest* release's dep list from the PyPI
+    JSON API; if an older version is resolved, its actual transitive deps
+    may differ slightly (tier-2 limitation).
+"""
+
+# -------------------------------------------------------------------------
+#
+# Standard Python modules
+#
+# -------------------------------------------------------------------------
+import ctypes
+import email.parser
+import functools
+import hashlib
+import importlib
+import importlib.metadata
+import io
+import json
+import logging
+import os
+import platform
+import re
+import ssl
+import subprocess
+import sys
+import urllib.request
+import zipfile
+
+# -------------------------------------------------------------------------
+#
+# Gramps modules
+#
+# -------------------------------------------------------------------------
+from ..const import GRAMPS_LOCALE as glocale
+
+_ = glocale.translation.sgettext
+
+LOG = logging.getLogger(__name__)
+
+_PYPI_JSON_URL = "https://pypi.org/pypi/{package}/json"
+_PURE_TAGS = {"py3-none-any", "py2.py3-none-any"}
+
+# Cached result of _compatible_tags() so it is only computed once.
+_COMPAT_TAGS: tuple[set[str], set[str], set[str]] | None = None
+
+# Mapping from Python import names to their PyPI distribution names.
+# Many packages have an import name that differs from the PyPI package name.
+# Addon authors declare requires_mod using the import name; this table lets
+# install_package() and pip look up the correct installable name.
+#
+# Only add entries where the import name would either:
+#  - return a 404 from PyPI (e.g. "cv2", "yaml"),
+#  - resolve to a completely different package (e.g. "serial" → a JSON lib),
+#  - or resolve to a dead/stub package with no wheels (e.g. "PIL", "sklearn").
+#
+# Do NOT add entries where the import name already maps correctly, even with
+# different capitalisation — PyPI normalises case and underscores/hyphens.
+_IMPORT_TO_PYPI: dict[str, str] = {
+    # Image processing
+    "PIL": "Pillow",
+    "cv2": "opencv-python",
+    # Data / science
+    "sklearn": "scikit-learn",
+    "yaml": "PyYAML",
+    "dateutil": "python-dateutil",
+    # Web / network
+    "bs4": "beautifulsoup4",
+    "serial": "pyserial",
+    "usb": "pyusb",
+    # Cryptography
+    "nacl": "PyNaCl",
+    "Crypto": "pycryptodome",
+    "OpenSSL": "pyOpenSSL",
+    # GUI (unlikely in headless addons, but common mismatch)
+    "wx": "wxPython",
+}
+
+# Architecture sets for manylinux legacy tag aliases (PEP 513/571/599).
+_MANYLINUX1_ARCHS: frozenset[str] = frozenset({"x86_64", "i686"})
+_MANYLINUX2010_ARCHS: frozenset[str] = frozenset({"x86_64", "i686"})
+_MANYLINUX2014_ARCHS: frozenset[str] = frozenset(
+    {"x86_64", "i686", "aarch64", "armv7l", "ppc64", "ppc64le", "s390x"}
+)
+
+# Compiled pattern for parsing a PEP 508 requirement string.
+# Captures: name, optional extras (without brackets), version spec, optional marker.
+_REQ_RE: re.Pattern[str] = re.compile(
+    r"^\s*"
+    r"(?P<name>[A-Za-z0-9](?:[A-Za-z0-9._-]*[A-Za-z0-9])?)"
+    r"\s*(?:\[(?P<extras>[^\]]*)\])?\s*"
+    r"(?P<spec>[^;]*)"
+    r"(?:;\s*(?P<marker>.+))?"
+    r"\s*$"
+)
+
+# Marker variables compared using version ordering rather than string ordering.
+_VERSION_VARS: frozenset[str] = frozenset(
+    {
+        "python_version",
+        "python_full_version",
+        "implementation_version",
+    }
+)
+
+
+# -------------------------------------------------------------------------
+#
+# Bundle detection
+#
+# -------------------------------------------------------------------------
+def is_frozen() -> bool:
+    """Return True when running inside a frozen bundle where sys.executable
+    is not a usable Python interpreter for spawning pip.
+
+    Most bundlers set sys.frozen, but some do not:
+    - macOS .app packagers may leave sys.frozen unset; we detect them by
+      checking whether sys.executable points into a .app bundle.
+    - Windows bundlers such as Nuitka and PyOxidizer produce a named app
+      binary (e.g. Gramps.exe) rather than a python*.exe interpreter; we
+      detect those by checking the executable basename.
+    """
+    if getattr(sys, "frozen", False):
+        return True
+    exe = sys.executable or ""
+    # macOS .app bundle: sys.executable is e.g. '/Foo.app' or
+    # '/Foo.app/Contents/MacOS/Gramps'
+    if exe.endswith(".app") or ".app/" in exe:
+        return True
+    # Windows bundle (Nuitka, PyOxidizer, …): sys.executable is the app
+    # binary (Gramps.exe), not a python*.exe interpreter.
+    if sys.platform == "win32":
+        basename = os.path.basename(exe).lower()
+        return not (basename.startswith("python") or basename.startswith("pypy"))
+    return False
+
+
+# -------------------------------------------------------------------------
+#
+# PyPIInstaller
+#
+# -------------------------------------------------------------------------
+class PyPIInstallError(Exception):
+    """Raised when a package cannot be installed from PyPI."""
+
+
+# -------------------------------------------------------------------------
+#
+# Internal helpers
+#
+# -------------------------------------------------------------------------
+def _ssl_context() -> ssl.SSLContext:
+    """
+    Return an SSL context that trusts the system certificate store.
+
+    On macOS the standard Python.org installer does not wire OpenSSL to the
+    macOS Keychain, so ``ssl.create_default_context()`` alone cannot verify
+    PyPI's certificate.  If the ``certifi`` package is importable its CA
+    bundle is used instead, which resolves the issue transparently.
+    """
+    try:
+        import certifi  # optional; absent in frozen bundles without certifi
+
+        return ssl.create_default_context(cafile=certifi.where())
+    except ImportError:
+        return ssl.create_default_context()
+
+
+def _ssl_cert_hint(exc: Exception) -> str:
+    """
+    Return a platform-specific hint when *exc* looks like an SSL cert error.
+
+    Returns an empty string for non-SSL errors.
+    """
+    cause = getattr(exc, "reason", exc)
+    if not (
+        isinstance(cause, ssl.SSLError) and "CERTIFICATE_VERIFY_FAILED" in str(cause)
+    ):
+        return ""
+    if platform.system() == "Darwin":
+        return _(
+            "\n\nOn macOS, Python may not have access to the system SSL "
+            "certificates.  Try one of the following:\n"
+            "  1. python.org installer: run 'Install Certificates.command' "
+            "in your Python installation folder "
+            "(e.g. /Applications/Python 3.x/).\n"
+            "  2. pyenv / Homebrew / non-system Python: set the SSL_CERT_FILE "
+            "environment variable before starting Gramps:\n"
+            "       export SSL_CERT_FILE=/etc/ssl/cert.pem\n"
+            "  3. Install the certifi package:  pip install certifi"
+        )
+    return _(
+        "\n\nSSL certificate verification failed.  This is common with "
+        "non-system Python installations (pyenv, conda) whose bundled "
+        "OpenSSL does not know where system certificates are stored.  "
+        "Try one of the following:\n"
+        "  1. Install the certifi package:  pip install certifi\n"
+        "  2. Set SSL_CERT_FILE to your system CA bundle before starting "
+        "Gramps, e.g.:\n"
+        "       export SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt  "
+        "(Debian/Ubuntu)\n"
+        "       export SSL_CERT_FILE=/etc/pki/tls/certs/ca-bundle.crt    "
+        "(RHEL/Fedora)"
+    )
+
+
+@functools.cache
+def _glibc_version() -> tuple[int, int] | None:
+    """Return ``(major, minor)`` of the system glibc, or ``None`` if not glibc.
+
+    Tries the ctypes ``gnu_get_libc_version`` symbol first (most reliable),
+    then falls back to ``platform.libc_ver()``.
+    """
+    # Primary: call gnu_get_libc_version() via ctypes.
+    try:
+        process_namespace = ctypes.CDLL(None)
+        gnu_get_libc_version = process_namespace.gnu_get_libc_version
+        gnu_get_libc_version.restype = ctypes.c_char_p
+        ver_bytes = gnu_get_libc_version()
+        if ver_bytes:
+            parts = ver_bytes.decode("ascii").split(".")
+            return int(parts[0]), int(parts[1])
+    except Exception:
+        pass
+    # Fallback: platform.libc_ver() reads the libc binary.
+    try:
+        libc_name, libc_ver = platform.libc_ver()
+        if libc_name == "glibc" and libc_ver:
+            parts = libc_ver.split(".")
+            return int(parts[0]), int(parts[1])
+    except Exception:
+        pass
+    return None
+
+
+@functools.cache
+def _musl_version() -> tuple[int, int] | None:
+    """Return ``(major, minor)`` of musl libc, or ``None`` if not a musl system.
+
+    Detects musl by looking for its dynamic linker at
+    ``/lib/ld-musl-{arch}.so.1`` and running it to parse the version line.
+    """
+    machine = platform.machine()
+    # musl uses "arm" for all 32-bit ARM variants (armv7l, armv6l, …).
+    musl_arch = "arm" if machine.startswith("armv") else machine
+    candidate = f"/lib/ld-musl-{musl_arch}.so.1"
+    if not os.path.exists(candidate):
+        return None
+    try:
+        result = subprocess.run(
+            [candidate],
+            capture_output=True,
+            text=True,
+            timeout=3,
+        )
+        output = result.stdout + result.stderr
+        m = re.search(r"Version\s+(\d+)\.(\d+)", output)
+        if m:
+            return int(m.group(1)), int(m.group(2))
+    except Exception:
+        pass
+    return None
+
+
+def _manylinux_platform_tags(
+    machine: str, glibc_major: int, glibc_minor: int
+) -> set[str]:
+    """Return all manylinux platform tags compatible with the given glibc version.
+
+    Generates PEP 600 perennial tags (``manylinux_X_Y_{arch}`` for all Y the
+    system satisfies) and the three legacy aliases where the architecture
+    qualifies (manylinux1, manylinux2010, manylinux2014).
+
+    :param machine: Value from ``platform.machine()``.
+    :param glibc_major: Major version of the system glibc.
+    :param glibc_minor: Minor version of the system glibc.
+    :returns: Set of compatible manylinux platform tag strings.
+    """
+    tags: set[str] = set()
+    if glibc_major != 2:
+        # glibc 3 does not exist yet; nothing to generate.
+        return tags
+    # PEP 600: include manylinux_2_Y for every Y the system satisfies.
+    # Oldest practical minimum is 2.5 (manylinux1).
+    for mn in range(5, glibc_minor + 1):
+        tags.add(f"manylinux_{glibc_major}_{mn}_{machine}")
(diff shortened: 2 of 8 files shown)


dsblank:
@Nick-Hall can we build an AIO to try this? @stevenyoungs could you test it to see if you can use an addon that uses `requires_mod` on Windows?

@jralls can you try on a Mac bundle?

Nick-Hall:
I have built a [Windows AIO](https://github.com/gramps-project/gramps/actions/runs/25933686167) for testing.

stevenyoungs:
> @stevenyoungs could you test it to see if you can use an addon that uses `requires_mod` on Windows?
> 
@dsblank do you know a suitable addon that I can use to test?

dsblank:
These two have "litellm" as a requires_mod:

```
ChatWithTree/ChatWithTree.gpr.py:    requires_mod=['litellm'],
GrampsChat/GrampsChat.gpr.py:    requires_mod=["litellm"],
```

After installing, in the Python Evaluation Gramplet, you should be able to "import litellm" without error.

jralls:
@dsblank:
This worked for TMG Backup, installing dbf and aenum.
 Chat With Tree Interactive Addon (ChatWithTree) fails with
```
No pure-Python wheel found for 'fastuuid'. Compiled packages cannot be installed by this installer.
```
Proving out another branch of code. Good so far.

If certificates are a problem (I had to reconfigure OpenSSL's openssldir parameter to point to `/etc/ssl`, it doesn't actually provide certificates for a non-system installation) it errors out with
```
Cannot reach PyPI for package 'dbf': <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1081)>
```
You might want to catch that exception and emit a prettier message.

dsblank:
Great progress, thank you!

emyoulation:
[NetworkChart.gpr.py](https://github.com/gramps-project/addons-source/blob/86f8bf836bb45b0a52be9ac88c81ba24919617e3/NetworkChart/NetworkChart.gpr.py#L39)
has 2 requires_mod items... if you want to verify a list of installs works as well as a single install:
"networkx", "pygraphviz"

and [lifelinechartview.gpr.py](https://github.com/gramps-project/addons-source/blob/86f8bf836bb45b0a52be9ac88c81ba24919617e3/LifeLineChartView/lifelinechartview.gpr.py#L36) has 2 registrations with 2 requires, which are repeated across another registration. (So you can check if the 1st installs are recognized and prevents a redundant attempt to install.)

They is a good examples of Addons that gets less attention because many users cannot resolve the prerequisites.

dsblank:
I wasn't happy not supporting compiled dependencies. If we support those too, then it opens up the entire world of Python. So I added them. I also added some tests for Mac and Windows. They don't test actual bundles, but do test the infrastructure surrounding the system.

dsblank:
> You might want to catch that exception and emit a prettier message.

Done!

dsblank:
@Nick-Hall, @stevenyoungs, @jralls can you test again using any and all Python modules that you can think of? Should be able to download compiled Python modules now. :crossed_fingers:

jralls:
@dsblank, will do. Note that we need to actually test the addons: Simply getting them installed [isn't enough](https://github.com/gramps-project/addons-source/pull/762#issuecomment-3259238614).

BTW, there's also [a bug](https://gramps-project.org/bugs/view.php?id=14074) about installing dependencies with addon manager.

Nick-Hall:
I've just built another [Windows AIO](https://github.com/gramps-project/gramps/actions/runs/25971781837) which includes the new commits.

jralls:
@dsblank,
Attempt to install ChatWithTree failed:
```
24172: ERROR: grampsapp.py: line 188: Unhandled exception
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/site-packages/gramps/gui/plug/_windows.py", line 308, in __on_install_clicked
    subprocess.check_output(
    ~~~~~~~~~~~~~~~~~~~~~~~^
        [
        ^
    ...<8 lines>...
        stderr=subprocess.STDOUT,
        ^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/subprocess.py", line 473, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
               **kwargs).stdout
               ^^^^^^^^^
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/subprocess.py", line 555, in run
    with Popen(*popenargs, **kwargs) as process:
         ~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/subprocess.py", line 1039, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                        pass_fds, cwd, env,
                        ^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
                        gid, gids, uid, umask,
                        ^^^^^^^^^^^^^^^^^^^^^^
                        start_new_session, process_group)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/john/Desktop/Gramps.app/Contents/Resources/lib/python3.14/subprocess.py", line 1990, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
PermissionError: [Errno 13] Permission denied: '/Users/john/Desktop/Gramps.app'
```

jralls:
@dsblank, the mac bundle doesn't set "frozen" so loading now tries to use `system.executable`, which predictably raises a privs exception. Just add
```
import sys
setattr(sys, "frozen", True)
```
to mac/gramps_launcher.py.

I worked around the wrong branch problem by calling `setattr(sys, "frozen", True) in the Python Shell gramplet, and proceeded to test:
* ChatWithTree failed: `Cannot reach PyPI for package 'mdurl~': HTTP Error 404: Not Found`
* LifeLineAncestorChart succeeds, after which the LifeLineDescendantsChart Requirements dialog shows green check marks on its dependencies. After restarting Gramps and widening the window (we need to make the charts toolbar scrollable...) the view is present and works.
* S3MediaUpload succeeds, but [this seems to be a Gramps-web thing](https://www.grampsweb.org/install_setup/s3/) so I didn't test it.
* SharedPostgresSQL fails: `No compatible wheel found for 'psycopg2' on this platform.`  [psycopg2](https://pypi.org/project/psycopg2/#files) is only built for `win-amd64`.

That's all of the addons that show up with requirements in the addons manager besides TMGImporter. 

One other comment: I don't think we want a function that's called by all platforms in a file named _windows.py. That's sure to drive us crazy.

dsblank:
@jralls:

> Note that we need to actually test the addons: Simply getting them installed isn't enough.

The new system will at least import them, but yes the addon should have some actual tests. I saw that @GaryGriffin had mentioned that before. Once we get this working, we can build some tests.

> BTW, there's also [a bug](https://gramps-project.org/bugs/view.php?id=14074) about installing dependencies with addon manager.

Maybe this won't be a problem with this mini-installer :shrug: 

Regarding:

> PermissionError: [Errno 13] Permission denied: '/Users/john/Desktop/Gramps.app'

I added some additional cases to try to catch this one. When you get chance, please try the latest.

dsblank:
@jralls (just saw your post before my last post). I think I fixed the frozen issue.

> One other comment: I don't think we want a function that's called by all platforms in a file named _windows.py. That's sure to drive us crazy.

Fixed!

> I worked around the wrong branch problem by calling `setattr(sys, "frozen", True) in the Python Shell gramplet, and proceeded to test:

...

Fix applied: _direct_dependencies() now uses 

```python
re.split(r"[><=!~\[@]", ...) 
```
instead of a chained .split() sequence. The `~=` (compatible release) specifier left a trailing `~` on the package name (e.g. mdurl~), causing a bogus PyPI 404. The regex handles all PEP 508 specifier operators in one pass and also correctly strips extras ([extra]).
  
**psycopg2** — not fixable here: psycopg2 only publishes compiled wheels for win_amd64. On macOS or Windows ARM frozen bundles the error is accurate and unavoidable — our installer can't compile from source. The right fix belongs in the SharedPostgreSQL addon itself: switching requires_mod to psycopg2-binary (which has wheels for macOS, Linux, and more Windows targets), or listing both and letting the first available one win. That's an addon change, not a pypi.py change.

jralls:
@dsblank,
> When you get chance, please try the latest.

Works. ChatWithTrees installs and runs, though there are some wrinkles about configuring it that have nothing to do with this 

>>BTW, there's also [a bug](https://gramps-project.org/bugs/view.php?id=14074) about installing dependencies with addon manager.

>Maybe this won't be a problem with this mini-installer 🤷

I think this fixes it. @GaryGriffin please try https://sourceforge.net/projects/gramps/files/Unstable/Gramps-Arm-6.1.0-mini-inst-1.dmg/download

> Once we get this working, we can build some tests.

Are you contemplating tests in one or more of the addons that would be run somehow from a running Gramps instance?

dsblank:
> Are you contemplating tests in one or more of the addons that would be run somehow from a running Gramps instance?

Yes, we should be able to have the bundles run all of the addons' tests programmatically. Although that is nothing more than a desire at this point.

GaryGriffin:
Tested the custom Gramps 6.1

- Still got the glade warning msg. I thought this was fixed. Will wait for beta2 for this.
- Addon Manager -- added Lifeline Ancestor Chart. This loaded 2 requirements (life_line_chart, svgwrite) and **addon worked**. Created the lib folder and populated it.
- Addon Manager -- added TMGimporter. This requires dbf, which requires aenum. Both loaded and **addon worked**.
- Addon Manager -- added ChatWithTree. Warnings on missing module botocore. **addon worked**.
- Addon Manager -- added S3 Uploaded. This has a requirement of boto3. **Loaded fine but not tested**
- Once S3 loaded its required boto3, ChatWithTree no longer produced warnings.

One surprise - I went to Addon Manager --> Settings and selected Allow gramps to install required Python modules. This was not effective until gramps restarted. Requires Restart should added to the Settings window (like in Gramps --> Settings --> Data) . I will file a bug.

Summary - GREAT JOB!

dsblank:
Thanks for all of the feedback and testing team! Looking forward to all of the things we can do :smile_cat:

dsblank:
We need to also check on Windows @Nick-Hall, @stevenyoungs (or anyone you might know that can test a new installer).

@Nick-Hall this is definitely a bug fix; let me know what branch you want to target once fully tested.

jralls:
@GaryGriffin 
> Still got the glade warning msg. I thought this was fixed. Will wait for beta2 for this.

Nope, the `__slots__` definition in glade.py is [still there](https://github.com/gramps-project/gramps/blob/ecbab0514135a974c96b3a259a4b822aa6f536b0/gramps/gui/glade.py#L62) (see [#14153](https://gramps-project.org/bugs/view.php?id=14153) for the diagnosis. I'll do a PR with the fix soon.

Nick-Hall:
> let me know what branch you want to target once fully tested.

Bug fixes should go into the _gramps61_ branch please.

dsblank:
This PR also fixes two other related bugs:

1. Handles `depends_on` properly
2. Updates the whole UI (rather than just the row) after an install because a tgz file may contain many addons

dsblank:
## Known limitations of the stdlib wheel installer

Eight things the current installer cannot handle, in rough order of likelihood to affect a real addon:

---

### 1. Source-only distributions (no wheel on PyPI)

The installer only unpacks `.whl` files. If a package ships only a `.tar.gz` sdist, `_pick_wheel` raises `PyPIInstallError` because there is nothing to pick.

**Examples:** `psycopg2` (use `psycopg2-binary` instead); `lxml` on platforms where no binary wheel was uploaded; any niche package where the author never ran `python -m build --wheel`.

---
### 2. <s>Pre-releases incorrectly ranked above stable releases</s>

<s>
`_version_sort_key` strips non-numeric suffixes, so `"2.0a1"` becomes `(2, 0, 0)` and `"1.9"` becomes `(1, 9)`. The alpha therefore sorts higher and `_find_satisfying_version` would pick it over the stable release.

```
releases: {"2.0a1": [...], "1.9.0": [...]}
constraint: >=1.0
resolver picks: 2.0a1  ← should be 1.9.0
```

This only affects the resolver (passes 1–2). The non-resolver path uses `meta["urls"]`, which PyPI always points at the latest *stable* release, so it is unaffected.
</s>
---

### 3. Namespace packages skipped by the wrong top-level check

`_install_one` reads `top_level.txt` from the wheel and skips installation if any listed name is already importable. Namespace packages list their root namespace, not the specific sub-package.

**Examples:** `google-cloud-storage` lists `google`; `azure-storage-blob` lists `azure`; `zope.interface` lists `zope`. If any sibling package from that namespace is already installed (e.g. `google-auth`), `import google` succeeds and the installer concludes the target package is already present, skipping extraction entirely.

---

### 4. Import names not in `_IMPORT_TO_PYPI`

Addons that declare `requires_mod` with a Python import name that is not in the mapping table will attempt to install a non-existent PyPI package.

**Examples not currently in the table:** `skimage` → should be `scikit-image`; `gi` → system package, not installable from PyPI at all. When an unmapped name is passed, `_pypi_metadata` either gets a 404 (raises `PyPIInstallError`) or returns a completely different package that happens to share the name.

---

### 5. Constraint gathering uses the latest version's dependency list

`_gather_requirements` reads `meta["info"]["requires_dist"]`, which is the *latest* release's metadata. If the resolver picks an older version, the transitive deps that were gathered are those of the newer version — the actual deps of the resolved version may differ.

**Example:** constraint `requests>=1.0,<2.0` resolves to `1.9.2`, but transitive deps were gathered from `requests 2.31.0` (latest), which declares `urllib3>=1.21.1,<3`; the actual `1.9.2` deps have different bounds.

---

### 6. <s>Yanked releases not skipped</s>
<s>
PyPI can mark a release as "yanked" (known-bad). The PyPI JSON API includes yanked releases in `meta["releases"]` and `_find_satisfying_version` iterates them without checking the `yanked` field.
</s>

---

### 7. <s>`Requires-Python` not checked</s>

<s>
The `Requires-Python` field in wheel METADATA is not validated. A wheel declaring `Requires-Python: >=3.12` with `py3-none-any` filename tags (compatible with any Python) will be extracted on Python 3.9 without complaint and then fail at import time.
</s>

---

### 8. No support for VCS URLs, local paths, or custom indexes

```
git+https://github.com/org/pkg@main   # not supported
file:///home/user/mypackage            # not supported
--index-url https://my.internal.pypi  # not supported
```

Only `https://pypi.org/pypi/{package}/json` is consulted.

---

### Summary

| # | Limitation | Likely to affect addons? | Workaround |
|---|---|---|---|
| 1 | Source-only sdist (no wheel) | Occasionally | Use `-binary` variant if available |
| 2 | <s>Pre-releases ranked above stable<s> | Rarely | None currently |
| 3 | Namespace packages skipped | Rarely | None currently |
| 4 | Missing import→PyPI name mapping | Occasionally | Add entry to `_IMPORT_TO_PYPI` |
| 5 | Stale transitive deps for old resolved version | Rarely | Tier-2 fix (future) |
| 6 | <s>Yanked releases not skipped</s> | Very rarely | None currently |
| 7 | <s>`Requires-Python` not validated</s> | Rarely | None currently |
| 8 | No VCS / local / private-index support | Essentially never | None (by design) |

dsblank:
@Nick-Hall it would be great to start testing this in Gramps 6.1. If it does get included, we should checkout to make sure that any older addons with dependencies don't break, and that the new ones work as designed.

Nick-Hall:
Rebased and squashed.

Added 2 missing files to `POTFILES.skip`.

Nick-Hall (approved):

PR #2310, opened 2026-05-16
Spanish relationship calculator overhaul

niedon:
Complete overhaul of Spanish relationship calculator:
- Added specific grandparent and grandchild Spanish words until generation +/- 13.
- Improved letter ordinals from 21 to 100.
- Changed the compound formulation of "second cousin of a great-grandparent" an so on into a direct Spanish accepted naming. The old one was technically correct, but Spanish supports a more direct naming.
- Fixed step descendant and ascendant step relationship applying correct Spanish names (e.g. bisnietastro) instead of the previously generated placeholder "segundo nietastro" (-astro is the Spanish suffix for step relationships, thus the non-step-relationship "segundo nieto" is not a valid form, it would be "bisnieto").
- Fixed a wrong generation in the section "testing nieces" and "testing nephews" from step relationships, as it was being generated with the naming tía/tío (aunt/uncle) instead of sobrino/sobrina (niece/nephew) for generation from +3.
- Fixed in-law generation as it was generating "político" for females, whose word would be "política".
- Fixed some specific Spanish words with suffix -astro for close in-law relatives.
- Fixed more complex but close Spanish relationship like consuegro/consuegrastro concuñado/concuñadastro.
- Translated partner types.
- Fixed decimonono -> decimonoveno (decimonono was not wrong but it is less used than the latter).

### DIFF
--- a/gramps/plugins/rel/rel_es.py
+++ b/gramps/plugins/rel/rel_es.py
@@ -9,7 +9,7 @@
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation; either version 2 of the License, or
-# (at your option) any latr version.
+# (at your option) any later version.
 #
 # This program is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


Nick-Hall:
@jcsaaver Does this change look good to you?

jcsaaver:
> @jcsaaver Does this change look good to you?

The terms 'pentabuelo', 'hexabuelo', etc. are more or less new, so not widely used. At least in academic contexts, forms like 'quinto abuelo' are preferred.

niedon:
@jcsaaver I know, but I had no chance as those ordinal forms would produce relations like "tío quinto abuelo segundo" or "tío abuelo quinto segundo". With this change I pushed those cases back 10 generations.

PR #2313, opened 2026-05-17
Don't use slots on GObject-derived classes.

jralls:
It conflicts with the PyGobject memory management, causing crashes especially with PyGobject 3.55.0 and later, see [PyGobject Issue 734](https://gitlab.gnome.org/GNOME/pygobject/-/issues/734). PyGobject 3.55.3 and later [deletes `__slots__` namespaces with a warning](https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/508). 

This PR removes the `__slots__` declaration from `class Glade`. To preserve the `__slots__` behavior of preventing ad-hoc member variable creation it adds a `__setattr__` override that raises an `AttributeException` if a new name isn't one of the coded in values. 

Fixes [#14153](https://gramps-project.org/bugs/view.php?id=14153).

### DIFF
--- a/gramps/gui/glade.py
+++ b/gramps/gui/glade.py
@@ -59,7 +59,14 @@ class Glade(Gtk.Builder):
     Glade class: Manage glade files as Gtk.Builder objects
     """
 
-    __slots__ = ["__toplevel", "__filename", "__dirname"]
+    def __setattr__(self, name, value):
+        if not hasattr(self, name) and name not in [
+            "__toplevel",
+            "__filename",
+            "__dirname",
+        ]:
+            raise AttributeError(f"Ad-hoc attribute {name} is not permitted.")
+        super().__setattr__(self, name, value)
 
     def __init__(self, filename=None, dirname=None, toplevel=None, also_load=[]):
         """


lordemannd:
it works for me on Windows 6.0.8 and on debian with 6.0.7

dsblank (approved):
Nice workaround!

gramps/gui/glade.py
```diff
@@ -59,7 +59,14 @@ class Glade(Gtk.Builder):
     Glade class: Manage glade files as Gtk.Builder objects
     """
 
-    __slots__ = ["__toplevel", "__filename", "__dirname"]
+    def __setatter__(self, name, value):
```
hgohel:
```suggestion
    def __setattr__(self, name, value):
```

Is it a typo?

jralls: Yes. One I made about 50 times while writing this PR, no less.

Fixed. Thanks for catching it.

petr-fedorov:
@Nick-Hall @jralls 
It appears that it does not  work:
Снимок экрана от 2026-06-02 19-02-46

Probably it should not work 'as is' because of [attribute name mangling](https://docs.python.org/3/reference/expressions.html#index-5) ?
Снимок экрана от 2026-06-02 19-02-10

hgohel:
+1 to the exception reported by @petr-fedorov in a `gramps61` branch build `6.1.0-beta1-d6a61b942f` on Windows MSYS2 UCRT environment, with Python 3.14.

PR #2314, opened 2026-05-17
Skip bsddb import tests when neither berkeleydb nor bsddb3 is installed

eduralph:
## Root cause
`db_load` in `gramps/plugins/test/imports_test.py` calls `make_database("bsddb")`, which raises a generic `Exception: can't load database backend: 'bsddb'` when neither `berkeleydb` (the maintained successor) nor `bsddb3` (deprecated since Python 3.6, unavailable on Windows) is installed. The catch-all `except Exception` in `db_load` swallows that and reports it as `AssertionError: Cannot open database …`, so the legacy BSDDB import tests fail on every SQLite-default install instead of skipping cleanly.

## Fix
Probe `berkeleydb` then `bsddb3` before invoking `make_database` for the `bsddb` backend; if neither is importable, call `self.skipTest(...)`. Add `except SkipTest: raise` ahead of the catch-all so the skip isn't reconverted into an assertion failure. Behaviour is unchanged on systems with either backend installed.

## Verified against
- gramps-project/gramps@6235c3ba3a:gramps/plugins/db/bsddb/bsddb.py:32-35 — the plugin already tries \`berkeleydb\` first and falls back to \`bsddb3\`, so the probe order in the test mirrors what the plugin loader does.

## Test
The change is to the test file itself. Verified locally on a Linux Python 3.12 install with neither package installed:
- **Before**: \`FAILED (failures=5, skipped=56)\` — the five \`test_imp_{3_4_5,4_1_3,4_2_8,5_0_1bsd,5_1_2bsd}_zip\` tests fail with \`AssertionError: Cannot open database … can't load database backend: 'bsddb'\`.
- **After**: \`OK (skipped=61)\` over 32499 tests; the same five now report \`skip: neither berkeleydb nor bsddb3 is installed — skipping legacy bsddb import path\`.

Note: 6.1 has a related Windows-only skip (`29a0bedaa0 Skip bsddb tests on Windows`) that this PR generalises; a forward-merge from \`maintenance/gramps60\` will need to drop that hunk in favour of this one.

### DIFF
--- a/gramps/plugins/test/imports_test.py
+++ b/gramps/plugins/test/imports_test.py
@@ -26,6 +26,7 @@
 # **********************************************
 
 import unittest
+from unittest import SkipTest
 import os
 import sys
 import re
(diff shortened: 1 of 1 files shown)


eduralph:
Closing as superseded. `maintenance/gramps61` already carries what this PR proposed: `gramps/plugins/test/imports_test.py` there already skips the bsddb backend (`if dbid == "bsddb" and win()`) and already has the `SkipTest` import plus the `except SkipTest: raise` re-raise handling. The only remaining delta would be changing the skip condition from `win()` to a berkeleydb/bsddb3 import probe, which isn't worth a standalone PR.

hgohel:
@eduralph I'm running all UTs on Ubuntu 24.02 locally and the same five tests are failing as you described:

> `test_imp_{3_4_5,4_1_3,4_2_8,5_0_1bsd,5_1_2bsd}_zip` tests fail with `AssertionError: Cannot open database … can't load database backend: 'bsddb'`.

> Closing as superseded. 

Yes, in gramps61 branch the tests are skipped on Windows but this change would be needed for Linux, no?

eduralph:
@hgohel - let me verify, I might have been too quick to dismiss this

PR #2315, opened 2026-05-19
Fix fan chart crash at startup with age-gradient background

eduralph:
## Root cause
`FanChartWidget.set_generations` and `FanChart2WayWidget.set_generations` initialised `self.data[i]` with a shared `(None,) * 4` tuple per slot, leaving the `userdata` element as `None`. When `_fill_data_structures` short-circuits (no `rootpersonh` yet, e.g. on startup with "Remember last view displayed" enabled, or after a tree rename), the slots are never replaced, and the `BACKGROUND_GRAD_AGE` / `BACKGROUND_GRAD_PERIOD` paths in `prepare_background_box` then crash with `AttributeError: 'NoneType' object has no attribute 'append'`.

## Fix
Replace the buggy initialiser with a list comprehension that gives each slot its own empty `userdata` list, in both fan-chart variants. The invariant — every `data[i][j][3]` is a list, never `None` — now holds regardless of whether `_fill_data_structures` runs to completion. The list-comprehension form also avoids the latent shared-reference bug the `[(...,) * 4] * 2**i` pattern would have caused if any code path relied on per-slot mutation.

## Verified against
- gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/gui/widgets/fanchart.py:1517 (ancestor variant — buggy line)
- gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/gui/widgets/fanchart2way.py:228 (2-way variant — same buggy pattern)
- gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/gui/widgets/fanchartdesc.py:204 (descendant variant — already safe; uses `[]` literal)

## Test
New `gramps/gui/widgets/test/fanchart_test.py` with three cases:
1. `TestFanChartWidgetSetGenerations.test_userdata_is_list_in_every_data_slot` — asserts the invariant on the ancestor chart.
2. `TestFanChartWidgetSetGenerations.test_each_data_slot_has_its_own_userdata_list` — guards against the shared-reference latent bug.
3. `TestFanChart2WayWidgetSetGenerations.test_ascendance_userdata_is_list_in_every_data_slot` — same invariant on the 2-way chart's ascendance side.

Tests use `FanChartWidget.__new__(cls)` to bypass `Gtk.DrawingArea.__init__` so they run headlessly under `python3 -m unittest gramps.gui.widgets.test.fanchart_test`.

Verification matrix:
| State | Result |
|---|---|
| Pre-fix (stash the source changes) | 2 assertion failures + 1 error reproducing the exact production stack: `AttributeError: 'NoneType' object has no attribute 'append'` |
| Post-fix | 3/3 pass |

## Related Mantis reports
Same NoneType.append fingerprint via different triggers — this fix should resolve them too, but I've left them for separate verification rather than auto-closing here:
- Issue #12932 — "When 'remember last view displayed' is enabled" (same path)
- Issue #13591 — "Cannot launch Gramps anymore"
- Issue #14076 — closed; same fingerprint via tree rename

### DIFF
--- a/gramps/gui/widgets/fanchart.py
+++ b/gramps/gui/widgets/fanchart.py
@@ -1513,8 +1513,12 @@ class FanChartWidget(FanChartBaseWidget):
         elif self.form == FORM_QUADRANT:
             self.rootangle_rad = [math.radians(180), math.radians(270)]
         for i in range(self.generations):
-            # person, parents?, children?
-            self.data[i] = [(None,) * 4] * 2**i
+            # person, parents?, children?, userdata
+            # Each slot gets its own empty userdata list so that
+            # prepare_background_box can safely call userdata.append()
+            # even when _fill_data_structures short-circuits (e.g. on
+            # startup before rootpersonh is set). See bug 0013395.
+            self.data[i] = [(None, None, None, []) for _ in range(2**i)]
             self.angle[i] = []
             angle = self.rootangle_rad[0]
             portion = 1 / (2**i) * (self.rootangle_rad[1] - self.rootangle_rad[0])
--- a/gramps/gui/widgets/fanchart2way.py
+++ b/gramps/gui/widgets/fanchart2way.py
@@ -224,8 +224,12 @@ class FanChart2WayWidget(FanChartWidget, FanChartDescWidget):
         self.angle = {}
         self.data = {}
         for i in range(self.generations_asc):
-            # name, person, parents?, children?
-            self.data[i] = [(None,) * 4] * 2**i
+            # person, parents?, children?, userdata
+            # Each slot gets its own empty userdata list so that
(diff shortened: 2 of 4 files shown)


hgohel:
@eduralph Did you mean to include AGENTS.md in this PR?

eduralph:
@hgohel  - no, will fix this

eduralph:
@hgohel  - it was a stray update in my fork, sorry for that. Corrected now and ready for completion.

hgohel (commented):
While I didn't dive into the code fix, the change fixes the bug.

Nick-Hall:

PR #2316, opened 2026-05-19
Fix list view crash on column-sort click after closing the database

eduralph:
## Root cause
When the database is closed the framework switches the active model to an empty state by clearing `self.tree`, so the root entry `self.tree[None]` no longer exists. `TreeBaseModel.reverse_order` did the lookup unconditionally, so the next column-header click in any list view backed by a `TreeBaseModel` raised `KeyError: None`.

## Fix
Short-circuit `reverse_order` when the tree has no root entry: there is nothing to reverse in a cleared state, and the user's expected behaviour is "no action because no tree is open". The internal `__reverse` flag is still toggled, so the user's sort-direction intent is captured next time data is loaded.

## Verified against
- gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/gui/views/treemodels/treebasemodel.py:727 (buggy unconditional `self.tree[None]` lookup)
- gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/gui/views/listview.py:759 (column_clicked entry point; reaches the model via `self.model.reverse_order()`)

## Test
New `gramps/gui/views/treemodels/test/treebasemodel_test.py` exercises `reverse_order` on a `__new__`-constructed `TreeBaseModel` with empty `self.tree`.

Verification matrix:
| State | Result |
|---|---|
| Pre-fix | `KeyError: None` at `treebasemodel.py:735` — same line and same exception as the reported crash |
| Post-fix | passes |

## Related Mantis reports
Same `KeyError: None` fingerprint, left for separate verification rather than auto-closing:
- Issue #13610
- Issue #13967 (closed; marked as duplicate of #13214 in Mantis)
- Issue #12448
- Issue #2092 — umbrella "Problems when no database is open"

### DIFF
--- a/gramps/gui/views/treemodels/test/treebasemodel_test.py
+++ b/gramps/gui/views/treemodels/test/treebasemodel_test.py
@@ -1,0 +1,70 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Gramps developers
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <https://www.gnu.org/licenses/>.
+#
+
+"""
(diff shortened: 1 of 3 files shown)


hgohel:
@eduralph This PR also includes AGENTS.md, should it?

eduralph:
@hgohel - no absolutely not. I'll remove it from the PR ...

eduralph:
@hgohel - same as #2315 ; ready now

hgohel (approved):
Looks good to me!

I'm not understanding how the commit linked in "Verified against" section of the notes is related to this PR.

Nick-Hall:

PR #2317, opened 2026-05-19
Graph plugins: preserve date modifier in years-only mode (bug 4658)

eduralph:
Fixes https://gramps-project.org/bugs/view.php?id=4658 (MantisBT bug 0004658).

## Root cause

When the "Limit dates to years only" report option is on, both Family Lines Graph and Relationship Graph rendered modified or compound dates as a bare year — `before 1923` became `1923`, `between 1923 and 1925` became `1923`, `estimated about 1923-06-15` became `1923`. The defect is `Date(date.get_year())`, which invokes the `Date` constructor's int-source branch and builds a fresh date with `MOD_NONE`/`QUAL_NONE` and a single year, discarding the modifier (before/after/about), the quality (estimated/calculated), and the second stop of compound dates.

## Fix

Introduce a private `_year_only_date` helper in each plugin that copies the original date via `Date.copy_ymd` (the existing `Date` API for "copy with new y/m/d", which preserves modifier and quality) and zeroes month and day; for compound dates it additionally truncates the second stop to year-only via `set2_yr_mon_day`. Replace all four `Date(date.get_year())` call sites:

- `gramps/plugins/graph/gvfamilylines.py` — birth, death, wedding label strings.
- `gramps/plugins/graph/gvrelgraph.py` — `get_date_string` for any event when `event_choice` is in {4, 5}.

The helper is intentionally **duplicated per plugin** rather than placed in a new shared module under `gramps/plugins/graph/` — there is no precedent for `_*.py` helpers in plugin directories today, and a one-line difference between the two copies would silently regress one of the plugins. A `test_helpers_agree` test guards against the copies drifting.

The sister Relationship Graph plugin is included in this PR because gburto01 flagged in [the bug's 2011-02-27 note](https://gramps-project.org/bugs/view.php?id=4658#c0017855) that the buggy idiom was originally written in Relationship Graph and ported across to Family Lines Graph for harmonisation. Fixing only the named plugin would leave the same defect alive in the sibling.

## Verified against

- gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/gen/lib/date.py:665-709 — `Date.__init__` int-source branch sets `modifier = MOD_NONE`, `quality = QUAL_NONE`, single-stop only. Confirms the root cause.
- gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/gen/lib/date.py:1507-1515 — `Date.copy_ymd` is the existing API for "copy with new year/month/day" and preserves modifier/quality via the `Date(self)` copy.
- gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/gen/lib/date.py:1428-1436 — `set2_yr_mon_day` is the API for setting the second stop of a compound date.
- gburto01's [comment on 2011-02-27](https://gramps-project.org/bugs/view.php?id=4658#c0017855) confirming the Relationship → Family Lines port.

## Test

Regression test at `gramps/plugins/graph/test/year_only_date_test.py` — 10 cases:

- Modifier preservation: `MOD_BEFORE`, `MOD_AFTER`, `MOD_ABOUT`.
- Quality preservation: `QUAL_ESTIMATED`, `QUAL_CALCULATED`.
- Month/day truncation (year preserved).
- Compound `MOD_RANGE` / `MOD_SPAN` second-stop preserved.
- Plain unmodified date regression guard (behaviour unchanged for the common case).
- Both plugin helpers agree byte-for-byte via `Date.serialize` across all of the above.

Run:

```
  python3 -m unittest gramps.plugins.graph.test.year_only_date_test
```

End-to-end check with `gramps.gen.datehandler.displayer`:

```
before 1923                          -> before 1923                  (was: 1923)
between 1923-06-15 and 1925-09-20    -> between 1923 and 1925        (was: 1923)
estimated about 1923-06-15           -> estimated about 1923         (was: 1923)
```

### DIFF
--- a/gramps/plugins/graph/gvfamilylines.py
+++ b/gramps/plugins/graph/gvfamilylines.py
@@ -98,6 +98,31 @@ _CORNERS = [
     {"name": _("Both"), "value": "fm"},
 ]
 
+
+# ------------------------------------------------------------------------
+#
+# Helper functions
+#
+# ------------------------------------------------------------------------
+def _year_only_date(date: Date) -> Date:
+    """
+    Return a copy of *date* truncated to its year, preserving the
+    modifier (before/after/about) and quality (estimated/calculated).
+
+    Unlike ``Date(date.get_year())`` -- which builds a fresh date with
+    ``MOD_NONE``/``QUAL_NONE`` and drops the second stop of compound
+    dates -- this preserves the original modifier and quality, and
+    for ranges/spans preserves the second-stop year as well. That is
+    what bug 4658 asks for: "before 1923" must stay "before 1923",
+    not collapse to "1923".
+    """
+    if date.is_compound():
+        result = date.copy_ymd(date.get_year(), 0, 0, remove_stop_date=False)
+        result.set2_yr_mon_day(date.get_stop_year(), 0, 0)
+        return result
+    return date.copy_ymd(date.get_year(), 0, 0)
+
+
 # ------------------------------------------------------------------------
 #
 # A quick overview of the classes we'll be using:
@@ -896,7 +921,7 @@ class FamilyLinesReport(Report):
             if bth_event and self._incdates:
                 date = bth_event.get_date_object()
                 if self._just_years and date.get_year_valid():
-                    birth_str = self.get_date(Date(date.get_year()))  # localized year
+                    birth_str = self.get_date(_year_only_date(date))
                 else:
                     birth_str = self.get_date(date)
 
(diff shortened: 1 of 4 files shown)


Nick-Hall:

Nick-Hall:
@eduralph Please use a special keyword in the [bug tracker integration](https://gramps-project.org/wiki/index.php/Committing_policies#Bug_tracker_integration) rather than "Closes".

eduralph:
> @eduralph Please use a special keyword in the [bug tracker integration](https://gramps-project.org/wiki/index.php/Committing_policies#Bug_tracker_integration) rather than "Closes".

You found the one where it's wrong :-) , will take extra care going forward

PR #2318, opened 2026-05-19
Graph reports: raise ReportError when dot/gs write fails (bug 6556)

eduralph:
Fixes https://gramps-project.org/bugs/view.php?id=6556 (MantisBT bug 0006556) — bug-class symptom only; the reporter's additional ask in note ~0038649 for a pre-flight "try open output for write" check remains a separate feature ask.

## Root cause

Graphviz-based reports (Family Lines Graph, Relationship Graph, Hourglass Graph and the other GV*Doc formats) failed *silently* when the chosen output file could not be written — typically because another process (a viewer on Windows) held a lock on it, but also under permission denied / disk full / missing binary. The user saw no error and no file, and had no way to learn the report had not been generated. Text and graphical reports already surfaced this via `Could not create %s` (`ReportError`); only the graphviz path was silent.

[graphdoc.py](gramps/gen/plug/docgen/graphdoc.py) shells out to `dot` (and to `gs` for the multi-page PDF path) via `os.system(...)` at 11 sites, and **discards the return code at every one of them**. Any failure to write the output is invisible to the rest of the function.

## Fix

Route every `os.system` invocation through a new module-level helper:

```python
def _check_rc(rc: int, target: str) -> None:
    if rc != 0:
        raise ReportError(_("Could not create %s") % target)
```

The wording matches the message uniformly used by the other docgen backends ([rtfdoc.py:104](gramps/plugins/docgen/rtfdoc.py#L104), [svgdrawdoc.py:92](gramps/plugins/docgen/svgdrawdoc.py#L92), [odfdoc.py:1194](gramps/plugins/docgen/odfdoc.py#L1194)), so downstream error-handling and the user-visible dialog see the same text regardless of which output format the user picked.

The 11 call sites span 8 doc classes:
- `GVPsDoc`, `GVSvgDoc`, `GVSvgzDoc`, `GVPngDoc`, `GVJpegDoc`, `GVGifDoc`, `GVPdfGvDoc` — one `os.system` each (dot → final output).
- `GVPdfGsDoc` — four sequential `os.system` calls (dot → tmp ps, then `gs` in single-page / per-piece / multi-page-merge variants).

Each site keeps its existing cleanup-then-check ordering so temp files generated *before* the failure point still get removed where they were already being removed. The minimum-delta approach preserves the file's existing cleanup semantics; one pre-existing inconsistency (the single-page PdfGs path leaks `tmp_dot`) is intentionally out of scope.

## Verified against

- `gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/gen/plug/docgen/graphdoc.py:804,838,873,908,943,978,1018,1064,1078,1133,1140` — the 11 `os.system(...)` callsites that discarded the return code at the time of the bug report (lines listed at the base SHA).
- `gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/gen/errors.py:70-85` — `ReportError(value, value2="")` is the class other docgen backends raise on write failure; same exception type used by gramps' user-visible dialog handling.
- `gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/plugins/docgen/rtfdoc.py:104`, `:gramps/plugins/docgen/svgdrawdoc.py:92`, `:gramps/plugins/docgen/odfdoc.py:1194` — canonical "Could not create %s" message wording matched by this fix.
- PaulFranklin's [comment on 2014-10-12](https://gramps-project.org/bugs/view.php?id=6556#c0038588) confirming the OS file-locking diagnosis and asking for at minimum a user-visible error message — which is exactly what this fix delivers.

## Test

New `gramps/gen/plug/docgen/test/graphdoc_test.py` — two groups, 6 cases:

- `CheckRcTest` — four cases asserting the helper is silent on `rc=0`, raises `ReportError` on non-zero, names the target in the message, and treats *any* non-zero (incl. POSIX-encoded high-byte exits like 256 and signal-encoded negatives) as failure.
- `GraphdocStaticAssertions` — two source-grep guards on `graphdoc.py`: (1) no bare `os.system(...)` callsites remain, and (2) every `rc = os.system(...)` assignment is paired with a `_check_rc(rc, ...)` call. These catch a regression where a future edit adds an `os.system` call without the check, which would silently reopen this bug.

Run:

```
  python3 -m unittest gramps.gen.plug.docgen.test.graphdoc_test
```

End-to-end check (manual repro of the bug + fix):

```
_check_rc(0,   "/tmp/ok.svg")     -> silent       (success path; was: silent, unchanged)
_check_rc(256, "/tmp/locked.svg") -> ReportError: "Could not create /tmp/locked.svg"
                                                  (was: silent, no file, no error)
```

### DIFF
--- a/gramps/gen/plug/docgen/graphdoc.py
+++ b/gramps/gen/plug/docgen/graphdoc.py
@@ -47,6 +47,7 @@ from subprocess import Popen, PIPE
 from ...const import GRAMPS_LOCALE as glocale
 
 _ = glocale.translation.gettext
+from ...errors import ReportError
 from ...utils.file import search_for, where_is
 from . import BaseDoc
 from ..menu import NumberOption, TextOption, EnumeratedListOption, BooleanOption
@@ -127,6 +128,22 @@ else:
 
 def esc(id_txt):
     return id_txt.replace('"', '\\"')
+
+
+def _check_rc(rc: int, target: str) -> None:
+    """
+    Raise :class:`ReportError` if *rc* (the return code from
+    ``os.system`` running a graphviz/ghostscript invocation) is
+    non-zero, so that the user sees a "Could not create <file>" error
+    instead of a silently-missing output (Mantis bug 0006556).
+
+    The non-zero exit typically means the target file was locked by
+    another process (the historical Windows symptom), but it also
+    covers permission denied, disk full, an absent ``dot``/``gs``
+    binary, or a malformed temporary DOT file.
+    """
+    if rc != 0:
+        raise ReportError(_("Could not create %s") % target)
 
 
 # ------------------------------------------------------------------------------
@@ -800,9 +817,10 @@ class GVPsDoc(GVDocBase):
             self.vpages * self.hpages
         ) > 1:
             command = command.replace(":cairo", "")
-        os.system(command)
+        rc = os.system(command)
         # Delete the temporary dot file
         os.remove(tmp_dot)
+        _check_rc(rc, self._filename)
 
 
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


Nick-Hall:

PR #2321, opened 2026-05-19
Russian dates: fix stale .f attribute name in dd_dformat05 Lexeme guard (bug 14231)

eduralph:
Fixes https://gramps-project.org/bugs/view.php?id=14231 (MantisBT bug 0014231).

## Root cause

jralls's [commit `7f8455c867`](https://github.com/gramps-project/gramps/commit/7f8455c867) "Extract translation classes from grampslocale.py" (2023-07-14) renamed the `Lexeme` attribute used to access inflected forms from `.f` to `.forms`. That commit updated every format-spec usage in the Russian date handler and every other guard:

| Location | Updated to `.forms` in 7f8455c867? |
|---|---|
| `_date_ru.py:142` (long-month guard in `dd_dformat04`) | ✅ |
| `_date_ru.py:147` (long-month format spec) | ✅ |
| `_date_ru.py:170` (short-month format spec) | ✅ |
| **`_date_ru.py:165` (short-month guard in `dd_dformat05`)** | ❌ **missed** |

The short-month guard still reads:

```python
elif not hasattr(short_months[date_val[1]], "f"):  # not a Lexeme
```

After the rename, `Lexeme` has `.forms` but no `.f`. So `hasattr(any_Lexeme, "f")` always returns `False`, the `not …` is always `True`, and execution always falls into the no-Lexeme str-fallback branch at lines 166-168. The inflected branch at lines 170-172 — which the same commit *specifically* updated to use `.forms[Р]` — is **dead code**.

## Effect

Russian short-format date display silently skips grammatical inflection even when the Russian translations are fully installed. Users see e.g. `9 декабрь 1945` (nominative abbreviation) instead of the grammatically-correct `9 декабря 1945` (genitive case).

This is the *opposite* failure mode of [bug 12283 / PR #2320](https://github.com/gramps-project/gramps/pull/2320) (which fixed the *missing*-translation case in Finnish): here the guard fires too eagerly when it shouldn't, rather than failing to fire when it should.

## How this was discovered

While auditing the bug-12283 pattern across all locale handlers — checking whether any other locale shared Finnish's missing-guard hole. The audit found:

- Only `_date_fi.py` and `_date_ru.py` use the `.forms[X]` format spec at all.
- The bug-12283 pattern itself doesn't appear in Russian (the day-zero path in `dd_dformat04` routes through `format_long_month_year` which is independently guarded).
- But the stale-attribute-name defect at `_date_ru.py:165` turned up during the same read-through.

No corresponding Mantis ticket existed when this was discovered. Filing one separately so the fix is cross-referenced upstream.

## Fix

Change `"f"` to `"forms"` on `_date_ru.py:165`. One-character change in the source; the resulting guard matches the sibling at line 142 word-for-word.

## Verified against

- `gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/gen/datehandler/_date_ru.py:165` — the stale `"f"` guard this PR fixes.
- `gramps-project/gramps@7f8455c86745a5c2fd1fb1bf05601f46b16fe979` — the 2023 rename commit that missed this line. The commit's own diff against `_date_ru.py` correctly updated lines 142, 147, 170 to `.forms`, but didn't touch 165.
- `gramps-project/gramps@6235c3ba3abe25940ca7beb2eed261d18d6641b4:gramps/gen/utils/grampstranslation.py:159-162` — `Lexeme.forms` is declared via `@property`, and there is no `.f` attribute on the class. Confirms the `hasattr("f")` check is unconditionally False post-rename.

## Test

New `gramps/gen/datehandler/test/date_ru_test.py` — 4 cases driving `dd_dformat05` and `dd_dformat04` directly:

- `test_full_date_uses_genitive_form_when_lexemes_present` — the primary bug regression: with a `Lexeme`-bearing `short_months` tuple, the genitive `Р` form must appear in the output.
- `test_genitive_form_across_all_months` — same regression across all 12 months via `subTest`.
- `test_full_date_with_non_lexeme_short_months_falls_back` — regression guard for the legitimate str-fallback case (no translation installed).
- `test_dd_dformat04_long_form_unaffected` — regression guard for the sibling long-month path that was already correct in 7f8455c867.

The test discriminates between the str-fallback and inflected paths by placing the *nominative* form first in the `Lexeme`'s ordered dict — since `Lexeme` inherits from `str` and uses its first form as the str value, the bug-path (str fallback) renders the nominative, while the fix-path (inflected `forms[Р]`) renders the genitive. Without that ordering choice both paths render the same text and the test couldn't catch the regression.

Pre-fix verification: with the stash-and-rerun pattern, the test correctly produces 13 failures (1 + 12 subtests) of the form `AssertionError: 'GEN12' not found in '9 NOM12 1945'` — exactly the user-visible symptom.

Run:

```
  python3 -m unittest gramps.gen.datehandler.test.date_ru_test
```

4/4 pass post-fix; the broader datehandler test suite (52 existing tests) is also still green.

### DIFF
--- a/gramps/gen/datehandler/_date_ru.py
+++ b/gramps/gen/datehandler/_date_ru.py
@@ -162,7 +162,7 @@ class DateDisplayRU(DateDisplay):
                 )
         elif date_val[1] == 0:  # month is zero but day is not (see 8477)
             return self.display_iso(date_val)
-        elif not hasattr(short_months[date_val[1]], "f"):  # not a Lexeme
+        elif not hasattr(short_months[date_val[1]], "forms"):  # not a Lexeme
             return "{day:d} {short_month} {year}".format(
                 day=date_val[0], short_month=short_months[date_val[1]], year=year
             )
--- a/gramps/gen/datehandler/test/date_ru_test.py
+++ b/gramps/gen/datehandler/test/date_ru_test.py
@@ -1,0 +1,167 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Gramps Development Team
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <https://www.gnu.org/licenses/>.
+#
+
+"""
+Unit test for the Russian short-format date display.
+
+In jralls's 2023-07-14 commit ``7f8455c867`` "Extract translation
+classes from grampslocale.py", the ``Lexeme`` attribute used to
+access inflected forms was renamed from ``.f`` to ``.forms``. That
+commit updated every format-spec usage and almost every guard --
+but missed the guard at ``_date_ru.py:165`` in
+``DateDisplayRU.dd_dformat05``. The result is that ``hasattr(lexeme,
+"f")`` always returns False (no ``Lexeme`` has ``.f`` after the
+rename), so the no-Lexeme fallback always runs and the inflected
+``.forms[Р]`` branch is dead code. Russian short-format dates
+therefore skip inflection even when the Russian translations are
+installed -- the *opposite* failure mode of bug 12283 (which is
+about the *missing*-translation case).
+
+The same idiom in the sibling ``dd_dformat04`` (line 142) was
+correctly migrated, as were the format-spec accesses on lines 147
+and 170. Only line 165 was forgotten.
+"""
+
+import unittest
+
+from ...utils.grampslocale import GrampsLocale
+from ...utils.grampstranslation import Lexeme
+from .._date_ru import DateDisplayRU
+
+
+class RussianDateDFormat05InflectionTest(unittest.TestCase):
+    """
+    Exercise ``DateDisplayRU.dd_dformat05`` with a ``Lexeme`` short
+    month list (the translation-installed scenario) and verify the
+    Russian-genitive (``Р``) form is selected, not the default str
+    coercion.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        # The Russian translation catalog isn't required for this test
+        # since we feed dd_dformat05 a synthetic Lexeme tuple directly.
+        cls.dd = DateDisplayRU(blocale=GrampsLocale(lang="ru"))
+
+    @staticmethod
+    def _lexeme_months(genitive_prefix="GEN", nominative_prefix="NOM"):
+        """Build a tuple of 13 entries (index 0 unused, 1..12 are
+        Lexemes carrying a Russian-genitive (``Р``) form and a
+        nominative fallback).
+
+        The nominative form is intentionally placed *first* in the
+        ordered dict, so that ``str(lexeme)`` (the bug path's
+        fallback view, since Lexeme inherits from str and uses its
+        first form as the str value) yields the nominative -- making
+        the bug-path output distinguishable from the inflected-path
+        output that uses the genitive ``.forms[Р]``. Without this
+        ordering, both paths would render the same string and the
+        test couldn't catch the regression.
+        """
+        return ("",) + tuple(
+            Lexeme(
+                [
+                    ("И", f"{nominative_prefix}{m}"),
+                    ("Р", f"{genitive_prefix}{m}"),
+                ]
+            )
+            for m in range(1, 13)
+        )
+
+    def test_full_date_uses_genitive_form_when_lexemes_present(self):
+        """
+        Bug regression: with a Lexeme-bearing ``short_months`` and a
+        day-non-zero date, ``dd_dformat05`` must emit the genitive
+        (``Р``) form. Pre-fix the guard at ``_date_ru.py:165`` checked
+        ``hasattr(..., "f")`` which never matches post-2023 Lexemes,
+        so the function silently returned the nominative.
+        """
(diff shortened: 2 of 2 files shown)


Nick-Hall:
@petr-fedorov Does this look OK to you?

dsblank:
@eduralph If I have understood @Nick-Hall correctly, bug fixes should target `maintenance/gramps61` now.

petr-fedorov:
> @petr-fedorov Does this look OK to you?

@Nick-Hall  While the justification of this patch appears to be correct, in reality it is not needed.   Short months names are not inflected in Russian, so the proposed change will not change anything -  `.forms` have the same month names for all inflections for all months, for example:
Снимок экрана от 2026-05-20 12-49-54

eduralph:
> > @petr-fedorov Does this look OK to you?
> 
> @Nick-Hall While the justification of this patch appears to be correct, in reality it is not needed. Short months names are not inflected in Russian, so the proposed change will not change anything - `.forms` have the same month names for all inflections for all months, for example: Снимок экрана от 2026-05-20 12-49-54

@petr-fedorov - apologies, I went off the code as I don't speak Russian. I worked off the code logic, not the language. The blanket "Russian short-format date display silently skips grammatical inflection."  is an overclaim and you rightly pushed back on. 

It did challenge me to look closer at it and  I found one case where it becomes an issue: May.

The inflected forms[Р] branch is dead code. For 11 of 12 months this has no visible effect — po/ru.po gives those months identical short forms across all cases (янв/янв/янв …), because abbreviating truncates the word before its case ending. May is the exception: po/ru.po stores May as И=май|Р=мая|Т=маем — the word is too short to truncate, so the genitive мая survives. A May date in the abbreviated-month format therefore renders 9 май 1945 (nominative) instead of the correct genitive 9 мая 1945. One month, one date format — which is why it went unreported.

Dead code branch for all 12 months, user-visible difference only for May - Lol.

eduralph:
> @eduralph If I have understood @Nick-Hall correctly, bug fixes should target `maintenance/gramps61` now.

@dsblank - is there a reference page where I can check such things? I worked of the Mantis Bug Tracking which has a 6.0.9 release entry so I assumed bug fixed would still flow into there. 

I'm happy to rebase into the 6.1 branch, so no worries - just looking for a way to stay ontop of such things.

petr-fedorov:
> > > @petr-fedorov Does this look OK to you?
> > 
> > 
> > @Nick-Hall While the justification of this patch appears to be correct, in reality it is not needed. Short months names are not inflected in Russian, so the proposed change will not change anything - `.forms` have the same month names for all inflections for all months, for example: Снимок экрана от 2026-05-20 12-49-54
> 
> @petr-fedorov - apologies, I went off the code as I don't speak Russian. I worked off the code logic, not the language. The blanket "Russian short-format date display silently skips grammatical inflection." is an overclaim and you rightly pushed back on.
> 
> It did challenge me to look closer at it and I found one case where it becomes an issue: May.
> 
> The inflected forms[Р] branch is dead code. For 11 of 12 months this has no visible effect — po/ru.po gives those months identical short forms across all cases (янв/янв/янв …), because abbreviating truncates the word before its case ending. May is the exception: po/ru.po stores May as И=май|Р=мая|Т=маем — the word is too short to truncate, so the genitive мая survives. A May date in the abbreviated-month format therefore renders 9 май 1945 (nominative) instead of the correct genitive 9 мая 1945. One month, one date format — which is why it went unreported.
> 
> Dead code branch for all 12 months, user-visible difference only for May - Lol.

@eduralph This is wrong:  И=май|Р=**мая**|Т=маем ). Libreoffice Calc, for example, does not inflect short_dates in Russian:
Снимок экрана от 2026-05-20 22-12-09

eduralph:
> > > > @petr-fedorov Does this look OK to you?
> > > 
> > > 
> > > @Nick-Hall While the justification of this patch appears to be correct, in reality it is not needed. Short months names are not inflected in Russian, so the proposed change will not change anything - `.forms` have the same month names for all inflections for all months, for example: Снимок экрана от 2026-05-20 12-49-54
> > 
> > 
> > @petr-fedorov - apologies, I went off the code as I don't speak Russian. I worked off the code logic, not the language. The blanket "Russian short-format date display silently skips grammatical inflection." is an overclaim and you rightly pushed back on.
> > It did challenge me to look closer at it and I found one case where it becomes an issue: May.
> > The inflected forms[Р] branch is dead code. For 11 of 12 months this has no visible effect — po/ru.po gives those months identical short forms across all cases (янв/янв/янв …), because abbreviating truncates the word before its case ending. May is the exception: po/ru.po stores May as И=май|Р=мая|Т=маем — the word is too short to truncate, so the genitive мая survives. A May date in the abbreviated-month format therefore renders 9 май 1945 (nominative) instead of the correct genitive 9 мая 1945. One month, one date format — which is why it went unreported.
> > Dead code branch for all 12 months, user-visible difference only for May - Lol.
> 
> @eduralph This is wrong: И=май|Р=**мая**|Т=маем ). Libreoffice Calc, for example, does not inflect short_dates in Russian: Снимок экрана от 2026-05-20 22-12-09

@petr-fedorov - I'm hardly in the position to argue about your language, right? I worked off [Wiktionary](https://en.wiktionary.org/wiki/%D0%BC%D0%B0%D0%B9#Russianl) & [Russificate](https://russianschoolrussificate.com/tip/russian-genitive-nominative-case-tell-date/) when looking into this and it said you have to use Genitiv when using that construction, but to be honest - no idea and if you and Libreoffice say that's the right declension in that situation, who am I to argue? 

I was interested in the dead code branch, not trying to tell a native speaker how is language works, lol.

I close this and the bug report.

dsblank:
> is there a reference page where I can check such things? I worked of the Mantis Bug Tracking which has a 6.0.9 release entry so I assumed bug fixed would still flow into there.

I thought @Nick-Hall declared as much somewhere, but I can't find it. Only he knows.

Nick-Hall:
Correct.  The _gramps60_ branch is now closed.

I would have created the 6.0.9 version in Mantis as part of the 6.0.8 release procedure.  We can move bugs that target 6.0.9 to 6.1.0.

PR #2322, opened 2026-05-20
Fall back to uninflected date format on an unknown inflection (bug 14100)

eduralph:
## Root cause

`DateDisplay.format_long_month_year()` and its short / month-only siblings
look up the modifier-derived `inflect` key in `FORMATS_long_month_year` /
`FORMATS_short_month_year` with a bare subscript. `display_formatted()`
builds that key from a translatable string (`_("", "about-date")` and
friends), and Finnish `fi.po` translates the `about-date` context to the
native word "noin" instead of the English keyword the dict is keyed by — so
displaying an `about <Month> <Year>` date under a long-month format
evaluates `FORMATS_long_month_year["noin"]` and raises `KeyError: 'noin'`.

## Fix

Look the inflection key up with `dict.get()`, falling back to the
uninflected (`""`) format that is always present in both dicts, so a
mistranslated locale degrades to a readable (if grammatically plain) date
instead of aborting report or view rendering.

## Verified against

- `gramps/gen/datehandler/_datedisplay.py:182` and `:244` — the `""`
  (uninflected) key is unconditionally defined in `FORMATS_long_month_year`
  and `FORMATS_short_month_year`, so it is a safe fallback key.
- `gramps/gen/datehandler/_datedisplay.py:505-509` — `display_formatted()`
  derives the key from `_("", "about-date")`, whose translator note asks for
  the English keyword or no translation, not a native-language word.
- `gramps/plugins/view/relview.py:1440` — `info_string()` reaches the same
  `get_date()` unguarded, so the Relationship View crashes on such a date
  too, not only the Detailed Ancestral Report named in the report.

## Test

`gramps/gen/datehandler/test/datedisplay_test.py` —
`DateDisplayInflectionFallbackTest`: an inflection key absent from the
`FORMATS_*` dicts must fall back to the uninflected format. The four cases
(long/short month-year and long/short month) raise `KeyError` before this
change and pass after it.

Fixes [#14100](https://gramps-project.org/bugs/view.php?id=14100).

### DIFF
--- a/gramps/gen/datehandler/_datedisplay.py
+++ b/gramps/gen/datehandler/_datedisplay.py
@@ -577,24 +577,29 @@ class DateDisplay:
             return "{long_month} {year}".format(
                 long_month=long_months[month], year=year
             )
-        return self.FORMATS_long_month_year[inflect].format(
-            long_month=long_months[month], year=year
-        )
+        # bug 14100: an inflection key not present in the dict (e.g. from a
+        # locale whose .po mistranslated the inflection context) falls back
+        # to the uninflected format ("") instead of raising KeyError.
+        return self.FORMATS_long_month_year.get(
+            inflect, self.FORMATS_long_month_year[""]
+        ).format(long_month=long_months[month], year=year)
 
     def format_short_month_year(self, month, year, inflect, short_months):
         if not hasattr(short_months[1], "forms"):  # not a Lexeme: no inflection
             return "{short_month} {year}".format(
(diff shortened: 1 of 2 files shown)


hgohel:
Fixes the unhandled exception. Thanks.

Nick-Hall:
Yes.  It's a good idea to protect against translators not following instructions.

Nick-Hall:
Rebased.

PR #2325, opened 2026-05-22
Hourglass Graph: link spouse in descendant traversal (bug 9628)

eduralph:
## Root cause

`HourGlassReport.traverse_down` (`gramps/plugins/graph/gvhourglass.py:163`)
linked focal → family → children but never added the family's **other**
parent. `traverse_up` already handles both — its father-section and
mother-section each add the parent node and an edge to the family. The
descending side was asymmetric and every descendant generation's spouse
was invisible. Mantis bug 9628.

Concrete repro on `example.gramps`, center person `I00044` (Lewis Anderson
Garner), `maxdescend=2, maxascend=0`, format `dot`:

- Pre-fix DOT: Lewis → F00017 → eight children, then each child's
  family + grandchildren — but no Luella Jacques Martel and no spouse
  for any of the eight children either.
- 7 spouses absent: Luella (I00045), Frances Reed (I00053), Viola Taylor
  (I00641), George Яковлев (I00645), Maude García (I00650), Cora Jackson
  (I01107), Clarence Robinson (I00654).

## Fix

In `traverse_down`, after adding the family node and the focal-to-family
edge, derive the family's other parent (whichever of
`family.get_father_handle()` / `get_mother_handle()` isn't the current
`person`), add that person to the graph, and link them to the family
with the same arrow style as the focal-to-family edge. The existing
`__used_people` set keeps the spouse from being re-added if a
cousin-marriage path brings them through later.

## Verified against

- `gramps/plugins/graph/gvhourglass.py:163` — `traverse_down` body
- `gramps/plugins/graph/gvhourglass.py:193` — `traverse_up` (the existing
  symmetric father+mother section this fix mirrors)

## Test

`gramps/plugins/test/reports_test.py` adds
`test_hourglass_graph_includes_spouse_mantis_9628`. It invokes the
report on `data.gramps` with `pid=I0027` (Smith Ingeman, has spouse
Ericsdotter Marta + child Smith Martin), `maxdescend=1`, `maxascend=0`,
and asserts that the produced DOT contains `Ericsdotter`.

- Pre-fix: fails with `'Ericsdotter' not found in <DOT>` — the DOT
  contains only Ingeman, the family node, and the child Martin.
- Post-fix: passes — Marta appears as a person node linked to F0001.

Fixes #9628

### DIFF
--- a/gramps/plugins/graph/gvhourglass.py
+++ b/gramps/plugins/graph/gvhourglass.py
@@ -175,6 +175,30 @@ class HourGlassReport(Report):
                 head=self.arrowheadstyle,
                 tail=self.arrowtailstyle,
             )
+            # Add the spouse (the other parent of this family) so the
+            # descendant tree shows both parents of each generation,
+            # mirroring traverse_up's symmetric father+mother linking.
+            # Without this, only the focal-side ancestor line appears
+            # and the spouse of every descendant is invisible (Mantis
+            # 9628).
+            father_handle = family.get_father_handle()
+            mother_handle = family.get_mother_handle()
+            if father_handle == person.handle:
+                spouse_handle = mother_handle
+            elif mother_handle == person.handle:
+                spouse_handle = father_handle
+            else:
+                spouse_handle = None
+            if spouse_handle and spouse_handle not in self.__used_people:
+                self.__used_people.append(spouse_handle)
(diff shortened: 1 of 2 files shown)


Nick-Hall:
Rebased.

PR #2326, opened 2026-05-22
keep gen/db/upgrade.py off gramps.gui/cli imports (bug 6085)

eduralph:
## Summary
**User impact:** When Gramps upgrades an old family tree, the end-of-upgrade "Upgrade Statistics" message popped up as a stray window not attached to the main Gramps window, and even the text-only command-line version of Gramps had to load graphical code it should never need — a long-standing tangle where the shared core reached up into the graphical and command-line layers.

This untangles the core so it no longer depends on those layers: the statistics message is now shown by the graphical layer attached to the main window, and the command line skips the popup entirely.

Reported in Mantis [#6085](https://gramps-project.org/bugs/view.php?id=6085).

## What to look at
The shared core (`gramps/gen/`) is meant to have no dependence on the graphical or command-line front-ends; this removes the last two such imports. To see the user-visible half: run a database upgrade in the GUI and confirm the "Upgrade Statistics" dialog now appears parented to the main window; run the same upgrade on the command line and confirm no dialog pops up. A source-grep test asserts `gen/` imports nothing from `gramps.gui`/`gramps.cli`.

## Root cause
bmcage's 2012 audit (Mantis 6085) listed five files under `gramps/gen/` that imported from `gramps.gui` or `gramps.cli` — the gen submodule is the GUI-agnostic core and was never meant to depend on either. Four sites were cleared over the years; `gen/db/upgrade.py` kept two:

- `from gramps.cli.clidbman import NAME_FILE` (line 39) — the `"name.txt"` constant.
- `from gramps.gui.dialog import InfoDialog` (line 44) — the end-of-upgrade statistics dialog popped from `gramps_upgrade_16`, carrying a `# TODO no-parent` comment from whoever first wrote it.

## Fix
**`NAME_FILE` moves to `gramps/gen/db/dbconst.py`** (the existing database-constants module) and joins its `__all__`. `gramps/cli/clidbman.py` keeps a `from gramps.gen.db.dbconst import NAME_FILE` re-export so the existing callers — `gramps/gui/dbman.py`, `gramps/plugins/db/bsddb/bsddb.py`, and any third-party plugin importing from there — keep working unchanged.

**`gramps_upgrade_16`** now stores the upgrade-statistics text on `self.upgrade_summary` instead of popping a dialog. The GUI caller `gramps/gui/dbloader.py`, after `db.load(…)` returns successfully, reads `db.upgrade_summary` and displays it via `InfoDialog` — and because the call site is now in the GUI layer the dialog gets a proper transient parent (the main window), fixing the historical `# TODO no-parent`. The CLI flow never touches `db.upgrade_summary` and so gets the cleaner no-dialog behaviour it should have had all along.

After the change, `grep -rn -E '^(from gramps\.(gui|cli)|import gramps\.(gui|cli))' gramps/gen/` returns no matches — `gen/` is fully clean for the first time since bmcage filed the bug.

## Verification
- **Claim:** `gramps/gen/` imports nothing from `gramps.gui` or `gramps.cli`; the upgrade-statistics text is handed to the GUI to display (parented to the main window) and left undisplayed on the CLI.
- **Checked:** on master — `gramps/gen/db/upgrade.py:1322` (`gramps_upgrade_16` summary assignment, was `InfoDialog(_("Upgrade Statistics"), txt, …)`), `gramps/gui/dbloader.py:213` (display point after `db.load()`), `gramps/cli/clidbman.py:71` (re-export of `NAME_FILE` from gen), `gramps/gen/db/dbconst.py:48` (`NAME_FILE` declaration). After the change, `grep -rn -E '^(from gramps\.(gui|cli)|import gramps\.(gui|cli))' gramps/gen/` returns no matches.
- **Test:** `gramps/gen/db/test/gen_db_upgrade_imports_test.py` — architectural guard that walks every `.py` under `gramps/gen/` (excluding `test/` subdirs) and asserts no `from gramps.(gui|cli)` / `import gramps.(gui|cli)` statement exists. Pre-fix it fails with `['gen/db/upgrade.py:39: from gramps.cli', 'gen/db/upgrade.py:44: from gramps.gui']`; post-fix it passes. It is intentionally a source-grep rather than an upgrade-run test — the bug class is import architecture, and a runtime test would need a v15-schema database fixture (not in the test data) and a heavyweight db-upgrade harness that does not exist; the grep catches the same regression class with O(file-count) speed and no fixtures.

Fixes [#6085](https://gramps-project.org/bugs/view.php?id=6085)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2,6 +2,83 @@
 
 This document specifies rules and conventions that agents should follow when making changes to the Gramps codebase.
 
+## Commands
+
+**Run from source (no install):**
+```bash
+python3 Gramps.py
+```
+
+**Build wheel:**
+```bash
+python3 -m build --wheel
+```
+
+**Run all tests:**
+```bash
+export GDK_BACKEND=-
+export GRAMPS_RESOURCES=build/share
+export LANGUAGE= 
+export LANG=en_US.utf-8 
+python3 -m unittest discover -p "*_test.py"
+```
+
+**Run a single test module or method:**
+```bash
+python3 -m unittest gramps.gen.db.test.db_test
+python3 -m unittest gramps.gen.db.test.db_test.TestGenericDb.test_get_person
+```
+
+**Type checking:**
+```bash
+mypy
+```
+
+**Code formatting (must pass CI):**
+```bash
+black .
+# Check only:
+black --check .
+```
+
+## Architecture
+
+Gramps is a GTK4 genealogy application. The codebase is split into four main packages:
+
+### `gramps/gen/` — Core (no GUI dependency)
+- `lib/` — Data model: `Person`, `Family`, `Event`, `Place`, `Source`, `Citation`, `Repository`, `Media`, `Note`, `Tag`. All inherit from `PrimaryObject` (has a database handle) or `SecondaryObject`.
+- `db/` — Database abstraction layer. `DbReadBase` / `DbBase` are the abstract interfaces. `generic.py` is the DBAPI-based implementation used by SQLite, DuckDB, MySQL, PostgreSQL backends. Transactions via `DbTxn`; undo/redo built-in.
+- `plug/` — Plugin system core: `BasePluginManager` (singleton), `PluginRegister`, plugin metadata, and base `Plugin` class.
+- `filters/`, `datehandler/`, `display/` — Shared utilities with no GUI imports.
+
+### `gramps/gui/` — GTK4 frontend
+- `editors/` — Modal dialogs for editing each primary object type.
+- `views/` — Navigation views (People, Families, Events, etc.) that populate the main window.
+- `grampsgui.py` — Main window and view-switching logic.
+- `dbman.py` — GUI database manager (open/close/import/export).
+- `configure.py` — Preferences dialog.
+- `glade/` — GTK UI definition files (`.glade` / `.ui`).
+
+### `gramps/plugins/` — All plugin implementations
+- `db/` — Database backends (bsddb legacy, dbapi/SQLite default, duckdb, mysql, postgresql).
+- `gramplet/` — ~50 dashboard widgets (mini read-only views). Each is a self-contained gramplet.
+- `importer/`, `export/` — GEDCOM, XML, CSV, and other formats.
+- `view/`, `tool/`, `textreport/`, `drawreport/`, `quickview/`, `sidebar/` — Other plugin categories.
+
+### `gramps/cli/` — Command-line interface
+- `grampscli.py` — Non-GUI entry point (import/export/script automation).
+- `arghandler.py` — Argument parsing and dispatch.
+
+### Plugin system
+Plugins are discovered from `~/.local/share/gramps/grampsx.x/plugins/` (user) and the system prefix. Each plugin registers itself via a `gpr.yaml` file or a `register()` call. The `BasePluginManager` singleton loads and manages them. Categories: Importers, Exporters, DB backends, Gramplets, Views, Tools, Reports, Quick views, Sidebars, Map services.
+
+### Database backends
+The default backend is DBAPI (SQLite via `gramps/plugins/db/dbapi/`). All backends implement the `DbReadBase`/`DbBase` interface from `gramps/gen/db/`. Schema migrations live in `gramps/gen/db/upgrade.py`.
+
+## CI
+
+The CI pipeline (`.github/workflows/gramps-ci.yml`) runs: build wheel → `mypy` → `unittest discover`. The `black.yml` workflow checks formatting on every push. Both must pass before merging.
+
 ## Code Style & Formatting
 
 ### Black
@@ -24,7 +101,16 @@ All functions and methods must have type hints using Python 3.10+ syntax:
 
 ### Docstrings
 
-All functions and methods must have docstrings using Sphinx format:
+All functions and methods must have docstrings. Keep them concise — a short description is almost always enough:
+
+```python
+def close(self):
+    """
+    Close the specified database.
+    """
+```
+
+Use the full Sphinx format (`:param:`, `:type:`, `:returns:`, `:rtype:`) only when the parameters or return value are non-obvious and not already expressed by type hints:
 
 ```python
 def get_person_from_gramps_id(self, gramps_id: PersonGrampsID) -> Person | None:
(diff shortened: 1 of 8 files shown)


dsblank:
## Code Review

### Overview

This PR fixes a long-standing architectural violation (Mantis 6085): `gramps/gen/db/upgrade.py` imported both `gramps.cli.clidbman.NAME_FILE` and `gramps.gui.dialog.InfoDialog`, which broke the invariant that `gramps/gen/` must be GUI-agnostic. The fix is two-pronged:

1. Move `NAME_FILE` into `gramps/gen/db/dbconst.py` with a backward-compat re-export from `clidbman.py`.
2. Replace the `InfoDialog(...)` call in `gramps_upgrade_16` with `self.upgrade_summary = txt`, and let the GUI caller (`dbloader.py`) display the dialog after `db.load()` returns.

An architectural guard test is included to prevent regression.

---

### ✅ What's Done Well

- **Root-cause analysis is thorough** — the PR description is unusually clear and traces the history back to bmcage's 2012 audit.
- **`NAME_FILE` placement is correct** — `dbconst.py` is exactly the right home for a database-layer filename constant.
- **Backward-compat re-export** — `gui/dbman.py`, `plugins/db/bsddb/bsddb.py`, and third-party code importing `NAME_FILE` from `clidbman` continue to work unchanged.
- **`# TODO no-parent` is finally resolved** — the dialog now correctly receives a transient parent via `self.uistate.window`.
- **Architectural guard test** is well-structured, fast, and the failure message is diagnostic (shows the offending file:line).
- **mypy passes** on all changed files.
- **The new test passes** and verifies the fix.

---

### ⚠️ Issues

#### 1. `upgrade_summary` is not declared on the database class — medium

`gramps_upgrade_16` sets `self.upgrade_summary = txt` dynamically on the db instance (`gramps/gen/db/upgrade.py:1328`), but this attribute is not declared anywhere in `DbGeneric.__init__` or `DbReadBase`. The consumer in `dbloader.py` works around this with `getattr(db, "upgrade_summary", None)` (line 227), which is a code smell acknowledging that the attribute may not exist.

**Suggestion:** Declare `upgrade_summary: str | None = None` in `DbGeneric.__init__` (or `DbReadBase.__init__`). Then `dbloader.py` can use `db.upgrade_summary` directly and reset it to `None` without `getattr`. This makes the contract explicit and signals to future maintainers that this is an intentional communication channel.

#### 2. CLI silently discards upgrade statistics — minor

The PR description says "The CLI flow never touches `db.upgrade_summary` and so gets the cleaner no-dialog behaviour." That's correct, but the upgrade statistics (object counts, merge suggestion) are now silently dropped in CLI use. Previously, `InfoDialog` from CLI context would have been unreachable on headless runs in any case, so this is not a regression. Still, it could be useful to at least log the summary at `LOG.info(...)` inside `gramps_upgrade_16` so CLI users see it in verbose output.

#### 3. Missing copyright line in test file — minor

Per project convention, every new `.py` file must include `# Copyright (C) <YEAR>  <Author Name>` in the file header. `gramps/gen/db/test/gen_db_upgrade_imports_test.py` has the GPL boilerplate but the copyright line is absent.

#### 4. `# noqa: F401` comment in `clidbman.py` — trivial

The `# noqa: F401 (re-export)` comment suppresses a flake8 unused-import warning. The project uses pylint rather than flake8. The cleanest solution would be to add `NAME_FILE` to an `__all__` list in `clidbman.py`, which makes the re-export intent explicit to both linters and readers without any suppression comment.

---

### Summary

The fix is architecturally sound and correctly resolves the 14-year-old violation. The main thing worth addressing before merge is declaring `upgrade_summary` formally on the db class rather than relying on `getattr` with a default — that's the only change that touches correctness/API clarity. The missing copyright and noqa comment are small style nits. The CLI-silent-drop is intentional per the author and acceptable.

eduralph:
Summary of what changed in response to the four review items:

- Medium — [generic.py:662-668](vscode-webview://1igs668572jq5jpgsdq2oabm2sdudu4tj6b0kboh0fuk4iue863l/gramps/gramps/gen/db/generic.py#L662-L668) declares self.upgrade_summary: str | None = None; [dbloader.py:227-235](vscode-webview://1igs668572jq5jpgsdq2oabm2sdudu4tj6b0kboh0fuk4iue863l/gramps/gramps/gui/dbloader.py#L227-L235) now reads db.upgrade_summary directly without getattr.
- Minor — [upgrade.py:1328](vscode-webview://1igs668572jq5jpgsdq2oabm2sdudu4tj6b0kboh0fuk4iue863l/gramps/gramps/gen/db/upgrade.py#L1328) adds LOG.info(txt) so CLI/headless runs surface the summary at verbose log level.
- Minor — [gen_db_upgrade_imports_test.py:4](vscode-webview://1igs668572jq5jpgsdq2oabm2sdudu4tj6b0kboh0fuk4iue863l/gramps/gramps/gen/db/test/gen_db_upgrade_imports_test.py#L4) gets the Copyright (C) 2026 Eduard Ralph line.
- Trivial — [clidbman.py:78](vscode-webview://1igs668572jq5jpgsdq2oabm2sdudu4tj6b0kboh0fuk4iue863l/gramps/gramps/cli/clidbman.py#L78) uses from … import NAME_FILE as NAME_FILE (PEP 484 explicit re-export); # noqa: F401 is gone. (Used PEP 484 form instead of introducing an __all__ to a module that previously had none — same effect, no wildcard-import semantics change. Easy to swap to __all__ if the reviewer prefers that mechanism after seeing it.)

dsblank:
Thanks for the quick iteration — the second commit addresses the main concerns from the previous round (declaring `upgrade_summary` formally in `DbGeneric.__init__`, the PEP 484 re-export form, and the missing copyright line). A few smaller items remain:

---

**1. Verbose comment block in `gramps/cli/clidbman.py` (lines 70–77)**

The 8-line comment before the re-export reads more like a commit message than a code comment. The essential information fits in two lines:

```python
# Canonical location is gramps.gen.db.dbconst (Mantis 6085).
# ``as NAME_FILE`` is the PEP 484 explicit re-export form for mypy/pylint/flake8.
from gramps.gen.db.dbconst import NAME_FILE as NAME_FILE
```

---

**2. Missing class header comment in the test file (`gen_db_upgrade_imports_test.py`, line 73)**

Per project convention, every class needs a navigation header:

```python
#------------------------------------------------------------
#
# GenImportBoundaryTest
#
#------------------------------------------------------------
class GenImportBoundaryTest(unittest.TestCase):
```

---

**3. Test `"test"` directory skip only matches the leaf directory name (line 85)**

```python
if os.path.basename(dirpath) == "test":
    continue
```

This correctly skips `gramps/gen/foo/test/` but would not skip files in `gramps/gen/foo/test/fixtures/` (or any subdirectory of a `test/` dir). A slightly more robust check would be:

```python
if "test" in os.path.relpath(dirpath, GEN_DIR).split(os.sep):
    continue
```

This is low-risk today since the test dirs have no subdirectories, but worth tightening.

---

**4. `change_database()` fires before the upgrade dialog is shown (`dbloader.py`, lines 219 vs 227–234)**

`change_database()` emits signals that notify the rest of the UI the database is open. The upgrade statistics `InfoDialog` is then shown afterward, so the main window briefly appears fully open before the blocking dialog appears. The original `InfoDialog` call (inside `db.load()`) actually showed *before* `change_database()`. Consider moving the `if db.upgrade_summary:` block to just before `self.dbstate.change_database(db)` — the summary is already set on `db` at that point and moving it would restore the original UX ordering.

---

**5. `LOG.info(txt)` for upgrade statistics**

The upgrade statistics text is multi-line and verbose — more appropriate for `LOG.debug()` than `LOG.info()`. Using `INFO` means it appears in default log output for every user who upgrades, whereas `DEBUG` keeps it available for those actively diagnosing. The surrounding upgrade log calls (`LOG.debug([data_upgradeobject])` at line 1306, `LOG.debug(...)` at line 1230) all use `DEBUG`.

eduralph:
@dsblank - cleaned up as asked

Nick-Hall:
Please rebase on the _master_ branch and add missing files to `POTFILES.skip` or `POTFILES.in`.

PR #2327, opened 2026-05-23
GEDCOM import: warn on bare-numeric DATE values (bug 9298)

eduralph:
## Root cause
GEDCOM 5.5.1 requires three-letter month abbreviations (`7 NOV 1959`), but tools such as ancestry.com export bare numeric `D/M/YYYY` dates the spec does not define. `libgedcom.GedcomDateParser` hardcodes `dhformat = "%m/%d/%y"` (en_US locale, `libgedcom.py:925`), so a bare numeric date is unconditionally read MM/DD/YYYY. When the source actually uses DD/MM/YYYY and the day is ≤ 12 (both readings syntactically valid), the importer silently swaps day and month and emits no warning — `GEDCOM import report: No errors detected` despite a wrong date being written to the database. Day > 12 already falls back to text-only storage (visually flagged in the Events view as bold), so only the day ≤ 12 case is silently dangerous.

This was reported in 2016 and re-confirmed in 2022/2023 across multiple discourse threads; multiple developers have discussed narrower-scope fixes (warn-only, source-detect ancestry, switch in import dialog) but none have shipped. No code-side parse-direction change is possible without breaking US users who currently rely on the existing MM/DD/YYYY default for legitimately-shaped numeric input.

## Fix
The smallest defensible change: emit an advisory warning per ambiguous bare-numeric date during import, no parse-direction change. US users see no behaviour change; European users importing ancestry.com-shaped GEDCOMs get a per-event audit list with both the original GEDCOM text and the chosen interpretation.

- `BARE_NUMERIC_DATE` regex added next to the other module-level GEDCOM patterns (`libgedcom.py:894`).
- `GedcomParser.__add_warn` (new) wraps `__add_msg` with the `number_of_errors -= 1` pattern already used at the end of `__finish_import` for the unknown-references note; the summary line stays "No errors detected" when the warnings are advisory.
- `GedcomParser.__warn_if_ambiguous_numeric_date` (new) is called from `__event_date` after the date object is attached to the event. Suppresses on `MOD_TEXTONLY` (already visually flagged) and on non-numeric inputs. `line` is intentionally not passed through `__add_warn` because `__add_msg` truncates the problem string to 66 columns when a line is supplied, which drops the actionable half of the message — and the warning text already contains the original GEDCOM date string verbatim.

Scope is limited to `__event_date` (the only date sink the bug report cites). `__name_date` / `__lds_ord_date` / `__addr_date` / `__citation_date` / `__media_date` could call the same helper in a follow-up if a maintainer asks; left out here to keep the diff focused on the cited symptom.

## Verified against
- `gramps/plugins/lib/libgedcom.py:925` — the hardcoded `dhformat = "%m/%d/%y"` that drives the silent swap.
- `gramps/plugins/lib/libgedcom.py:3797` — existing precedent for the `__add_msg` then `number_of_errors -= 1` pattern used to attach an advisory note without polluting the error count.
- `gramps/gen/lib/date.py:590` — `Date.MOD_TEXTONLY = 6` (the fallback used when the parser can't make sense of the input, recognised in the helper to suppress the warning).

## Test
`gramps/plugins/importer/test/importgedcom_ambiguous_date_test.py` imports an Ancestry-shaped GEDCOM fixture via `import_as_dict` with a `CapturingUser` that intercepts the summary dialog text. Four assertions:

1. `7/11/1959` (day ≤ 12, the silent-swap case) triggers an "Ambiguous numeric date" warning naming the original text and the chosen interpretation.
2. `5 MAR 1899` (spec-compliant) stays silent.
3. `25/3/1934` (day > 12, text-only fallback) stays silent.
4. The summary still reads "GEDCOM import report: No errors detected".

Pre-fix, test 1 fails (`'Ambiguous numeric date' not found in 'GEDCOM import report: No errors detected'`); the other three pass. Post-fix all four pass.

Fixes #9298

### DIFF
--- a/gramps/plugins/importer/test/importgedcom_ambiguous_date_test.py
+++ b/gramps/plugins/importer/test/importgedcom_ambiguous_date_test.py
@@ -1,0 +1,130 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <https://www.gnu.org/licenses/>.
+#
+
+"""Mantis 9298 — GEDCOM importer must warn on bare-numeric DATE values.
+
+GEDCOM 5.5.1 requires three-letter month abbreviations (``7 NOV 1959``),
+but some tools (notably ancestry.com exports) emit bare numeric
+``D/M/YYYY`` dates that the spec does not define.  ``libgedcom`` parses
+them as ``MM/DD/YYYY`` per :class:`GedcomDateParser.dhformat_changed`,
+which silently swaps day and month for sources that use ``DD/MM/YYYY``.
+The fix adds a warning per affected event so the user can audit; it does
+not change the parse direction (US users still get the same result they
+always did).
+
+Asserts:
+
+  * Bare numeric date with day ≤ 12 (the silent-swap case) triggers an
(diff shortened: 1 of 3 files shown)


Nick-Hall:

PR #2329, opened 2026-05-23
Add regression test for RelationshipPathBetween filter rule (bug 13820)

eduralph:
## Root cause
`RelationshipPathBetween.init_list` regressed in 1280aa45a5f47581badfb655021ce1d430f8a581 when `for person_handle in firstList & secondList` became `for person_handle in firstList and secondList` — Python's short-circuit `and` returns `secondList`, so `firstMap[person_handle]` raised `KeyError` on the first handle reachable from the second root but not the first. The rule has no direct test (only the `RelationshipPathBetweenBookmarks` wrapper is tested), which is why the regression slipped through and could reappear.

## Fix
Test-only. Adds `test_relationshippathbetween` to `gramps/gen/filters/rules/test/person_rules_test.py`, exercising the rule with two example.gramps persons (`I0044`, the home person, and `I2127`, a member of the unrelated Δεληπέτρου family) that share no ancestor — the configuration that triggered the original `KeyError`.

## Verified against
- `gramps/gen/filters/rules/person/_relationshippathbetween.py:127` — current `init_list` uses `firstSet & secondSet`, confirming the fix is live on this branch.
- Pre-fix file (checked out at `48a6cbfb0541d429a3e3fba778c611f6e1843a6a^`): the new test fails with `KeyError` at `_relationshippathbetween.py:130` — same line and exception class as in Mantis 13830's traceback.
- Regression-introducing commit: 1280aa45a5f47581badfb655021ce1d430f8a581 ("Refactor, fix, and optimize filters/rules").
- Fix commit (already on this branch): 48a6cbfb0541d429a3e3fba778c611f6e1843a6a ("Fix regression in relationship path between people filter").

## Test
`test_relationshippathbetween` — passes on current branch (88/88 in `person_rules_test.py` pass), fails on the pre-fix file with the original `KeyError`.

Issue #13830.

### DIFF
--- a/gramps/gen/filters/rules/test/person_rules_test.py
+++ b/gramps/gen/filters/rules/test/person_rules_test.py
@@ -101,6 +101,7 @@ from ..person import (
     PersonWithIncompleteEvent,
     ProbablyAlive,
     RegExpName,
+    RelationshipPathBetween,
     RelationshipPathBetweenBookmarks,
 )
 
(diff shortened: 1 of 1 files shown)


Nick-Hall:
Rebased.

PR #2330, opened 2026-05-24
Fix AttributeError on iconlist when GalleryTab parent dialog closes (bug 13326)

eduralph:
## Root cause
`GalleryTab.clean_up()` calls `super().clean_up()`, which removes
`self.iconlist` via `track_ref_for_deletion("iconlist")`, but never
disconnects the `'selection-changed'` signal handler on the iconlist
widget. When the parent dialog tears down on Cancel — the Forms
addon's GalleryTab-in-a-dialog path exercises this reliably — GTK
dispatches a final `selection-changed` to the still-live widget; the
handler calls `self.get_selected()` → `self.iconlist.get_selected_items()`
and raises `AttributeError: 'GalleryTab' object has no attribute
'iconlist'`.

## Fix
Capture the `'selection-changed'` handler id in `build_interface`,
then disconnect it at the head of `clean_up()` — before
`super().clean_up()` removes the attribute. Guard the disconnect
with `GObject.signal_handler_is_connected` so the already-disposed
case (the reason the original disconnect was removed in 37395da262)
stays silent.

## Verified against
- `gramps/gui/editors/displaytabs/gallerytab.py:232,659-660` — the
  unguarded `connect` and the no-op `clean_up` that the reported
  trace fingers.
- `gramps/gui/editors/displaytabs/buttontab.py:319` — the call site
  in the inherited `_selection_changed` callback that reaches into
  `self.get_selected()`.
- The 2012 disconnect (svn r20849 / `51a53ccebd`) — same defect
  class, same fix shape; removed in `37395da262` to silence a GTK
  warning on the normal close path. The new
  `signal_handler_is_connected` guard keeps that path silent while
  restoring teardown-race safety.

## Test
`gramps/gui/editors/displaytabs/test/gallerytab_test.py` covers both
the bug and the warning the original disconnect was removed to
silence:

- `test_clean_up_disconnects_selection_changed` constructs a real
  `GalleryTab`, records the `'selection-changed'` signal id, and
  asserts via `GObject.signal_handler_find` that no handler remains
  after `clean_up()`. Pre-fix `42 != 0`; post-fix passes.
- `test_clean_up_after_iconlist_destroyed_emits_no_warning` destroys
  the iconlist widget before `clean_up()` — the normal close path
  where commit 37395da262 was seeing the GTK critical — and asserts
  via `warnings.catch_warnings` that no `instance '...' has no
  handler with id '...'` warning is recorded. Mutating the fix to
  drop the `signal_handler_is_connected` guard surfaces exactly that
  warning, so future removals of the guard regress here.

Both need a display (`Gtk.IconView`), so run under `xvfb-run`:

```
xvfb-run -a python3 -m unittest \
    gramps.gui.editors.displaytabs.test.gallerytab_test -v
```

Fixes [#13326](https://gramps-project.org/bugs/view.php?id=13326).

### DIFF
--- a/gramps/gui/editors/displaytabs/gallerytab.py
+++ b/gramps/gui/editors/displaytabs/gallerytab.py
@@ -229,7 +229,9 @@ class GalleryTab(ButtonTab, DbGUIElement):
         self.iconlist.set_selection_mode(Gtk.SelectionMode.SINGLE)
 
         # connect the signals
-        self.iconlist.connect("selection-changed", self._selection_changed)
+        self._sel_changed_handler = self.iconlist.connect(
+            "selection-changed", self._selection_changed
+        )
         self.iconlist.connect("button_press_event", self.double_click)
         self.iconlist.connect("key_press_event", self.key_pressed)
         self._connect_icon_model()
@@ -657,4 +659,14 @@ class GalleryTab(ButtonTab, DbGUIElement):
         return self.get_data().index(obj)
 
     def clean_up(self):
+        # Disconnect selection-changed before super() deletes self.iconlist via
+        # track_ref_for_deletion. Otherwise a late selection-changed emission
+        # (e.g. when the parent dialog tears down on Cancel) re-enters
+        # _selection_changed -> get_selected() and crashes because self.iconlist
+        # no longer exists. Bug 13326.
+        sel_handler = getattr(self, "_sel_changed_handler", None)
+        if sel_handler and hasattr(self, "iconlist"):
+            if GObject.signal_handler_is_connected(self.iconlist, sel_handler):
+                self.iconlist.disconnect(sel_handler)
+            self._sel_changed_handler = None
         super(ButtonTab, self).clean_up()
--- a/gramps/gui/editors/displaytabs/test/gallerytab_test.py
+++ b/gramps/gui/editors/displaytabs/test/gallerytab_test.py
@@ -22,6 +22,7 @@
 # ------------------------
 # Python modules
 # ------------------------
+import os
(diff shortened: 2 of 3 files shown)


romjeromealt:
Is it also related to [bug report #12289]( https://gramps-project.org/bugs/view.php?id=12289)?

eduralph:
Thanks for the pointer to 0012289 — it's relevant to understand the story behind this one. 

The key distinction is severity and path. 12289 was a cosmetic GObject warning (instance ... has no handler with id ...) on the normal Person Editor close. 13326 is different in kind: an unhandled AttributeError crash on a specific sequence — add an image to the gallery, then Cancel the dialog — where a selection-changed callback fires after iconlist has been torn down. 

This resolves the crash without re-touching the part of the cleanup behavior that the earlier reports were wary about. Happy to adjust the approach if there's a side effect I've missed — the guard is specifically there to respect the original concern,

romjeromealt:
Do not need to worry too much, I often make relations, this does not mean something should be improved. I am fine with this PR. As far as I remember the root of 12289 was pointed during a tracking down session around memory leaks. So, even cosmetic GalleryTab or some gramplets, generated some minor memory issues. Now, with the help of AI, we can just provide something more "professional[1]", quick and maybe more efficient. That's all. 
[1] my workflow was to run Gramps 5.1.3 with an OS using less than 512Mo of ram... and I closed 12289 as "wont't fix" because today, such config would not run Gramps 6 and later.

Nick-Hall:

We probably should run our CI unit test suite using Xvfb, but that can be done in a separate PR.

PR #2333, opened 2026-05-25
FamilySearch integration: bulk import, merge window, login UX, and bug fixes

SourceAirbender:
## Summary

This PR extends the FamilySearch integration with several new user-facing
features, resolves bugs found during testing, and addresses all findings
from three rounds of code review.

## New features

- **Bulk import**: a new import tool lets users pull an entire FamilySearch
  tree into an empty Gramps database. After signing in with an empty tree
  open, users are prompted to start a bulk import automatically.
- **Merge window**: a new merge-overview dialog (modelled on the FamilySearch
  website) is shown before a record is merged, giving users a clearer picture
  of what will change.
- **Sign-in from the status-bar icon**: clicking the FS icon when not signed
  in now offers to open the sign-in dialog instead of only displaying a
  warning. If the Gramps Foundation access key is missing, the icon opens
  Preferences → Integrations directly.
- **Enable / disable the integration**: a toggle in Preferences → Integrations
  hides or shows all FamilySearch UI elements without requiring a restart.
- **Progress feedback**: a counter dialog is shown while the compare window
  fetches sources for people with many FS records, preventing the UI from
  appearing frozen.
- **CSS segregation**: FamilySearch-specific styles have been moved from
  `gramps.css` into a dedicated `data/familysearch.css` file.

## Bug fixes

- Restore child handles in `Family.get_referenced_handles()` — they were
  accidentally dropped, breaking callman.py UI tracking for child changes.
- Replace `time.mktime(parsedate(…))` with `calendar.timegm(parsedate(…))`
  in `tree.py` and `cache.py` so that HTTP `Last-Modified` timestamps are
  treated as UTC rather than local time; guard against `parsedate()` returning
  `None` on a malformed header.
- Fix `_datemod` → `_datmod` typo in `gui/fs/compare/window.py` so the
  early-return optimisation in `compare_fs_to_gramps` actually fires.
- Stop `add_persons()` from overwriting already-present `_etag`/`_last_modified`
  values in `Tree._persons` when iterating the global deserialiser index.
- In `_find_best_fs_fact_match`, prefer a fact that has a date over a dateless
  one as the type-only fallback when the Gramps event has a date.
- Fix the couple loop in `compare_spouse_notes` / `compare_spouses` to skip
  couples where neither member has an FS ID, eliminating spurious rows.
- Log (not swallow) network and auth exceptions from concurrent FS fetches
  in `Tree.add_persons()`.
- Fix `tags.py` attribute lookup: replace the ad-hoc alias set with the
  canonical `FS_ATTR_CANON`/`FS_ATTR_OLD`/`FS_ATTR_HUMAN` constants from
  `actions.py`, and fix `get_string()` → `str()` which was silently catching
  `AttributeError` and skipping every attribute.
- Fix a startup crash (private-attribute guard on a GObject-derived class)
  introduced after the maintenance/gramps61 rebase.
- Fix asyncio event loop handling so that compare and import work correctly
  on Windows builds.
- Fix compare window incorrectly reading cached FS values due to a stale
  cache entry.
- Fix import failures when a parent had a child outside the current
  relationship node.

## Tests

Added regression tests covering all of the above fixes (75 new tests across
comparators, cache mixin, tags, tree, importer, and family referenced-handles
modules, all passing).

## Code quality

- Resolved all mypy errors introduced by the new code (type annotations for
  `_last_modified`, `_etag`, and outcome values on `Person`).
- Addressed all findings from three rounds of code review (twenty items total).
- `.gitignore` updated to exclude FS JSON cache files.


dsblank (dismissed):
## Code Review — PR #2333 (FS upgrades & bug fixes for 6.1)

Reviewed at high effort across three angles (line-by-line scan, removed-behavior audit, cross-file tracer). Ten findings follow, ranked by severity.

---

### 1. 🔴 Child-triple consumed by the wrong FS family ()

In the refactored `compare_spouses` inner child-matching loop, the new condition is:

```python
other_parent_id = _other_parent_id_for_child_rel(fsid, triple)
if (
    other_parent_id is not None
    and triple.child.resourceId == child_fsid
):
    fs_children.remove(triple)
    break
```

This removes the triple as soon as `fsid` is any parent of child X, without checking that `other_parent_id == fs_spouse_id`. The old code required both parents to match before consuming the triple.

**Scenario:** Person A is a parent in two Gramps families — one with spouse B, one with spouse C — and child X appears in both FS families. When iterating for family A–B, the loop finds the A–C triple first, removes it, and shows an orange mismatch. Family A–C then finds no triple for X and shows it as yellow (only in Gramps). One child silently misattributed, one silently missing.

---

### 2. 🔴 Post-loop FS-only child scan absorbs unknown-parent children into wrong family ()

The post-loop scan at the end of each family block:

```python
if _other_parent_id_for_child_rel(fsid, triple) == fs_spouse_id:
```

When a Gramps family has no FS-linked spouse (`fs_spouse_id == ''`), and a FS triple has an absent other-parent slot (`_other_parent_id_for_child_rel` returns `''`), the condition `'' == ''` is True. Every FS child with an unknown other parent is claimed as a yellow3 ("Only in FamilySearch") row under the first family with no linked spouse, regardless of whether it actually belongs there.

---

### 3. 🔴 `add_child()` silently drops single-parent families when one parent is outside the downloaded subtree ()

Old behavior: if `parent1` exists in FS but its resourceId is absent from `FS_INDEX_PEOPLE`, `father_h = None` and a single-parent family is still created with the known parent.

New behavior: `_parent_handles_from_refs` calls `_ensure_imported_person_handle` for every parent. Any parent whose FS id is not in `FS_INDEX_PEOPLE` **and** not in `deserialize.Person.index` goes into `missing`, and `add_child()` returns immediately — creating no family at all. The child ends up parentless in the local DB, and the only feedback is a `LOG.warning` that is invisible without `--debug`.

---

### 4. 🟠 `_apply_selected_merge_rows` commits a partially-mutated `Person` object when a row handler raises ()

```python
for row in rows:
    try:
        if row.kind == "gender":
            changed += int(self._merge_gender_row(gr, fs_person))
        # ...
    except Exception:
        continue

db.commit_person(gr, txn)   # always runs
```

Handlers like `_merge_gender_row` and `_merge_name_row` mutate `gr` in-place. If a later handler raises (e.g. `HandleError` from a stale event handle), the exception is swallowed and the loop continues. `db.commit_person(gr, txn)` then commits whatever partial state `gr` is in. The user sees a "N item(s) copied" success dialog with no indication that some rows failed silently.

---

### 5. 🟠 `get_session()` priority flip can return an unauthenticated session and overwrite the cached authenticated one ()

```python
# Old:
sess = _SESSION or get_active_session()
# New:
sess = get_active_session() or _SESSION
```

`Session` instances register themselves in class/module-level variables during `__init__`, before authentication completes. If `get_active_session()` returns such a freshly-constructed but not-yet-authenticated `Session` object while `_SESSION` already holds an authenticated one, the new code picks the unauthenticated session and overwrites `_SESSION` with it. All subsequent FS operations find no `access_token` and fail — even though a working session existed.

---

### 6. 🟠 asyncio parallelism removed from `add_persons()` with no replacement — severe slowdown for bulk imports ()

The deleted asyncio block used `run_in_executor` to fetch multiple FS persons concurrently. The replacement is a plain sequential `for`-loop. Each `add_person(fid)` call makes at least one blocking HTTP request. For N=200 persons at 200ms average latency, old code pipelined (finishing in ~200ms); new code takes ~40 seconds. No alternative parallelism (threading, concurrent.futures) was introduced. The UI progress bar appears frozen during person loading on large bulk imports.

---

### 7. 🟡 `_FS_PHOTO_CACHE` permanently stores `None` on portrait fetch failure ()

```python
self.__class__._FS_PHOTO_CACHE[cache_key] = None
return None
```

The class-level dict has no TTL or eviction mechanism. Any person whose FS portrait was transiently unavailable (HTTP 503, timeout) will always show a placeholder avatar for the rest of the Gramps process session, even after network recovery. Only positive hits (successfully loaded pixbufs) should be cached.

---

### 8. 🟡 `print()` for error reporting in `_hydrate_relative_payloads` ()

```python
print(f"[FS Cache] failed to fetch {endpoint} for {fsid}: {exc}")
print(f"[FS Cache] failed to deserialize {endpoint} for {fsid}: {exc}")
```

Project guidelines require `LOG = logging.getLogger(__name__)` and forbid `print()` for diagnostic output. `cache.py` defines no module-level logger. When Gramps is launched from a desktop launcher, stdout is suppressed — these errors are completely invisible, making diagnosis of missing relatives in the compare view impossible without a terminal.

---

### 9. 🟡 `_update_familysearch_ui()` silently skips `update_menu()` when called before `uimanager` is ready ()

```python
if hasattr(self, "uimanager"):
    self.uimanager.update_menu()
```

This method is called from `__build_main_window` before `__create_action_groups` sets `self.uimanager`, and also from the `familysearch.enable` config-key callback. If the callback fires during startup before the UI manager is ready, `update_menu()` is silently skipped. The old code called `update_menu()` unconditionally in `close_database` and `change_page`. FS menu items could remain visible/sensitive after database close.

---

### 10. 🟡 Blocking FS search on the GTK main thread in `_empty_tree_start_person_dialog` ()

```python
while Gtk.events_pending():
    Gtk.main_iteration()
# then:
results = _search_familysearch_people(session, raw_query)   # blocking HTTP
```

The search blocks the GTK main thread for the full HTTP round-trip. The `Gtk.main_iteration()` flush allows window-manager close events to fire during the flush loop. If the user dismisses the dialog via the window manager while `main_iteration` is running, the dialog's GTK widgets are destroyed before `_search_familysearch_people` returns, and subsequent accesses to `store`, `status_label`, or `search_button` crash with a GObject-already-destroyed error. The search should run on a worker thread.

emyoulation:
Is there a document with the targets for the FamilySearch integration? 

After having been back and forth through Emails, PRs and Commits; I still have not isolated any introductory information or rudimentary roadmap of features.

Nick-Hall:
There is no roadmap.  At this stage, we are getting the integration reviewed by FamilySearch and then addressing the rquested changes.  The timescale will depend on the availability of @SourceAirbender and to some extent @sam-m888.

emyoulation:
Thanks, that saves me a lot more searching.

So perhaps I could ask a question then?

One of the features noted was a bulk import and there are 3 FS flavors of that which would save me a lot of time. I wonder if they are on the horizon?

When starting conversations with a related researcher, I try to stitch in their connection to me. This gives me a good place to log them as a Gramps Repository for any original source documents in their possession which I lack. 

FamilySearch offers a feature to view my relationship to a contributor, to a selected person in their world tree, and connections to registered visitors to RootsTech.



I tediously trace these charts for where my research log tree overlaps and where there are gaps.

I'd love a feature to bulk download the missing linkages from these relationship charts.

Is a scraper of a specific relationship  one of the planned Bulk Download options?

data/gramps.css
emyoulation, line 1: # Request that the FS specific CSS be split into a separate daisy-chained stylesheet.

Segregated CSS would make for easier theme enhancement experimenting and simplify rolling in any approved suggestions. (One assumes that beta testing with light/dark theming will discover that some FS theme changes will be needed.)

An _experimental_ 6.1 [ Theme2 addon](https://github.com/emyoulation/CuratedGrampsPlugins/tree/main/gramps61/source/Themes2) streamlines testing variants of CSS layers. (It still needs more evolving before releasing for End-User use.)  But is viable for Developer and Beta testing of CSS overlays. 

The sample below has Blue and Green color-scheme patches of the Addon Manager lozenges. (Mainly to see how well overlay switching worked for live Gramps interface.) 

I do not yet have access to the FS interface to experiment with its appearance. The FamilySearch CSS patch from `gramps.css` was cloned in anticipation of that GUI becoming available.   
Image

SourceAirbender:
rebased on correct branch & added a suggestion from Nick

dsblank:
Fixed many of the remaining issues, and added more tests. Also, I split the CSS into a familysearch-specific file. 

I'll review the whole PR in an hour or two.

CC: @emyoulation @Nick-Hall

dsblank:
There are a lot of changes, and fixes in this PR. I have no way to test it against familysearch, but I added a lot of unit tests. Changes include big, nasty bugs, and mere line-ending issues.

dsblank:
Import error...

dsblank:
I have now tested a lot of the whole FS system, and have fixed all the issues I have encountered. They are all in this PR.

Approved by me and Claude :)

Nick-Hall:
@dsblank Thank you.  If @SourceAirbender agrees, I'll create another release over the weekend and submit it to FamilySearch for certification.

SourceAirbender:
one small fix and did (what I consider) a visual improvement on the "sync to FS" window. I'd like to test everything is working on windows with the AIO version so we don't run into any unexpected bugs when presenting to FS. After that I think we're good to go!
@Nick-Hall thoughts?

PR #2337, opened 2026-05-28
Make the Share/Select existing records labels consistent

DaveSch-gramps:
from the Discourse discussion

https://gramps.discourse.group/t/alignment-of-help-text-for-icon/9551

### DIFF
--- a/gramps/gui/editors/displaytabs/buttontab.py
+++ b/gramps/gui/editors/displaytabs/buttontab.py
@@ -70,7 +70,7 @@ class ButtonTab(GrampsTab):
         "add": _("Add"),
         "del": _("Remove"),
         "edit": _("Edit"),
-        "share": _("Share"),
+        "share": _("Select Existing"),
         "jump": _("Jump To"),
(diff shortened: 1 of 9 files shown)


dsblank (approved):
Looks good!

Nick-Hall:
Rebased and squashed commits.

emyoulation:
A question about harmonizing Tooltips...  

Does the the elimination of a translatable string variant effect Weblate.  Does it trigger an elimination of the "Source" string from the Weblate Gramps component?

Nick-Hall:
I updated the `gramps.pot` file in a separate commit.

emyoulation:
Ok.  I see `po/gramps.pot` commit.  Switching question to there.  (But I think the answer would be more 'findable' in this thread and useful information for those who want to help make tooltips more consistent.  If there is another step needing to be done before the change ALSO reduces the Weblate burden, that's important.)  

https://github.com/gramps-project/gramps/commit/64634be80a4a03f2caf8f06bbb4fd154c6b266b8

PR #2339, opened 2026-05-28
Fix ResourceWarning from unclosed /dev/null in Mac defaults lookup

eduralph:
## Root cause
`mac_setup_localization` passed `open("/dev/null", encoding="utf8")` as
`stderr` to `subprocess.Popen`. Popen's context manager closes only the
pipes it creates itself, not a file object the caller supplies, so that
`/dev/null` handle leaked and raised a `ResourceWarning` whenever the
`defaults` lookup ran (i.e. on every unittest run on macOS).

## Fix
Use `subprocess.DEVNULL` for the discarded stderr — Popen manages it
internally and there is no Python-level file object to leak. The
`encoding` on a discarded stream was meaningless. No change to the
captured stdout.

## Verified against
- `gramps/gen/utils/maclocale.py`:100-102 — the leaking `Popen` call;
  the surrounding `_mac_get_gramps_defaults` closure is the only caller.
- `gramps/gen/utils/grampslocale.py`:45-46 — `maclocale` is imported
  only under `sys.platform == "darwin"`, so the leaking line is
  unreachable off macOS and the warning only ever surfaces there. (The
  handle-leak mechanism is platform-independent — `open("/dev/null", …)`
  is evaluated as a `Popen` argument, so it would leak before `Popen`
  runs if the line executed at all — but the module never loads off Mac.)

## Test
`gramps/gen/utils/test/maclocale_test.py` — stubs `subprocess.Popen`,
drives the public `mac_setup_localization`, and captures the `stderr`
argument the `defaults` call receives. Asserts it is not a leakable
`io.IOBase` file object (and pins it to `subprocess.DEVNULL`). The stub
makes the otherwise macOS-only path reachable on every platform.
Pre-fix both assertions fail (capturing `<_io.TextIOWrapper
name='/dev/null' ...>`); post-fix both pass.

Fixes #14243.

### DIFF
--- a/gramps/gen/utils/maclocale.py
+++ b/gramps/gen/utils/maclocale.py
@@ -98,7 +98,7 @@ def mac_setup_localization(glocale):
 
             answer = None
             with subprocess.Popen(
-                args, stderr=open("/dev/null", encoding="utf8"), stdout=subprocess.PIPE
+                args, stderr=subprocess.DEVNULL, stdout=subprocess.PIPE
             ) as proc:
                 answer = proc.communicate()[0]
             if not answer:
--- a/gramps/gen/utils/test/maclocale_test.py
+++ b/gramps/gen/utils/test/maclocale_test.py
@@ -1,0 +1,166 @@
+# -*- coding: utf-8 -*-
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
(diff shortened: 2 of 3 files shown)


Nick-Hall:
Rebased.

PR #2208, opened 2026-03-27
Major upgrade to Latexdoc (prculley PR#1061, rebased + fixed + tested)

dsblank:
## Summary

This PR brings in Paul Culley's (prculley) [Major upgrade to Latexdoc (#1061)](https://github.com/gramps-project/gramps/pull/1061), rebased against current master with all merge conflicts resolved and several bugs fixed.

### What prculley's original PR adds

- New `VerticalFineTuning` mixin class for fine-grained vertical spacing adjustments per style name
- Reworked table state machine: `handle_table()` with `CELL_BEG/TEXT/END`, `ROW_BEG/END`, `TAB_BEG/END` constants
- `repack_row()` for transposing multi-row cell structures
- `discover_col_widths()` and `discover_multcol_widths()` for column width calculation
- `mk_complete_row()` and `mk_splitting_row()` for row assembly
- `mapping_to_latex()`: CSV-based transliteration/hyphenation mapping loaded at open time
- Updated `add_media()` with image cropping support and improved error handling

### Bug fixes applied during rebase

- Restored constants dropped in conflict resolution: `HI_STRUT/LO_STRUT/HI_SPACE/LO_SPACE`, `FLOAT_PAT`, `STYLE_DETAIL`, `HASH_STRIP`
- Restored `mapping_to_latex()` function
- Restored `calc_latex_widths()` (called by master's `handle_table`)
- Restored missing `LaTeXDoc` class-level variables (`curr_style_name`, `curr_tab_style`, `stick_next`, `kind_of_pict`, `space_above/below_paragr`)
- Fixed `str_incr()`: `for i in reversed(lili)` used chars as list indices (TypeError); corrected to `range(len-1, -1, -1)`
- Fixed `write_table()`: `splitted_row` typo, missing `complete_row` assignment, missing subsequent-rows loop, `put_addit()` → direct `.addit`
- Fixed `repack_row()`: `tabmem.add_row()` → `tabmem.rows.append()`
- Made `mk_complete_row(row, last=None)` to handle both call forms
- Fixed mypy errors: annotated `gramps_to_latex`, `tail_vals`, `cmpl_vals`

### Tests added (60 total)

**`gramps/plugins/docgen/test/latexdoc_test.py`** (47 unit tests)
- `latexescape` / `latexescapeverbatim`
- `get_charform` / `get_numform` roundtrip and error cases
- `str_incr` increment, carry, double-carry, exhaustion
- `map_font_size`
- `TabCell`, `TabRow`, `TabMem` construction
[...]

PR #1061, opened 2020-05-14
Major upgrade to Latexdoc

prculley:
See https://gramps-project.org/bugs/view.php?id=11166#c59480  Harald has rewritten the latexdoc plugin to deal with several issues.

PR #2341, opened 2026-05-29
Fix LaTeX report crash when a table has two or more multicolumns

eduralph:
## Root cause
`str_incr` in `latexdoc.py` generates the column identifiers used for LaTeX
table multicolumns. Its increment loop, `for i in reversed(lili)`, iterates the
character list's *elements* rather than its indices, so `lili[i]` indexes the
list with a string and raises `TypeError: list indices must be integers or
slices, not str`. The first `next()` returns at the `yield` before the loop, so
the crash appears on the second value — reached whenever a table has two or more
multicolumns, such as a styled note containing subscript/strikeout (bug 13418).

## Fix
Iterate the indices instead of the elements:

```python
-        for i in reversed(lili):
+        for i in reversed(range(len(lili))):
```

The carry behaviour is unchanged: `aaa, aab, … aaz, aba, …`.

## Verified against
- `gramps/plugins/docgen/latexdoc.py:496-517` — the `str_incr` generator and its
  carry/`else` branch.
- This is independent of bug 13417 (styled-note markup): that fix is commit
  8b935f1551 and does not touch `str_incr`.

## Test
`gramps/plugins/docgen/test/latexdoc_test.py` adds a headless `StrIncrTest`
asserting `str_incr` yields successive ids (including a carry) without raising.
The second yielded value reproduces the `TypeError` on the unfixed code; both
tests pass after the change.

Fixes #13418.

### DIFF
--- a/gramps/plugins/docgen/latexdoc.py
+++ b/gramps/plugins/docgen/latexdoc.py
@@ -509,7 +509,7 @@ def str_incr(str_counter):
                     )
                 )
             )
-        for i in reversed(lili):
+        for i in reversed(range(len(lili))):
             if lili[i] < "z":
                 lili[i] = chr(ord(lili[i]) + 1)
                 break
--- a/gramps/plugins/docgen/test/latexdoc_test.py
+++ b/gramps/plugins/docgen/test/latexdoc_test.py
@@ -1,0 +1,62 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026       Eduard Ralph
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <https://www.gnu.org/licenses/>.
+#
+
+"""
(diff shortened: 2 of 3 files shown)


dsblank:
Does https://github.com/gramps-project/gramps/pull/2341 have any overlap with https://github.com/gramps-project/gramps/pull/2208 ?

Yes, there is significant overlap. Both PRs touch the same three files, and they share a direct code fix:

Shared files:
  - gramps/plugins/docgen/latexdoc.py
  - gramps/plugins/docgen/test/__init__.py
  - gramps/plugins/docgen/test/latexdoc_test.py
  
Shared fix — the str_incr bug:
  - PR #2341 (small, targeted): fixes for i in reversed(lili) → for i in reversed(range(len(lili))) and adds a StrIncrTest in latexdoc_test.py
  - PR #2208 (major upgrade): fixes the exact same bug (noted in its description: "Fixed str_incr(): for i in reversed(lili) used chars as list indices (TypeError)") as part of a full rewrite of latexdoc.py, plus adds 60 tests

Key difference in scope:
  - #2341 is a one-line surgical fix for bug #13418 (two-column LaTeX table crash)
  - #2208 is a full major upgrade rebasing prculley's PR #1061 — new table state machine, column width calculation, CSV transliteration mapping, etc.
  
  If #2208 were merged, it would supersede #2341 entirely. If #2341 is merged first, #2208 would need to be rebased and the str_incr fix reconciled (it's already there in #2208, so it would just be a conflict to resolve, not a duplicate patch).

eduralph:
Hi dsblank, you are right, there is an overlap.  Sorry, I wasn't aware of the complete scope of 2208 and the bugfix wasn't referenced, so a quick search was too little. 

The implementation is a minimalist bug fix for the bug 13418 targeting gramps61; the PR 2208 is a bigger enhancement which would go directly into master (I'm assuming based on what jrails told mit a while back) . Depending on when your feature lands it might make sense to live with the conflict or bury the bug fix .

Nick-Hall:

PR #2343, opened 2026-05-30
Suppress xgettext python brace format warnings

Nick-Hall:
We use a non-standard format for date inflections.  The fatal errors produced by `msgfmt -c` and the corresponding Weblate messages are not helpful in this case.

### DIFF
--- a/gramps/gen/datehandler/_datedisplay.py
+++ b/gramps/gen/datehandler/_datedisplay.py
@@ -179,128 +179,148 @@ class DateDisplay:
             #
             # Not moving to DateStrings, as this is part of display code only,
             # coupled tightly with the formats used in this file.
-            "": _("{long_month} {year}"),
+            "": _( # xgettext: no-python-brace-format
+            "{long_month} {year}"),
             "from"
             # first date in a span
             # Translators: If "from <Month>" needs a special inflection
             # in your language, translate to "{long_month.forms[X]} {year}"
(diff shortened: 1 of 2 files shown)


dsblank:
Lint says you gotta get those spaces in the right place :smile:

Nick-Hall:
The problem is that if I format the file with `black`, then `xgettext` doesn't recognise the directive.  Any ideas?

dsblank:
Root cause: xgettext associates a # xgettext: flag comment with the next string it sees in the token stream. Black was collapsing 
```
_(
   # comment
   "string"
) 
```

into 
```
_("string"),  # comment
```
which puts the comment after the string — so it applies to the wrong string.
  - Fix: Add # fmt: off / # fmt: on markers around the bodies of both FORMATS_long_month_year and FORMATS_short_month_year dictionaries.
  Black recognizes these as explicit opt-outs and leaves the _(  # xgettext: no-python-brace-format\n"string") form untouched.

PR at: https://github.com/Nick-Hall/gramps/pull/7

Nick-Hall:
@dsblank Thanks.  I opted for `# fmt: skip`  instead which I think is slightly neater.

A similar directive in [configure.py](https://github.com/gramps-project/gramps/blob/master/gramps/gui/configure.py#L868) survived the formatting because it was a multi-line string.  I remembered that the directive had to be before the string though. :)

PR #2348, opened 2026-06-01
Fix Top Surnames gramplet opening report for the wrong surname

brianmcaudill:
Fixes [#11101](https://gramps-project.org/bugs/view.php?id=11101).

## Problem

In the *Top Surnames* gramplet (Dashboard), double-clicking a surname opens a Same Surnames report for a **different** surname. Depending on the tree it shows up on the top entry or the 2nd/3rd. Reproducibility is "always" for affected trees; also reported on the [forum](https://gramps.discourse.group/t/top-surnames-gramplet-reports-wrong-surname/3870).

## Root cause

`TopSurnamesGramplet.main()` counts each person under **every** group name from their primary *and* alternate names, and sets `representative_handle[surname] = person.handle` for each. The stored representative is therefore the last person iterated who carries that surname among *any* of their names — possibly someone whose **primary** surname is different (they matched only via an "also known as" alternate name).

The `samesurnames` quick view (category `CATEGORY_QR_PERSON`) re-derives the surname from that person:

```python
rsurname = person.get_primary_name().get_group_name()
```

If the representative's primary surname differs from the clicked surname, the report is for the wrong surname. This matches the analysis in #11101 (note ~0065258): the triggering person had redundant "Also known as" alternate names matching the top surname, while her *primary* surname differed.

## Fix

Prefer a representative whose **primary** surname matches the group name, so the quick view's re-derivation matches what was clicked. Surnames that appear only as alternate names (no primary carrier) use a deterministic first-seen fallback and remain best-effort, since `samesurnames` always derives from the primary name.

The per-person tally is extracted into a module-level `record_surnames()` helper so the selection logic is testable without GTK or a database; the gramplet's incremental `yield` behaviour is unchanged.

## Tests

Adds `gramps/plugins/gramplet/test/topsurnamesgramplet_test.py` covering surname counting across primary/alternate names, representative selection regardless of iteration order, and the alternate-only fallback. Built with `gramps.gen.lib` objects — no display or database required. `black --check` clean.

### DIFF
--- a/gramps/plugins/gramplet/test/topsurnamesgramplet_test.py
+++ b/gramps/plugins/gramplet/test/topsurnamesgramplet_test.py
@@ -1,0 +1,156 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Brian Caudill
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <https://www.gnu.org/licenses/>.
+#
+
+"""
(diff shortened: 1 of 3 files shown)


Nick-Hall:
This fix should target the _gramps61_ branch.

brianmcaudill:
Thanks @Nick-Hall — rebased onto `maintenance/gramps61` and retargeted the PR to that branch. The fix applied cleanly with no conflicts.

Nick-Hall:
Added the new `__init__.py` file to `POTFILES.skip`.

PR #2354, opened 2026-06-05
Delay gtk for headless ut

hgohel:
This draft PR is for comments on the direction that AI was steering a change to run unit tests that attempt to import modules which depend on GTK/GDK, and end up crashing during test execution (see PR #2154 for background). See commit [bc11a6d](https://github.com/gramps-project/gramps/pull/2354/commits/bc11a6d87418b41e1338465892d9bf4a4d3100f0) for detailed notes.

This is not intended for merging because several other tests are failing with these changes.

### DIFF
--- a/gramps/gui/clipboard.py
+++ b/gramps/gui/clipboard.py
@@ -46,7 +46,7 @@ import cairo
 # gramps modules
 #
 # -------------------------------------------------------------------------
-from gramps.gen.const import URL_MANUAL_PAGE
+from gramps.gen.const import URL_MANUAL_PAGE, ICON
 from gramps.gen.lib import NoteType
 from gramps.gen.datehandler import get_date
 from gramps.gen.display.place import displayer as place_displayer
@@ -57,7 +57,7 @@ from .managedwindow import ManagedWindow
 from .glade import Glade
 from .ddtargets import DdTargets
 from .makefilter import make_filter
-from .utils import is_right_click, no_match_primary_mask
+from .utils import has_gtk_display, is_right_click, no_match_primary_mask
 from gramps.gen.const import GRAMPS_LOCALE as glocale
 from gramps.gui.widgets.multitreeview import MultiTreeView
 
@@ -78,8 +78,18 @@ clipdb = None  # current db to avoid different transient dbs during db change
 #
 # -------------------------------------------------------------------------
 
-theme = Gtk.IconTheme.get_default()
-LINK_PIC = theme.load_icon("stock_link", 16, 0)
+theme = None
+if has_gtk_display():
+    try:
+        theme = Gtk.IconTheme.get_default()
+    except Exception:
+        theme = None
+
+try:
+    LINK_PIC = theme.load_icon("stock_link", 16, 0) if theme else None
+except Exception:
+    LINK_PIC = None
+
 OBJ2ICON = {
     "media": "gramps-media",
     "note": "gramps-notes",
@@ -106,7 +116,16 @@ def obj2icon(target):
 
 ICONS = {}
 for name, icon in OBJ2ICON.items():
-    ICONS[name] = theme.load_icon(icon, 16, 0)
+    if theme:
+        try:
+            ICONS[name] = theme.load_icon(icon, 16, 0)
+            continue
+        except Exception:
+            pass
+    try:
+        ICONS[name] = GdkPixbuf.Pixbuf.new_from_file(ICON)
+    except Exception:
+        ICONS[name] = None
 
 
 # -------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


dsblank (changes requested):
## Review: PR #2354 — Delay GTK for headless UT

### What the PR is trying to fix

The real root cause is one line: `DbManager.BUSY_CURSOR` is a **class-level attribute** initialized at class-definition time via `Gdk.Cursor.new_for_display(Gdk.Display.get_default(), ...)`. Under `GDK_BACKEND=-`, `get_default()` returns `None`, so `new_for_display` raises `TypeError` the moment the module is imported. Everything else cascades from that.

---

### What the PR actually does (and where it goes wrong)

**The good part:** Making `BUSY_CURSOR` lazy via `_get_busy_cursor()` is the right fix for the root cause. This single change is sound.

**The over-engineering:**

**1. `widgets/__init__.py` — silently hides all widgets in headless mode**

```python
if has_gtk_display():
    from .basicentry import *
    ...
else:
    __all__ = []
```

Any code that does `from gramps.gui.widgets import SomeThing` in a headless context will silently get `ImportError` or `NameError`. This trades a loud crash for a silent failure, which is much harder to debug. The `widgets` package has nothing to do with the `BUSY_CURSOR` crash.

**2. `persistenttreeview.py` — runtime base-class switching is a design smell**

```python
class PersistentTreeView(Gtk.TreeView if Gtk else object):
```

This creates two completely different classes depending on runtime state. Code that calls `set_vexpand()`, `get_selection()`, etc. on a headless `PersistentTreeView` will raise `AttributeError` with no warning. `isinstance(x, Gtk.TreeView)` also breaks. None of this is needed — `PersistentTreeView` is already mocked in the test.

**3. `DbManager.__init__()` early return — creates half-initialized objects**

```python
if not self._gtk_display_available:
    return
```

This means `DbManager` can be constructed but in an invalid partial state. Methods that depend on `self.glade`, `self.top`, `self.selection`, etc. will fail in unexpected ways if called on a headless instance. The 15 lines of `self.foo = None` before the guard are there to compensate. A half-initialized object is a worse design than just failing on import.

**4. `has_gtk_display()` called at module import time**

```python
# in clipboard.py and widgets/__init__.py
if has_gtk_display():
    ...
```

This bakes the display-availability decision into the module's loaded state permanently. In production, display is always available; in tests it never is. But this pattern prevents testing with a mocked display after the module is already loaded.

**5. `clipboard.py` — broad exception swallowing hides real errors**

```python
try:
    LINK_PIC = theme.load_icon("stock_link", 16, 0) if theme else None
except Exception:
    LINK_PIC = None
```

`except Exception` at module level will silently hide genuine failures (wrong icon name, permission error, corrupted theme) even when a display IS available.

---

### The minimal, correct fix

The test only exercises `_select_next_after_deletion` — pure selection logic. It already mocks `Glade`, `PersistentTreeView`, `User`, `ManagedWindow.setup_configs`, and `DbManager._select_default`. The only thing that prevented the module from even importing was `BUSY_CURSOR` at class definition time.

**Make `BUSY_CURSOR` a lazy class property (one change to `dbman.py`):**

```python
# Replace the class-level:
BUSY_CURSOR = Gdk.Cursor.new_for_display(Gdk.Display.get_default(), Gdk.CursorType.WATCH)

# With:
_busy_cursor: "Gdk.Cursor | None" = None

@classmethod
def _get_busy_cursor(cls) -> "Gdk.Cursor | None":
    if cls._busy_cursor is None and Gdk.Display.get_default() is not None:
            Gdk.Display.get_default(), Gdk.CursorType.WATCH
        )
    return cls._busy_cursor
```

Then update the one call site. That's it — no changes to `widgets/__init__.py`, `clipboard.py`, `persistenttreeview.py`, or `pluginmanager.py`.

This also fixes the headless case beyond tests: CLI usage, scripts, or any other context that imports `DbManager` without a display will work cleanly without half-initialized objects or silently unavailable widget classes.

dsblank:
@hgohel I made a PR on your branch: https://github.com/hgohel/gramps/pull/1

stevenyoungs:
> ```python
> @classmethod
> def _get_busy_cursor(cls) -> "Gdk.Cursor | None":
>     if cls._busy_cursor is None and Gdk.Display.get_default() is not None:
>         cls._busy_cursor = Gdk.Cursor.new_for_display(
>             Gdk.Display.get_default(), Gdk.CursorType.WATCH
>         )
>     return cls._busy_cursor
> ```

This is not thread safe. Is that a problem?

dsblank:
I'm going to close this one, in favor of one based on https://github.com/hgohel/gramps/pull/1

PR #2355, opened 2026-06-06
Fix underscores and escaping in menu labels

ztlxltl:
This PR fixes the display of underscores in the menu, especially for user-defined names, like database or tag names. 

In GTK, an underscore in menu labels is used to mark the following letter as keyboard accelerator called a mnemonic. The letter appears underlined, or becomes underlined when pressing the `Alt` key. In case the user defines a database name or a tag name that includes an underscore, the underscore will not be shown in the menu. By using `.replace("_", "__")`, the underscore is escaped so it will appear in the menu. The same applies for bookmarked people with an underscore in their formatted name (e.g. many names in the [`royal92.ged`](https://github.com/emyoulation/example-Gramps-Trees/blob/main/royal92.ged)).

While checking other custom strings in the menu, I also noticed that plugin names used in the Reports and Tools menus are not escaped using `html.escape()` or `xml.sax.saxutils.escape`. While I don't know a plugin name that requires this, devs of new plugins could have a hard time debugging the resulting error message.

I also saw that less thorough cleaning up of plugin IDs happens when handling "unsupported items" in `ViewManager.build_plugin_menu()`. I'm not sure what those unsupported items can be, but from my limited understanding their IDs have a similar risk of having invalid chars as the IDs of other plugins. I wasn't able to test this, so I would appreciate it if someone who has more knowledge of this part could have a closer look. 

In summary, this PR does the following:
- Recent databases: add `.replace("_", "__")`
- Reports and Tools menus: add `html.escape()` and `.replace("_", "__"))`
  - For "unsupported" items: additionally add `valid_action_name()`
- Bookmarks: add `.replace("_", "__")`
- Tags: add `.replace("_", "__"))`

### DIFF
--- a/gramps/gui/displaystate.py
+++ b/gramps/gui/displaystate.py
@@ -5,6 +5,7 @@
 # Copyright (C) 2008       Brian G. Matherly
 # Copyright (C) 2010       Nick Hall
 # Copyright (C) 2011       Tim G L Lyons
+# Copyright (C) 2026       ztlxltl
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


Nick-Hall:
Did you intend this to target the _gramps61_ branch?

ztlxltl:
I wasn't sure which version this bug should ideally be fixed in. I'm happy to change it if you prefer 6.1.

Nick-Hall:
Rebased.

PR #2363, opened 2026-06-09
Make GDK cursor class attributes lazy to fix headless import

dsblank:
## Summary

`DbManager.BUSY_CURSOR`, `DisplayState.BUSY_CURSOR`, `RemoveUnused.BUSY_CURSOR`, and `PedigreeView.FLEUR_CURSOR` were all class-level attributes evaluated at import time via `Gdk.Cursor.new_for_display(Gdk.Display.get_default(), ...)`. Under `GDK_BACKEND=-` (headless/CI), `get_default()` returns `None`, so `new_for_display` raises `TypeError` the moment any of these modules is imported.

Each is replaced with a `_get_busy_cursor()` / `_get_fleur_cursor()` classmethod that defers creation until first use, which only occurs when a real window is visible. The guard `if display is not None` is used rather than `except Exception` so that a headless environment doesn't retry on every call.

## Test plan

- [ ] `env GDK_BACKEND=- python3 -m unittest gramps.gui.test.dbman_selection_test` passes
- [ ] Normal (display-present) startup unaffected — cursors are created on first use as before

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/gramps/gen/filters/rules/test/person_rules_test.py
+++ b/gramps/gen/filters/rules/test/person_rules_test.py
@@ -1570,7 +1570,10 @@ class FilterMatchMissingFilterTest(unittest.TestCase):
         """Apply a filter with the given rule directly, no base filter registered."""
         filter_ = GenericFilter()
         filter_.add_rule(rule)
-        return set(filter_.apply(self.db))
+        with self.assertLogs(".filter", level="WARNING") as log:
+            result = set(filter_.apply(self.db))
+        self.assertTrue(any("Can't find filter" in m for m in log.output))
+        return result
 
     def test_IsDescendantOfFilterMatch_missing_filter(self):
         rule = IsDescendantOfFilterMatch(["_no_such_filter_"])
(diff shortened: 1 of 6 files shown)


dsblank:
Merge this before #2364 and then rebase #2364.

dsblank:
The lazy getters introduced in this PR have a thread-safety issue. The bare check-then-set pattern:

```python
if cls._busy_cursor is None:
    cls._busy_cursor = Gdk.Cursor.new_for_display(...)
```

leaves a race window where two threads can both observe `None`, then both call `Gdk.Cursor.new_for_display` (or `Gdk.atom_intern`) and both write the result. This is the same issue in all three getters: `DbManager._get_busy_cursor`, `DisplayState._get_busy_cursor`, and `_DdType.atom_drag_type`.

The fix is double-checked locking — an outer `None` check avoids lock overhead on every call after initialisation, and an inner check inside the lock prevents duplicate initialisation when two threads both pass the outer check before either acquires the lock:

```python
if cls._busy_cursor is None:
    with cls._busy_cursor_lock:
        if cls._busy_cursor is None:
            display = Gdk.Display.get_default()
            if display is not None:
                    display, Gdk.CursorType.WATCH
                )
```

A \`threading.Lock\` class variable (or instance variable for \`_DdType\`) is added alongside each cached value. I've pushed a commit with this fix to the branch.

PR #2364, opened 2026-06-09
Improved Manage Family Trees behavior when deleting trees: preserve sort column and better next item selection

dsblank:
This is a rebase of #2154 by @hgohel, with the test updated to run under `GDK_BACKEND=-` (headless CI). All credit for the feature goes to @hgohel.

---

Manage Family Trees list unexpectedly (1) reverts the sort column selection to Name when a tree is deleted from the list and (2) always selects the most recently accessed tree as the current selection after a tree is deleted, rather than the next/previous item.

There are two commits in this PR for easy review:
1. Preserve sort order when the list is rebuilt after a tree is deleted.
2. Remember the item that was next after a deleted item (or previous if there's no next), and select that after the list is rebuilt.

## Testing

With sort order set to name, last accessed etc., repeat each step below and verify that the selection after delete is reasonable (next item, previous item, or none if there are no more items):

- delete item at top of list
- delete item at bottom of list
- delete item in middle of list

To run the three new unit tests:

```
env GDK_BACKEND=- python3 -m unittest gramps.gui.test.dbman_selection_test.TestDbManSelection
```

**Note:** this branch is rebased on top of #2363, which fixes GTK cursor and drag-target atoms being initialized at import time (required for headless test execution). Once #2363 merges, the diff here will reduce to just the sort/selection changes.

Fixes #12187.

AI Usage: KwaiKAT's KAT-Coder-Pro V1 used within Cline host to fix the bugs, with manual tweaking before commit. Further changes and unit tests by GitHub Copilot. Rebase and headless test fix by Claude Code.

### DIFF
--- a/gramps/gui/dbman.py
+++ b/gramps/gui/dbman.py
@@ -157,9 +157,18 @@ class DbManager(CLIDbManager, ManagedWindow):
         CLIDbManager.ICON_OPEN: "document-open",
     }
 
-    BUSY_CURSOR = Gdk.Cursor.new_for_display(
-        Gdk.Display.get_default(), Gdk.CursorType.WATCH
-    )
+    _busy_cursor: "Gdk.Cursor | None" = None
+
+    @classmethod
+    def _get_busy_cursor(cls) -> "Gdk.Cursor | None":
+        """Return the busy cursor, creating it lazily on first use."""
+        if cls._busy_cursor is None:
+            display = Gdk.Display.get_default()
+            if display is not None:
+                cls._busy_cursor = Gdk.Cursor.new_for_display(
+                    display, Gdk.CursorType.WATCH
+                )
+        return cls._busy_cursor
 
     def __init__(self, uistate, dbstate, viewmanager, parent=None):
         """
@@ -1034,7 +1043,7 @@ class DbManager(CLIDbManager, ManagedWindow):
         message
         """
         self.msg.set_label(msg)
-        self.top.get_window().set_cursor(self.BUSY_CURSOR)
+        self.top.get_window().set_cursor(self._get_busy_cursor())
         while Gtk.events_pending():
             Gtk.main_iteration()
 
--- a/gramps/gui/dbman.py
+++ b/gramps/gui/dbman.py
@@ -243,6 +243,27 @@ class DbManager(CLIDbManager, ManagedWindow):
             self.selection.select_path(tree_path)
             self.dblist.scroll_to_cell(tree_path, None, 1, 0.5, 0)
 
+    def _select_next_after_deletion(self, next_item_name):
+        """
+        Select the item that would have been next after deletion.
+        
+        Args:
+            next_item_name: The name of the item that should be selected
+        """
+        if self.model is None or not next_item_name:
+            return
+            
+        # Find the item with the matching name in the rebuilt model
+        iter = self.model.get_iter_first()
+        while iter:
+            name = self.model.get_value(iter, NAME_COL)
+            if name == next_item_name:
+                path = self.model.get_path(iter)
+                self.selection.select_path(path)
+                self.dblist.scroll_to_cell(path, None, 1, 0.5, 0)
+                break
+            iter = self.model.iter_next(iter)
+
     def __connect_signals(self):
         """
         Connects the signals to the buttons on the interface.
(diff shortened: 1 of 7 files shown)


dsblank:
Note that this PR also contains the same fixes from #2363 so that this can pass the tests. But review and merge #2363 first, and those changes won't appear here with a rebase.

hgohel (approved):
@dsblank Thanks for taking this over the finish line. All UTs pass, and I've manually tested the behavior - works as described. LGTM.

@Nick-Hall Recommend squash-and-merge on this PR. Thanks.

dsblank:
@hgohel thanks for all your work on this, and for the testing!

hgohel:
@stevenyoungs had asked in a previous PR whether `_get_busy_cursor()` was thread-safe, so I asked DeepSeek, and it responded "no" with the following analysis:

# Thread Safety Analysis of `DbManager._get_busy_cursor()`

**No, this method is not thread-safe.**

Here's the code in question (lines 162-171):

```python
@classmethod
def _get_busy_cursor(cls) -> "Gdk.Cursor | None":
    """Return the busy cursor, creating it lazily on first use."""
    if cls._busy_cursor is None:
        display = Gdk.Display.get_default()
        if display is not None:
                display, Gdk.CursorType.WATCH
            )
    return cls._busy_cursor
```

## The Problem: Unsynchronized Lazy Initialization

This is a classic check-then-act (TOCTOU) race condition:

| Time | Thread A | Thread B |
|------|----------|----------|
| T1 | `if cls._busy_cursor is None:` → `True` | |
| T2 | | `if cls._busy_cursor is None:` → `True` |
| T3 | `cls._busy_cursor = Gdk.Cursor.new_for_display(...)` (Cursor A) | |
| T4 | | `cls._busy_cursor = Gdk.Cursor.new_for_display(...)` (Cursor B) |

- **Thread A** reads `None`, enters the branch.
- **Thread B** reads `None` before Thread A has written, enters the branch too.
- Both threads create a `Gdk.Cursor`, and the last write wins.
- The other cursor object is orphaned (minor memory leak).

While Python's GIL prevents two threads from executing Python bytecodes *simultaneously*, it does **not** prevent interleaving — a thread can be suspended between the `if` check and the assignment. The GIL is not a substitute for proper synchronization.

## Why It Matters Here

1. **Correctness**: The method could create more than one cursor object. In practice this is benign (GDK cursor creation is idempotent), but it's still a correctness bug.

2. **Performance**: Unnecessary repeated cursor creation on contention.

3. **GTK thread-safety**: GTK/GDK operations are generally expected to run on the main thread. If this method is called from a background thread, creating a `Gdk.Cursor` off the main thread could lead to more serious issues.

## Potential Fix

The standard Python fix for lazy initialization in a classmethod is to use a `threading.Lock`:

```python
import threading

class DbManager(CLIDbManager, ManagedWindow):
    _busy_cursor: "Gdk.Cursor | None" = None
    _busy_cursor_lock = threading.Lock()

    @classmethod
    def _get_busy_cursor(cls) -> "Gdk.Cursor | None":
        """Return the busy cursor, creating it lazily on first use."""
        if cls._busy_cursor is None:
            with cls._busy_cursor_lock:
                # Double-checked locking: re-check inside the lock
                if cls._busy_cursor is None:
                    display = Gdk.Display.get_default()
                    if display is not None:
                            display, Gdk.CursorType.WATCH
                        )
        return cls._busy_cursor
```

This uses the **double-checked locking** pattern: the cheap `None` check outside the lock avoids acquiring the lock on every call, and the second check inside the lock ensures only one thread performs the creation.

## Verdict

| Aspect | Safe? |
|--------|-------|
| Python-level thread safety | ❌ No — race condition on lazy init |
| Practical crash risk | Low (GIL + benign reassignment) |
| GTK thread-safety | Depends on caller — if called off-main-thread, risk is higher |

The method is **not thread-safe** due to the unsynchronized check-then-act pattern. Whether this causes observable problems depends on whether it's actually called from multiple threads, but the code as written does not protect against that scenario.

dsblank:
The lazy getters introduced in this PR have a thread-safety issue. The bare check-then-set pattern:

```python
if cls._busy_cursor is None:
    cls._busy_cursor = Gdk.Cursor.new_for_display(...)
```

leaves a race window where two threads can both observe \`None\`, then both call \`Gdk.Cursor.new_for_display\` (or \`Gdk.atom_intern\`) and both write the result. This is the same issue in all three getters: \`DbManager._get_busy_cursor\`, \`DisplayState._get_busy_cursor\`, and \`_DdType.atom_drag_type\`.

The fix is double-checked locking — an outer \`None\` check avoids lock overhead on every call after initialisation, and an inner check inside the lock prevents duplicate initialisation when two threads both pass the outer check before either acquires the lock:

```python
if cls._busy_cursor is None:
    with cls._busy_cursor_lock:
        if cls._busy_cursor is None:
            display = Gdk.Display.get_default()
            if display is not None:
                    display, Gdk.CursorType.WATCH
                )
```

A \`threading.Lock\` class variable (or instance variable for \`_DdType\`) is added alongside each cached value. I've pushed a commit with this fix to the branch.

dsblank:
Thanks @hgohel for mentioning @stevenyoungs' comment. I tried to find it yesterday and couldn't. And thanks @stevenyoungs !

emyoulation:
Is this Merge still pending?  It has missed 6.1 but can it be in 6.2 ?

Kari Kujansuu has a further minor improvement (as an addon) that ensures that the active/selected row is centered after a sort order change.

https://github.com/kkujansuu/gramps/tree/master/addons/DbManPatch

Nick-Hall:
> It has missed 6.1 but can it be in 6.2 ?

Yes.

PR #2368, opened 2026-06-10
Enhance simple book title page to be listed on TOC and can be placed anywhere in the book

dave-khuon:
This enhancement is to make this title page more flexible, as it can be used not just at the beginning of the book, but throughout it, as chapter separators for different sections of the families.

The appearance of the page on the TOC is controlled by a check button.  Its Subject is determined by the presence of Title, or Subtitle, Image filename, or a "Not Provide" if all the above fail.

### DIFF
--- a/gramps/plugins/textreport/simplebooktitle.py
+++ b/gramps/plugins/textreport/simplebooktitle.py
@@ -107,17 +107,25 @@ class SimpleBookTitle(Report):
     def write_report(self):
         """Generate the contents of the report"""
         # Skip the whole page if we have no contents to output
-        if not self.title_string and not self.media_filename:
+        if (
+            not self.title_string
+            and not self.subtitle_string
+            and not self.footer_string
+            and not self.media_filename
+        ):
             return
 
         self.doc.start_paragraph("SBT-Title")
         if self.show_in_toc:
-            # Insert to TOC if requested
+            # Insert to TOC if requested using whatever is available
             if self.title_string:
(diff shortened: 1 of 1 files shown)


dave-khuon:
This updated code has been successfully tested: see output above.
[sample_book.pdf](https://github.com/user-attachments/files/28816204/sample_book.pdf)

To avoid confusion between PR features, it is a lot easier for me to start a new branch from scratch (=master).  This new PR is totally independent.  The old PR 2365 will be closed (= abandoned), since it was stacked against an uncleaned PR 2300, which also was stacked against another uncleaned PR 1873.

I have implemented all changes as suggested by DSBank for PR 2365.  
DSBank recommendations (PR2365):
<<
  simplebooktitle.py
    self.toc_title is unnecessary — self.title_string is already available; the alias adds nothing.

  Verbose emptiness checks — len(self.subtitle_string.strip()) > 0 should be just self.subtitle_string.strip().

The core change (skipping empty subtitle/footer paragraphs and adding the TOC mark) is correct and a good improvement. 
>>

Thanks, and sorry for all the confusions.

dsblank:
@dave-khuon can you add a description?

dave-khuon:
Just did.  Thanks.

dave-khuon:
Here is the resulting Book output, using this book item list:
image
[sample_book.pdf](https://github.com/user-attachments/files/28861229/sample_book.pdf)

PR #2371, opened 2026-06-11
Add Visual Studio Code config to gitignore

hgohel:
Add Visual Studio Code's configuration folder to gitignore, allowing developers to have local files which don't get added to commits.

For now the entire folder is ignored, but it would make sense to provide a list of recommended extensions, launch configurations, etc. and include those files in the repo.


Nick-Hall:
Rebased.

PR #2235, opened 2026-04-16
Change the default gramps IDs to %05d

stevenyoungs:
The previous default of %04d is insufficient for larger trees.

Further discussion here: https://gramps.discourse.group/t/update-the-default-id-formats-for-6-1/9341/40

PR #2372, opened 2026-06-12
Add Event and Citation table support to CSV importer

dsblank:
## Summary

- Adds Event and Citation support to both the **CSV importer** and **CSV exporter**, so events and citations can be round-tripped through CSV files.

### Importer (`importcsv.py`)

- Adds two new table types: **Event** (detected by `eventtype` header) and **Citation** (detected by `citation` header).
- Both follow the existing importer conventions: bracketed Gramps IDs (`[E0001]`) to update existing objects, sheet-local refs for cross-row linking, and `get_or_create_*` helpers.
- All type columns (`eventtype`, `note_type`, `role`) accept any standard value or fall back to a custom type for unknown strings, using the same `_DATAMAP` pattern as `place_type`.
- `place_type` dict updated to use `_DATAMAP` directly so both translated and untranslated type names are accepted in all locales.

**Event table columns:** `event`, `eventtype`, `date`, `place`, `description`, `source`, `note`, `note_type`, `role`, `tag`, `person`, `family`, `media`, `media_description`, `media_date`

**Citation table columns:** `citation`, `source`, `page`, `date`, `confidence`, `note`, `note_type`, `tag`, `person`, `family`, `event`, `place`, `media`, `media_description`, `media_date`

### Exporter (`exportcsv.py`)

- Adds **Include events** and **Include citations** checkboxes to the export options dialog (both default on).
- Event table: exports all events with full metadata. When an event is linked to multiple persons or families, additional rows are written under the same bracketed event ID so all EventRefs are preserved on re-import.
- Citation table: exports all citations with full metadata. Backlinks to person, family, event, and place objects each get their own row under the same bracketed citation ID.

## Test plan

- [ ] 38 new unit tests in `gramps/plugins/importer/test/importcsv_test.py` covering creation, type resolution (standard + custom), note types, person/family/place/event linking, media attachment, update-via-bracket-ref, and error/skip cases.
- [ ] Run with `python3 -m unittest gramps.plugins.importer.test.importcsv_test`
- [ ] Export a database via File → Export → CSV and verify events and citations appear in the output file.
- [ ] Import the exported CSV back and verify events and citations are recreated correctly.


emyoulation:
The `.gpr.py` files for exporter and importer are not on the "Files changed" list.  Were they updated for version and copyright?

gramps/plugins/test/exports_test.py
emyoulation, line 1: Does not 6.2 have a change from 4 to 5 digit IDs across all categories? (e.g., `I%04d` to `I%05d`)  Does this test need to conform to that?  

https://github.com/gramps-project/gramps/pull/2235

dsblank:
> The .gpr.py files for exporter and importer are not on the "Files changed" list. Were they updated for version and copyright?

I don't know how that works in gramp core. Do I need to manually update those?

Nick-Hall:
> Do I need to manually update those?

No.  We don't update the versions of core plugins.  The release version is sufficient.

emyoulation:
While we are discussing CSV import export, Serge kindly added a [PR#1314 5.2 Enhancement](https://github.com/gramps-project/gramps/pull/1314) that expands delimiter options.  

It felt like PR feedback requesting fine-tuning the interface (just GUI labeling changes at that point) were straining the Developer patience. (see comments [1354132063](https://github.com/gramps-project/gramps/pull/1314#issuecomment-1354132063) and [1354194088](https://github.com/gramps-project/gramps/pull/1314#issuecomment-1354194088) ) 
So rather than risk the PR being withdrawn and losing the feature, I stopped suggesting ANY changes.

However, the interface for selecting the delimiter was added to the **local level: in any list view's View -> Configuration** tab even though the effect was application-wide.  Global settings ought to be in **Preferences**, not View **Configurations**. This was a more far-reaching change compared to Labeling changes.

I'd like feedback as to whether changes to the Delimiter GUI would be considered for implementation before committing the energy in writing up a format request. If such a request will simply irritate and not have potential for results, there are better ways to spend the energy.
 
image

CSV Export dialog

## Import
Preferences Import tab
image

Import dialog with manual CSV selections. (Note that there is no `.tsv` file extension support for autodetection of tab separated value delimiters.)
CSV Import dialog

dsblank:
@emyoulation asking for feedback on an unrelated issue is out of scope for this PR. Please make a bug report and leave out extraneous comments and be direct in your request. You can link to it from here.

emyoulation:
The related part of the question is... whether a different developer working on the related feature would be open to working on that other piece. If it will be batted back to the originator, then it is a dead end.

emyoulation:
adapted the changed plugins to be Addons and ran some tests under 5.2 (after disabling the built-ins).  Discovered that 4 object types ignored the ID specified and used the Default ID instead.  Passed the report through Claude with this response:

> Let me read the relevant import parsing sections carefully.Now I have a complete picture. The bug is clear and consistent across all object types. Let me confirm the exact ordering problem:The pattern is completely clear now. Here is the precise diagnosis:
>

## Root Cause: ID set *after* `db.add_*()` instead of *before*

In Gramps, `db.add_*(obj, trans)` assigns the auto-generated Gramps ID to the object **at call time**, overwriting whatever is in `obj.gramps_id`. The correct pattern — used correctly by `get_or_create_family` — is to call `set_gramps_id()` on the object **before** passing it to `db.add_*()`. The new parsers for events and citations both get this backwards, and the person and place parsers have a subtler version of the same bug.

There are **four distinct cases**:

### 1. `_parse_event` — ID never set at all
`event_ref` (e.g. `[E0042]`) is read from the CSV, but `set_gramps_id()` is never called on the new `Event()` object. The sequence is:
```python
event = Event()
# ← set_gramps_id() call is missing here
self.db.add_event(event, self.trans)   # auto-ID assigned, e.g. E0001
self.storeup("event", event_ref, event) # too late — ID is already E0001
```

### 2. `_parse_citation` — ID never set at all
Identical problem: `Citation()` is committed to the DB with no `set_gramps_id()` call before `db.add_citation()`.

### 3. `_parse_person` — ID set *after* `create_person()` which calls `db.add_person()` internally
```python
person = self.create_person()       # db.add_person() fires here → auto-ID set
...
if grampsid is not None:
    person.gramps_id = grampsid     # too late; must commit_person() to persist this
elif person_ref is not None:
    person.gramps_id = ...          # same — needs a follow-up commit
```
The person ID *is* eventually saved because `db.commit_person(person, self.trans)` is called at the end of `_parse_person`, so this one actually works — but it's fragile and relies on the final commit rescuing it.

### 4. `_parse_place` — same late-assignment pattern as person
`place.gramps_id = ...` is set after `create_place()` → `db.add_place()`, but is rescued by the final `db.commit_place()`. Same fragile pattern as person.

---

## The Fix

The pattern from `get_or_create_family` is the model: set the ID **before** `db.add_*()`. Here is the corrected `importcsv.py` with all four cases fixed:**Fix 1: `_parse_event` — set gramps_id before `db.add_event()`****Fix 2: `_parse_citation` — set gramps_id before `db.add_citation()`****Fix 3: `create_person` — accept and apply optional gramps_id before `db.add_person()`**Now update `_parse_person` to pass the ID into `create_person()` and remove the late assignments:**Fix 4: `create_place` — same pattern, accept optional gramps_id**Now update `_parse_place` to pass the ID into `create_place()` and remove the late assignment:Now verify the diff is exactly what we intended and nothing else changed:The diff is exactly and only the four targeted fixes. No unintended changes.Here's the complete explanation of what was wrong and what was changed.

---

## The Bug: ID Set After `db.add_*()`

In the Gramps database API, `db.add_*(obj, trans)` assigns the auto-generated Gramps ID to the object **at the moment of the call**, overwriting anything already in `obj.gramps_id`. The ID from the CSV must therefore be placed on the object **before** `db.add_*()` is called. The correct pattern is shown in the pre-existing `get_or_create_family()`:

```python
family.set_gramps_id(id_)   # ← BEFORE
self.db.add_family(family, self.trans)
```

All four new object types got this backwards.

### Fix 1 & 2 — `_parse_event` and `_parse_citation` (IDs never set at all)

Both parsers read the `[E0042]`-style ref from the CSV but never called `set_gramps_id()` anywhere. Added a `set_gramps_id()` call immediately after constructing the object and before `db.add_event()` / `db.add_citation()`:

```python
event = Event()
event.set_type(...)
if event_ref is not None and event_ref.startswith("[") and event_ref.endswith("]"):
    event.set_gramps_id(self.db.eid2user_format(event_ref[1:-1]))  # ← NEW, before add
self.db.add_event(event, self.trans)
```

### Fix 3 — `_parse_person` (ID set too late, rescued by accident)

`create_person()` called `db.add_person()` internally, then `_parse_person` assigned `person.gramps_id` afterwards. This happened to work only because `db.commit_person()` at the end of `_parse_person` wrote the corrected ID back. Removed the late assignments and resolved the ID upfront, passing it into `create_person(gramps_id=resolved_id)` so it is set before `db.add_person()`.

### Fix 4 — `_parse_place` (same late-assignment pattern as person)

`place.gramps_id = ...` was assigned after `create_place()` → `db.add_place()`. Same fix: `create_place(gramps_id=resolved_id)` sets it first.

emyoulation:
applying the Claude-suggestd changes did not resolve the ID problem.  (The test is importing into a new, empty tree)

dsblank:
Thanks @emyoulation; I fixed those issues and added more tests.

emyoulation:
Do you want to add a `gramps/images/hicolor/scalable/actions/gnome-mime-application-x-csv.svg` to the distribution?
gnome-mime-application-x-csv

dsblank:
> Do you want to add a gramps/images/hicolor/scalable/actions/gnome-mime-application-x-csv.svg to the distribution?

Sure, what needs to be done?

emyoulation:
That's an excellent question... And, after searching for a few hours, I have no idea!

In the NEWS file, the section "2021-07-26 Version 5.1.4", there is a section speaking of changes to the Icon/MIME type handling. But it is not linked to a PR or Issue. 
 https://github.com/gramps-project/gramps/blob/aef9f35ec64b67f5912c5d19543060d43f270a9a/NEWS#L1407-L1411

There's a 2021 'confirmed' but not resolved 	[0011209](https://gramps-project.org/bugs/view.php?id=11209): Media type showing unknown on TIF file type

Searching for MIME issues file is not enlightening
https://gramps-project.org/bugs/search.php?project_id=0&search=mime&sticky=on&sort=last_updated&dir=DESC&hide_status=-2&match_type=0

There's a section of the thumbnailing utility that seems to associate MIME types to icon image files based on the filename (rather file extension):  https://github.com/gramps-project/gramps/blob/aef9f35ec64b67f5912c5d19543060d43f270a9a/gramps/gen/utils/thumbnails.py#L187-L201

There are some Gramps `APP_` Constants that seem related to MIME types and follow the same pattern as the rest of the `gramps/images/hicolor/scalable/actions/gnome-mime-application-x-` filenaming pattern

emyoulation (approved):
I spun out the upgraded CSV importer/exporters to addon plugins. After hiding the built-ins with Plugin Manager Enhanced, they worked nicely.
 
The original requesting user asked on 29 Jun 2026 if this PR will be in 6.1? I had to reply that it'd already missed that bus before the 1st request 18d ago. Patching is beyond his comfort level. So provided the functionality as that test addon --- even though Hiding built-ins strains his comfort level. 

However, what can be done to ensure that PR#2372 will be in 6.2?

PR #2202, opened 2026-03-27
Unit Tests for Windows [was Invariant unittest environment]

stevenyoungs:
Update 2026-03-29: this PR has broadened into allowing unit tests to run correctly on Windows. Some of the changes are not Windows specific.

Do not load the user config file when running unittests.

Previously the user config settings were used to set db prefixes which could cause tests to fail. For example an iprefix of "I%05d" would cause the imported gramps_id to be I00044 rather than the default of I0044

To reproduce locally

1. Configure gramps to use a person ID pattern of I%05d
Start gramps
Edit > Preferences > ID Formats
Change the value for Person to I%05d
2. run the unittest `gramps.gen.filters.rules.test.person_rules_test.BaseTest.test_hasidof_matching` using the command `python -m unittest gramps.gen.filters.rules.test.person_rules_test.BaseTest.test_hasidof_matching`

Observe that the test fails with
```
AssertionError: Items in the second set but not the first:
'GNUJQCL9MD64AM56OH'
```
This [test](https://github.com/gramps-project/gramps/blob/eff3f9ae63110a6ac239467ae8c75f7c978fe544/gramps/gen/filters/rules/test/person_rules_test.py#L1096-L1108) is running a filter with the rule `HasIdOf(["I0044"])` which is failing because internally the `gramps_id` of the Person is "I00044"
This is just an example. There are many tests in the full suite that fail.

PR #2373, opened 2026-06-12
Skip legacy-bsddb import tests when no backend is loadable

eduralph:
## Summary
**User impact:** On a computer where Gramps is installed without the older BerkeleyDB database support — the normal situation on many Linux systems (for example Ubuntu 24.04) — running the project's own test suite reports outright failures for a handful of tests it should simply skip. A perfectly healthy install looks broken to anyone checking it.

This makes those tests skip whenever the database backend genuinely cannot be loaded on the running machine, instead of reporting failures — exactly as they already do on Windows.

Reported in Mantis [#13260](https://gramps-project.org/bugs/view.php?id=13260) ([Linux Mint] "Exception: can't load database backend: 'bsddb'") as the underlying backend-unavailability issue. This PR fixes the *test-suite* symptom — backend-less installs reporting failures that should be skips — not the runtime install error itself, so it uses the non-closing `Bug #13260` trailer rather than `Fixes`. Reported upstream in gramps#2314.

## What to look at
The whole change is one guard: the five legacy-bsddb import tests should skip — not fail — when neither BerkeleyDB backend is importable. To try it, on a default SQLite checkout with no backend installed, run `GRAMPS_RESOURCES=. python3 -m unittest gramps.plugins.test.imports_test`: the five tests now report `skipped` instead of `FAILED`, and wherever a backend *is* installed the tests run unchanged.

## Root cause
On a default SQLite install where neither `berkeleydb` nor `bsddb3` is importable
(the common Linux case, e.g. Ubuntu 24.04), the five legacy-bsddb import tests fall
through to `make_database("bsddb")` at `gramps/plugins/test/imports_test.py:248`,
which raises `can't load database backend: 'bsddb'`; the catch-all at
`gramps/plugins/test/imports_test.py:266-275` reconverts that into an assertion
failure. The existing skip guard fires only under `win()`
(`gramps/plugins/test/imports_test.py:245-246`, added in
`29a0bedaa03eae52cb2e678f9a1a22121a0a7fc5`), so the backend-less Linux case reports
`FAILED` instead of skipping.

## Fix
Replace the Windows-only guard with a real backend-loadability probe, so the tests
skip whenever the backend is genuinely unavailable on the running platform — exactly
as they already do on Windows — and run unchanged wherever a backend is installed.

- New `gramps/plugins/test/_bsddb_avail.py:38-59` — `bsddb_backend_available()`,
  a test-harness helper whose probe order (try `berkeleydb`, then `bsddb3`,
  anything else → unavailable) mirrors the plugin loader at
  `gramps/plugins/db/bsddb/bsddb.py:32-35`.
- `gramps/plugins/test/imports_test.py:245-246` — the `dbid == "bsddb" and win()`
  guard becomes `dbid == "bsddb" and not bsddb_backend_available()`, with a skip
  message naming the real reason.
- `gramps/plugins/test/imports_test.py:59-65` — the helper is imported defensively
  (a `try/except ImportError` fallback returns `True`, i.e. do not skip), so a
  missing prober never masks a real load failure. No `win()` / OS check is
  reintroduced anywhere.

The `except SkipTest: raise` re-raise (`gramps/plugins/test/imports_test.py:263-265`,
also from `29a0bedaa03eae52cb2e678f9a1a22121a0a7fc5`) is what propagates the skip out
of `db_load`'s catch-all and is left untouched. The `while True` upgrade loop, the
catch-all exception block, the sqlite (`dbid != "bsddb"`) path, and
`gramps/plugins/db/bsddb/` are all unchanged.

## Verification
- **Claim:** the five legacy-bsddb import tests skip whenever the backend is genuinely unavailable on the running platform, and run unchanged wherever a backend is installed.
- **Checked:** `gramps/plugins/test/imports_test.py:245-246` on maintenance/gramps61 — confirmed the pre-fix guard was `win()`-only, so a non-Windows host with no backend never reached the skip; `:248` — the `make_database(dbid)` the "available" branch still falls through to, unchanged, so an installed backend runs the real import/compare as before; `:266-275` — the catch-all that turned the loader exception into `AssertionError: Cannot open database … can't load database backend: 'bsddb'` pre-fix; `gramps/plugins/db/bsddb/bsddb.py:32-35` — the loader's `berkeleydb`→`bsddb3` probe the new helper mirrors, keeping "available" in lockstep with what `make_database("bsddb")` actually finds.
- **Test:** no new test file — the regression evidence is the existing `gramps/plugins/test/imports_test.py` tests flipping from failures to skips. On a backend-less Linux checkout (`GRAMPS_RESOURCES=. python3 -m unittest gramps.plugins.test.imports_test`): representative `test_imp_5_1_2bsd_zip` goes pre-fix `FAILED … can't load database backend: 'bsddb'` → post-fix `skipped 'bsddb backend not loadable on this platform (neither berkeleydb nor bsddb3 is installed)'`; whole suite pre-fix `FAILED (failures=5, …)` → post-fix `OK (skipped=…)`. The "backend installed" half is non-regressive by construction: the diff changes only the guard condition and message, and on non-Windows the original `win()` guard was already `False`, so the run-the-real-import path is byte-identical to today's behaviour.

Bug [#13260](https://gramps-project.org/bugs/view.php?id=13260)

### DIFF
--- a/gramps/plugins/test/_bsddb_avail.py
+++ b/gramps/plugins/test/_bsddb_avail.py
@@ -1,0 +1,53 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026 Gramps Development Team
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <https://www.gnu.org/licenses/>.
+#
+
+"""Test-harness helper: probe whether the bsddb database backend is loadable.
+
+The legacy-bsddb import tests in :mod:`gramps.plugins.test.imports_test` must
+*skip* — not *fail* — when the bsddb backend cannot be loaded, on every
+platform rather than only on Windows.  This helper answers the exact question
+the plugin loader asks at ``gramps/plugins/db/bsddb/bsddb.py`` lines 32-35:
+is :mod:`berkeleydb` (the maintained successor) importable, or failing that
+:mod:`bsddb3` (deprecated since Python 3.6)?  If neither is, the backend is
+unavailable and ``make_database("bsddb")`` would raise.
+"""
+
+
+def bsddb_backend_available():
+    """Return ``True`` iff the bsddb backend is loadable on this platform.
+
+    Mirrors the loader's probe order at ``gramps/plugins/db/bsddb/bsddb.py``
+    lines 32-35 (try ``berkeleydb`` then ``bsddb3``) so the test harness's
+    notion of "bsddb is available" matches what ``make_database("bsddb")``
+    will actually find.  Any import error — a missing package or a failed
+    shared library — counts as unavailable, exactly as the loader tolerates.
+    """
+    try:
+        from berkeleydb.db import DB  # noqa: F401
+
+        return True
+    except Exception:
+        pass
+    try:
+        from bsddb3.db import DB  # noqa: F401
+
+        return True
+    except Exception:
+        pass
+    return False
--- a/gramps/plugins/test/imports_test.py
+++ b/gramps/plugins/test/imports_test.py
@@ -64,6 +64,16 @@ from gramps.gen.db.exceptions import (
     DbConnectionError,
 )
 
+try:
+    from gramps.plugins.test._bsddb_avail import bsddb_backend_available
+except ImportError:  # pragma: no cover - defensive fallback
+
+    def bsddb_backend_available():
+        # If the availability prober cannot be imported, do not skip: let the
+        # real backend load attempt surface any problem rather than mask it.
+        return True
+
+
 # logger = logging.getLogger(__name__)
 
 # the following defines where to find test error diffs and import result XML files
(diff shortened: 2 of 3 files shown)


prculley:
I think this is better than just saying Windows won't do the tests.
I have been out of this for a while, so I just spent some time setting up a UCRT environment to do my testing and possibly development.  I used the AIO build script (mostly) to set up the environment.  While doing some of my routine tests for new releases, I was surprised that the bsddb trees I have would not load/upgrade.  So I went back and looked at the UCRT repo at MSYS2 to see if they had dropped bsddb and python-bsddb.  The repo has recently updated versions of both so I installed them with pacman.  The AIO build did not include these installs.

I could then load/upgrade the old bsddb trees with Family Tree manager.  I next tried running the import-tests to see if they would pass the db part with the old bsddb trees, and discovered that someone had decided to exclude these if the tests were run on Windows.  I did disable the check for win() and the tests successfully passed with the old bsddb dbs.

I found that there was some history that someone had decided that the  python-bsddb was no longer supported; apparently that was when this test change was implemented.  Its not clear to me why we could not continue using python-bsddb as long as it passed our tests; after all an upgrade only dumps out the db records, and probably doesn't run through most of the db code.

However I then disabled python-bsddb and tried to install the berkeleydb adapter that was supposed to replace python-bsddb.  It is on pypy so I used pip, only to discover that it would not build on the UCRT environment.  I then spent too much time figuring out how to fix the build script (setup.py) so it would install and finally  got it installed.

As expected it performs exactly the same as python-bsddb for our upgrade, and passed the bsddb db import tests.

I don't know how often users come back to Gramps after a long break, only to discover that they cannot load their old trees.  But it seems like including enough to load/upgrade these into the AIO would cut down some frustration/support when it does occur.

If this seems agreeable, I can do a PR putting the bsddb files back into the AIO build script.  Comments welcome.

Nick-Hall:
> If this seems agreeable, I can do a PR putting the bsddb files back into the AIO build script.

Yes, that sounds like a good idea.  Thank you.

eduralph:
@prculley  - the change is in 29a0beda (Steve Youngs, 2026-04-19), one month after the AIO migration, with the message "bsddb is no longer supported by MSYS2 UCRT64 builds".

stevenyoungs:
> @prculley - the change is in [29a0bed](https://github.com/gramps-project/gramps/commit/29a0bedaa03eae52cb2e678f9a1a22121a0a7fc5) (Steve Youngs, 2026-04-19), one month after the AIO migration, with the message "bsddb is no longer supported by MSYS2 UCRT64 builds".

Here's some background on this change:
PR #2202 
bsddb support was broken in the official AIO releases 6.0.3 to 6.0.7. https://github.com/gramps-project/gramps/pull/2202#issuecomment-4149057645. If we can get these tests working on Windows again with minimal effort, I'm all in favour of that. I think it is reasonable to assume that most Windows users were not impacted by the broken bsddb support in 6.0.3 onwards; there were no bug reports that I'm aware of.

@eduralph's change looks good to me and I think is worth committing independently of if we get bsddb support working on Windows again.

PR #2377, opened 2026-06-13
Add (i) info button to date editor with calendar and syntax help

dsblank:
## Summary

- Adds a flat `(i)` information button to the upper-left of the Calendar row in the Edit Date dialog.
- Clicking it opens a popover that explains:
  - **Calendar** — what the calendar selector does and when it is disabled (Free text mode, Dual-dated active, or a validation error)
  - **Dual-dated dates** — what slash-year notation means and how to enter it (e.g. `1735/6`, alternate New Year codes)
  - **Date entry syntax for this locale** — the existing per-locale two-column table of formats and live examples
- Fixes a bug where the calendar selector was always disabled when the dialog first opened (a spurious `DateError` from `Date.set()` running a round-trip sanity check on an empty `(0,0,0,False)` date value). (made a separate PR)
- Fixes `show_on_calendar()` to bail early for empty dates, preventing a spurious GTK Calendar `day-selected` signal. (made a separate PR)
- Fixes GTK translation detection in `grampsgui.py` to search `XDG_DATA_DIRS` and `/usr/share/locale-langpack`, so `language-pack-gnome-*` packages installed on Ubuntu are found when running from a virtualenv or conda environment. (made a separate PR)

## Motivation

The motivation for this PR is two-fold:

1. Explain some esoteric date-related things that are hard to remember
2. Provide examples of how to enter dates in a particular locale

In English:

image

In French (the text has not been translated yet):

image

I hope to add additional :information_source: buttons with short helpful advice in various places in Gramps.

## Test plan

- [ ] Open the Edit Date dialog and confirm the calendar selector is enabled (not greyed out) on first open.
- [ ] Click the `(i)` button and confirm the popover shows Calendar, Dual-dated, and syntax sections.
- [ ] Set date type to "Free text" and confirm the calendar selector becomes disabled; the popover text explains why.
- [ ] Check "Dual dated" and confirm the calendar selector becomes disabled; the popover explains this too.
- [ ] Enter an invalid date and confirm the calendar selector disables with an error in the status bar.
- [ ] Switch calendar to Hebrew/Julian/etc. and confirm the syntax table updates for that locale.
- [ ] On Ubuntu with `language-pack-gnome-en` installed, confirm no "GTK translations missing" warning on startup.

### DIFF
--- a/gramps/gui/editors/editdate.py
+++ b/gramps/gui/editors/editdate.py
@@ -584,28 +584,30 @@ class EditDate(ManagedWindow):
         return [
             {
                 "title": _("Calendar"),
-                "body": _(
-                    "Choose the calendar system: **Gregorian**, **Julian**, "
-                    "**Hebrew**, **French Republican**, **Persian**, **Islamic**, "
-                    "or **Swedish**. "
-                    "The selector is disabled when the date type is "
-                    "**Free text**, when **Dual dated** is active, or when "
-                    "the date contains a validation error (shown in the "
-                    "status bar at the bottom of this dialog)."
-                ),
+                # fmt: off
+                # Translators: this string may contain markdown formatting; preserve them
+                "body": _("Choose the calendar system: **Gregorian**, **Julian**, "
+                          "**Hebrew**, **French Republican**, **Persian**, **Islamic**, "
+                          "or **Swedish**. "
+                          "The selector is disabled when the date type is "
+                          "**Free text**, when **Dual dated** is active, or when "
+                          "the date contains a validation error (shown in the "
+                          "status bar at the bottom of this dialog)."),
+                # fmt: on
             },
             {
                 "title": _("Dual-dated dates"),
-                "body": _(
-                    "Slash dates such as `Jan 23, 1735/6` mark a historic transition "
-                    "between New Year conventions (e.g., March 25 vs. January 1). "
-                    "Enter a slash between years to create one: `1721/2`, `1719/20`, "
-                    "`1799/800`. Dual-dated dates use the Julian calendar. "
-                    "An alternate New Year day can be added in parentheses after "
-                    "the calendar name: `Jan 20, 1750 (Julian,Mar25)` or "
-                    "`Feb 23, 1710/1 (Mar25)`. Valid New Year codes: "
-                    "`Jan1`, `Mar1`, `Mar25`, `Sep1`."
-                ),
+                # fmt: off
+                # Translators: this string may contain markdown formatting; preserve them
+                "body": _("Slash dates such as `Jan 23, 1735/6` mark a historic transition "
+                          "between New Year conventions (e.g., March 25 vs. January 1). "
+                          "Enter a slash between years to create one: `1721/2`, `1719/20`, "
+                          "`1799/800`. Dual-dated dates use the Julian calendar. "
+                          "An alternate New Year day can be added in parentheses after "
+                          "the calendar name: `Jan 20, 1750 (Julian,Mar25)` or "
+                          "`Feb 23, 1710/1 (Mar25)`. Valid New Year codes: "
+                          "`Jan1`, `Mar1`, `Mar25`, `Sep1`."),
+                # fmt: on
             },
             {
                 "title": _("Date entry syntax for this locale"),
--- a/gramps/gui/editors/editdate.py
+++ b/gramps/gui/editors/editdate.py
@@ -64,6 +64,7 @@ from gramps.gen.const import URL_MANUAL_SECT1
 from ..display import display_help
 from ..managedwindow import ManagedWindow
 from ..glade import Glade
+from ..widgets.markdown_render import render_md, setup_tags
 
 LOG = logging.getLogger(".EditDate")
 _ = glocale.translation.sgettext
@@ -583,10 +584,11 @@ class EditDate(ManagedWindow):
             {
                 "title": _("Calendar"),
                 "body": _(
-                    "Choose the calendar system: Gregorian, Julian, Hebrew, "
-                    "French Republican, Persian, Islamic, or Swedish. "
+                    "Choose the calendar system: **Gregorian**, **Julian**, "
+                    "**Hebrew**, **French Republican**, **Persian**, **Islamic**, "
+                    "or **Swedish**. "
                     "The selector is disabled when the date type is "
-                    '"Free text", when "Dual dated" is active, or when '
+                    "**Free text**, when **Dual dated** is active, or when "
                     "the date contains a validation error (shown in the "
                     "status bar at the bottom of this dialog)."
                 ),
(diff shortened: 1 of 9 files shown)


emyoulation:
I feel that it is unwise to add Help as (paragraph or page sized) chunks of of text as hard-coded strings in Python modules. It seems like such text should be external files in a format that non-programmers can easily change. 

And that the Weblate system is better suited to individual words, phrases or sentences; not paragraphs or pages.  

Can the Info system be converted to a re-usable module that can be pointed at a JSON, XML, .htm HTML, .md markdown or .txt text file.  So that an Info GUI item can be added elsewhere in Gramps too.

dsblank:
> I feel that it is unwise to add Help as (paragraph or page sized) chunks of of text as hard-coded strings in Python modules. It seems like such text should be external files in a format that non-programmers can easily change.

They are not hardcoded.

> And that the Weblate system is better suited to individual words, phrases or sentences; not paragraphs or pages.

It will be filled-in by Weblate. 

I took a survey and people voted to translate larger sections of text:

image

Nick-Hall:
In Weblate, smaller strings are generally easier to handle.  Read the translator comments on the string containing [command line examples](https://hosted.weblate.org/translate/gramps-project/gramps/en/?checksum=0e1d973140c5e966) for example.

The string length in this PR seems OK to me though.

emyoulation:
>   I feel that it is unwise to add Help as (paragraph or page sized) chunks of of text as hard-coded strings in Python modules. It seems like such text should be external files in a format that non-programmers can easily change.
>> They are not hardcoded.

I do not understand this response.  “They are not hardcoded” does not seem accurate in the literal sense. If text is embedded in a Python module, it is hard-coded, even when marked translatable. The following part of the module has the ext in your screen capture above. (But the table of examples  

https://github.com/dsblank/gramps/blob/2e4bc474b045660a2e2bb3b4535f721d47eee2a7/gramps/gui/editors/editdate.py#L676-L702 

>    And that the Weblate system is better suited to individual words, phrases or sentences; not paragraphs or pages.
>> It will be filled-in by Weblate.
>>  I took a survey and people voted to translate larger sections of text:

Sending Paragraphs/Documents to Weblate is a misinterpretation of the response. (Regardless of interpretation, 5 respondents in a survey is too small to be statistically significant in any direction)  As 1 of the 4 respondents supporting "Entire Document", my intent was that the translation should be based on externalized reference document... which provides context and conversational flow... not in translatation-marked words, phrases or sentences. And words, phrases or sentences are the only situation where Weblate should be used to translate. An externalized file would be more translator friendly.  

Also, full-fledged **Help** Utility would not be use-friendly in the style of a super-sized roll-over (or click activated) tooltip anchored to a ℹ️ or . It has too many drawbacks. First, the content obscures the interface it explains. You cannot drag it around to reveal both elements simultaneously. (Overly detailed Tooltips often cause a similar, but not equivalent,  aggravation. A large tooltip for one field will obscures disambiguating data above/below/adjacent. Forcing the user to move the mouse pointer to somewhere that is not a tooltip rollover area.)


I've been working a different approach. When a features exceeds the purposes that can be satisfied by tooltips, the developer can link a <kbd>Help</kbd>, ℹ️ (or any of the other many Info/help/Faq icons in the Adwaita and Gnome themes).  Clicking that will pop a Window that is fed from the README.md (or other designated) file for its content... where GUI icons, hotlinks, bitmap illustration URLs, and text styling can augment the content. (Not limited to Headers, HorzRules, tables and such. It allows 'Help' content to be decoupled from the coding yet leverages the GitHub standard README.md that Developers are already writing.

dsblank:
> I've been working a different approach.... a Window that is fed from the README.md (or other designated) file for its content.

That's not a bad idea. But it doesn't exist.

We do have a mechanism that does exist (Weblate) and @Nick-Hall believes that these strings are not too long.

Feel free to propose or wish-design a different mechanism (but it still has to be translated... how will translators find it and translate it?)

emyoulation:
> That's not a bad idea. But it doesn't exist.

It is fully functional ... tested as an addon in 5.2, 6.0, and 6.1 version. I have just been holding it back because I was duplicating the same Markdown code in 5 of my addons already.  So I spun it out into a bespoke MarkdownUtils (with most GitHub Flavored Markdown plus a bunch of Gramps flavor markdown) module. 

I'm not even CLOSE to being a software architect. So it needs some input (probably re-design) on where the MarkdownUtils.py could live in the Core.  And the MarkdownDash plug living in the core would open a host of possibilities. (Like replacing the "Welcome" gramplet, FAQ gramplet with more attractive variants.)

image

dsblank:
> I have just been holding it back because I was duplicating the same Markdown code in 5 of my addons already.

I'd be glad to work with you on this, but this is independent from this PR. Of course once it exists, we can switch to it. (But we still have to figure out how to get it translated!)

emyoulation:
> (but it still has to be translated... how will translators find it and translate it?)

It currently uses a system where it will use the target filename to feed the content.  But if the same filename exists in the `locale` folder for the current locality, it takes precedence. So `README.md` in the English locale and `locale/fr/LC_MESSAGES/README.md` on a French system, falling back to English if it does not exist.

I am not confident that this is the best place for the system.

emyoulation:
> I'd be glad to work with you on this, but this is independent from this PR. Of course once it exists, we can switch to it. (But we still have to figure out how to get it translated!)

I would love (absolutely LOVE) your help to make this useful to the whole community. 

It is certain that I have been going down the Garden path with a few ideas. And that the code is suffering from organic sprawl. Most importantly, if the use becomes as widespread as I expect, it needs to be made as efficient as possible.

emyoulation:
I will (cringe) post the proof-of-concept source to my public 6.1addon repository and start the Ideas discussion. For the last week, I've been wrestling with the PluginMgrPlus addon ... which has will have the most pervasive use of the MarkdownUtils. It uses it for the content panels, the interactive help panel and the pop-up Help buttons windows. 

My thought was that getting all that working smoothly might save some revising of the MarkdownUtils. So far, no tweaks have been necessary.

dsblank:
I refactored the info-text:

* could be used for a search function
* could be used to generate documentation
* can use and render markdown

Some formatting really helps readability:

image

gramps/gui/editors/editdate.py
```diff
@@ -585,7 +585,7 @@ def get_help_topics(cls) -> list[dict]:
             {
                 "title": _("Calendar"),
                 # fmt: off
-                # Translators: this string may contain markdown formatting; preserve them
+                # Translators: this string may contain markdown formatting; preserve it
```
Nick-Hall, line 588: We could also add the `md-text` flag to these strings in Weblate.

dsblank: Yes, but that would be a manual process. Trying to automate everything that we can! :smile: 

But also, if we tag them as `md-text`, Weblate will "validate" them, and we might want to deviate slightly from pure markdown (for example, links that resolve to UI components).

Nick-Hall: Weblate checks can often be useful because translators sometimes ignore instructions.  I don't really mind though - I was just making you aware of the option.

dsblank: Thanks; appreciate it!

emyoulation:
Please consider choosing a layout standard where the placement for each ℹ️ gadget should be to the far-right of the featured textbox/combo box/menu.

Placement at right is logical in that (in left-to-right reading) it is "lower priority" visually and  thus easier to ignore when unneeded. Which make sense because needing "help" should be the exception rather than the norm.

PR #2380, opened 2026-06-13
Fix GTK translation detection missing Ubuntu locale-langpack paths

dsblank:
## Summary

On Ubuntu, GTK translations installed via `language-pack-gnome-*` live in
`/usr/share/locale-langpack`, not in the Python prefix's `share/locale`.
When running Gramps from a virtualenv or conda environment,
`gettext.find('gtk30')` searches only the Python prefix and returns `None`,
causing a spurious **"GTK translations missing"** warning dialog even when
the translations are correctly installed.

Replace the single `gettext.find(GTK_GETTEXT_DOMAIN)` call in
`grampsgui.py` with a `_gtk_translations_found()` helper that searches:
- every directory in `XDG_DATA_DIRS` (the standard freedesktop path list)
- `/usr/share/locale-langpack` (Ubuntu-specific fallback)

This silences the false-positive warning for users who have GTK
translations installed via the normal Ubuntu package manager.

## Test plan

- [ ] On Ubuntu with `language-pack-gnome-en` (or any language pack)
  installed, run Gramps from source (`python3 Gramps.py`) and confirm
  no "GTK translations missing" warning appears on startup.
- [ ] Remove all `language-pack-gnome-*` packages and confirm the warning
  still appears correctly.
- [ ] Confirm the fix has no effect on non-Linux platforms (the warning is
  already guarded by `lin()`).

### DIFF
--- a/gramps/gui/grampsgui.py
+++ b/gramps/gui/grampsgui.py
@@ -28,6 +28,7 @@
 # -------------------------------------------------------------------------
 import sys
 import os
+import gettext
 import logging
 
 LOG = logging.getLogger(".grampsgui")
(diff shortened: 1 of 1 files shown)


Nick-Hall:
This approach may actually work with the Snap package removing the need to exclude it.

PR #2382, opened 2026-06-15
Silence noisy traceback in FamilySearch tree test

dsblank:
## Summary

- `test_add_persons_one_fetch_raises_other_still_imported` intentionally raises a `RuntimeError` to simulate a network failure, but didn't patch `LOG`
- This caused `LOG.warning(..., exc_info=True)` inside `add_persons` to emit a full traceback to stderr on every test run
- Fix: wrap the `add_persons` call with `patch("gramps.gen.fs.tree.LOG")`, matching the pattern already used by the companion test `test_add_persons_fetch_exception_is_logged_not_raised`

## Test plan

- [x] Run `python3 -m unittest gramps.gen.fs.test.tree_test -v` — all 12 tests pass with no traceback output


Nick-Hall:
Rebased.

PR #2383, opened 2026-06-16
Prevent race condition between JSON and blob metadata

DavidMStraub:
It turned out that my fix https://github.com/gramps-project/gramps/pull/2029 was not complete: when two concurrent processes access the data simultaneously, a reace condition can occur that leads to a database in a broken state, with partially populated JSON column. See https://github.com/gramps-project/gramps-web-api/issues/880, in particular this analysis: https://github.com/gramps-project/gramps-web-api/issues/880#issuecomment-4711947501.

It would be great if this could be included in the next 6.0 maintenance release. Thanks!

### DIFF
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -811,16 +811,15 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
             # Note: downgrade check only works from v5.1.2 and later on sqlite
             self.has_changed = 1  # to make sure genderstats gets saved
             if self.use_json_data():
-                # Write JSON first so concurrent readers always find valid
-                # json_data (avoids NULL json_data race window).
                 self.set_serializer("json")
                 self._set_all_metadata()
-            self.set_serializer("blob")
-            self._set_all_metadata()
-            if self.use_json_data():
-                # Restore the JSON serializer; the schema uses json_data columns
-                # and subsequent commits must target that column.
+                # Write blob second, then restore JSON as the active serializer.
+                self.set_serializer("blob")
+                self._set_all_metadata()
                 self.set_serializer("json")
+            else:
+                self.set_serializer("blob")
+                self._set_all_metadata()
             self.has_changed = 0  # number of commits
 
         self.db_is_open = True
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -817,6 +817,10 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
                 self._set_all_metadata()
             self.set_serializer("blob")
             self._set_all_metadata()
+            if self.use_json_data():
+                # Restore the JSON serializer; the schema uses json_data columns
+                # and subsequent commits must target that column.
+                self.set_serializer("json")
             self.has_changed = 0  # number of commits
 
         self.db_is_open = True
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -809,12 +809,14 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
             # for new db we always need blob metadata to allow prior Gramps versions
             # to make it to the downgrade version check
             # Note: downgrade check only works from v5.1.2 and later on sqlite
-            self.set_serializer("blob")
             self.has_changed = 1  # to make sure genderstats gets saved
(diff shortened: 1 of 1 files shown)


DavidMStraub:
Looks like I don't have permissions to trigger a copilot review here.

Nick-Hall:
Please rebase this on the _gramps61_ branch.  There will be no further 6.0 maintenance releases.

stevenyoungs:
@DavidMStraub is this fix sufficient to fix all race conditions?
e.g. `need_to_set_metadata` is initialised some time before the metadata is actually written. It is easy to imagine two or more threads could compute `need_to_set_metadata = True`. There's no guarantee when the last of these threads will actually commit the metadata. By that time, could another worker have already modified the metadata, in which case it would be overwritten? I readily admit that this requires some unusual scheduling of the various threads but I've been around multi-threaded systems long enough to know that this occasionally happens.

stevenyoungs:
I've no knowledge of gunicorn. Are workers independent processes or threads? If processes, what is the behaviour of the db layer when it comes to committing writes to disk. i.e. if worker process A calls `_set_all_metadata()` when will process B be able to read this. If gunicorn uses worker threads, does the db driver handle this sequencing?

Nick-Hall:
The case where Gramps crashes should also be taken into account.  We get regular reports of users losing metadata after a crash.  Perhaps we should not be caching metadata.

DavidMStraub:
> @DavidMStraub is this fix sufficient to fix all race conditions?

Probably not, because the correct architecture is to make the write atomic, but that's beyond the scope of my PR. I'm trying to fix a concrete bug with a simple change.

> The case where Gramps crashes should also be taken into account. We get regular reports of users losing metadata after a crash. Perhaps we should not be caching metadata.

Also out of scope here. I'd suggest to discuss the overall architecture after the bug is fixed.

DavidMStraub:
> There will be no further 6.0 maintenance releases.

Is this decision up for debate?

https://github.com/gramps-project/gramps-web-api/issues/880 is still open and I just received another bug report for the same issue.

This is a serious blocker for people getting started with Gramps Web.

According to the [6.2 Roadmap](https://www.gramps-project.org/wiki/index.php/6.2_Roadmap), it won't be released until October. I can't leave this bug in Web API open for another 3 months.

dsblank:
> Is this decision up for debate?

This wouldn't be an issue at all if Gramps 6.1 were released, right? Then everyone can just upgrade and get all the bug fixes.

In my opinion, we should put all bug fixes in the currently available version. Right now we are in limbo: all bug fixes are going to 6.1, which is unavailable. If we are not going to change this, then we have to release versions faster.

DavidMStraub:
> This wouldn't be an issue at all if Gramps 6.1 were released, right? Then everyone can just upgrade and get all the bug fixes.

In my opinion no, because 6.1 is a feature release with breaking changes including a database schema migration. This needs planning for downstream projects like Web API, and we don't have even have a release roadmap with reliable dates for it.

Nick-Hall:
> In my opinion, we should put all bug fixes in the currently available version. Right now we are in limbo: all bug fixes are going to 6.1, which is unavailable.

This is an unusual situation.  We should have released v6.1 by now, but since we have to rely on volunteers this has been delayed.

If I can't get this resolved quickly then I will ask Brian for admin access.

emyoulation:
Can we re-assess the roadmap?

How about publicly releasing a 6.1 that has the FamilySearch code but does not have a the middleware active.  We can use that to field test that none of the additions cause problems/slowdowns when no connect exists. (Like: bogging down as it repeatedly tries to check for a net connection. Or has problems because the macOS SSL validation is missing bits and pieces.)

That will unblock 6.1.1 for bugfixes.

Then release a 6.2 when the FamilySearch connection testing issues are worked out.

And I've seen a least one reference to the master branch as 7.0 ... maybe that should be the new designation?

Nick-Hall:
> How about publicly releasing a 6.1 that has the FamilySearch code but does not have a the middleware active.

Yes.  I have asked for a progress report on the FamilySearch middleware installation.  I will either ask for admin access to our web server and do the job myself, or perhaps just release v6.1 anyway.

> And I've seen a least one reference to the master branch as 7.0

I think that the _master_ branch will probably become v6.2.

DavidMStraub:
Is 6.2 still planned for October? In this case, I don't think I'll release a Web API version that's compatible with 6.1. It would require a database migration for all users with zero benefits.

PR #2388, opened 2026-06-19
Prompt new users to import the Example Family Tree on startup

dsblank:
## Summary

- On startup whenever the Example Family Tree doesn't exist, gramps will show a dialog offering to import `example/gramps/data.gramps` into a new tree called "Example Family Tree", with a "Don't ask again" checkbox — modelled on the existing Tips of Day dialog.
- After importing, seed the user's OSM tile cache with bundled zoom-0–2 tiles so the Geography view works immediately on first launch without network access.
- Fix `{GRAMPS_RESOURCES}` media-path expansion by adding the key to the `ENV` dict in `const.py`, so example media resolves correctly in installed packages.

## New / changed files

| File | Change |
|------|--------|
| `gramps/gui/exampleimport.py` | New module: `ExampleImportDialog`, `maybe_prompt_example_import()`, `_do_import()`, `_seed_map_tiles()` |
| `gramps/gui/grampsgui.py` | Call `maybe_prompt_example_import(self._vm)` after the TipOfDay block in `post_init_interface()` |
| `gramps/gen/config.py` | Register `behavior.show-example-prompt` (default `True`) |
| `gramps/gen/const.py` | Export `DOC_DIR`; add `GRAMPS_RESOURCES` key to `ENV` dict |
| `scripts/build_hook.py` | Copy `data/maps/` → `build/share/gramps/maps/` |
| `data/maps/openstreetmap/{0,1,2}/` | 21 bundled OSM PNG tiles (zoom 0–2, ~216 KB total) |
| `po/POTFILES.in` | Add `gramps/gui/exampleimport.py` |

## Behaviour details

- The prompt appears **only** when `behavior.show-example-prompt` is `True` **and** no tree named "Example Family Tree" exists.
- Clicking **Skip** or checking **Don't ask again** sets the config key to `False` permanently.
- Clicking **Import Example Tree** imports the tree and opens it, but leaves the config key `True` — so if the user later deletes or renames the tree the prompt reappears on next startup.
- If `data.gramps` is not present (e.g. a minimal snap install), `maybe_prompt_example_import()` returns silently.

## Test plan

- [ ] Fresh install (or reset `show-example-prompt = True` in `gramps.ini`): dialog appears on startup.
- [ ] Click "Import Example Tree": tree is created, opened, and media paths resolve.
- [ ] Geography view (world zoom): tiles display without network.
- [ ] Click "Skip": dialog does not reappear.
- [ ] Check "Don't ask again" + Skip: dialog does not reappear.
- [ ] Rename or delete the Example Family Tree, restart: dialog reappears.


emyoulation:
> with a "Don't ask again" checkbox — modelled on the existing Tip of Day dialog.

Selection state of the checkbox in the Tip of the Day dialog has a Edit -> Preferences : General tab checkbox labeled "Display Tip of the Day".

Will there be a Preferences checkbox for whether to offer to import a sample tree during startup?

Or will recursion testers need to reset this softswitch .ini setting with a text editor?

dsblank:
> Will there be a Preferences checkbox for whether to offer to import a sample tree during startup?

No, I don't want to make Gramps more complicated.

> Or will recursion testers need to reset this softswitch .ini setting with a text editor?

I have no idea what half those words mean in this context, but testers can set the value on the command-line:

```
python Gramps.py --config=behavior.show-example-prompt:True
```

PR #2393, opened 2026-06-20
Add image source option to genealogytree Tree reports

eduralph:
## Summary
**User impact:** When you make a genealogy Tree chart report, the photos in it come out low-quality (small thumbnails), and there is no way to get full-resolution pictures — the generated files also carry machine-specific thumbnail filenames that make them hard to share or identify. This affects anyone who wants print-quality images in a Tree report.

This adds a user-selectable option to choose full-resolution original images instead of the cached thumbnails.

Reported in Mantis [#13888](https://gramps-project.org/bugs/view.php?id=13888).

## What to look at
The change adds an "Images" choice to the Tree report's Node Options: thumbnails (default, smaller PDF) or original full-resolution images. To try it: generate a genealogytree Tree report, find the new Images option, switch it to "Original images (full resolution)", and confirm the output uses the full-size photos; with the default, the output uses thumbnails but records the original path in a LaTeX comment so the `.tex` stays identifiable.

## Root cause
Since PR 1620 (core commit 7335883f68), Tree reports embed only cached thumbnails to avoid unreasonably large PDFs and LaTeX parser failures on special-character paths. However, this reduced image quality (3–12 kB thumbnails vs 500 kB–2 MB originals) and machine-tied filenames make the generated `.tex` files non-standalone and unidentifiable.

## Fix
This adds a user-selectable `images` option to `gramps/gen/plug/docgen/treedoc.py` allowing users to choose between cached thumbnails (default, preserving PR 1620) and original full-resolution images (pre-1620 behaviour, now opt-in).

**Changes in `gramps/gen/plug/docgen/treedoc.py`:**

- New `_IMAGES` constant (around line 65) defining two items: `"thumbnail"` (default, _("Thumbnails (smaller PDF)")) and `"original"` (_("Original images (full resolution)")).
- `TreeOptions.__init__` (line 169): Add `self.images = None` alongside `self.detail`.
- `TreeOptions.add_menu_options` (line 201): Add `EnumeratedListOption("Images", "thumbnail")` in the Node Options category, immediately after the detail option, with help text explaining the size/standalone trade-off.
- `TreeDocBase.__init__` (line 379): Read the option with `self.images = get_option("images").get_value()`.
- `write_node` (line 593–604): Compute the full media path once, then branch on `self.images`:
  - When `"original"`: emit the full-resolution path (pre-1620 behaviour).
  - When `"thumbnail"` (default): emit the cached thumbnail path plus a LaTeX comment `% original image: <path>` so the `.tex` stays identifiable.

**Changes in `po/POTFILES.skip`:**

- Register the new test file `gramps/gen/plug/docgen/test/treedoc_test.py` in the `gen.plug.docgen.test` section (line 241), as a test with no translatable strings.

## Verification
- **Claim:** users can opt into full-resolution original images in Tree reports, with cached thumbnails (the PR 1620 safe behaviour) remaining the default.
- **Checked:** on `maintenance/gramps61` — `gramps/gen/plug/docgen/treedoc.py:65-69` (`_IMAGES` defines both items, defaults to thumbnail), `:169` (`self.images` initialized in `TreeOptions.__init__`), `:201-202` (option added to Node Options after detail, translatable label + help), `:379` (`self.images` fetched in `TreeDocBase.__init__`), `:593-604` (`write_node` branches: full path when `"original"`, thumbnail path + `% original image:` comment when `"thumbnail"`); `po/POTFILES.skip:241` registers the new test.
- **Test:** `gramps/gen/plug/docgen/test/treedoc_test.py` (new) — drives the production `write_node` path via a real `TreeOptions` menu, a concrete `TreeGraphDoc` subclass, a stub database and a media reference. Asserts `original` → `.tex` contains `image = {<original media path>}` with no `% original image:` comment; `thumbnail` (default) → `.tex` contains `image = {<cached thumbnail path>}` (not the original) plus a `% original image: <original path>` comment. It stubs `get_thumbnail_path` (an unchanged collaborator that would load Gtk in the headless runner) while keeping branch routing and comment emission in production code. Green-with-fix / red-without-fix on clean upstream (both the option and the branching logic are required).

cc @hgohel — this reconciles the requirements outlined in 13888 as left over (PR 1620 safety with user opt-in for original resolution + identifiability), defaulting to the current safe behaviour. I don't have arzdev's github handle and am also not 100% sure this is exactly what you had discussed, so I'd appreciate some feedback here.

Fixes [#13888](https://gramps-project.org/bugs/view.php?id=13888)

### DIFF
--- a/gramps/gen/plug/docgen/test/treedoc_test.py
+++ b/gramps/gen/plug/docgen/test/treedoc_test.py
@@ -1,0 +1,162 @@
+#
+# Gramps - a GTK+/GNOME based genealogy program
+#
+# Copyright (C) 2026  Gramps Development Team
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <https://www.gnu.org/licenses/>.
+#
+
+"""
+Unit tests for the image-source option of
+``gramps.gen.plug.docgen.treedoc`` (bug 0013888).
+
+Since PR 1620 (core commit 7335883f68) the genealogytree (Tree) reports
+embed only cached thumbnails. This adds a user-selectable ``images``
+option restoring the pre-1620 full-resolution emission as an opt-in, and
+- when a thumbnail is used - annotates the node with the original
+filename as a LaTeX comment so the ``.tex`` stays identifiable.
+
+These tests drive the production ``TreeDocBase.write_node`` path (via the
+concrete ``TreeGraphDoc`` subclass and a real options menu built by
+``TreeOptions``) and assert both branches:
+
+* ``images = "original"`` -> ``image = {<original media path>}`` and no
+  ``% original image:`` comment;
+* ``images = "thumbnail"`` (default) -> the cached-thumbnail path (not the
+  original) plus a ``% original image: <path>`` comment naming the source.
+
+Import-safety
+-------------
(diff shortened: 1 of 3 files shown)


hgohel:
Thanks for the heads up. Tagging @azrdev for review

PR #2396, opened 2026-06-20
Refresh sidebar filter Type selectors on custom-type changes

eduralph:
## Summary
**User impact:** If you create a new custom "Type" (for example, a Note type made automatically when you import a GEDCOM with errors) while your family tree is already open, that new type does not show up in the sidebar filter's Type drop-down. You have to close and re-open the view before you can filter by it. The repository filter also showed the wrong list of types entirely.

This makes the sidebar filter's Type drop-downs refresh from the live database so newly added custom types appear right away.

Reported in Mantis [#13716](https://gramps-project.org/bugs/view.php?id=13716).

## What to look at
The sidebar filter's Type selectors now re-read the database whenever it changes or when you open the drop-down, instead of only once when the widget was first built. To try it: open the Notes category, note the options in the sidebar Type filter, add a Note with a new custom type via the Note editor, then re-open the same Type drop-down (without recreating the view) — the new type is now offered. A headless regression test drives this same refresh path.

## Root cause
A view's sidebar-filter "Type" selector (the same widget the Filter Gramplet hosts) reads the database's custom types exactly once at widget construction time via `dbstate.db.get_note_types()` (and similar per-filter methods), storing the result in `self.custom_types`. The selector's model (`StandardCustomSelector`) is built once from that snapshot and never rebuilt. When custom types are added to the already-open database after the filter widget is constructed (e.g., a "GEDCOM import" Note type created by importing a GEDCOM with errors), the new types never appear until the filter widget is torn down and rebuilt. This differs from the editor dialogs, which rebuild their type selector from the database "anew every time" they are shown (Mantis note 3).

## Fix
**New GUI-free module: `gramps/gui/filters/sidebar/_typefilterlist.py` (lines 1–82)**

Introduces `TypeFilterList` — a registry of `(fetch, apply)` pairs that orchestrates the refresh of a sidebar filter's type selectors without GUI imports. The `refresh()` method re-calls each registered `fetch` (which re-reads the *live* database) and passes the result to `apply` (which rebuilds the selector widget in place).

**Shared path: `gramps/gui/filters/sidebar/_sidebarfilter.py`**

- Line 178: Imports `TypeFilterList`
- Lines 186–187: `SidebarFilter.__init__` initializes `self._type_filters` and calls the new `_register_type_filters()` hook
- Line 195: `_db_changed` calls `self._type_filters.refresh()` to update selectors when the database changes
- Lines 203–244: Three new methods in `SidebarFilter`:
  - `_register_type_filters()` (lines 203–212): Hook for subclasses; default does nothing
  - `add_type_filter(monitored, db_method)` (lines 214–236): Registers a type selector; resolves `db_method` on the *live* database, connects the combo's `notify::popup-shown` signal for refresh-on-open
  - `_type_popup_shown(combo, pspec)` (lines 238–244): Signal handler that calls `refresh()` when a selector's drop-down opens

**Sidebar filter subclasses: `_notesidebarfilter.py`, `_eventsidebarfilter.py`, `_familysidebarfilter.py`, `_personsidebarfilter.py`, `_placesidebarfilter.py`, `_reposidebarfilter.py`**

Each now overrides `_register_type_filters()` and calls `add_type_filter()` once per type selector:
- `_notesidebarfilter.py:85–87`: Registers Note's Type selector with `get_note_types`
- `_eventsidebarfilter.py:84–85`: Registers Event's Type selector
- `_familysidebarfilter.py:98–100`: Registers Event and Family Relation type selectors
- `_personsidebarfilter.py:126–128`: Registers Event type selector
- `_placesidebarfilter.py:99–101`: Registers Place type selector
- `_reposidebarfilter.py:73, 92–94`: **Fixes pre-existing bug** — reads `get_repository_types()` (not `get_event_types()`) for the repository Type selector; registers the correct method

**Refactor for in-place rebuild: `gramps/gui/autocomp.py` and `gramps/gui/widgets/monitoredwidgets.py`**

- `autocomp.py:154`: Stores `self.completion_store` for later reuse
- `autocomp.py:179–225`: Extracts model-filling logic into `_fill_menu()` and `_fill_list()` (clearing first, then re-appending), callable on existing store objects
- `autocomp.py:258–272`: New `rebuild()` method re-fills the selector's model in place, preserving the widget object the GUI is bound to (critical for live update of an open drop-down)
- `monitoredwidgets.py:588–600`: New `rebuild()` method on `MonitoredDataType` preserves the current selection while rebuilding via `StandardCustomSelector.rebuild()`

**POTFILES: `po/POTFILES.skip`**

Lines 611–613: Registers the new files (no translatable strings):
- `gramps/gui/filters/sidebar/_typefilterlist.py`
- `gramps/gui/filters/sidebar/test/__init__.py`
- `gramps/gui/filters/sidebar/test/_sidebarfilter_test.py`

## Verification
- **Claim:** a custom type added to the already-open database is offered by the sidebar Type selector after a refresh, without tearing down and rebuilding the view; and the repository filter reads repository types rather than event types.
- **Checked:** target branch `gramps-project/gramps @ upstream/maintenance/gramps61` (commit b679c084f6). Red→green proof (C4 verify gate, clean upstream/maintenance/gramps61): GREEN (fix applied) — headless unit test runs 4 tests OK; RED (production reverted, test kept) — test fails because the `TypeFilterList` repopulate path does not exist.
- **Test:** shipped regression test `gramps/gui/filters/sidebar/test/_sidebarfilter_test.py` (part of `patch.diff`) drives the production repopulate path (`TypeFilterList.refresh()` + `SidebarFilter.add_type_filter()` wiring) without a display, building filter instances via `__new__` to avoid `__init__`'s Gtk widget creation, with test doubles (`_RecordingMenu`, `_FakeDb`) for the selector and database: `test_type_selector_reflects_new_custom_type_after_refresh()` (lines 490–513) verifies a new custom Note type is offered after refresh and absent beforehand (the bug); `test_refresh_reflects_removed_custom_type()` (lines 515–526) tracks removed types; `test_typefilterlist_refetches_live_source()` (lines 528–541) verifies re-reading the source on every refresh; `test_repo_filter_reads_repository_types_not_event_types()` (lines 553–575) locks in the repository-filter fix. Advisory testbed interface repro `engine/interface/test_bug_0013716_sidebar_filter_type.py` (GUI/AT-SPI layer, NOT in `patch.diff`) opens the Notes category, adds a custom type via the Note editor, and asserts the sidebar Type combo now offers it without view recreation; it skips gracefully when dogtail/AT-SPI is unavailable — the headless unit test is the gating proof.

Depends on #2357

Fixes [#13716](https://gramps-project.org/bugs/view.php?id=13716)

### DIFF
--- a/gramps/gui/autocomp.py
+++ b/gramps/gui/autocomp.py
@@ -151,6 +151,7 @@ class StandardCustomSelector:
         else:
             self.store = self.create_list()
             completion_store = self.store
+        self.completion_store = completion_store
 
         self.selector.set_model(self.store)
         self.selector.set_entry_text_column(1)
@@ -178,6 +179,15 @@ class StandardCustomSelector:
         menu.
         """
         store = Gtk.TreeStore(int, str, bool)
+        self._fill_menu(store)
+        return store
+
+    def _fill_menu(self, store):
+        """
+        Fill (or refill) the given tree store with the menu, clearing it first
+        so it can be reused when the custom values change (bug 13716).
+        """
+        store.clear()
         for heading, items in self.menu:
             if self.active_key in items:
                 parent = None
@@ -192,13 +202,21 @@ class StandardCustomSelector:
                 key, value = self.get_key_and_value(event_type)
                 store.append(parent, row=[key, value, True])
 
-        return store
-
     def create_list(self):
         """
         Create a model and fill it with a sorted flat list.
         """
         store = Gtk.ListStore(int, str)
+        self._fill_list(store)
+        return store
+
+    def _fill_list(self, store):
+        """
+        Fill (or refill) the given list store with a sorted flat list, clearing
+        it first so it can be reused when the custom values change (bug 13716).
+        """
+        store.clear()
+        self.active_index = 0
         keys = sorted(self.mapping, key=self.by_value)
         index = 0
         for key in keys:
@@ -216,7 +234,21 @@ class StandardCustomSelector:
                     self.active_index = index
                 index += 1
 
-        return store
+    def rebuild(self, additional):
+        """
+        Refill the selector's models with a new set of custom (additional)
+        values, in place so an open drop-down updates live.
+
+        This keeps a database-derived type selector consistent with the
+        database's current custom types without recreating the widget
+        (bug 13716).
+        """
+        self.additional = additional
+        if self.menu:
+            self._fill_menu(self.store)
+            self._fill_list(self.completion_store)
+        else:
+            self._fill_list(self.store)
 
     def by_value(self, val):
         """
--- a/gramps/gui/filters/sidebar/_eventsidebarfilter.py
+++ b/gramps/gui/filters/sidebar/_eventsidebarfilter.py
@@ -92,6 +92,9 @@ class EventSidebarFilter(SidebarFilter):
 
         SidebarFilter.__init__(self, dbstate, uistate, "Event")
 
+    def _register_type_filters(self):
+        self.add_type_filter(self.event_menu, "get_event_types")
+
     def create_widget(self):
         cell = Gtk.CellRendererText()
         cell.set_property("width", self._FILTER_WIDTH)
--- a/gramps/gui/filters/sidebar/_familysidebarfilter.py
+++ b/gramps/gui/filters/sidebar/_familysidebarfilter.py
@@ -112,6 +112,10 @@ class FamilySidebarFilter(SidebarFilter):
 
         SidebarFilter.__init__(self, dbstate, uistate, "Family")
 
+    def _register_type_filters(self):
+        self.add_type_filter(self.event_menu, "get_event_types")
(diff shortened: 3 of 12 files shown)


prculley:
How does this recognize custom type changes from other editors, imports etc?
As near as I can tell db_changed would only update when switching to a different family tree.

Been a while since I tried to fix this as part of the big places PR so I could be remembering incorrectly.

eduralph:
> How does this recognize custom type changes from other editors, imports etc?

There's no subscription to "a type was created" events from editors or imports. The fix is pull-on-open, not push-on-change:

Each Type selector's combo has notify::popup-shown connected (_sidebarfilter.py:236).
When you open the drop-down, _type_popup_shown fires and calls _type_filters.refresh().
refresh() re-runs the selector's fetch, which is getattr(self.dbstate.db, db_method)() against the live database (_sidebarfilter.py:227-229) — not a snapshot.

PR #2296, opened 2026-05-09
Show date format examples in Preferences Data tab date format menu

dsblank:
## Summary

- In the Preferences → Data tab, the date format drop-down now shows a rendered example alongside each format name, e.g. `ISO (2001-04-21)` or `Month Day, Year (April 21, 2001)`.
- The example date used is 2001-04-21, the first public release of GRAMPS.
- A temporary displayer instance is constructed for each format so the global displayer's state is never mutated.

## Original author

This work was written by @emyoulation (commits `0a86f9f` and `608985d` from [PR #2287](https://github.com/gramps-project/gramps/pull/2287)). This PR is a clean rebase of that work — original authorship is preserved in the commits.

## Why a new PR was needed

PR #2287 was originally submitted against `maintenance/gramps60`. When @Nick-Hall asked for a rebase onto `master`, @emyoulation used GitHub's web UI base-branch switcher, which only changes what the PR *compares against* — it does not actually rebase commits. As a result, GitHub recalculated the diff and the PR picked up several unrelated commits from `maintenance/gramps60` that are not in `master` (Windows AIO build changes by @hgohel and @edwardralph), while the two actual feature commits were buried.

Since there is no way to push a force-corrected branch to someone else's fork, this new PR cherry-picks only @emyoulation's two commits onto a fresh branch from `master`.

## Test plan

- [ ] Open Preferences → Data tab and confirm the date format drop-down shows format names with rendered examples.
- [ ] Change the selected format and confirm the preference is saved and applied.
- [ ] Confirm no other Preferences behavior is affected.

PR #2406, opened 2026-06-22
Accept compact 8-digit dates without separators in date parser

dsblank:
## Summary

- Adds a `_compact8` regex that matches exactly 8 consecutive digits and parses them as a full date (day, month, year) according to the locale's field order.
- **DMY locales** (e.g. German): `01121720` → day=01, month=12, year=1720 (displayed as `01.12.1720`)
- **MDY locales** (e.g. US English): `01121720` → month=01, day=12, year=1720
- **YMD locales** (e.g. Japanese): unchanged — `_isotimestamp` already handles `YYYYMMDD`
- The new check runs before `_isotimestamp` to prevent that pattern from greedily misinterpreting `DDMMYYYY` input as a bogus year. If the compact digits don't form a valid date under the locale's field order, parsing falls through to `_isotimestamp` so genuine `YYYYMMDD` input still works.

Suggested by a user on the Gramps forum: https://gramps.discourse.group/t/whats-in-development-for-gramps-6-2/9641/6

## Tests

8 unit tests added to `gramps/gen/datehandler/test/dateparser_test.py` (`TestCompact8DateParsing`):

- DMY compact input: `01121720` → day=1, month=12, year=1720
- MDY compact input: `01121720` → month=1, day=12, year=1720
- YMD locale: `17201201` still handled correctly by `_isotimestamp` (unchanged path)
- YYYYMMDD fallthrough in DMY/MDY locales: compact8 rejects invalid month, `_isotimestamp` picks it up
- Invalid day for month (Feb 31) → stored as text, not silently mangled
- Invalid month (99) → stored as text
- Leading/trailing whitespace tolerated

## Test plan

- [ ] Type `01121720` in a date field on a DMY locale (e.g. German) — should display as `01.12.1720`
- [ ] Type `01121720` in a date field on an MDY locale (e.g. US English) — should display as `January 12, 1720`
- [ ] Type `17201201` — should still parse as year=1720, month=12, day=1 via `_isotimestamp` (YYYYMMDD)
- [ ] Type `31021720` (invalid: Feb 31) — should remain as text, not silently accepted
- [ ] Existing date formats with separators (`01.12.1720`, `12/01/1720`) still work unchanged

### DIFF
--- a/gramps/gen/datehandler/_dateparser.py
+++ b/gramps/gen/datehandler/_dateparser.py
@@ -579,6 +579,7 @@ class DateParser:
             r"(\d+)?\s+?%s\.?\s*((\d+)(/\d+)?)?\s*$" % self._smon_str, re.IGNORECASE
         )
         self._numeric = re.compile(r"((\d+)[/\.]\s*)?((\d+)[/\.]\s*)?(\d+)\s*$")
+        self._compact8 = re.compile(r"^\s*(\d{2})(\d{2})(\d{4})\s*$")
         self._iso = re.compile(r"(\d+)(/(\d+))?-(\d+)(-(\d+))?\s*$")
         self._isotimestamp = re.compile(
             r"^\s*?(\d{4})([01]\d)([0123]\d)(?:(?:[012]\d[0-5]\d[0-5]\d)|"
@@ -739,6 +740,22 @@ class DateParser:
             if check is None or check((d, m, y)):
                 return (d, m, y, False)
             return Date.EMPTY
+
+        # Compact 8-digit input without separators: DDMMYYYY (dmy) or
+        # MMDDYYYY (mdy). Must be checked before _isotimestamp, which would
+        # otherwise greedily match DDMMYYYY as a bogus YYYYMMDD. For ymd
+        # locales, _isotimestamp below already handles YYYYMMDD correctly.
+        if not self.ymd:
+            match = self._compact8.match(text)
+            if match:
+                a, b, y = int(match.group(1)), int(match.group(2)), int(match.group(3))
+                if self.dmy:
+                    d, m = a, b
+                else:
+                    m, d = a, b
+                if not check or check((d, m, y)):
+                    return (d, m, y, False)
+                # invalid under locale field order — fall through to _isotimestamp
 
         # Database datetime format, used in ex. MSSQL
         # YYYYMMDD HH:MM:SS or YYYYMMDD or YYYYMMDDHHMMSS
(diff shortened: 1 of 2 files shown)


emyoulation:
This is useful but adds some ambiguity for users and support.

Similar to the "Prefix" and "Suffix" fields in the Add Person dialog have hint text, could a blank Date field in an Edit Event dialog have a locale-specific hint? e.g. "YYYYMMDD" for Japanese, "TTMMJJJJ" for German, "MMDDYYY" for US English. The ISO format in Weblate shows as "ååååMMDD" for Danish, or "RRRMMDD" for Czech, or "ÉÉÉÉHHNN" for Hungarian

**Weblate** search for ["YYYY-MM-DD (ISO)"](https://hosted.weblate.org/search/gramps-project/gramps/?q=%22YYYY-MM-DD+%28ISO%29%22&sort_by=-priority%2Cposition&checksum=)

A Tooltip is an option but is less visible:
image

Suffix and Prefix are some of the few fields with visible hints when the value is a NULL:
image

dsblank:
> This is useful but adds some ambiguity for users and support.

We should definitely mention this in the docs, if merged. But there are so many variations already possible, I don't think it needs to be in the UI. We already don't mention most of the variations.

emyoulation:
There is a PR (waiting to be merged) that is specifically reduce the ambiguity of Date formats in Preferences. (Doubly necessary since the interface does not label the Date formats following standard display description notation.)
* Show date format examples in Preferences Data tab date format menu[#2296](https://github.com/gramps-project/gramps/pull/2296)

So while there is a PR to add feedback to reduce date format ambiguity in one part of the interface, this PR expands ambiguity without offsetting that with more feedback. 

## Reference:
Unnecessary ambiguity was initially created by adopting non-standard labeling of Date display format in Preferences.  More typical labeling would be like:
Perplexity suggests a practical mapping of the Gramps display labels to standard date-pattern notation.
```
    YYYY-MM-DD (ISO) (default) → yyyy-MM-dd
    Numerical → M/d/yyyy or d/M/yyyy, depending on locale
    Month Day, Year → MMMM d, yyyy
    MON DAY, YEAR → MMM d, yyyy if you mean a 3-letter month abbreviation
    Day Month Year → d MMMM yyyy
    DAY MON YEAR → d MMM yyyy
```

dsblank:
> So while there is a PR to add feedback to reduce date format ambiguity in one part of the interface, this PR expands ambiguity without offsetting that with more feedback.

FYI I don't think "ambiguity" is the right word. This PR creates no ambiguity: what worked before still works exactly as it did. What it does do is catch a common shortcut (entering a date without separators) and intelligently just "does the right thing". There is no ambiguity.

There are lots of options for entering dates not shown in the menu selection. BUT we can add this new format here:

https://github.com/gramps-project/gramps/pull/2377

But I don't think even that is necessary.

PR #2410, opened 2026-06-25
Include FS CSS in Windows build

SourceAirbender:
Windows executable only has files copied into share/gramps, and scripts/build_hook.py was only copying gramps.css, not familysearch.css, update for that. It was causing the windows version not to have the CSS styling.

What it should be:
image

image

What it was:
image

image


Nick-Hall (approved):

dsblank:
Good catch; thank you!

dsblank (approved):

PR #2412, opened 2026-06-25
Fix TypeError when RCS subprocesses write to stderr in dbman.py

eduralph:
## Summary
**User impact:** When you manage the archive/backup versions of a family tree, the operation could crash outright the moment the underlying archive tool printed any warning or diagnostic message — so an ordinary archive, restore, or rename step would fail instead of completing.

This PR decodes those diagnostic messages correctly so the operation no longer crashes when the archive tool writes to its error channel.

Reported in Mantis [#13518](https://gramps-project.org/bugs/view.php?id=13518).

## What to look at
The archive (RCS) helpers now read and decode the subprocess's error output as text before assembling a message, instead of trying to join raw bytes into a string. To try it: run an archive operation (create/check-in/check-out/rename/delete a revision) in a setup where the RCS tool emits a stderr diagnostic, and confirm it no longer crashes.

## Root cause
The five RCS (archive) helpers in `gramps/gui/dbman.py` open their subprocess with a plain `subprocess.Popen(cmd, stderr=subprocess.PIPE)` (binary mode), so `proc.stderr.readlines()` yields `bytes` objects. The code then attempts to join these bytes directly into a string with `"\n".join(...)`, which raises `TypeError: sequence item 0: expected str instance, bytes found` whenever the subprocess writes anything to stderr, crashing the operation.

## Fix
- Adds `gramps/gui/dbman_utils.py` with a single helper function `read_subprocess_messages(proc)` that reads the binary stderr pipe and decodes each line to `str` before joining. Uses `errors="replace"` to gracefully handle non-UTF-8 locale-encoded diagnostics.
- Updates `gramps/gui/dbman.py` to import this helper and replaces all five call sites (lines 637, 809, 1179, 1224, 1255) with calls to `read_subprocess_messages(proc)`.
- Registers both the new module and the regression test in `po/POTFILES.skip` (no translatable strings).

The extraction into a separate import-light module enables the message-assembly logic to be tested headless (without importing the GTK-dependent `dbman.py`), allowing the regression test to drive the production path directly with a stubbed subprocess.

## Verification
- **Claim:** subprocess stderr output is decoded to `str` and assembled into a message without raising `TypeError`, even for non-UTF-8 bytes.
- **Checked:** on `maintenance/gramps61`, all five call sites now route through `read_subprocess_messages(proc)`:
  - `gramps/gui/dbman.py:637` (`__rename_revision`) — was `message = "\n".join(proc.stderr.readlines())` joining bytes directly.
  - `gramps/gui/dbman.py:809` (`__delete_archive`) — same pattern.
  - `gramps/gui/dbman.py:1179` (`check_out`) — same pattern.
  - `gramps/gui/dbman.py:1224` (`_check_in`, archive creation) — same pattern.
  - `gramps/gui/dbman.py:1255` (`_check_in`, archive data) — same pattern.
- **Test:** `gramps/gui/test/dbman_test.py` (new) — a headless unittest that stubs a subprocess and drives `read_subprocess_messages()` through four cases: (1) the naive join raises `TypeError` (documents the bug), (2) the helper decodes bytes and returns a usable `str`, (3) empty stderr returns an empty string, and (4) non-UTF-8 bytes decode gracefully without raising `UnicodeDecodeError`. All four pass with the fix; the suite errors when the helper module is removed (red→green verification).

Fixes [#13518](https://gramps-project.org/bugs/view.php?id=13518)

### DIFF
--- a/gramps/gui/dbman.py
+++ b/gramps/gui/dbman.py
@@ -632,7 +632,9 @@ class DbManager(CLIDbManager, ManagedWindow):
 
         cmd = ["rcs", "-x,v", "-m%s:%s" % (rev, new_text), archive]
 
-        proc = subprocess.Popen(cmd, stderr=subprocess.PIPE)
+        proc = subprocess.Popen(
+            cmd, stderr=subprocess.PIPE, encoding="utf-8", errors="replace"
+        )
         status = proc.wait()
         message = "\n".join(proc.stderr.readlines())
         proc.stderr.close()
@@ -804,7 +806,9 @@ class DbManager(CLIDbManager, ManagedWindow):
 
         cmd = ["rcs", "-x,v", "-o%s" % rev, "-q", archive]
 
-        proc = subprocess.Popen(cmd, stderr=subprocess.PIPE)
+        proc = subprocess.Popen(
+            cmd, stderr=subprocess.PIPE, encoding="utf-8", errors="replace"
+        )
         status = proc.wait()
         message = "\n".join(proc.stderr.readlines())
         proc.stderr.close()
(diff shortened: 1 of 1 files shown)


QuLogic:
This way seems a bit overengineered; why not pass an explicit `encoding` and `errors` to the `Popen` call?

eduralph:
> This way seems a bit overengineered; why not pass an explicit `encoding` and `errors` to the `Popen` call?

@QuLogic - great idea, implemented

PR #2418, opened 2026-06-27
graph: fix Family Lines ancestor pruning by surname

eduralph:
## Summary
**User impact:** In the Family Lines Graph report, with "remove extra people and families" turned on, a genuine direct ancestor could be dropped from the chart. It happened when a family's surname spelling changed from one generation to the next (for example Smith → Smithe → Smyth): the older ancestor no longer matched the newer spelling, so the report treated a real bloodline ancestor as an outsider and left them out.

This makes the report keep anyone on the direct line of descent, so a surname that drifts in spelling no longer causes a real ancestor to disappear.

Reported in Mantis [#10415](https://gramps-project.org/bugs/view.php?id=10415).

## What to look at
The report now decides who belongs on the direct line by following parent-to-child links (actual lineage), not by comparing surname text. To try it: build a Family Lines Graph over a family whose surname spelling changes each generation, enable "remove extra people and families", and confirm the top-of-line ancestor and their married-in spouse are still shown. A new headless test drives exactly this on a four-generation line with a drifting surname.

## Root cause
The `remove_uninteresting_parents` method in `gramps/plugins/graph/gvfamilylines.py:664–822` keeps ancestors via a fixed set of heuristics to decide which people to include in the Family Lines Graph report when "remove extra people and families" is enabled. One heuristic (lines 772–791) checks if an ancestor's surname string matches any person of interest, with no lineage-based criterion independent of surname text.

A top-of-tree direct ancestor with a single child of interest, no further kept parents, and no person-of-interest status is kept *only* by the surname-equality test (lines 804–806); when that surname spelling drifts between generations, the legitimate bloodline ancestor is removed as "extra" despite being included by "follow parents".

## Fix
Extract the people/family-selection logic into a new GUI-free class `FamilyLinesSelection` (a testable seam: depends only on the database and option flags, not on the Report base or GUI machinery). The selection methods move verbatim from `FamilyLinesReport` to the new class with one change:

1. Add a new `_direct_ancestors()` method (line 260–293 in the patched file) that computes the transitive closure of parent links from the interest set, deciding membership purely by lineage (parent links), cached to avoid recomputation.

2. In `remove_uninteresting_parents`, check if a person's handle is in the direct-ancestor set *before* the surname heuristic (in the patched file, ahead of the surname-equality test). A direct-line ancestor is never "extra"; surname spelling drift no longer prunes the legitimate direct line.

3. `FamilyLinesReport.begin_report` now constructs a `FamilyLinesSelection`, calls its `select()` method, and copies the four results back onto `self` — production routes through the extracted unit; the report and the test drive the identical implementation.

## Verification
- **Claim:** a direct-line ancestor is never pruned as "extra" even when the surname spelling drifts between generations.
- **Checked:** on `maintenance/gramps61` — `gramps/plugins/graph/gvfamilylines.py:590–662` (`find_parents()` builds the initial direct-line set by recursively following parent-family links); `:664–822` (`remove_uninteresting_parents()` applies keep-heuristics; the new lineage-based criterion is checked before surname text matching (at the original lines 772–791); `:823–870` (`find_children()` extends the set downward, computed only from `find_parents` output); `po/POTFILES.skip` registers the new test file (no translatable strings, same style as `year_only_date_test.py`).
- **Test:** `gramps/plugins/graph/test/gvfamilylines_test.py` (new) drives the production `FamilyLinesSelection.find_parents()` + `remove_uninteresting_parents()` on a real in-memory database (`import_as_dict` from an inline fixture: a four-generation direct line where the paternal surname drifts each generation Smith → Smithe → Smyth, with a married-in spouse each generation). It asserts no direct-line ancestor is pruned, including I0005 "Smyth" (the drifted top-of-line great-grandfather) and I0006 "White" (the married-in great-grandmother). Import-light (`gvfamilylines` imports only `gramps.gen.*`, no `gramps.gui`), so it runs headless. Fails pre-fix, passes post-fix.

Fixes [#10415](https://gramps-project.org/bugs/view.php?id=10415)

### DIFF
--- a/gramps/plugins/graph/gvfamilylines.py
+++ b/gramps/plugins/graph/gvfamilylines.py
@@ -409,183 +409,74 @@ class FamilyLinesOptions(MenuReportOptions):
             self.justyears.set_available(False)
 
 
-# ------------------------------------------------------------------------
-#
-# FamilyLinesReport -- created once the user presses 'OK'
-#
-# ------------------------------------------------------------------------
-class FamilyLinesReport(Report):
-    """FamilyLines report"""
+class FamilyLinesSelection:
+    """
+    Determine which people and families a Family Lines report includes.
 
-    def __init__(self, database, options, user):
-        """
-        Create FamilyLinesReport object that eventually produces the report.
+    This is the people/family-selection half of :class:`FamilyLinesReport`,
+    factored out so it depends only on the database and the report's option
+    flags -- not on the GUI-entangled report machinery (the document writer,
+    the menu, the ``Report`` base class).  That makes the selection logic
+    (follow parents/children and the "remove extra people" pruning) directly
+    unit-testable without constructing the full report.
+    """
 
-        The arguments are:
-
-        database     - the Gramps database instance
-        options      - instance of the FamilyLinesOptions class for this report
-        user         - a gen.user.User() instance
-        name_format  - Preferred format to display names
-        incl_private - Whether to include private data
-        inc_id       - Whether to include IDs.
-        living_people - How to handle living people
-        years_past_death - Consider as living this many years after death
-        """
-        Report.__init__(self, database, options, user)
-
-        menu = options.menu
-        get_option_by_name = menu.get_option_by_name
-        get_value = lambda name: get_option_by_name(name).get_value()
-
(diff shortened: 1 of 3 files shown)


lordemannd:
works for me, tested in 6.0.8 on Windows

PR #2421, opened 2026-06-27
Make surname gramplets count unique surnames consistently

eduralph:
## Summary
**User impact:** Three different places that count surnames — Top Surnames, Surname Cloud, and Statistics — give three different totals for the very same family tree (the reporter saw 244 in one and 449 in another). A user has no way to know which number is right.

This makes all three use one shared counting rule, so they agree.

Reported in Mantis [#6793](https://gramps-project.org/bugs/view.php?id=6793).

## What to look at
The three tools each grew their own way of deciding what counts as a distinct surname, so they disagree. The change extracts the rule Top Surnames already used into one shared helper and points the other two at it. To try it: open the same tree's Top Surnames, Surname Cloud, and Statistics gramplets and confirm the surname totals now match. A new headless test drives all three paths against one fixture tree.

## Root cause
The Top Surnames, Surname Cloud, and Statistics gramplets each developed their own surname-counting approach independently: Top Surnames counted distinct group names (`topsurnamesgramplet.py:130-137`), Surname Cloud counted distinct non-empty `get_surname()` strings (`surnamecloudgramplet.py:101-117`), and Statistics used `len(set(database.surname_list))` which stores only primary-name surnames (`statsgramplet.py:189-191`). Because these three implementations count different things, the same tree produced three different totals (reporter observed 244 vs 449).

## Fix
Align the three divergent approaches by extracting Top Surnames' existing rule into a shared library and converging Surname Cloud and Statistics to use it. This gives the smallest behavioural change while eliminating the inconsistency:

- **New `gramps/plugins/lib/libsurnames.py`** — holds two functions:
  - `record_surnames(person, surnames, representative_handle)` — counts one person's surnames, moved verbatim from `topsurnamesgramplet.py:54-78`, already implementing the canonical rule.
  - `count_unique_surnames(db)` — wrapper that tallies all people and returns the distinct count; used by Statistics.
- **`gramps/plugins/gramplet/topsurnamesgramplet.py`** — imports `record_surnames` from libsurnames (`line 40`); count unchanged as it already used this rule.
- **`gramps/plugins/gramplet/surnamecloudgramplet.py`** — replaces inline enumeration (`lines 101-117`) with `record_surnames()` call and changes reported total from `len(namelist)` to `len(surnames)` (`line 183`), now aligning the count to the group names the cloud displays.
- **`gramps/plugins/gramplet/statsgramplet.py`** — imports `count_unique_surnames` and calls it instead of `len(set(database.surname_list))` (`lines 35, 191`), now counting alternate-name surnames.
- **`po/POTFILES.skip`** — registers the two new `.py` files per T2-potfiles gate.
- **New `gramps/plugins/gramplet/test/surnamecount_test.py`** — headless unittest verifying all three gramplets' paths agree on the same fixture tree.

**Why align on Top Surnames' approach:** It was already working (Top Surnames had no reported issues), matches the Surname Cloud's own keying (the cloud displays by group name, so the count should match its picture), and counts alternate-name surnames — which the Statistics index silently dropped. This choice gives the smallest behavioural change: Top Surnames unchanged, Surname Cloud aligned to its own visualization, only Statistics' count meaningfully increases (to include alternate names the old index excluded). Alternatives (standardizing on Statistics' rule) would require rewriting ~130 lines across both cloud gramplets versus the 2-line convergence here.

## Verification
- **Claim:** all three gramplets resolve to one and only one counting rule, so the same tree yields identical surname totals across them.
- **Checked:** on `maintenance/gramps61` — `gramps/plugins/lib/libsurnames.py` (new) exports `record_surnames` and `count_unique_surnames`; `gramps/plugins/gramplet/topsurnamesgramplet.py:40` imports the shared function; `gramps/plugins/gramplet/surnamecloudgramplet.py:30,45,183` calls `record_surnames` and reports the shared count; `gramps/plugins/gramplet/statsgramplet.py:35,191` imports and calls the wrapper; `po/POTFILES.skip:611,638` registers the new files (no translatable strings).
- **Test:** `gramps/plugins/gramplet/test/surnamecount_test.py` (new) drives the production counting routines on a fixture tree (6 people exercising shared surnames, alternate names, no-surname cases) and verifies: `count_unique_surnames` returns 8 distinct group names; the dict-building path (`record_surnames`) and the wrapper path yield the same total; all three gramplets import the same shared functions (via `assertIs`, `:232-264`); and alternate-name surnames are counted (the old Statistics index dropped these). Fails RED without the libsurnames module (gramplets cannot import the shared rule), passes GREEN with the patch (`python3 -m unittest gramps.plugins.gramplet.test.surnamecount_test`); existing `topsurnamesgramplet_test.py` still passes after the `record_surnames` relocation.

Fixes [#6793](https://gramps-project.org/bugs/view.php?id=6793)

### DIFF
--- a/gramps/plugins/gramplet/statsgramplet.py
+++ b/gramps/plugins/gramplet/statsgramplet.py
@@ -21,6 +21,7 @@
 #
 # ------------------------------------------------------------------------
 import os
+from collections import defaultdict
 
 # ------------------------------------------------------------------------
 #
@@ -32,6 +33,7 @@ from gramps.gen.utils.file import media_path_full
 from gramps.gen.datehandler import get_date
 from gramps.gen.lib import Person
 from gramps.gen.const import COLON, GRAMPS_LOCALE as glocale
+from gramps.plugins.lib.libsurnames import record_surnames
 
 _ = glocale.translation.sgettext
 
@@ -84,6 +86,11 @@ class StatsGramplet(Gramplet):
         unknowns = 0
         bytes_cnt = 0
         notfound = []
+        # Unique-surname tally, accumulated in the single people pass below so the
+        # figure shares the Top Surnames / Surname Cloud rule (libsurnames) without
+        # a second, non-yielding scan of the database (bug #6793).
+        surnames = defaultdict(int)
+        representative_handle = {}
 
         mobjects = database.get_number_of_media()
         mbytes = "0"
@@ -100,6 +107,7 @@ class StatsGramplet(Gramplet):
                 notfound.append(media.get_path())
 
         for cnt, person in enumerate(personList):
+            record_surnames(person, surnames, representative_handle)
             length = len(person.get_media_list())
             if length > 0:
                 with_media += 1
@@ -188,7 +196,7 @@ class StatsGramplet(Gramplet):
         self.append_text("\n")
         if hasattr(database, "surname_list"):
             self.link(_("%s:") % _("Unique surnames"), "Filter", "unique surnames")
-            self.append_text(" %s" % len(set(database.surname_list)))
+            self.append_text(" %s" % len(surnames))
             self.append_text("\n")
         self.append_text("\n%s\n" % _("Media Objects"))
         self.append_text("----------------------------\n")
--- a/gramps/plugins/gramplet/surnamecloudgramplet.py
+++ b/gramps/plugins/gramplet/surnamecloudgramplet.py
@@ -30,6 +30,7 @@ from collections import defaultdict
 from gramps.gen.plug import Gramplet
 from gramps.gen.config import config
 from gramps.gen.const import GRAMPS_LOCALE as glocale
+from gramps.plugins.lib.libsurnames import record_surnames
 
 _ = glocale.translation.sgettext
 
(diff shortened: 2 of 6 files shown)


eduralph:
Updated (force-pushed `9a11e46`) to address two issues found in review:

**1. The Statistics gramplet no longer scans the database twice.** The previous revision routed Statistics through a `count_unique_surnames(db)` helper that ran its *own* full `iter_people()` loop — a second pass on top of the one `StatsGramplet.main()` already makes, and (unlike that loop) it didn't `yield`, so on a large tree it both re-walked the database and blocked the UI. The unique-surname tally is now accumulated with the shared `record_surnames()` **inside the existing people pass**, and the figure is `len(surnames)` — the same single, yielding scan the gramplet already performs. Top Surnames and Surname Cloud already computed their count this way from their own single pass; Statistics now matches them. The now-unused `count_unique_surnames()` helper is removed, leaving `libsurnames` with the one canonical rule (`record_surnames`).

**2. The empty/no-surname is no longer counted as a unique surname.** `record_surnames()` now skips the empty group name, so a person with no surname contributes nothing to the total. Previously the empty string was tallied, inflating the count by one whenever a no-surname person existed — and it was inconsistent with Statistics' own treatment of an empty surname as an *incomplete name*. Because all three gramplets share the rule, the exclusion is uniform; Surname Cloud (which already excluded empties in its old count) keeps that behaviour.

**Tests.** `surnamecount_test.py` updated: the fixture's distinct count is now 7 (the no-surname person excluded), with a dedicated test that the empty surname isn't counted, and the "share one rule" check covers all three gramplets routing through `record_surnames`. Verified locally — the new test and the existing `topsurnamesgramplet_test.py` pass (9 tests green).

PR #2426, opened 2026-06-28
Update Dutch translation for 'Transcript'

coduinix:
The previous wordt that was used, "Vertaling" literally means "Translation". 
The word "Transcriptie" is a translation that's closer to the English word "Transcript" and also the common word that is used in the genealogy context.

### DIFF
--- a/po/nl.po
+++ b/po/nl.po
@@ -8935,7 +8935,7 @@ msgstr "Analyse"
 
 #: ../gramps/gen/lib/notetype.py:88
 msgid "Transcript"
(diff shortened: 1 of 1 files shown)


Nick-Hall:
We now use [Weblate](https://hosted.weblate.org/projects/gramps-project/gramps/) for our translations.  Please update the string ["Vertaling"](https://hosted.weblate.org/translate/gramps-project/gramps/nl/?checksum=beaa7ac116e44a17) in Weblate rather than submitting a pull request here.

PR #2441, opened 2026-07-01
Distinguish adopted relationships from step relationships

eduralph:
## Summary
**User impact:** An adopted child is labelled a "step" child in relationships and reports — most visibly the Narrative Web site — even though the family record clearly marks the child as adopted. Calling an adopted child a stepchild is both genealogically and legally wrong, and there was no way to get the correct wording.

This teaches the relationship calculator to treat "adopted" as its own kind of link, so adopted children (and their siblings and parents) read as "adopted" rather than "step", while genuine step relationships are unchanged.

Reported in Mantis [#10554](https://gramps-project.org/bugs/view.php?id=10554).

## What to look at
The relationship calculator learns a third class of link — birth / adopted / step — and threads the recorded adoption type through, so the wording matches the data. To try it: build a family with a birth child, an adopted child, and a stepchild, then view relationships (or generate a Narrative Web report) — the adopted child now reads "adopted …" instead of "step …", and the plain stepchild still reads "step …".

## Root cause
The relationship calculator (`gramps/gen/relationship.py:908-935`) defined relationships by a binary birth/non-birth distinction, and `__apply_filter` and `get_sibling_type` rendered every non-birth link with the `"step"` qualifier. Even though the database records an adopted child-reference explicitly with `ChildRefType.ADOPTED`, the calculator collapsed ADOPTED, STEPCHILD, FOSTER, and other non-birth types to one bucket, so reports (particularly the Narrative Web) call an adopted child a stepchild — genealogically and legally incorrect.

## Fix
The change threads the `ChildRefType` through the relationship paths as a third class of link (birth / **adopted** / step), keeping "step" behavior unchanged for non-adopted, non-birth links. Key modifications to `gramps/gen/relationship.py`:

1. **New path codes and sibling type** (`relationship.py:908-935`): Added `REL_MOTHER_ADOPT="n"`, `REL_FATHER_ADOPT="d"`, `REL_FAM_ADOPT="D"` to denote adopted parent links; introduced `ADOPT_SIB=5` for adopted siblings; added `ADOPT="adopted "` as the localizable qualifier.

2. **Path building in `__apply_filter`** (`relationship.py:1524-1650`): The method now emits the adopt code when a `ChildRefType` is `ADOPTED`, else the existing NOTBIRTH code for other non-birth types.

3. **Adopted parent detection** (`relationship.py:1204+` area): New helper method `_get_adopted_parent_list` mirrors `_get_nonbirth_parent_list` and returns parents linked by adoption only.

4. **Sibling qualifier logic** (`relationship.py:1124-1170`): `get_sibling_type` now calls a new `_step_or_adopt_sib` method to distinguish whether a non-birth sibling connection is via an adopted or plain-step parent; true stepsiblings (different birth parents known) still return `STEP_SIB`.

5. **Single-relationship string qualifier** (`relationship.py:2235+` area, `get_single_relationship_string` method): The qualifier is now `""` for birth, `ADOPT` when `only_adopt(path)` returns true (i.e., the path's non-birth hops are all adopted), else `STEP`.

6. **Dual-parent adoption** (`relationship.py:1843-1915`): `_famrel_from_persrel` now maps adopted-to-adopted links to `REL_FAM_ADOPT`; `get_one_relationship` (`relationship.py:1990-2061`) includes the new codes in the order list so `.index()` cannot raise.

7. **New `only_adopt()` method** (`relationship.py:1880+`): Returns true iff a path's non-birth hops are all adopted (no plain-step hop), used to select the adopted qualifier.

8. **Test registration** (`po/POTFILES.skip:301`): Registered the new test file (no translatable strings) per doc 16.

Locale safety: The 22 locale overrides derive `step` from the `only_birth` *flag* (verified in `rel_nl.py:769` as representative), not by parsing the path. An un-updated locale does not crash on the new path codes `n`/`d`/`D` and keeps current "step" rendering for adopted links until updated — exactly the "default sensibly, untranslated but not wrong/crashing" the design required.

## Verification
- **Claim:** the relationship calculator distinguishes adopted from step links, so an adopted child/parent/sibling reads as "adopted" while genuine step links still read "step", with no new codes crashing un-updated locales.
- **Checked:** on master (target branch) —
  - `gramps/gen/relationship.py:908-935` — relationship code constants (REL_* and ADOPT_SIB)
  - `gramps/gen/relationship.py:1124-1170` — `get_sibling_type` and new `_step_or_adopt_sib` logic
  - `gramps/gen/relationship.py:1204-1250` — `_get_nonbirth_parent_list`; `_get_adopted_parent_list` inserted below it
  - `gramps/gen/relationship.py:1524-1650` — `__apply_filter` path-building (added ADOPT check)
  - `gramps/gen/relationship.py:1843-1915` — `_famrel_from_persrel` and family-relation logic
  - `gramps/gen/relationship.py:1880-1910` — `only_birth` method; new `only_adopt` method appended
  - `gramps/gen/relationship.py:1990-2061` — `get_one_relationship` order list updated
  - `gramps/gen/relationship.py:2235-2350` — `get_single_relationship_string` qualifier logic
  - `po/POTFILES.skip:299-303` — test registration
- **Test:** `gramps/gen/test/relationship_test.py` (new) — builds an in-memory database with a father/mother couple and three sons (one birth, one adopted, one stepchild) and asserts: an adopted child is NOT labelled "step" and reads "adopted …"; a genuine stepchild is still "step"; the reciprocal adopted-parent relation is "adopted"; adopted siblings are NOT "step". It exercises the production path `get_one_relationship` (what the Narrative Web and every relationship consumer use), runs under the headless test runner (no GUI imports; `make_database("sqlite")` pattern from `gramps/gen/utils/test/alive_test.py`), and is RED without the production change (4 adopted assertions fail; step + birth guards pass) and GREEN with it (7/7 pass).

Fixes [#10554](https://gramps-project.org/bugs/view.php?id=10554)

### DIFF
--- a/gramps/gen/relationship.py
+++ b/gramps/gen/relationship.py
@@ -908,9 +908,12 @@ class RelationshipCalculator:
     REL_FATHER = "f"  # going up to father
     REL_MOTHER_NOTBIRTH = "M"  # going up to mother, not birth relation
     REL_FATHER_NOTBIRTH = "F"  # going up to father, not birth relation
+    REL_MOTHER_ADOPT = "n"  # going up to mother, adopted relation
+    REL_FATHER_ADOPT = "d"  # going up to father, adopted relation
     REL_SIBLING = "s"  # going sideways to sibling (no parents)
     REL_FAM_BIRTH = "a"  # going up to family (mother and father)
     REL_FAM_NONBIRTH = "A"  # going up to family, not birth relation
+    REL_FAM_ADOPT = "D"  # going up to family, adopted relation
     REL_FAM_BIRTH_MOTH_ONLY = "b"  # going up to fam, only birth rel to mother
     REL_FAM_BIRTH_FATH_ONLY = "c"  # going up to fam, only birth rel to father
 
@@ -922,10 +925,15 @@ class RelationshipCalculator:
     HALF_SIB_FATHER = 2  # same father, mother known to be different
     STEP_SIB = 3  # birth parents known to be different
     UNKNOWN_SIB = 4  # insufficient data to draw conclusion
+    ADOPT_SIB = 5  # connected through an adoptive (not birth) parent
 
     # sibling strings
     STEP = "step"
     HALF = "half-"
+    # Adopted links are not "step": an adoptee is treated as born into the
+    # family (issue 10554). Kept as a single localizable qualifier so the
+    # exact wording stays a cheap, per-locale choice.
+    ADOPT = "adopted "
 
     INLAW = "-in-law"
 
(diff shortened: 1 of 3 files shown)


petr-fedorov:
>  Calling an adopted child a stepchild is both genealogically and legally wrong, and there was no way to get the correct wording.

IMHO, calling an adopted child "adopted" is at least legally wrong too. The idea to add REL_MOTHER_ADOPT etc is good, but [`step = self.ADOPT`](https://github.com/gramps-project/gramps/pull/2441/changes#diff-c9c6f05ba0d3cd4941c85f4d429818f19cd245e2abc34a98c61a85be3c284fb0R2428) seems to be wrong. Instead it should probably be `step = ""`

eduralph:
> > Calling an adopted child a stepchild is both genealogically and legally wrong, and there was no way to get the correct wording.
> 
> IMHO, calling an adopted child "adopted" is at least legally wrong too. The idea to add REL_MOTHER_ADOPT etc is good, but [`step = self.ADOPT`](https://github.com/gramps-project/gramps/pull/2441/changes#diff-c9c6f05ba0d3cd4941c85f4d429818f19cd245e2abc34a98c61a85be3c284fb0R2428) seems to be wrong. Instead it should probably be `step = ""`

From a legal perspective you are right - an adopted child is just "child" but it makes it hard to distinguish that relationship. I think for genealogical purposes calling the child "adopted" is fine. In this case I followed the bug reports convention.

PR #2458, opened 2026-07-04
Guard tar/zip extraction against path-traversal (CVE-2007-4559)

dsblank:
## Summary
- The `.gpkg` importer (`gramps/plugins/importer/importgpkg.py`) extracted archive members with `tarfile.TarFile.extract()` without validating their paths, vulnerable to [CVE-2007-4559](https://nvd.nist.gov/vuln/detail/CVE-2007-4559) / "Zip Slip" (see also [python/cpython#73974](https://github.com/python/cpython/issues/73974)).
- The addon/plugin installer (`load_addon_file()` in `gramps/gen/plug/utils.py`) has the same class of bug in **two** places: the custom `Zipfile.extractall()` wrapper (used for `.zip` addon packages) and a bare `tarfile.TarFile.extractall()` call (used for `.tar.gz`/`.tgz` addon packages). This is arguably higher risk than the `.gpkg` case since addon packages are fetched from a URL (the Gramps addon server or a configured mirror) rather than a file the user deliberately imports.
- In each case a crafted archive could use an absolute path, a `../` path-traversal member name, or a symlink/hardlink to create or overwrite files outside the intended target directory (media directory for `.gpkg`, plugin directory for addons).
- Added a shared `gramps/gen/utils/safearchive.py` module (`is_safe_archive_member()`, `is_safe_tar_member()`) and applied it at all three extraction sites before any member is written to disk. This is a self-contained workaround that does not rely on Python 3.12+'s `filter=` extraction parameter (PEP 706), so it works across all Python versions Gramps supports (3.10+).

Fixes #12911.

## Test plan
- [x] `gramps/gen/utils/test/safearchive_test.py` — unit tests for the shared helpers (absolute paths, `../` traversal, symlink/hardlink escapes).
- [x] `gramps/plugins/importer/test/importgpkg_test.py` — integration-style test against a crafted malicious tar archive.
- [x] Manual checks: crafted malicious zip/tar member is rejected via `Zipfile.extractall()` (raises `ValueError`) and via the addon-installer's tar-member pre-check; a benign zip with a real directory entry still extracts correctly (no regression).
- [x] `python3 -m unittest gramps.gen.utils.test.safearchive_test gramps.plugins.importer.test.importgpkg_test -v` — all pass.
- [x] `black --check` on all changed files.
- [x] `mypy` (full project) — no issues.

### DIFF
--- a/gramps/gen/plug/utils.py
+++ b/gramps/gen/plug/utils.py
@@ -43,6 +43,7 @@ from ..config import config
 from ..const import GRAMPS_LOCALE as glocale
 from ..const import USER_PLUGINS
 from ..constfunc import mac
+from ..utils.safearchive import is_safe_archive_member, is_safe_tar_member
 from . import BasePluginManager
 from ._pluginreg import make_environment
 
@@ -119,8 +120,14 @@ class Zipfile:
     def extractall(self, path, members=None):
         """
         Extract all of the files in the zip into path.
+
+        Raises ``ValueError`` if an entry's path would resolve outside of
+        ``path`` (Zip Slip / CVE-2007-4559).
         """
         names = self.zip_obj.namelist()
+        for name in names:
+            if not is_safe_archive_member(name, path):
+                raise ValueError("Unsafe path in archive member: %s" % name)
         for name in self.get_paths(names):
             fullname = os.path.join(path, name)
             if not os.path.exists(fullname):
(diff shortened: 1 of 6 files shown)


emyoulation:
@DavidMStraub 
Is this kind of protection needed for the GEDCOM7 zip importer?  Does it use relative paths for media that might susceptible to this kind of attack?

DavidMStraub:
.gpkg is disallowed in Gramps Web API precisely because it's not safe.

There was a path traversal vulnerability in the media importer though, see https://github.com/advisories/GHSA-m5gr-86j6-99jp and https://github.com/gramps-project/gramps-web-api/commit/3ed4342711e3ec849552df09b1fe2fbf2ca5c29a.

dsblank:
Thank you @emyoulation for mentioning zip importer (in general). I have updated the PR to fix all known security issues in gramps zip/tar importing.

PR #2460, opened 2026-07-07
Add check for stale files in POTFILES.in

hgohel:
When` intltool-update -m` finds file entries in POTFILES.in which no longer exist, it logs those to `notexist`. If that file is found, CI will now treat that as an error in addition to the new file test.

It's interesting that the man page for intltool-update does not mention this scenario, but I found that to be the behavior and Claude confirmed it.

Please squash-and-merge.


hgohel:
@Nick-Hall CI check failed installing dependencies...a transient error?

PR #2465, opened 2026-07-12
Assert on messages logged during test execution

hgohel:
Some unit tests result in log messages (warnings) being emitted which can be seen in the output (see below). These messages are actually expected and do not represent failures.

However, messages being logged in the output of a test run means someone has to review them and determine whether they are real warnings, failures or errors.

In the modified tests these warnings were expected, so I used the Python capability to assert on log messages to capture and validate them, thus strengthening the test case and cleaning the output at the same time. This pattern was already in use in some places.

To run modified tests:

`python3 -m unittest gramps.gen.utils.test.pypi_test.TestResolvePypiName`
`python3 -m unittest  gramps.gen.utils.test.pypi_test.TestCompatibleTags.test_macos_empty_version_does_not_crash`
`python3 -m unittest gramps.gen.fs.fs_import.test.importer_test.TestAddChild`

Expand the `Run test suite` section in the [CI run from this PR which no longer shows warnings  in the output](https://github.com/gramps-project/gramps/actions/runs/29203342419/job/86678315740?pr=2465). For comparison, see the prior output below:

<details><summary>UT Run showing warnings interspersed</summary>
<p>

```
Run export GDK_BACKEND=-
  export GDK_BACKEND=-
  export GRAMPS_RESOURCES=build/share
  python3 -m unittest discover -p "*_test.py"
  shell: /usr/bin/bash -e {0}
/usr/lib/python3/dist-packages/requests/__init__.py:87: RequestsDependencyWarning: urllib3 (2.7.0) or chardet (4.0.0) doesn't match a supported version!
  warnings.warn("urllib3 ({}) or chardet ({}) doesn't match a supported "

(python3 -m unittest:4625): Gtk-CRITICAL **: 19:54:21.331: gtk_icon_theme_get_for_screen: assertion 'GDK_IS_SCREEN (screen)' failed
..........................................................................................................................................................................................................................................................Can't find filter _no_such_filter_ in the defined custom filters
.Can't find filter _no_such_filter_ in the defined custom filters
.Can't find filter _no_such_filter_ in the defined custom filters
.Can't find filter _no_such_filter_ in the defined custom filters
.Can't find filter _no_such_filter_ in the defined custom filters
.Can't find filter _no_such_filter_ in the defined custom filters
.Can't find filter _no_such_filter_ in the defined custom filters
.Can't find filter _no_such_filter_ in the defined custom filters
............................................................................................................Skipping child relationship with unresolved FS parent(s): ['P1', 'P2']
.....................................................................................................
(...many log lines removed...)
.......................ssssssssssssssssssssssssssssssssssssssssssssssssssssssss........................................................................................................................Could not determine macOS version from platform.mac_ver(); compiled wheels will not be matched.
.............................................................................................................................................requires_mod value 'bs4' is a Python import name, not a PyPI package name.  Please update the addon to use 'beautifulsoup4' instead.  The correct name has been applied automatically this time.
.requires_mod value 'cv2' is a Python import name, not a PyPI package name.  Please update the addon to use 'opencv-python' instead.  The correct name has been applied automatically this time.
...requires_mod value 'serial' is a Python import name, not a PyPI package name.  Please update the addon to use 'pyserial' instead.  The correct name has been applied automatically this time.
.requires_mod value 'sklearn' is a Python import name, not a PyPI package name.  Please update the addon to use 'scikit-learn' instead.  The correct name has been applied automatically this time.
..requires_mod value 'yaml' is a Python import name, not a PyPI package name.  Please update the addon to use 'PyYAML' instead.  The correct name has been applied automatically this time.
........................................................ss..............................................................................................................................................................................................................................................................................................................................................................................
Ran 32951 tests in 212.897s

OK (skipped=58)
```
</p>
</details>

### DIFF
--- a/gramps/gen/utils/test/pypi_test.py
+++ b/gramps/gen/utils/test/pypi_test.py
@@ -157,6 +157,7 @@ class TestIsPureWheel(unittest.TestCase):
 #
 # -------------------------------------------------------------------------
 class TestResolvePypiName(unittest.TestCase):
+
     def test_pil_maps_to_pillow(self):
         """PIL (current addon-source import name) maps to Pillow and warns."""
         with self.assertLogs("gramps.gen.utils.pypi", level="WARNING") as cm:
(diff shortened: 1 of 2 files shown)


dsblank:
Do you really need to review the warnings?

hgohel:
> Do you really need to review the warnings?

@dsblank When the UTs emit these messages it's not always evident that it's an ignorable warning,  and not something more serious. Addressing the issue results in clean output which is desirable, no?

gramps/gen/utils/test/pypi_test.py
```diff
@@ -157,6 +157,7 @@ def test_empty_filename_is_not_pure(self):
 #
 # -------------------------------------------------------------------------
 class TestResolvePypiName(unittest.TestCase):
+
     def test_pil_maps_to_pillow(self):
         """PIL (current addon-source import name) maps to Pillow and warns."""
         with self.assertLogs("gramps.gen.utils.pypi", level="WARNING") as cm:
```
hgohel, line 163: BTW, similar code was already in the same file, just not applied everywhere.

Nick-Hall:
> Addressing the issue results in clean output which is desirable, no?

Yes.  We should avoid extraneous output in unit tests.  Any noise makes identifying real problems more difficult.

PR #2468, opened 2026-07-14
Add UI language preference with restart-and-restore session state

dsblank:
## Summary

Gramps currently determines its UI language purely from `LANG`/`LANGUAGE`
environment variables set before launch — there's no in-app way to change it.
This adds a **Language** preference to the Preferences dialog, alongside the
4 existing settings already marked "Requires Restart" (date format, age
display precision, age-after-death display, sidebar text), which previously
only showed a passive warning telling the user to restart manually.

Changing any of these 5 settings now offers to restart Gramps automatically:

- Any open Person/Family/etc. editors with unsaved changes go through the
  existing per-editor save/discard prompt first.
- On confirmation, Gramps snapshots the current session (open tree, active
  view, selected record, open editors) to a small JSON file and relaunches
  itself via `os.execv` with a new `--restore-state <path>` flag.
- The relaunched process restores the tree, view, selection, and open
  editors instead of landing on the welcome screen.

## Design notes

- A live, in-process reload (no restart at all) was considered and rejected
  — it would require making `_()`/gettext lazy everywhere translated
  strings get cached at import time, plus a correctly-ordered teardown/
  rebuild of the entire GTK widget tree including third-party gramplets.
  `os.execv` trades that complexity for a guaranteed clean process restart.
- The UI language override only sets the `LANGUAGE` environment variable
  (gettext's own message-catalog selector, which accepts bare codes like
  `fr`), not `LANG` — `LANG` needs a full OS locale name for `setlocale()`
  to succeed, and setting it to a bare code risked falling back to `"C"`
  and degrading date/collation formatting along with the message language.
- Window geometry and gramplet/sidebar layout are intentionally *not*
  duplicated in the restore-state file — they're already persisted
  continuously to config independent of this feature.

## Follow-up (not in this PR)

Collation and date-format probably shouldn't be global, env-derived,
restart-gated settings at all — they likely belong per-tree, following the
existing pattern name-display-format already uses (`db.get_name_group_mapping`,
per-tree database metadata) rather than the global `ConfigManager`. Noted for
a future PR.

## Test plan

- [x] `black --check` on all changed files
- [x] `mypy` — no issues
- [x] `python3 -m unittest gramps.cli.test.argparser_test gramps.gen.utils.test.grampslocale_test gramps.gui.test.restartstate_test` — 52/52 passing
- [ ] Manual end-to-end verification in a GUI environment (open tree → edit
      a record → change a restart-required preference → confirm restart →
      confirm tree/view/selection/editor restored) — not yet done, this
      sandbox's GTK bindings are mismatched and can't launch the full app

Opening as a **draft** pending that manual GUI verification and review of
the state-restoration design.

### DIFF
--- a/gramps/cli/argparser.py
+++ b/gramps/cli/argparser.py
@@ -73,6 +73,7 @@ Help options
 
 Application options
   -O, --open=FAMILY_TREE                 Open Family Tree
+  --language=LANG_CODE                   Start Gramps in LANG_CODE for this run only
   --restore-state=STATE_FILE             Restore session state from a restart
   -U, --username=USERNAME                Database username
   -P, --password=PASSWORD                Database password
@@ -141,7 +142,11 @@ gramps -O 'Family Tree 1' -e output.gramps -f gramps
 10. To generate a web site into an other locale (in german):
 LANGUAGE=de_DE; LANG=de_DE.UTF-8 gramps -O 'Family Tree 1' -a report -p name=navwebpage,target=/../de
 
-11. Finally, to start normal interactive session type:
+11. To start Gramps in German for this run only, without changing the
+LANGUAGE preference used the next time Gramps is started:
+gramps --language=de
+
+12. Finally, to start normal interactive session type:
 gramps
 
 Note: These examples are for bash shell.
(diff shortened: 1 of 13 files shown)


sam-m888:
Feature request [6210](https://gramps-project.org/bugs/view.php?id=6210) Provide preference for localization settings

Thanks 👍

PR #2469, opened 2026-07-15
Warn when the UI language's system locale isn't installed

dsblank:
## Motivation

Currently, gramps displays the following when a language pack is not installed:

```
(Gramps.py:2401429): Gtk-WARNING **: 05:37:58.349: Locale not supported by C library.
	Using the fallback 'C' locale.
```

We add the following translatable text:

```
.gramps.gen.utils.grampslocale.WARNING: Gramps could not find the system locale 
fi_FI.UTF-8 needed for the language fi. Dates, numbers, and sorting may not display
correctly until it is installed. See 
https://gramps-project.org/wiki/index.php?title=Gramps_6.1_Wiki_Manual_-_Setting_up_language_environment 
for instructions.
```

Note: That URL doesn't exist, but I couldn't find a page that would be helpful. We should remove the URL or fill that page in.

## Summary
- Gramps resolves the chosen UI language to a full system locale (e.g. `fi_FI.UTF-8`) and writes it into `LANG` for Gtk to pick up.
- If that locale isn't actually installed, Gtk's own `setlocale(LC_ALL, "")` call silently falls back to the `'C'` locale, printing only a generic, unhelpful warning: `Locale not supported by C library. Using the fallback 'C' locale.`
- `GrampsLocale` now probes for this ahead of time (`_locale_is_installed()`), and once translations are available (they aren't yet this early in construction), logs a translatable warning naming the missing locale, the language that needed it, and a wiki link with instructions on how to install it.

## Test plan
- [x] `black --check` on changed files
- [x] `mypy` on changed files
- [x] `python3 -m unittest gramps.gen.utils.test.grampslocale_test` — all passing, including new `LocaleIsInstalledTest` cases
- [x] Manually reproduced: selecting a language whose locale isn't installed now logs `Gramps could not find the system locale fi_FI.UTF-8 needed for the language fi. Dates, numbers, and sorting may not display correctly until it is installed. See <wiki-url> for instructions.` instead of Gtk's opaque message; an installed language stays silent.

### DIFF
--- a/gramps/gen/utils/grampslocale.py
+++ b/gramps/gen/utils/grampslocale.py
@@ -380,16 +380,23 @@ class GrampsLocale:
             if len(check_lang) < 2 or check_lang[1] not in ("utf-8", "UTF-8"):
                 lang = ".".join((check_lang[0], "UTF-8"))
 
-        os.environ["LANG"] = lang
         if (
             check_lang[0] not in ("C", "en")
             and sys.platform not in ("darwin", "win32")
             and not _locale_is_installed(lang)
         ):
-            # Translations aren't available yet this early in construction
-            # (self.translation is set later, in __init__); remember the
-            # missing locale and warn about it once they are.
+            # Don't hand Gtk a LANG it can't satisfy -- that's exactly what
+            # triggers its own opaque "Locale not supported by C library"
+            # warning when it makes its own setlocale(LC_ALL, "") call.
+            # Leave LANG as whatever the environment already provided (it
+            # hasn't been touched above) so date/number/collation
+            # formatting keeps following that instead. Translations aren't
+            # available yet this early in construction (self.translation
+            # is set later, in __init__); remember the missing locale and
+            # warn about it, in the user's chosen language, once they are.
             self._missing_locale = lang
+        else:
+            os.environ["LANG"] = lang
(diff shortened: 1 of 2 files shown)


sam-m888:
> https://gramps-project.org/wiki/index.php?title=Gramps_6.1_Wiki_Manual_-_Setting_up_language_environment 
> for instructions.
> Note: That URL doesn't exist, but I couldn't find a page that would be helpful. We should remove the URL or fill that page in.

I've added a stub section on the settings page in the 6.0 user manual that already has a number of sections on languages that can be cleaned up and simplified later on for the 6.2 release.  I've also changed  the section name from _Setting_up_language_environment_   to *Setting_up_**the**_language_environment*

https://gramps-project.org/wiki/index.php/Gramps_6.0_Wiki_Manual_-_Settings#Setting_up_the_language_environment

When I roll over the user manual for 6.1 the link will work correctly and be available for the 6.2 release as 

https://gramps-project.org/wiki/index.php/Gramps_6.2_Wiki_Manual_-_Settings#Setting_up_the_language_environment

What information should go into that section please, has it been written?


I've created a windows section and duplicated the existing message for windows installs that don't install all the languages

> More information can be found at [Installing_Gramps_for_Windows_computers#Missing_other_languages](https://gramps-project.org/wiki/index.php/Installing_Gramps_for_Windows_computers#Missing_other_languages) page

PR #2471, opened 2026-07-15
Made so we ignore several Legacy import errors, and added useful debug

ghcaveman:
Made so we ignore several Legacy import errors, and added line number to the "Line ignored" message


hgohel:
Summary of changes by Claude:

1. **More informative "line ignored" message** — the error text now includes the actual line number (`"Line %d ignored "`) instead of a generic `"Line ignored "`, making it easier to trace problem lines in a GEDCOM import. A minor spacing tweak was also made to the formatted output.

2. **Silently skip Legacy-specific tags** — in `__ignore()`, lines with the tags `_PREF`, `_PAREN`, `_ITALIC`, or `_NAME` are now bypassed entirely rather than logged. The accompanying comments explain that Legacy Family Tree uses `_PREF` to mark "preferred" spouses/siblings/children for chart display, and uses `_PAREN`/`_ITALIC` for source-citation styling that Gramps already handles via its own Bibliography/Report engine — so these don't need to clutter the import log.

3. **New "enhanced silent intercept" block** in `__parse_record()` — detects Legacy's custom `_PLAC_DEFN` and `_EVENT_DEFN` record definitions and, when found, consumes and discards all subordinate lines until the next level-0 record is reached (pushing that record back for normal processing). This prevents these Legacy-specific definitions from generating a flood of "not understood"/"not supported" warnings during import.

4. The "Tag recognized but not supported" message was also changed to include the actual tag data and line number for better debuggability.

hgohel:
@ghcaveman Gramps core code expects formatting with Black (see wiki on [Coding Style](https://www.gramps-project.org/wiki/index.php/Developer_policies#Coding_style)). Integrating that into your IDE/toolset will make it easier to pass the CI tests.

ghcaveman:
Can I get someone to approve my workflow?  I am interested to know if I fixed my pipeline errors

kulath:
I think it's a **really bad idea** to silently ignore tags and blocks that are not imported by Gramps. Sure, a large number of warning messages may annoy you, but what about other users? They may be totally unaware that their carefully curated Legacy data has been ignored. It doesn't seem too much of an imposition for you to just silently ignore the warnings.

Making the "Line ignored" and the "Tag recognized but not supported" messages more informative is a very good idea. However, I thought that the __add_msg method already allowed for output of the line number, would it not be better to use the existing method?

ghcaveman:
@kulath - Tim, going to have to strongly disagree with you 100%.  Any log that produces hundreds or thousands of meaningless errors is terrible because it floods the user with so much information that anything useful is totally overlooked.  And this was the perfect example in my case.  I had a gedcom file with a mere 620 people in it, and I was legitimately getting 3000+ errors.  It wasn't until I quieted the totally useless messages that I was able to see that I indeed actually did have some corrupt entries in my gedcom file.

I researched every one of the tags that I silenced, and they are all either redundant, or implemented in gramps in some other fashion using the existing tags.

kulath:
The fact that you had problems with the number of errors is not a sufficient reason for silently suppressing all the warning messages for all other users. Other users may have no idea that their GEDCOM file was not fully imported.

Why don't you copy all the messages, and use
`grep -v - E "_PREF|_PAREN|_ITALIC|_NAME" myfile.txt`
(Note, I haven't tried this command, so there may be syntax errors, but it gives the idea).
This would be much simpler that changing Gramps as you suggest.

If you were to make a change to Gramps (which I strongly oppose) then the change should be specific to the file origin (i.e. test for "Legacy" as the source).

> implemented in gramps in some other fashion using the existing tags

If the tags are implemented in Gramps in another way, then surely the correct change is to implement those tags (but only for Legacy files).

ghcaveman:
I'm just going to stop trying to share my development efforts on gramps.  It's far less frustrating.

emyoulation:
> I'm just going to stop trying to share my development efforts on gramps. It's far less frustrating.

I was under the impression that the FamilySearch Pipeline was not open for use yet.  That the 6.1 beta1 version was only usable by the FamilySearch team within their sandboxed network.

There have been murmurs of a real beta test version for actual field testing since mid-June 

@ghcaveman could you eMail me at emyoulation at yahoo dot com ?

dsblank:
Ran the tests.

ghcaveman:
I decided to make changes so the tags are not entirely suppresed.  Basically for each tag, it prints the first one and tells the user subsequent tags will be suppressed.  All of these tags are custom tags that Gramps doesn't use, so throwing away the data, then polluting the log with notifications stating "we don't care about this" is silly.  A one-liner (for each suppressed tag) improves this with the best of both worlds.  Now instead of a 3000-line log that swamps my ability to see actual problems is now 12 lines, I still understand gramps threw away some data, and I can happily see the true issues with the external data.

kulath:
Outputting the first message is a big improvement on just silently ignoring all the messages, however, I still think that it is bad practice to suppress messages about incompatible tags. If there are too many messages, then they can be simply post processed to remove the ones you want to ignore.

More specifically (and I haven't actually tried using this changed software, so please forgive me if I have misunderstood some of the code):

1. Since this processing is specific to GEDCOM output by a specific application ("Legacy Family Tree"), then it would be better if the specific functionality related to that application were in a separate import addon, especailly as a recent change has been made to "sniff" the GEDCOM to detect which addon should be used.
2. If many messages are being ignored, then in addition to outputting the first, it would be helpful if the number of messages of that particular type that had been encountered were to be output at the end. (e.g. 27 occurrences of _PLAC_DEFN encountered).
3. If you are going to change the "Line ignored" message, then it would be better to allign to line number with the line number in all the other messages.
4. For better consistency with the style of the rest of libgedcom it would be better to add the specific Legacy Family Tree tokens to TOKENS instead of placing the literals in-line in the code `line.data in ("_PLAC_DEFN", "_EVENT_DEFN")`
5. Similarly it would be more stylistically consistent if the `_PLAC_DEFN` and `_EVENT_DEFN` top level instamces were added to the string of `elif` in `__parse_record`. This would also mean that if, at some future time, it were decided to do something with the ignored tags, this would be simple to add.
6. Why does the SUPPRESSED TAG INTERCEPT need to check line.data and "token_text" attribute and line.data.startswith? Does Legacy Family Tree have a particularly peculiar way of presenting these tags, which is different from most other GEDCOMs?
7. The change in formatting for "Tag recognized but not supported" seems unnecessary, because the line number and the whole lext of the line are already in the error message.
8. Also, judging by the test files, the text after the message seems to be the remainder of the line, which is already at the end of the error message.
9. Also not all occurrence seem to have been changed (again, I think any change is unecessary).
10. Not surviving a round-trip applies to other elements of GEDCOM import, so it is not clear why this is explicitly mentioned for the Legacy Family Tree elements.
11. It's not clear why warnings are LOGed for the newly processed tags, as well as adding the warning to the output error file. LOG does not seem to be used for other ignored tags.
12. I'm not sure what the logic is to adding a note about just the first example of each tag that is subsequently going to be ignored. Perhaps it is too complicated to suppress all note about these tags.
13. In __ignore, it might be cleaner to write
```
if line.token in (TOKEN_PREF, TOKEN_PAREN, TOKEN_ITALIC, TOKEN_NAME);
    some code
elif line.token == TOKEN_UNKNOWN:
    more code
Else
    other code
```

kulath:
Thanks very much for addressing the comments (though tiresomely, I would still prefer if messages were not supressed at all). The code is much cleaner now.

Just a few points

1. I'm not sure what the supressed tags messages at the end would look like - it doesn't seem to mention that there were import errors, but these were supressed, and I'm not sure if it would be clear that the count is the number of tags encountered (all except the first being supressed).
2. The title of this PR is all about the Legacy Family Tree application, but the comment for _EVENT_DEFN and _PLAC_DEFN refer to "MyHeritage". It would be clearer if they just refer to Legacy Family Tree (Just refering to :Legacy" is ambiguous, because it might just mean that the tags are a legacy from some old version of something). skip_subordinate_levels also refer to MyHeritage, and there may be other refernces to it - I haven't checked.
3. It would be better to use TOKEN_LEGACY_FT (or something similar (rather than TOKEN_UNKNOWN_SECTION), because these tags are not unknown, they are just handleed differently by Gramps. Or perhaps better TOKEN_EVENT_DEFN and TOKEN_PLAC_DEFN, similarly to (most) other custom tags, and test for either tag later on.
4. The comment about "these tags were previously mapped onto..." is probably unnecessary, because I think it refers to how you had previously coded this PR, which will now have entirely disappeared.

ghcaveman:
@kulath:
1. The end of the file test looks like this:
The following unsupported GEDCOM tags were not imported:_EVENT_DEFN (33), _ITALIC (1), _NAME (2), _PAREN (70), _PLAC_DEFN (391), _PREF (195).
2. Hopefully I clarified this
3. I disagree on this one - the way TOKEN_UNKNOWN_SECTION was meant to be used was generic use case that happened to not exist for skipping an entire section, much in the way that TOKEN_UNKNOWN skips a one-liner.
4. Hopefully this is better too.

Thanks for the feedback,

-Kevin

PR #2472, opened 2026-07-16
Add "Prefer birth parent family" preference for ancestor views

RhoMancer:
## Summary

When a person has multiple parent families (e.g., biological and adoptive), Gramps always uses the first family in the list for Pedigree View, Ancestor gramplet, and Person Details gramplet. This means the displayed ancestry may follow the adoptive line rather than the biological line, with no way for users to choose.

This PR adds a preference (**Edit > Preferences > General > "Prefer birth parent family in ancestor views"**) that, when enabled, selects the parent family where the child relationship is marked as `ChildRefType.BIRTH`.

### Changes

- **`gramps/gen/utils/db.py`**: New `get_preferred_parent_family_handle(db, person)` utility function that checks `ChildRefType` for a BIRTH relationship across all parent families
- **`gramps/gen/config.py`**: New config key `behavior.prefer-birth-parent-family` (default: `True`)
- **`gramps/gui/configure.py`**: Checkbox in the General preferences panel
- **`gramps/plugins/view/pedigreeview.py`**: Uses new function in `find_tree()` instead of `parent_families[0]`
- **`gramps/plugins/gramplet/ancestor.py`**: Uses new function instead of `get_main_parents_family_handle()`
- **`gramps/plugins/gramplet/persondetails.py`**: Uses new function in `display_parents()` and `get_has_data()`

### Design decisions

- **Default: enabled.** Most genealogists want biological ancestry. Users who prefer the legal/adoptive line can uncheck the box.
- **Utility function approach** rather than modifying `Person.get_main_parents_family_handle()`, since that method is used by many other components (finddupes, check, reports) where the current behavior is correct.
- **Only changes ancestor-tracing views**, not tools or reports.

### Testing

Tested with Gramps 6.0.8 (Flatpak). A person with:
- Family A: father (adoptive), mother — child relation: ADOPTED
- Family B: father (biological), mother — child relation: BIRTH

With preference ON: Pedigree View shows the biological father line.
With preference OFF: Pedigree View shows the first family (previous behavior).

## Test plan

- [ ] Open a family tree with a person who has multiple parent families (e.g., biological + adoptive)
- [ ] Verify Pedigree View follows the birth line with preference enabled
- [ ] Verify Pedigree View follows the first family with preference disabled
- [ ] Verify Ancestor gramplet reflects the same preference
- [ ] Verify Person Details gramplet shows birth parents when enabled
- [ ] Verify the checkbox appears in Edit > Preferences > General
- [ ] Verify preference persists across restarts

👾 Generated with [Letta Code](https://letta.com)

Co-Authored-By: Letta Code <noreply@letta.com>


emyoulation:
Hmmm.  What happens if there are multiple parent families but none with both a BIRTH father and mother?  (e.g., a child of violence where the Father is only married to a step-mother and the Mother is only married to a step-father. So there was no "relationship" between the Father and Mother other than abuser-victim. Or the product of a one-night stand.)

RhoMancer:
@emyoulation Good question. If no parent family has a BIRTH relationship for the child, the function falls back to `family_handles[0]` (the previous default behavior). So it only changes behavior when a BIRTH family **exists** — otherwise it is a no-op.

For the specific cases you mention:
- **Child of violence / one-night stand**: If the child only has one parent family, or if none are marked BIRTH, the first family is used (same as today).
- **Father in one family, mother in another**: The function iterates all parent families and returns the first one where `frel == BIRTH or mrel == BIRTH`. So if Family A has the father with `frel=BIRTH` and Family B has the mother with `mrel=BIRTH`, Family A wins (first match). The father's ancestry would be followed. This matches how Pedigree View already works — it follows one family branch (father + mother together), not individual parent paths.

The function also respects the preference toggle — disabled means **always** use `family_handles[0]`, regardless of ChildRefType.

Just pushed a fix for the CI failures (import path + formatting).

emyoulation:
Hadn't seen [Letta Code](https://letta.com/).  So tried setting up an agent.  It broke just as we had 'gotten to know each other'. (Throwing the same error in geekspeak in response to every prompt.) What has been your experience with it?

iand:
I'm not objecting to the PR but just wanted to point out this part:

> This means the displayed ancestry may follow the adoptive line rather than the biological line, with no way for users to choose.

The user has control over the ordering of families so they can arrange to have their preferred one to be first in the list.

RhoMancer:
@emyoulation Thanks for the conversation! Re: Letta Code — I've had a positive experience with it overall. It's an AI coding agent that runs locally and has persistent memory across sessions. Like any tool, it has rough edges, but it's been genuinely useful for projects like this one. Sorry yours broke — if you reach out to their Discord (discord.gg/letta) they're usually responsive about debugging issues.

@iand Thanks for pointing that out — that's exactly the information I needed. Manually reordering parent families via the Parents tab (right-click → Move Up/Down) puts the preferred family at position 0, which achieves the same result without any code changes.

I'll withdraw this PR in favor of that existing capability. Appreciate the guidance.

PR #2475, opened 2026-07-18
FamilySearch: fix 5 bugs, and 3 performance issues

dsblank:
## Bug fixes

- **Compare button used a stale, already-committed copy of the person instead of the one open in the editor.** After adding a FamilySearch ID attribute in the Edit Person dialog (but before saving), the Compare button correctly became enabled — but clicking it raised an error that the person didn't have the attribute. The button's enable/disable check (`_tick` → `_live_person_obj`) and the click handlers (`_require_ready` → `_editor_person_obj`) were reading two different copies of the person: one live from the editor, one re-fetched from the database. Unified both paths to always use the live, in-progress `Person` object from the open editor, so Compare/Sync/Import/Link all see attributes as soon as they're added, without needing to save first.
- **Bulk import could hang indefinitely while downloading notes/sources.** The FamilySearch `Session` class never set an HTTP request timeout, so a single stalled connection anywhere in a bulk import (which can issue thousands of sequential requests for a deep ancestor/descendant pull) would hang forever with no way to recover short of killing the app. Added a default 30s timeout (overridable via `GRAMPS_FS_API_TIMEOUT`), applied centrally in `_route_api_request` so it covers every API call. Timeouts are already handled gracefully by `get_jsonurl` (logs and returns `None`), so this doesn't introduce new crashes — a stalled request now just gets skipped instead of freezing the app.
- **Bulk import could freeze for a very long time (and eventually get OOM-killed) *after* the import itself finished.** The post-import comparison-status refresh called `compare_fs_to_gramps()` once per imported person, and that function opened its own database transaction on every call. For a large import this meant thousands of separate commits, each firing a live GUI update signal and adding a new entry to the undo history, instead of one batched update — this happens after "import done." is printed. Added an optional shared `txn` parameter to `compare_fs_to_gramps()`/`FSStatusDB.commit()` so the whole post-import refresh loop now runs inside a single transaction with signals suppressed, and the GUI is rebuilt once at the end instead of once per person.
- **Bulk import could trigger the OS's "app not responding" / Force Quit dialog during every download phase, even though the import was working normally.** Each download phase (persons, ancestors, descendants, spouses, notes/sources) fetches concurrently via a `ThreadPoolExecutor`, but the main GTK thread just blocked on `as_completed()` until the whole batch finished, never returning control to the event loop — long enough for the window manager to decide the app had hung. Added an optional `progress_callback` to `Tree.add_persons()` (and `add_parents`/`add_children`/`add_spouses`, which delegate to it), invoked once per completed fetch; the GUI layer passes in a callback that pumps pending GTK events, keeping the interface responsive throughout every download phase.
- **The same "not responding" freeze could also happen *after* "import done." was printed, during the post-import comparison-status refresh.** That pass closed the progress dialog and busy cursor *before* looping over every downloaded person, so there was no progress reporting or GTK event pumping at all during what can be a genuinely slow pass — `compare_fs_to_gramps()` can issue a synchronous FamilySearch HEAD request per person when `Last-Modified` wasn't already captured during download. The whole loop was also wrapped in a single bare `except Exception: pass`, so one failed request would silently abort comparison-refresh for every remaining person with no log output at all. Kept the progress dialog open through this pass so `progress.step()` keeps pumping GTK events, moved exception handling inside the loop so one failure doesn't cancel the rest, and added debug output: periodic done/total progress, a warning whenever a single person's comparison takes over a second (to surface slow/stalled FamilySearch requests), and a final summary of elapsed time plus compared/failed counts.

## Performance

- **Parallelized the notes/sources download phase of a bulk import.** This phase fetched notes, sources, and memories one HTTP request at a time — 3 requests per person plus 2 per couple relationship, fully serial — making it by far the slowest phase for a large import, unlike the ancestor/descendant/spouse phases which already fetch each generation concurrently via a thread pool in `Tree.add_persons`. Now fetches all the raw JSON payloads concurrently (up to 8 worker threads, matching `Tree.add_persons`), then deserializes them sequentially in the main thread since deserializing mutates shared tree state that isn't safe to touch from multiple threads at once. Also added periodic progress printing during this phase, since it previously printed nothing until the whole (potentially very long) phase finished, making it look stuck even when working normally.
- **Progress bar sat frozen for long stretches during the ancestor/descendant/spouse download phases.** Those phases advanced the bar by one tick per *generation*, regardless of how many people that generation involved — later generations of a deep pull can contain hundreds or thousands of people, so the bar (and console) went quiet for the entire time it took to fetch a large generation, even though the download was proceeding normally. Switched these passes to activity mode and pass `progress.step` itself as the `progress_callback`, so the bar now pulses once per person actually fetched. Also print each generation's frontier size before fetching it and update the dialog's header text per generation, so both the console and GUI show concrete, moving feedback throughout a large pull.
- **Comparison-status refresh was the slowest part of a large bulk import.** The new per-person debug output surfaced the cause directly: `compare_fs_to_gramps()` issues a synchronous FamilySearch HEAD request per person, one at a time, whenever `Last-Modified` wasn't already captured during download — confirmed on a real 1849-person import where many individual people were each taking 1–2.5s in this pass, entirely serially. Added `backfill_last_modified()`, which resolves `Last-Modified`/`Etag` (following FSID redirects) for a batch of people concurrently via a thread pool, matching the pattern already used for notes/sources downloads. The GUI importer now runs this once, concurrently, right before the comparison loop; `compare_fs_to_gramps()` is unchanged and still resolves it itself when called standalone (e.g. the single-person Compare button), but during bulk import it now finds the value already cached and skips its own per-person network call entirely.

## Feature additions

- **"Open Person in FamilySearch" button.** Added next to the "Link/Edit FamilySearch ID" button. Opens the linked person's page at `https://www.familysearch.org/<lang>/tree/person/details/<fsid>` in a web browser, using the current Gramps UI locale for `<lang>` (falling back to `en`). Only enabled once a FamilySearch ID is actually linked (same gating as Compare/Sync/Import).
- **Clickable FamilySearch logo banner.** The logo at the top of the FamilySearch Tools window now opens `https://www.familysearch.org/` when clicked, with a pointer cursor and tooltip on hover so it reads as clickable.

## Test plan

- [x] `black --check` passes on all changed files
- [x] `mypy` passes on all changed files
- [x] `gen/fs/test/tree_test.py`, `tags_test.py`, `compare/test/comparators_test.py`, `compare/test/datmod_attr_test.py`, `compare/test/aggregate_test.py`, `fs_import/test/importer_test.py`, and `person/mixins/test/cache_test.py` all pass
- [ ] Manually verified: add a FamilySearch ID attribute in Edit Person (unsaved), confirm Compare/Sync/Import work without saving first
- [ ] Manually verified: "Open Person in FamilySearch" opens the correct browser page
- [ ] Manually verified: clicking the logo banner opens familysearch.org
- [ ] Manually verified: bulk import of a large tree no longer hangs during notes/sources download or freezes/OOMs afterward during the comparison-status refresh, and completes noticeably faster
- [ ] Manually verified: bulk import of a large tree no longer triggers the OS "not responding" / Force Quit dialog during download or comparison-status refresh, and debug output in the console shows per-phase and per-person progress
- [ ] Manually verified: during a deep ancestor/descendant pull, the progress bar visibly pulses throughout each generation's download instead of sitting still until the whole generation finishes
- [ ] Manually verified: comparison-status refresh on a large tree no longer shows repeated 1s+ per-person delays in the debug output, and completes noticeably faster than before

### DIFF
--- a/gramps/gui/fs/fs_import/importer.py
+++ b/gramps/gui/fs/fs_import/importer.py
@@ -23,6 +23,11 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 import json
 from typing import Any
 
+import gi
+
+gi.require_version("Gtk", "3.0")
+from gi.repository import Gtk  # noqa: E402
+
 from gramps.gen.db import DbTxn
 from gramps.gui.dialog import WarningDialog
 from gramps.gui.utils import ProgressMeter
@@ -40,6 +45,19 @@ from gramps.gen.fs import utils as fs_utilities
 import gramps.gui.fs.person.fsg_sync as FSG_Sync
 from gramps.gui.fs.utils.index import build_fs_index
 from gramps.gen.fs.compare import compare_fs_to_gramps
+
+
+def _pump_gtk_events() -> None:
+    """
+    Process pending GTK events.
+
+    The bulk download phases block the main thread on network I/O in
+    ThreadPoolExecutor batches, so without this the window manager sees an
+    unresponsive main loop and offers to force-quit the app even though the
+    download is progressing normally.
+    """
+    while Gtk.events_pending():
+        Gtk.main_iteration()
 
 
 class FSToGrampsImporter(CoreFSToGrampsImporter):
@@ -93,7 +111,9 @@ class FSToGrampsImporter(CoreFSToGrampsImporter):
         )
         print(_("Downloading person…"))
         if self.FS_ID:
-            self.fs_TreeImp.add_persons([self.FS_ID])
+            self.fs_TreeImp.add_persons(
+                [self.FS_ID], progress_callback=_pump_gtk_events
+            )
         else:
             if signals_disabled:
                 caller.dbstate.db.enable_signals()
@@ -112,7 +132,12 @@ class FSToGrampsImporter(CoreFSToGrampsImporter):
                 break
             done |= todo
             print(_("Downloading %d generations of ancestors…") % (i + 1))
-            todo = self.fs_TreeImp.add_parents(set(todo)) - done
+            todo = (
+                self.fs_TreeImp.add_parents(
+                    set(todo), progress_callback=_pump_gtk_events
+                )
+                - done
+            )
 
         progress.set_pass(_("Downloading descendants… (5/11)"), self.desc)
         todo = set(self.fs_TreeImp._persons.keys())
@@ -123,7 +148,12 @@ class FSToGrampsImporter(CoreFSToGrampsImporter):
                 break
             done |= todo
             print(_("Downloading %d generations of descendants…") % (i + 1))
-            todo = self.fs_TreeImp.add_children(set(todo)) - done
+            todo = (
+                self.fs_TreeImp.add_children(
+                    set(todo), progress_callback=_pump_gtk_events
+                )
+                - done
+            )
 
         if self.include_spouses:
             progress.set_pass(
@@ -131,7 +161,7 @@ class FSToGrampsImporter(CoreFSToGrampsImporter):
             )
             print(_("Downloading spouses…"))
             todo = set(self.fs_TreeImp._persons.keys())
-            self.fs_TreeImp.add_spouses(set(todo))
+            self.fs_TreeImp.add_spouses(set(todo), progress_callback=_pump_gtk_events)
 
         if self.include_notes or self.include_sources:
             progress.set_pass(
@@ -193,6 +223,7 @@ class FSToGrampsImporter(CoreFSToGrampsImporter):
                             raw_payloads[url] = future.result()
                         except Exception as exc:
                             print(f"WARNING: failed to fetch {url}: {exc}")
+                        _pump_gtk_events()
 
             # Deserializing mutates shared tree state, so do it sequentially
             # in the main thread even though the fetches above ran
--- a/gramps/gui/fs/fs_import/importer.py
+++ b/gramps/gui/fs/fs_import/importer.py
@@ -21,6 +21,8 @@ from __future__ import annotations
 
 from concurrent.futures import ThreadPoolExecutor, as_completed
 import json
+import logging
+import time
 from typing import Any
 
 import gi
@@ -45,6 +47,12 @@ from gramps.gen.fs import utils as fs_utilities
 import gramps.gui.fs.person.fsg_sync as FSG_Sync
 from gramps.gui.fs.utils.index import build_fs_index
 from gramps.gen.fs.compare import compare_fs_to_gramps
+
+LOG = logging.getLogger(__name__)
+
+# Log (and print) any single person's comparison refresh that takes longer
+# than this, to help pinpoint slow/stalled FamilySearch requests.
+_SLOW_COMPARE_THRESHOLD_S = 1.0
 
 
 def _pump_gtk_events() -> None:
@@ -306,8 +314,6 @@ class FSToGrampsImporter(CoreFSToGrampsImporter):
         self.txn = None
 
         print("import done.")
-        caller.uistate.set_busy_cursor(False)
-        progress.close()
 
         # Keep existing post-import compare refresh behavior in the GUI
         # layer, batched into a single transaction with signals suppressed.
@@ -316,17 +322,38 @@ class FSToGrampsImporter(CoreFSToGrampsImporter):
         # commits, each firing a live GUI signal and growing the undo
         # history by one entry -- effectively O(N) commits/signals/undo
         # entries instead of one.
+        #
+        # This pass can be slow: compare_fs_to_gramps() may issue a
+        # synchronous FamilySearch HEAD request per person if the person's
+        # Last-Modified wasn't already captured during download. The
+        # progress dialog is kept open (rather than closed before this
+        # loop, as before) so progress.step() keeps pumping GTK events and
+        # the app doesn't look hung while this runs.
+        compare_people = list(self.fs_TreeImp.persons or [])
+        progress.set_pass(
+            _("Refreshing comparison status…"),
+            len(compare_people) or 1,
+        )
+        print(_("Refreshing comparison status for %d people…") % len(compare_people))
+        compare_start = time.monotonic()
+        compared = 0
+        failed = 0
         caller.dbstate.db.disable_signals()
         try:
             with DbTxn(
                 "FamilySearch: refresh comparison status", caller.dbstate.db
             ) as compare_txn:
-                for fs_person in list(self.fs_TreeImp.persons or []):
+                for done, fs_person in enumerate(compare_people, start=1):
+                    progress.step()
                     gr_handle = fs_utilities.FS_INDEX_PEOPLE.get(fs_person.id)
                     if not gr_handle:
                         continue
                     gr_person = caller.dbstate.db.get_person_from_handle(gr_handle)
-                    if gr_person:
+                    if not gr_person:
+                        continue
+
+                    person_start = time.monotonic()
+                    try:
                         compare_fs_to_gramps(
                             fs_person,
                             gr_person,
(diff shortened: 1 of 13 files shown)


dsblank:
CC: @SourceAirbender @Nick-Hall

gramps/gui/fs/fs_import/importer.py
```diff
@@ -101,35 +131,68 @@ def import_tree(self, caller, FSFTID):
...
         for i in range(self.asc):
-            progress.step()
             if not todo:
                 break
             done |= todo
-            print(_("Downloading %d generations of ancestors…") % (i + 1))
-            todo = self.fs_TreeImp.add_parents(set(todo)) - done
+            print(
+                _(
+                    "Downloading ancestor generation %(gen)d/%(total)d… (%(count)d people)"
```
Nick-Hall: You should use `ngettext` here so that the plural forms can be translated correctly.

gramps/gui/fs/fs_import/importer.py
```diff
@@ -101,35 +131,68 @@ def import_tree(self, caller, FSFTID):
...
         for i in range(self.desc):
-            progress.step()
             if not todo:
                 break
             done |= todo
-            print(_("Downloading %d generations of descendants…") % (i + 1))
-            todo = self.fs_TreeImp.add_children(set(todo)) - done
+            print(
+                _(
+                    "Downloading descendant generation %(gen)d/%(total)d… (%(count)d people)"
```
Nick-Hall: Another issue with plural forms.

gramps/gui/fs/fs_import/importer.py
```diff
@@ -138,39 +201,79 @@ def import_tree(self, caller, FSFTID):
...
+                for url in urls:
+                    payload = raw_payloads.get(url)
+                    if not payload:
+                        continue
+                    self._strip_unknowns(payload)
+                    deserialize.deserialize_json(self.fs_TreeImp, payload)
+                done += 1
+                if done == 1 or done % 25 == 0 or done == total:
+                    print(
+                        _("  …notes/sources: %(done)d/%(total)d")
```
Nick-Hall, line 285: Possible plural form issue for translators.

gramps/gui/fs/fs_import/importer.py
```diff
@@ -234,25 +337,95 @@ def _load_couple_extras(rel_id: str) -> None:
...
+        # Last-Modified wasn't already captured during download. The
+        # progress dialog is kept open (rather than closed before this
+        # loop, as before) so progress.step() keeps pumping GTK events and
+        # the app doesn't look hung while this runs.
+        compare_people = list(self.fs_TreeImp.persons or [])
+        progress.set_pass(
+            _("Refreshing comparison status…"),
+            len(compare_people) or 1,
+        )
+        print(_("Refreshing comparison status for %d people…") % len(compare_people))
```
Nick-Hall: Another plural form issue.

Nick-Hall (dismissed):
I'll let @SourceAirbender do the main review but I just saw a few plural form issues for translators.

dsblank:
Thanks @Nick-Hall for noticing the ngentext issues. I'll make a PR to update AGENTS.md so that we can automate nick :smile:

gramps/gui/fs/fs_import/importer.py
dsblank: Fixed — switched to `ngettext("%(count)d person", "%(count)d people", count) % {...}` for this and the other counts. Commit 16bbcb42c7.

gramps/gui/fs/fs_import/importer.py
dsblank: Fixed the same way as the ancestors message above (ngettext, commit 16bbcb42c7).

gramps/gui/fs/fs_import/importer.py
dsblank: Took a closer look at this one — "…notes/sources: %(done)d/%(total)d" doesn't actually have a countable noun in it (it's just a fraction, like "page 3/10"), so I don't think ngettext applies here the way it does for the "N people"/"N generations" strings elsewhere in this diff. Left it as-is, but flag if you had something else in mind.

While auditing the rest of the file for the same class of issue you flagged, I did find two more real instances that weren't called out: "Downloading spouses for %d people…" and the two new "...timestamp(s)..." strings in the Last-Modified backfill step. Fixed those too, in commit 16bbcb42c7.

gramps/gui/fs/fs_import/importer.py
dsblank: Fixed — same ngettext treatment as the others (commit 16bbcb42c7). The line moved a bit in the meantime since I added a Last-Modified backfill pass just before this loop, but the string itself is unchanged: "Refreshing comparison status for %d person/people…".

gramps/gui/fs/fs_import/importer.py
Nick-Hall: OK.  I thought that the translation of "notes/sources" may change depending on the number, but maybe not.

SourceAirbender:
Thanks Doug, I really appreciate you digging into the larger import performance problems.

It all looks good to me, just a small point in the Compare flow, I think the live editor `Person` may be getting committed as a side effect. `CompareWindow` receives the live editor object, then calls `link_gramps_fs_id()`, which commits the person. 
Could that save other unsaved Edit Person changes before the user clicks OK?

Also, with regards to the gramps middleware, with the new concurrent import/backfill requests, the middleware may see several proxied requests arrive at once when the FamilySearch token is expired. 
I wonder if `ensure_fs_access_token()` needs a per-login refresh lock so multiple threads do not try to refresh the same token simultaneously.

dsblank:
Thanks for the careful read on both of these, @SourceAirbender — both were real issues. Fixed in dbcc07f7f9.

**Compare flow / live editor commit**: confirmed. `CompareWindow._open()` was calling `link_gramps_fs_id(db, self.person, self.fsid)` on whatever `Person` it was handed, which for the Compare action is the live, in-progress object from an open Edit Person dialog. `link_gramps_fs_id` unconditionally commits, so that could indeed persist any other unsaved field edits sitting in the editor before the user clicked OK. Fixed by having `CompareWindow` load a fresh copy of the person from the db by handle and commit that instead of `self.person`, so the live editor object itself is never written to the database. (Falls back to the passed-in object only if the db lookup fails, e.g. a person that genuinely isn't saved yet — same as before for that edge case.)

**Token refresh race**: `ensure_fs_access_token()` itself lives in the Foundation middleware service, so I can't touch that directly from this repo, but you're right that the client side had the same hole. `Session.refresh_access_token()` had no locking, and the new `ThreadPoolExecutor`-based concurrent fetches (notes/sources, and the existing ancestor/descendant/spouse phases) all share a single `Session` instance — so several worker threads could each hit a 401 around the same time and each fire their own refresh call. Added a `threading.Lock` around the refresh, with a double-check: a thread that loses the race sees the token another thread already refreshed and reuses it instead of requesting a second one. That at least stops this client from being the source of the simultaneous refresh requests you were worried the middleware would see.

dsblank:
I think in a future version, we should probably not piggyback off of a live-edited person and should instead have a specific modal dialog. The connection between FSID and linked items could get corrupted if changing the FSID attribute in an arbitrary way. (Maybe even make that attribute read-only in the normal GUI).

SourceAirbender:
Looks good!

Nick-Hall:
OK.  I'll merge this.  How is the middleware installation progressing?

dsblank:
> OK. I'll merge this. How is the middleware installation progressing?

@SourceAirbender if you need help let me know!

SourceAirbender:
@Nick-Hall No response from Sam yet. Still available to install the middleware myself if that's easier

Nick-Hall:
I got no response from my last email suggesting that you were given admin access to our web server.  The next step is to ask @bmatherly to give you access.  I don't have permissions to do this myself.

Nick-Hall (approved):

Nick-Hall:
Rebased.

PR #2476, opened 2026-07-18
Clarify when ngettext is required in agent guidelines

dsblank:
## Summary

- The `AGENTS.md` Internationalization section said *what* to use (`ngettext(singular, plural, n)`) but not *when* a string actually needs it. That gap let several count-embedding strings ship with a plain `_()` call in a recent FamilySearch import PR (#2475) — e.g. `_("%d people") % 1` reads as "1 people" for translators, with no way to fix it downstream since the English source string is already wrong.
- Adds a concrete trigger (a formatted count next to a noun whose English form changes between singular and plural), a code example matching the codebase's dominant `%` formatting style, an explicit counter-example for ordinal/fraction displays that don't need it (e.g. `"generation %(gen)d/%(total)d"`), and a note to also export `ngettext` alongside a centralized `_` alias in package `__init__.py` files.

## Test plan

- [x] Docs-only change; no code affected.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -202,7 +202,31 @@ All user-visible strings must be wrapped with `_()` for translation support:
 raise ValueError(_("Invalid handle: %s") % handle)
 ```
 The alias `_(string , context)` is preferred to `pgettext(context, message)`.
-Use `ngettext(singular, plural, n)` for plural forms.
+
+Use `ngettext(singular, plural, n)` instead of `_()` whenever a formatted
+string embeds a count next to a noun whose English form changes between 1
+and 2+ (person/people, entry/entries, generation/generations, match/matches,
+timestamp/timestamps, etc.):
+
+```python
+ngettext(
+    "%(count)d person",
+    "%(count)d people",
+    count,
+) % {"count": count}
+```
+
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -202,7 +202,31 @@ All user-visible strings must be wrapped with `_()` for translation support:
 raise ValueError(_("Invalid handle: %s") % handle)
 ```
 The alias `_(string , context)` is preferred to `pgettext(context, message)`.
-Use `ngettext(singular, plural, n)` for plural forms.
+
+Use `ngettext(singular, plural, n)` instead of `_()` whenever a formatted
```
Nick-Hall: We still need to use `ngettext` even when the English plural form is the same as the singular or the case where n is always greater than 1.  Many languages have more complex plural forms than in English.

dsblank: Good catch, that framing was itself wrong. Reworded to drop the "English form changes" test entirely — the rule is now "any string that counts a noun" uses ngettext, and I called out explicitly that identical English singular/plural and n-always->1 are *not* valid reasons to skip it, since other languages have plural rules (2-4 vs 5+, dual forms, etc.) English doesn't. Commit 4485f48c41.

PR #2482, opened 2026-07-27
Add last changed timestamp to metadata

DavidMStraub:
(Feature PR against **6.2** branch.)

This PR adds a single new metadata entry, namely a last change timestamp.

Currently, the file modification date of the empty file `meta_data.db` in the database directory is used for this purpose. Gramps Web API's undo transaction log relies heavily on this timestamp.

Storing the timestamp in the database instead is much more robust.

This would allow the Gramps Web Sync addon to rely on it, too, making time-based sync more reliable.

A database migration is not needed as this change is backward compatible.

Note that the timestamp is stored as float for higher accuracy, not as int as the object change column. Consequently, it is guaranteed to always be larger than the largest change timestamp in the database.

### DIFF
--- a/gramps/gen/db/generic.py
+++ b/gramps/gen/db/generic.py
@@ -880,6 +880,7 @@ class DbGeneric(DbWriteBase, DbReadBase, UpdateCallback, Callback):
         """
         # Save metadata
         self._set_metadata("version", str(self.VERSION[0]))
+        self._set_metadata("last_changed", time.time())
         self._set_metadata("name_formats", self.name_formats)
         self._set_metadata("researcher", self.owner)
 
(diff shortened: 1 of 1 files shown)


DavidMStraub:
To clarify: the timestamp is set exactly in the same instances as the `meta_data.db` one, plus at database creation. The `meta_data.db` timestamp is not removed, so we stay backward compatible.

dsblank:
@DavidMStraub master will become Gramps 6.2; Gramps 6.1 is maintenance/gramps61 and is only excepting bug fixes.

DavidMStraub:
Thanks for the heads up.

It would be great if somebody could bring the 6.1 and 6.2 roadmaps up to date. They are still skeletons.

PR #2484, opened 2026-07-30
Add from/to date modifiers to DTD/RNG schema.

Greunlis:
The PR #1124 added TO and FROM modifiers. They have, however, not been added to the validation schema, parsing a Gramps XML for my own report generator failed.

### DIFF
--- a/data/grampsxml.dtd
+++ b/data/grampsxml.dtd
@@ -450,12 +450,12 @@ SHARED ELEMENTS
 
 <!ELEMENT dateval EMPTY>
 <!ATTLIST dateval
-        val       CDATA                  #REQUIRED
-        type      (before|after|about)   #IMPLIED
-        quality   (estimated|calculated) #IMPLIED
-        cformat   CDATA                  #IMPLIED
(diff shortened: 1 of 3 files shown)


dsblank:
@Greunlis thanks for this! Would you also be able to add a Python test that verifies that this did not pass before the PR, and now it passes?

Greunlis:
The schema_validation_test.py validates ‎example/gramps/data.gramps and fails without the changes to RNG.
Or was there something else you had in mind?

dsblank:
> The schema_validation_test.py validates ‎example/gramps/data.gramps and fails without the changes to RNG. Or was there something else you had in mind?

I wonder why the CI tests were not failing before?

dsblank:
> I wonder why the CI tests were not failing before?

Oh, I see: you updated the gramps example tree so that it would fail without the fix. Looks good to me!

dsblank (approved):
Thank you; looks good. Approved.

Greunlis:
Thank you!

Nick-Hall:
Rebased onto the _gramps61_ branch.

PR #2485, opened 2026-07-30
Fix broken layout in the Multiple Surnames editor

dsblank:
## Summary
- Column widths in the Person Editor's "Multiple Surnames" table (Prefix, Surname, Connector, Origin, Primary) did not persist across dialog opens. `SurnameTab` appends its Origin and Primary columns *after* calling the base `EmbeddedList.build_columns()`, but that method restored saved widths before those extra columns existed, so Origin/Primary always reset to their hard-coded defaults.
- The Origin column used `GROW_ONLY` sizing instead of `FIXED`, which made its resize handle (between Origin and Primary) resist shrinking and feel "stuck."
- Default widths for Prefix, Connector, and Origin were quite small (150/100/150px); widened slightly (170/130/160px) so they aren't cramped out of the box.

## Fix
- Moved `restore_column_size()` in `embeddedlist.py` to run after `build_columns()` fully completes (in `EmbeddedList.__init__`), instead of partway through, so subclasses that append extra columns (only `SurnameTab` does this) get all their saved widths restored correctly.
- Set `Gtk.TreeViewColumnSizing.FIXED` on the Origin column in `surnametab.py`, matching the other columns, so drag-resizing behaves consistently.
- Bumped the default column widths.

## Test plan
- [x] `black --check` passes on changed files and repo-wide
- [x] Manually verified in a running instance: resized all 5 columns, closed and reopened the Person Editor's Multiple Surnames table, confirmed all widths (including Origin and Primary) persisted
- [ ] Reviewer: please verify on your own family tree

Fixes 0013365.

### DIFF
--- a/gramps/gui/editors/displaytabs/embeddedlist.py
+++ b/gramps/gui/editors/displaytabs/embeddedlist.py
@@ -121,6 +121,12 @@ class EmbeddedList(ButtonTab):
         self.col_icons = {}
         self.columns = []
         self.build_columns()
+        # restore_column_size() is called here rather than at the end of
+        # build_columns() because some subclasses (e.g. SurnameTab) append
+        # additional columns of their own after calling the base
+        # build_columns(); restoring saved widths must wait until all
+        # columns exist, or the extra columns never get their sizes back.
+        self.tree.restore_column_size()
 
         if self._DND_TYPE:
             self._set_dnd()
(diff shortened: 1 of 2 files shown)


emyoulation:
Thank you very much for this fix

dsblank:
It seems so very long ago :smile_cat:

PR #2488, opened 2026-08-05
Add HasNotAttributeBase filter

soig:
This enables to look for persons that do NOT have a particular attribute (eg: looking for people w/o occupation in order to know which ones we need to find that info in eg: their mariage act)

### DIFF
--- a/gramps/gen/filters/_filterparser.py
+++ b/gramps/gen/filters/_filterparser.py
@@ -243,6 +243,7 @@ old_names_2_class = {
     "Has the personal event": rules.person.HasEvent,
     "Has the family event": rules.person.HasFamilyEvent,
     "Has the personal attribute": rules.person.HasAttribute,
+    "Has not the personal attribute": rules.person.HasNotAttribute,
     "Has the family attribute": rules.person.HasFamilyAttribute,
     "Has source of": rules.person.HasSourceOf,
     "Matches the filter named": rules.person.HasSourceOf,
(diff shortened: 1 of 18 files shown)


dsblank:
Hey @soig, how is this different using a filter for people that have an attribute, and using the checkbox "Return values that do not match the filter rules"?

soig:
@dsblank  It's not the same b/c inverting results means that I cannot combine it wither other filters.
See my workflow : I'm creating a filter with that rules (not having "occupation") plus 2 rules : 

1. born between eg 1860 and 1930
2. ascendants of XX

This enables me to look for a subset of people and find out who miss that info, and thus who I need to doble check their mariage in order to find their occupation.
I usually start from one people (me, my wife, one my cousins), choose him/her as the ascendant, then go up by batches (hence the "born between X & Y")
This makes the search manageable, eg one batch a week.

If I'd use the existing rule "has an attribute", then if I seltected "Return values that do not match the filter rules", it wouldn't work with the other rules "ascendants of X" & "born between Y & Z", it would basically return every people but for those in that very small subset.

dsblank:
@soig, you can nest custom filters inside other custom filters. So you can make one filter that negates a rule, and then use that filter from others. In isn't convenient (we're working on that) but that is the way Gramps was designed. Otherwise we'd have double the number of filters just to negate them.

soig:
This is not user friendly but that works indeed.

PR #2490, opened 2026-08-06
Use geocode-glib-2.0

City-busz:
It uses `libsoup3` instead of the old, unmaintained `libsoup2`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ Python bindings for Oracle Berkeley database. Only needed to upgrade older Gramp
  to a place name. This is used if you already have osmgpsmap installed.
  If installed, when you add or link a place from the map, you have a red line
  at the end of the table for selection.
- The package name is usually gir1.2-geocodeglib-1.0 or geocode-glib.
+ The package name is usually gir1.2-geocodeglib-2.0 or geocode-glib.
 
 * [**fontconfig**](https://www.freedesktop.org/wiki/Software/fontconfig/)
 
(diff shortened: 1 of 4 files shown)


Nick-Hall:
The [osm-gps-map](https://nzjrs.github.io/osm-gps-map/) widget used in our geography views still requires `libsoup2`.

hgohel:
For context, see #2133 in which I made a change to the Windows build to use geocode-glib-2.0. Even though the official osm-gps-maps distribution depends on libsoup2 as Nick correctly pointed out, on Windows the MSYS2 team has pulled in a patch to make osm-gps-maps work with libsoup3. Could a similar approach be used for Linux and MacOS?

dsblank:
We could just fork [osm-gps-map](https://nzjrs.github.io/osm-gps-map/) until they update.

City-busz:
Alpine Linux, Arch Linux, Debian and openSUSE already patched the osm-gps-map package to make it work with libsoup3. But ideally there should be a new upstream release for osm-gps-map as libsoup2 is unmaintained and it has known security vulnerabilities.

jralls:
> We could just fork [osm-gps-map](https://nzjrs.github.io/osm-gps-map/) until they update.

Just fork it and plan to be the new maintainer, updates aren't likely. The About block on the top right says "LOOKING FOR A NEW MAINTAINER".

dsblank:
@Nick-Hall and I poked them today, and looks like they are going to release a version. BTW, @Nick-Hall offered to be that maintainer.

dsblank:
https://github.com/nzjrs/osm-gps-map/issues/102#issuecomment-5227460006

PR #2492, opened 2026-08-08
Gedcom warning improvements

ghcaveman:
Added a bit more debug to the log output when importing gedcom files.  Also improved the unit tests so the log output can be captured and tested.


emyoulation:
@ghcaveman 
There are a few things in the GEDCOM importer that reduce its capacity for version comparison:

* there are 2 options in Import tab of Preferences that change the "last modified" timestamp for records. This complicates compares when Gramps no longer shows which record is truly most recently edited in the exporting program. (Maybe the import can force the timestamp for the modified record back to what the GEDCOM indicates after adding the Gramps import context items?)   
  1) Add tag on import
  2) Add default source on GEDCOM import. (This is a "neat" feature but it does not go far enough to be useful. It would be useful if the Source was more explicit about the Source GEDCOM.  The date could be the export date of the GEDCOM.  Source author could be the Author of the Gedcom and the Publisher be the Software that generated the the export. A source media object could be the filepath/filename of the GEDCOM file.  The Citation could use the Vol/Page for the line count of the imported object and the Date for the import date.)

* When generating a Note for import failure items. The generated Note is Formatted to be human readable. Which renders that data useless for automated Post Processing. It can only be Post Processed by hand. The failed GEDCOM tag should remain in GEDCOM format. It could have the formatted content contained in an appended GEDCOM comment markers. The Note type could be set to a custom type that indicates the GEDCOM version of the GEDCOM source file instead of just "GEDCOM import".  (Attaching the generated Note also changes the timestamp of the Parent object. The last modified time for the Note is sufficient. Could that Parent object timestamp be reset?)

ghcaveman:
@emyoulation Interesting concept, but what I am after should not affect this portion of the GEDCOM importer at all.  Are you looking for help resolving this issue?  Is there an issue which describes this?  I don't mind looking into this as a separate issue.

emyoulation:
True. But you asked elsewhere if there were any comparison tools. 

These import flaws will interfere with the accuracy of what analysis/comparison tools exist. (Which are mostly for XML camparison/ harmonization.) 

And they will frustrate the efforts of contributors attempt to enhance such tools.

ghcaveman:
Ah, I see.  I was confused because the comparison tool comment was on a different thread.

As it stands, this PR does not do any data modifications in the slightest.

Nick-Hall:
Please revert the changes to the CI workflow and ensure that your code works with our standard workflow prior to merging.

PR #2499, opened 2026-08-19
Extend Hebrew date parser to accept more input formats

avma:
The Hebrew date parser (he_IL) now accepts numeric dates using a wider range of field orders and separators, in addition to the formats already supported. Users can enter dates such as 15-06-1888, 1888-15-06, or 06-15-1888 using '.', '-', or '/' as a separator; day and month are distinguished by magnitude when one value exceeds 12, falling back to the existing default order when both are ambiguous.

Abbreviated Hebrew month names are now recognized with a trailing geresh character, matching common Hebrew typographic convention for abbreviations (e.g. "15 יונ' 1888").

A bare two-year period such as "1893-1894" or "1893–1894" is now understood as a span from the first year to the second, equivalent to writing "מ־1893 עד 1894". The span and range keywords "מ" and "ל" are also recognized when followed by a plain hyphen, not only the special maqaf character.

None of this functionality previously existed for any locale in Gramps, including English and the base date parser.

### DIFF
--- a/gramps/gen/datehandler/_date_he.py
+++ b/gramps/gen/datehandler/_date_he.py
@@ -120,10 +120,24 @@ class DateParserHE(DateParser):
 
     def init_strings(self):
         DateParser.init_strings(self)
-        self._modifier = re.compile(r"%s\s*(.*)" % self._mod_str, re.IGNORECASE)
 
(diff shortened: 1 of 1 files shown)


emyoulation:
Could you review a discussion thread requesting support of Hebrew month names with Orthodox spelling variants in addition to modern spellings?

https://gramps.discourse.group/t/hebrew-month-names-spelling-change/9753/2

PR #2505, opened 2026-08-24
Make maclocale_test independent of the host locale module

dsblank:
## Summary
- `locale.LC_MESSAGES` is a POSIX-only constant CPython's `locale` module never defines on Windows.
- `maclocale.py` is fine to assume it exists: `mac_setup_localization()` is only ever called behind `if sys.platform == "darwin":` in `grampslocale.py`, so it never actually runs on Windows in production.
- `maclocale_test.py` deliberately calls `mac_setup_localization()` directly on any host OS (stubbing `subprocess.Popen` to fake being on a Mac), which let a Windows CI runner reach the `LC_MESSAGES` lookup and crash with `AttributeError` before the `Popen` stub was ever exercised.
- This failed the `maclocale_test` suite in the "Windows AIO" workflow: https://github.com/gramps-project/gramps/actions/runs/32670650282
- Fix: stand `LC_MESSAGES` in for the duration of the stubbed run in the test, rather than adding Windows-specific handling to the mac-only production module.

## Test plan
- [x] `python3 -m unittest gramps.gen.utils.test.maclocale_test -v` passes
- [x] Verified the test passes even with `locale.LC_MESSAGES` removed from the real `locale` module (simulating Windows), with `maclocale.py` unchanged from upstream
- [x] `black --check gramps/gen/utils/test/maclocale_test.py`
- [x] `mypy gramps/gen/utils/test/maclocale_test.py`

### DIFF
--- a/gramps/gen/utils/test/maclocale_test.py
+++ b/gramps/gen/utils/test/maclocale_test.py
@@ -26,6 +26,7 @@ Tests for the Mac localization helpers in :mod:`gramps.gen.utils.maclocale`.
 # Python modules
 # ------------------------
 import io
+import locale
 import os
 import subprocess
 import unittest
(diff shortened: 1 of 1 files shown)


jralls:
Sorry, trying to test maclocale.py on Windows doesn't make any sense.

Suggestion: Make gramps-ci reusable and run it on multiple runners including macOS and Windows with platform-specific tests running only on the appropriate platforms.

Nick-Hall:
OK.  I'll close this and replace it with PR #2515 which skip Mac specific tests unless running macOS.

We can also consider running the unit test suite on other platforms.

PR #2512, opened 2026-08-28
Add thumbnails to ancestry and descendant reports

ghcaveman:
Add an option to add a small thumbnail to the ancestry and descendant report dialogs so that the primary image associated with an individual may be put on the report, much like all of the major players in the family tree business.  Also included a "mask" option so that uers can make the image oval shaped, or add a frame.


ghcaveman (commented):
I think this is ready

Nick-Hall:
There seems to be some duplication of code in the two reports. Consider using a library function.

Where did you obtain the images?  Are they open source?

ghcaveman:
Will take a look at adding a library for both reports.

Images were started with ChatGPT then heavily modified.  The simpler images are my original work.

emyoulation:
I noticed that all the included masks are .png files.  Perhaps you could use .SVG format vector data instead so they scale nicely for printing?

[https://github.com/ghcaveman/gramps/tree/f89aa18fb03fbf987fdc183554e3561ff0c8633c/images/masks](https://github.com/ghcaveman/gramps/tree/f89aa18fb03fbf987fdc183554e3561ff0c8633c/images/masks)

The obvious 1st two targets for SVG format would be:
* `images/masks/oval_frame_black_line.png`
* `images/masks/oval_frame_empty.png`

I suspect that converting some intricate PNgs would not be viable. The bitmapped PNGs to embedded base64-encoded raster data SVGs would make them huge.  But some of more simplistic images (like `images/masks/oval_frame_plain_gold.png` ) should be easily re-created with shape oriented data.   

It might be possible to control the background color of SVGs in code.... or to overlay SVG transparency masks.  And that would render theme colors and dark mode into non-issues.

ghcaveman:
The graphical tools I have do not support svg files, and I am not familiar with them enough to make comment on them.  I have no issue making such a change if they work seamlessly on all platforms, and you were willing to assist with conversion.  I chose the .png format for high compression/losslessness, and the largest image is less than 300K.
None of the images are required.  I would suggest if you want .svg files, perhaps adding extra images as alternatives would be a good option in addition to the .png files.

emyoulation:
Brainstorming follows:

The assumption is that the image being overlaid with these frames will already be generated as a Normal and Large thumbnails as soon as it is added the Media collection. Which makes preemptively generating Thumbnail-sized overlay Masks files attractive. The thumbnailers take variable file formats to consistently generate `.png` format at the same resolutions. And that consistency should make compositing easier.

I generated some .svg icons as Thumbnails and discovered a serious flaw. They took the default format to generate a bitmap and then scaled that. So instead it scaled up the preview bitmap to give Gramps crappy data.  

So I have to discover how to fix the quality before it'd be worth handing off to your overlay feature.

image

The data it could have been working: 
image

PR #2513, opened 2026-08-28
Add a descendantview, so we have the opposite view from pedigreeview.

ghcaveman:
In the charts menu we previously only had an upstream view for linage.  I added a view so we are now able to see downstream/top-down as well as the orignal upstream/bottom-up view.

### DIFF
--- a/gramps/plugins/view/descendantview.py
+++ b/gramps/plugins/view/descendantview.py
@@ -953,9 +953,7 @@ class DescendantView(NavigationView):
                         )
                         outbound_stub.set_vexpand(True)
                         outbound_stub.set_valign(Gtk.Align.FILL)
-                        table.attach(
-                            outbound_stub, grid_column - 1, grid_row, 1, 1
-                        )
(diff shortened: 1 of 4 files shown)


gramps/plugins/view/descendantview.py
```diff
@@ -0,0 +1,1584 @@
...
+        return True
+
+    def on_delete(self) -> None:
+        """Save config on shutdown."""
+        self._config.save()
+        NavigationView.on_delete(self)
+
+    def on_help_clicked(self, dummy) -> None:
+        """Display the relevant portion of Gramps manual."""
+        display_url(WIKI_PAGE)
```
Nick-Hall: Use `display_help` rather than `display_url` for manual pages.

Nick-Hall:
I don't see a problem with adding this as a built-in, but we would usually trial it as a third-party addon first.

gramps/plugins/view/descendantview.py
ghcaveman: I changed this.

PR #2521, opened 2026-09-17
Restore CI coverage reporting, fix stale Codecov badge

chrisdebian:
Fixes p:gramps:14333.

Coverage hasn't actually been measured since Travis CI was removed
in 2022 (120372e88c) — Codecov was only ever wired up through
Travis's own upload step, and nothing replaced it when the project
moved to GitHub Actions. Since then, each time a new maintenance
branch was cut, the badge URLs in README.md were mechanically
updated to reference the new branch name without anyone checking
the underlying Codecov data was still live. The "42%" on the badge
today is a frozen number from whenever Travis last ran, not
anything current.

This PR:
- Runs the existing unit test suite under coverage.py and uploads
  the report to Codecov via codecov/codecov-action.
- Leaves fail_ci_if_error: false, since this repo doesn't currently
  have a CODECOV_TOKEN secret set — happy to adjust once one exists.
- Points the README's CI/Codecov badges at master rather than the
  stale maintenance/gramps60 reference.

I don't have a way to add repo secrets, so a maintainer would need
to add CODECOV_TOKEN (from the project's Codecov settings) for
uploads to start actually succeeding — the workflow change is safe
to merge either way since it won't fail CI in the meantime.

Verified locally: `coverage run -m unittest ...` followed by
`coverage xml` produces a valid report; couldn't run the full suite
end-to-end in my environment (needs bsddb3/icu/a real display), so
this is mechanically verified rather than a full green local run.

Drafted with an AI assistant (Claude Code), reviewed manually.


chrisdebian:
CI failure on the first push was in report generation, not the tests — all 32,951 tests passed. `coverage xml` choked on `CliMerge.gpr.py`, an addon downloaded into the runner's home directory (outside the checked-out repo) by the earlier install-addons step; tests exercise it, coverage traces it, but the XML step can't find its source afterward. Pushed a fix using `--ignore-errors`, coverage.py's documented flag for exactly this case.

PR #2523, opened 2026-09-21
Grizard tools plugin

ghcaveman:
This is a plugin which is intended to help merge similar GEDCOM GED and XML files as well as gramps files.


hgohel:
@ghcaveman One of my commits ([aef9f35](https://github.com/gramps-project/gramps/pull/2523/commits/aef9f35ec64b67f5912c5d19543060d43f270a9a)) is showing up in this PR; you may want to start with a clean branch so that only your commits are included.

